PR #509, opened 2015-02-25
Add ATLAS-style ?geadd function

grisuthedragon:
In contrast to standard BLAS and LAPACK the ATLAS project includes a set of functions `ATL_?geadd` which realize the matrix-matrix addition `C := alpha * A + beta * C`  with the following Fortran signature:

    ATL_?geadd(M, N, ALPHA, A, LDA, BETA, C, LDC) 

I implemented the one without further optimizations on top of `?axpby` for inside OpenBLAS. The calling sequence is 

    ?GEADD(M, N, ALPHA, A, LDA, BETA, C, LDC) 

in case of the Fortran interface. A cblas wrapper `cblas_?geadd` is supplied as well.


xianyi:
@grisuthedragon , Thank you for the patch. I fixed a bug with DYNAMIC_ARCH=1.

grisuthedragon:
Thank you for fixing this. I did not recognize it. I hope adding some more of this "BLAS - Extentions" that are made by other BLAS vendors like, ATLAS or Intel, are useful to have. When I need the next of this routines which is not included in OpenBLAS yet, I will add them again.

PR #524, opened 2015-03-27
added arm64 core types and optimized gemm kernels

wernsaar:
Compiling for target XGENE1 does not work. There are wrong symbols ( v0.d ) in sgemm_kernel_4x4.S. I don't know, who wrote this kernel.


cbergstrom:
delete this immediately - it should not be publicly available

wernsaar:
deleted.

cbergstrom:
Not yet - you have to delete the whole repo and refork to remove the commit. github doesn't gc commits so it stays online

PR #534, opened 2015-04-07
added optimized saxpy- and daxpy-kernel for haswell and sandybridge

wernsaar:
added optimized saxpy- and daxpy-kernel for haswell and sandybridge.
For Ivybridge, I can't help, because I don't have access to such a machine


xianyi:
Awesome! OpenBLAS uses Sandy Bridge kernels for Ivy Bridge.

wernsaar:
Hi,

I cannot run tests on a IVYBRIDGE machine.
But I think that ivybrige is the successor of sandybridge, so
it should work.

Best regards
Werner

xianyi:
@wernsaar  Yes, I agree with you. We can use same kernels.

PR #565, opened 2015-05-07
Add vecLib benchmarks

andreasnoack:
@wernsaar This one builds on my machine, but be aware of the ugly hard coding of the library path to libgfortran. I'm not a Makefile expert so I don't know a better way to fix this right now.

### DIFF
--- a/Makefile.system
+++ b/Makefile.system
@@ -23,6 +23,7 @@ CC = gcc
 UNAME_S := $(shell uname -s)
 ifeq ($(UNAME_S),Darwin)
      CC = clang
+     EXTRALIB += -Wl,-no_compact_unwind
 endif
 endif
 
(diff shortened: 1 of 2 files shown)


xianyi:
@andreasnoack , please try the develop branch. I delete `no_compact_unwind` flag and detect the path of `-lgfortran`.

andreasnoack:
I've just tried it and it works fine. I had to add the path and the `no_compact_unwind` to make it compile, but initially. Is it

https://github.com/xianyi/OpenBLAS/blob/6743beb7480d8e31e11211cb4fce5954749ee70e/f_check#L41

that makes a difference?

xianyi:
Yes, the `f_check` should detect the path of `-lgfortran` and store it at `FEXTRALIB`. The line 41 fixed the bug.

PR #595, opened 2015-06-13
Fix test execution when USE_OPENMP=0

tanderson92:
The standard way to disable OpenMP support is to set USE_OPENMP=0,
as indicated by other checks to see if USE_OPENMP equals 1. The
problem is obviously then that `ifdef USE_OPENMP` is very much not
what we want to test for. This causes tests to fail when no OpenMP
library is installed.

### DIFF
--- a/Makefile.system
+++ b/Makefile.system
@@ -591,7 +591,7 @@ else
 FCOMMON_OPT += -m32
 endif
 endif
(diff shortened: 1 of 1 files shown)


xianyi:
Thank you for the patch!

PR #612, opened 2015-08-04
ppc64le platform support (ELF ABI v2)

mtbrandy:


xianyi:
Thank you for the patch.

Issue #632, opened 2015-09-06
The build system doesn't know about SkyLake


yuyichao:
I'm trying to compile OpenBLAS on a SkyLake desktop and see many compile errors very similar to the one reported in https://github.com/xianyi/OpenBLAS/issues/332 . I can understand that it would take some time to tune the kernels on a newly release hardware but it would be nice to be able to compile first.

yuyichao:
I'm adding `TARGET=HASWELL` to make it compile now. (Tried `BROADWELL` and `SKYLAKE` but they don't seem to be acceptable names yet.)

martin-frbg:
"TARGET=BROADWELL" should work on the "develop" branch (which you will probably want to try rather than the already somewhat dated 0.2.14 release)

yuyichao:
I'm pulling from the default branch from github (which should be develop) and `TARGET=BROADWELL` doesn't seem to work. The commit I'm on is 40a3fed and the full make command line argument I have is

```bash
_config=(
  'SMP=1'
  'USE_OPENMP=1'
  'USE_THREAD=1'
  'MAJOR_VERSION=3'
  'NO_LAPACK=0'
  'NUM_THREAD=8'
  'TARGET=BROADWELL'
  'DYNAMIC_ARCH=1'
)
```

martin-frbg:
My mistake, seems the fix from #583 only added the cpuid mapping for Broadwell, not a BROADWELL target as such.

yuyichao:
That's what it seems like for me after a `git grep` as well. =)

xianyi:
@yuyichao , I didn't optimize skylake yet. Therefore, please use haswell kernels. make TARGET=HASWELL

yuyichao:
Generic binary with DYNAMIC_ARCH recognize skylake as Prescott. Would probably be good enough for now to just fix this first.

yuyichao:
Since this issue is about recognising SkyLake rather than optimize for it, I'll close this now.

PR #635, opened 2015-09-09
Support Intel Skylake by Haswell kernels.

yuyichao:
Ref #632 

This makes OpenBLAS compiles and also recognize my computer as Haswell instead of Prescott. I didn't find a complete list of CPUID's for skylake processors and only adding the model/family numbers for 6700K so there's almost certainly missing ones.

### DIFF
--- a/cpuid_x86.c
+++ b/cpuid_x86.c
@@ -1150,6 +1150,16 @@ int get_cpuname(void){
 #endif
           else
 	    return CPUTYPE_NEHALEM;
+        case 14:
+	  // Skylake
+          if(support_avx())
+#ifndef NO_AVX2
+            return CPUTYPE_HASWELL;
+#else
+	    return CPUTYPE_SANDYBRIDGE;
+#endif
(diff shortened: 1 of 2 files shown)


yuyichao:
And this is the Julia `versioninfo()` showing that `OpenBLAS` recognize the CPU as Haswell instead of Prescott now....

```
julia> versioninfo()
Julia Version 0.4.0-pre+7407
Commit fc936a0* (2015-09-09 08:12 UTC)
Platform Info:
  System: Linux (x86_64-unknown-linux-gnu)
  CPU: Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz
  WORD_SIZE: 64
  BLAS: libopenblas (DYNAMIC_ARCH NO_AFFINITY Haswell)
  LAPACK: libopenblas
  LIBM: libm
  LLVM: libLLVM-3.6.2
```

xianyi:
Thank you for the patch

Issue #597, opened 2015-06-15
OpenBLAS/lapack-netlib 'make' numerical errors in ARMv7 multi-threaded only


Jagmn:
When running a 'make' in the 'lapack-netlib' directory of OpenBLAS on an ARMv7 target, I get different numerical failures  when using multi-threading.

```
--- ARMv7 (Raspberry Pi 2) x 4 (no OpenMP), built with 'make TARGET=ARMV7'
            -->   LAPACK TESTING SUMMARY  <--
        Processing LAPACK Testing output found in the TESTING direcory
SUMMARY                 nb test run     numerical error     other error  
================    =========== =================   ================  
REAL                528599      11  (0.002%)    0   (0.000%)    
DOUBLE PRECISION    605768      1   (0.000%)    0   (0.000%)    
COMPLEX             434073      114 (0.026%)    0   (0.000%)    
COMPLEX16           522687      70  (0.013%)    0   (0.000%)    

--> ALL PRECISIONS  2091127     196 (0.009%)    0   (0.000%)    


--- ARMv7 (Raspberry Pi 2) x 4 (no OpenMP), OPENBLAS_NUM_THREADS=1 
            -->   LAPACK TESTING SUMMARY  <--
        Processing LAPACK Testing output found in the TESTING direcory
SUMMARY                 nb test run     numerical error     other error  
================    =========== =================   ================  
REAL                1109189     0   (0.000%)    0   (0.000%)    
DOUBLE PRECISION    1110527     0   (0.000%)    0   (0.000%)    
COMPLEX             581782      0   (0.000%)    0   (0.000%)    
COMPLEX16           582594      0   (0.000%)    0   (0.000%)    

--> ALL PRECISIONS  3384092     0   (0.000%)    0   (0.000%)    
```

Jagmn:
I should have commented: There are no numerical errors (or any other kinds of errors) on any of my x86_64 targets regardless of number of threads.

martin-frbg:
Is this with the 0.2.14 release or current git "develop" branch ? (If the former, could you retest with the latter just in case ?)

Jagmn:
Both, I'm afraid. Though the 0.2.14 release deadlocks occasionally during the test (or appears to).

xianyi:
@wernsaar , 

I just run lapack test on our lab's ARM Cortex-A15 board.

The result of 4 threads (without USE_OPENMP)

```
			-->   LAPACK TESTING SUMMARY  <--
[...]

Issue #246, opened 2013-07-07
stack overflow bug in getrf_parallel


xianyi:
This bug is track julialang/julia#3630

When NUM_THREADS is very large, e.g. 128 or 256, it causes the stack overflowing in getrf_parallel.

There is a similar bug in potrf_parallel.

PR #696, opened 2015-11-23
Cortex A57 fixes and Lapack 3.6.0

ashwinyes:
- lapack-test fixes for Cortex A57
- Update to lapack 3.50


martin-frbg:
Do not see this on my various Linux/x86 systems either, suspect this is actually an ARM (or at most OS_ANDROID) thing, not Linux in general, cf. #597

xianyi:
@martin-frbg , @jeromerobert 

Please see #246

I think the problem is on big machine. For `getrf`, the current threshold `GETRF_MEM_ALLOC_THRESHOLD` is 80. When the number of cores is 80, the `job_t` array will be 80k on the stack.

martin-frbg:
Hmm. Wonder if it could be made conditional on "high" number of cores then (I see that the code initializing blas_cpu_number comes immediately after the function, so would only need to be shuffled up a few lines it seems) ?

PR #733, opened 2016-01-05
Do not use vsub to clear the register values

yuyichao:
A summary (to the best of my knowledge) of different ways to zero a fp register on arm (also available in the code comment https://github.com/xianyi/OpenBLAS/commit/594b9f4c73d314a35ed56a484c42c08a1967c283#diff-ea578ff593f8cc88b2eaceaa1003b051R918)

* `vsub`, this doesn't work for `NaN`, and `+-Inf`
* `vmov.fn` (immediate): doesn't work for zero
* `vmov.In` (immediate) and `veor`: these are NEON instructions
* `vldr` (pseudo-instruction): This let the assembler to generate the constant pool automatically and is used whenever possible in this patch. However, the assemble can only put the pool at the end of the file and therefore doesn't work if the function is larger than 1KB (maximum `pc` relative load offset).
* `vldr`: manually create the constant. This is used when the pseudo instruction doesn't work.
* I've also thought about `vmov.fn` a `1.0` immediate and do `vsub`. Not sure if it's faster.

Close https://github.com/JuliaLang/julia/issues/10260

### DIFF
--- a/kernel/arm/asum_vfp.S
+++ b/kernel/arm/asum_vfp.S
@@ -368,11 +368,11 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 	.align 5
 
 #if defined(DOUBLE)
-	vsub.f64                d0 , d0 , d0
-	vsub.f64                d1 , d1 , d1
+	vldr.f64                d0 , =0
+	vldr.f64                d1 , =0
 #else
-	vsub.f32                s0 , s0 , s0
-	vsub.f32                s1 , s1 , s1
+	vldr.f32                s0 , =0
+	vldr.f32                s1 , =0
 #endif
 
 	cmp	N, #0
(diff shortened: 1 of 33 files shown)


yuyichao:
The AppVeyor failure looks unrelated. The `linalg/dense` test in julia passes with this patch. I'm currently running the full julia linalg test suite on this. but it will take some time (single process due to julia bugs, will take a few hours).

Also, how can I run the OpenBLAS test?

@xianyi

xianyi:
@yuyichao , Thank you for the patch.

Please refer https://github.com/xianyi/OpenBLAS/wiki/Developer-manual#run-openblas-test for the OpenBLAS Test.

yuyichao:
Thx. The Julia linalg tests passes (apart form an issue with julia itself that we are investigating). I'll try the OpenBLAS tests.

yuyichao:
I haven't try the BLAS-Tester yet but `make ctest` `make test` and `make lapack-test` doesn't seems to be doing anything.

xianyi:
@yuyichao , OK. I merged the patch.

yuyichao:
thx

PR #743, opened 2016-01-18
re enable Fortran optimization flag on windows

tkelman:
partial revert of https://github.com/xianyi/OpenBLAS/commit/299cdcdc29999d591fcb300630d50b2986bfb6fc
from #696, was not explained why that was needed cc @wernsaar

### DIFF
--- a/Makefile.system
+++ b/Makefile.system
@@ -986,11 +986,7 @@ COMMON_OPT = -O2
 endif
 
 ifndef FCOMMON_OPT
-ifeq ($(OSNAME), WINNT)
-FCOMMON_OPT = -O0 
-else
 FCOMMON_OPT = -O2 -frecursive
-endif
 endif
 
 


xianyi:
@tkelman , did you run lapack test on windows?

tkelman:
Not yet. I'll try to remember to do that when I have a moment. Keep reminding me every few days until I do.

tkelman:
I couldn't figure out how to run lapack's tests directly, but all of Julia's tests pass on an i7-6700HQ, windows 64 bit build of openblas commit 5fef0d1

martin-frbg:
"make lapack-test" probably, or are you using an IDE rather than a plain makefile environment ?

tkelman:
using makefiles, but with the build configuration customized as julia needs. I'm also cross compiling from cygwin which the test scripts might not be written in a way that can handle, but I'll give it a try.

tkelman:
```
                        -->   LAPACK TESTING SUMMARY  <--
SUMMARY                 nb test run     numerical error         other error
================        ===========     =================       ================
REAL                    1169109         0       (0.000%)        0       (0.000%)
DOUBLE PRECISION        1169921         0       (0.000%)        0       (0.000%)
COMPLEX                 633152          0       (0.000%)        0       (0.000%)
COMPLEX16               633964          0       (0.000%)        0       (0.000%)

--> ALL PRECISIONS      3606146         0       (0.000%)        0       (0.000%)
```

Issue #730, opened 2016-01-02
Suprisingly slow `dsyrk` on Mac


jakirkham:
Here is an example on Mac OS 10.9 on a Late 2013 MacBook Pro 15" with a 2.3 GHz Intel Core i7-4850HQ using OpenBLAS 0.2.15, Python 2.7.11, NumPy 1.10.2, SciPy 0.16.1. I have verified that NumPy and SciPy are properly linked to the version of OpenBLAS specified. OpenBLAS has built with the following flags `make DYNAMIC_ARCH=1 BINARY=64 NO_LAPACK=0 NO_AFFINITY=1 NUM_THREADS=1` no other options or modifications were made.

```
>>> import numpy
>>> from scipy.linalg import blas
>>> a = numpy.ones((100,101), dtype=numpy.float32)
>>> %timeit blas.ssyrk(1, a)
10000 loops, best of 3: 37.8 µs per loop
>>> a = numpy.ones((100,101), dtype=numpy.float64)
>>> %timeit blas.dsyrk(1, a)
1000 loops, best of 3: 520 µs per loop
```

I wouldn't be surprised to see it takes twice as a long due to the fact that it is double versus single precision; however, taking over an order of magnitude longer seems to be a bit much.

Following the same build procedure on a Linux VM on the same machine (uses VirtualBox 5.0.12), I get a much more reasonable time for `dsyrk` (around double `ssyrk`). I have no idea whether this carries over to Windows or not. If someone is able to reproduce a similar example using C or Fortran, please share your steps.

After further discussion, we found this was array size dependent. Below is a graph showing this dependence. More details about how the graph was made can be found in this comment. ( https://github.com/xianyi/OpenBLAS/issues/730#issuecomment-171545950 )

![performance_of_openblas_syrk_subroutines](https://cloud.githubusercontent.com/assets/3019665/12317033/00983468-ba5b-11e5-9d10-fe25db36df99.png)

For comparison, one can look at the time taken by `sgemm` and `dgemm`, but one will not see this behavior.

![performance_of_openblas_gemm_subroutines](https://cloud.githubusercontent.com/assets/3019665/12418718/f348683c-be7e-11e5-8a98-24bfc2249ed4.png)

brada4:
Are you sure result is from OpenBLAS and not Accelerate Framework running on (float64-impaired, HD-Iris) GPU?
Single line versions of test cases:
```
[...]

PR #745, opened 2016-01-19
BENCH: Minor fixes in SciPy benchmarks

jakirkham:
* Allocate a normal zeroed array instead of a random one.
* Write results into the array.
* Drop unneeded semicolons
* Use environment Python.
* Fix PEP8 issues.
* Use Fortran arrays.

### DIFF
--- a/benchmark/scripts/SCIPY/dsyrk.py
+++ b/benchmark/scripts/SCIPY/dsyrk.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
 
 import os
 import sys
--- a/benchmark/scripts/SCIPY/dsyrk.py
+++ b/benchmark/scripts/SCIPY/dsyrk.py
@@ -16,7 +16,7 @@ def run_dsyrk(N, l):
 
     start = time.time()
     for i in range(0, l):
-        C[...] = blas.dsyrk(1.0, A)
+        blas.dsyrk(1.0, A, c=C, overwrite_c=True)
     end = time.time()
 
     timediff = (end - start)
--- a/benchmark/scripts/SCIPY/dsyrk.py
+++ b/benchmark/scripts/SCIPY/dsyrk.py
@@ -16,7 +16,7 @@ def run_dsyrk(N,l):
 
 	start = time.time()
 	for i in range(0,l):
-		C = blas.dsyrk(1.0,A)
+		C[...] = blas.dsyrk(1.0,A)
 	end = time.time()
 	
 	timediff = (end -start) 
--- a/benchmark/scripts/SCIPY/dsyrk.py
+++ b/benchmark/scripts/SCIPY/dsyrk.py
@@ -14,7 +14,7 @@ def run_dsyrk(N,l):
 	C = zeros((N,N), dtype='float64')
 
 
-	start = time.time();
+	start = time.time()
 	for i in range(0,l):
 		C = blas.dsyrk(1.0,A)
 	end = time.time()
@@ -49,7 +49,7 @@ if __name__ == "__main__":
 	if 'OPENBLAS_LOOPS' in os.environ:
 		p = os.environ['OPENBLAS_LOOPS']
 		if p:
-			LOOPS = int(p);
+			LOOPS = int(p)
 
 	print("From: %d To: %d Step=%d Loops=%d" % (N, NMAX, NINC, LOOPS))
 	print("\tSIZE\t\t\tFlops\t\t\t\t\tTime")
--- a/benchmark/scripts/SCIPY/dsyrk.py
+++ b/benchmark/scripts/SCIPY/dsyrk.py
@@ -4,13 +4,14 @@ import os
 import sys
 import time
 import numpy
+from numpy import zeros
 from numpy.random import randn
 from scipy.linalg import blas
 
 def run_dsyrk(N,l):
 
 	A = randn(N,N).astype('float64')
-	C = randn(N,N).astype('float64')
+	C = zeros((N,N), dtype='float64')
 
 
 	start = time.time();
--- a/benchmark/scripts/SCIPY/dsyrk.py
+++ b/benchmark/scripts/SCIPY/dsyrk.py
@@ -8,52 +8,51 @@ from numpy import zeros
 from numpy.random import randn
 from scipy.linalg import blas
 
-def run_dsyrk(N,l):
 
-	A = randn(N,N).astype('float64')
-	C = zeros((N,N), dtype='float64')
+def run_dsyrk(N, l):
 
+    A = randn(N, N).astype('float64')
+    C = zeros((N, N), dtype='float64')
 
-	start = time.time()
-	for i in range(0,l):
-		C[...] = blas.dsyrk(1.0,A)
-	end = time.time()
-	
-	timediff = (end -start) 
-	mflops = ( N*N*N) *l / timediff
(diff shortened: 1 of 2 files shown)


xianyi:
I am not the expert in Python.  I think the preallocation C is for reducing the overhead in timing BLAS function.
@wernsaar any ideas?

jakirkham:
I am nearly certain that SciPy is still doing this allocation for you. In an event, if C has an array beforehand we are not writing into it here. It will simply be deallocated and a new array will be allocated for the return result.

xianyi:
@jakirkham , thank you for the answer.

jakirkham:
If we wanted to reuse `C`, we might need something like `dsyrk(1.0, A, c=C, overwrite_c=True])` for instance.

jakirkham:
Python variables could be thought of like C pointers with some reference counting attached and possible garbage collection. All Python variables are pointers to Python objects. So, if I assign a new value to a Python variable, it merely points to a new Python object. The underlying Python object is unaffected by this assignment.

jakirkham:
The only other points I would add are `C` is allocated before timing begins. So, it isn't factoring into the timing. Also, as `C` is being filled with random values this is a bit slow. One would prefer `numpy.empty` to allocate an array that has not been zeroed. If one wants to overwrite that array, replace `C =` with `C[...] =`.

zerothi:
However, C gets allocated over and over.
As will be clear from below, empty shouldn't be used as it does it like C
(empty is a _reservation_, not _allocation_).

The scipy implementation recognizes if the variable is pre-allocated.
However, ONLY if doing array assignments.

What should be done is:

C = np.zeros([N,N],np.float32)
C[:,:] = blas.ssyrk(...)


You can check the memory profiling using these small scripts:
#####
import numpy as np
from scipy.linalg import blas
from memory_profiler import profile

N = 1000
A = np.random.randn(N,N).astype('float64')

@profile
def blas1():
    C = np.zeros([N,N],np.float64,order='F')
    C = blas.dsyrk(1.0,A)
    del C

@profile
def blas2():
    C = blas.dsyrk(1.0,A)
    del C

@profile
def blas3():
    C = np.zeros([N,N],np.float64,order='F')
    C[:,:] = blas.dsyrk(1.0,A)
    del C


blas1()
blas1()
blas2()
blas2()
blas3()
blas3()

print("Mem: {} MB".format(N**2*8./1024**2))

#####
Then use the memory_profiler module to check how it allocates/de-allocates.
Cleary method 3 is the best in terms of memory. Only on the first call will
it do some allocation in the blas library.

Hence to get the best timing, one should:

1. Pre-allocate all arrays
2. Do a pre-initialization call to the BLAS routine
3. Perform stress test, WITH array assignments




2016-01-19 18:54 GMT+01:00 jakirkham <notifications@github.com>:

> The only other points I would add are C is allocated before timing
> begins. So, it isn't factoring into the timing. Also, as C is being
> filled with random values this is a bit slow. One would prefer numpy.empty
> to allocate an array that has not been zeroed. If one wants to overwrite
> that array, replace C = with C[...] =.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/xianyi/OpenBLAS/pull/745#issuecomment-172933037>.
>



-- 
Kind regards Nick

jakirkham:
I am not following by what you mean in terms of reservation vs. allocation here, @zerothi. Can you please clarify?

zerothi:
When you do:
`np.empty(..)`
all numpy is doing is _reserving_ a memory block to be later used.
Only when you start assigning values to the variable will the actual allocation take place.

Hence, if you do:
```
C = np.empty(...)
tic
C[:,:] = 1.
toc
```
you will effectively also time the allocation of the variable. Essentially...

Take this code:
```
C = np.empty([N,N],np.float64,order='F')
C[:,:] = 1.
del C
```
the memory profiler yields:
```
Line #    Mem usage    Increment   Line Contents
================================================
    27     47.9 MiB      0.0 MiB       C = np.empty([N,N],np.float64,order='F')
    28     55.4 MiB      7.5 MiB       C[:,:] = 1.
    29     55.4 MiB      0.0 MiB       del C
```
However, if you used `zeros` you would get:
```
Line #    Mem usage    Increment   Line Contents
================================================
    27     55.1 MiB      7.5 MiB       C = np.zeros([N,N],np.float64,order='F')
    28     55.1 MiB      0.0 MiB       C[:,:] = 1.
    29     55.1 MiB      0.0 MiB       del C
```
Empty has it usages, but I would not use it for benchmarks.

jakirkham:
Interesting, so if I do the following.

    >>> C = np.empty([N,N],np.float64,order='F')
    >>> print(C)

Does this result in allocation when printing `C`?

zerothi:
There are corner cases I am not fully sure about. In this case it depends on what `print` does to C.

It also depends on the allocation routines that is linked in to Python/numpy.
So with Python it is a bit difficult to know exactly what happens. :)

jakirkham:
After some initial issues with `memory_profiler`, it seems this is the answer.

    Line #    Mem usage    Increment   Line Contents
    ================================================
         4     34.4 MiB      0.0 MiB   def test():
         5     34.4 MiB      0.0 MiB       N = 100
         6     34.4 MiB      0.0 MiB       C = np.empty([N,N],np.float64,order='F')
         7     34.6 MiB      0.2 MiB       print(C)

In short, allocation does not occur until the `print`. I had always assumed that the allocation had already occurred because `print` would show an array with non-zeroed values. Thanks for helping me realize that `empty` is only *reserving* data, @zerothi.

zerothi:
Oh, remark that your PR should still do `C[:,:] = blas....` calls, otherwise it will de-allocate/allocate `C` on every call.

jakirkham:
@zerothi, do you know why this `dsyrk(1.0, A, c=C, overwrite_c=True)` doesn't work? Based on what I have seen in general BLAS documentation and the little bit in SciPy, I would figure this would write the result in `C`, but it doesn't.

zerothi:
No, I do not know why. You could perhaps create an issue with `scipy`? :)
Besides, the other way is (in my opinion) more pythonic:  
`C[:,:] = blas.*syrk....`

jakirkham:
Added ( https://github.com/scipy/scipy/issues/5738 ). Will change. Though I think it would be desirable to benefit from this option.

jakirkham:
Turns out it will work, but only on Fortran arrays.

jakirkham:
How does this look?

wernsaar:
Hi,

dsyrk is defined as:

C := alpha*A'*A + beta*C

I use the benchmark scripts to compare the performance against
the benchmarks written C.

If you don't preallocate and initialize the matrix C, then you will have
malloc and page faults between the start and stop time and page faults
are very expensive on some platforms.

Best regards
Werner

jakirkham:
Ok, so, we are still initializing the array in advance. It is just a bit more cleaned up. Also, we are making sure this array is passed in as `C` so that memory is available to be written into.

BTW, have you seen this ( https://github.com/xianyi/OpenBLAS/issues/730 ), @wernsaar? We are kind of at a loss for why this weird behavior is happening.

wernsaar:
Hi,

could you please benchmark sgemm and dgemm with scipy on your platform.
I want to know, if there is a similar behavior.

Best regards

Werner

jakirkham:
So, I did, but I don't see similar behavior. It is the second graph in the description. Or is there another way that you would like me to benchmark these?

PR #749, opened 2016-01-22
illumos fixes

### LINKED ISSUE #700: Compiling on Solaris failed - memory.c: error: constructor priorities are not supported
Hi Xianyi,

I am trying to compiling the latest version 0.2.15 and the develop version on Sun Solaris using gcc 4.7.3.

However, it stop at
...
memory.c:1342:1: error: constructor priorities are not supported
memory.c:1397:1: error: destructor priorities are not supported
...

I saw a similar problem with Mac. Do you have any idea how to fix it?

Thank
Archy

lotheac:
Fixes #700 and a few other issues with the build on illumos (OmniOS to be exact).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,7 +85,7 @@ endif
 
 shared :
 ifndef NO_SHARED
-ifeq ($(OSNAME), Linux)
+ifeq ($(OSNAME), $(filter $(OSNAME),Linux SunOS))
 	@$(MAKE) -C exports so
 	@-ln -fs $(LIBSONAME) $(LIBPREFIX).so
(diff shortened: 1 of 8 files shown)


lotheac:
This one was required too, but I don't know if it's going to break any other OSes so I didn't include it in this PR. https://github.com/lotheac/OpenBLAS/commit/66dcd655b665beebb1a7995a0c86cfd7af990dce

lotheac:
I amended this with the f_check patch modification that you suggested in #700. It still builds fine, but I should note generally about this PR that while both Solaris and illumos identify as 'SunOS' in uname, there may be differences and I do not have a Solaris system to test these with.

PR #771, opened 2016-02-02
Fix the source paths and use CMAKE_C_COMPILER_ID to properly set -fopenmp

byzhang:


xianyi:
Thank you for the cmake patch.

PR #820, opened 2016-03-24
Affinity shared memory area must be at least a page

antonblanchard:
If the system page size is greater than 4kB (or 32kB in the case of
BIGNUMA), then the affinity code will fail to initialise.

Ensure the shared memory area is at least one page in size.


xianyi:
Any problems?

PR #847, opened 2016-04-22
MIPS P5600(32 bit) and I6400(64 bit) cores support added.

sva-img:
Seperated mips and mips64 files.
Configurations support for mips 32 bit.
MSA optimizations to follow in subsequent patches.

Signed-off-by: Shivraj Patil <shivraj.patil@imgtec.com>


xianyi:
@sva-img , thank you for the patch.

Is it possible to give me remote MIPS machine access? Then, I can test the code.

Issue #857, opened 2016-04-27
building v0.2.18 fails depending on CC variable.


vondele:
/data/vjoost/toolchain-tst/build/OpenBLAS-0.2.18> make -j 12 CC=/data/vjoost/toolchain-tst/install/gcc-6.1.0/bin/gcc  FC=/data/vjoost/toolchain-tst/install/gcc-6.1.0/bin/gfortran >& log

/data/vjoost/toolchain-ar  -ru ../libopenblas_nehalemp-r0.2.18.a saxpy.o [...]
make[1]: /data/vjoost/toolchain-ar: Command not found
../Makefile.tail:40: recipe for target 'libs' failed

The reason seems to be that this leads to a definition in Makefile.conf of 

CROSS_SUFFIX=/data/vjoost/toolchain-

which causes other problems later on. This comes from c_check which is called as

perl ./c_check Makefile.conf config.h /data/vjoost/toolchain-tst/install/gcc-6.1.0/bin/gcc

and somehow gets confused by the path name in the CC definition, which includes dashes.

martin-frbg:
Looks a lot like #605 (which xianyi failed to reproduce back then)

vondele:
yes, looks like the same issue. Should be easy to reproduce, but the problem is not a dash in the path to openblas, but the dash in the path of CC. The detection in c_check would be somewhat more robust if the path were stripped first.

buffer51:
Well the problem was that in your case, the path has one but the compiler executable doesn't, which isn't the case usually. Could you try [my commit](https://github.com/buffer51/OpenBLAS/commit/20b0ed1da598a98b2b5036bc0decbd4aae79aa20) to confirm that it fixes the issue?

I tried a bunch of combinations with dashes in the path or the executable itself, it seemed to handle them correctly.

xianyi:
@buffer51 , thank you for the patch. Actually, I am not good at perl :)

vondele:
@buffer51, unfortunately, it doesn't quite work. Since CC=/data/vjoost/toolchain-tst/install/gcc-6.1.0/bin/gcc this results in CROSS_SUFFIX=/data/vjoost/toolchain-tst/install/gcc-6.1.0/bin/ which causes the build to fail as this leads to executing /data/vjoost/toolchain-tst/install/gcc-6.1.0/bin/ar which doesn't exist. Note that I'm not cross-compiling, this is just gcc being installed in a non-standard place.

buffer51:
[...]

PR #858, opened 2016-04-27
Fixed cross-suffix detection for path that contains dashes

### LINKED ISSUE #605: c_check assumes nonempty $cross_suffix if run in a path with a dash
When executing make in a path with a dash it in, the `c_check` program automatically detects cross_suffix, but it seems like anything with a dash is assumed to be a cross compile. I don't know whether this is intentional or not, but perhaps a README warning is in order.
```
make[1]: /idi/yesimon-ar: Command not found
../Makefile.tail:40: recipe for target 'libs' failed
make[1]: *** [libs] Error 127
make[1]: Leaving directory '/idi/yesimon-scratch/yesimon/tmp/openblas20150724-14846-n6dn4n/OpenBLAS-0.2.14/interface'
Makefile:141: recipe for target 'libs' failed
make: *** [libs] Error 1
```
From `c_check`: https://github.com/xianyi/OpenBLAS/blob/3f1b57668e3d73944981d7bda194ca133f4660b7/c_check#L28

buffer51:
A more robust approach, which can handle the special case when the path has dashes but the compiler name doesn't.
Also prevents CROSS_SUFFIX from being set while CROSS=0 just because a specific path is set for CC.
Fixes #605 and #857.

### DIFF
--- a/c_check
+++ b/c_check
@@ -1,5 +1,7 @@
 #!/usr/bin/perl
 
+use File::Basename;
+
 # Checking cross compile
 $hostos   = `uname -s | sed -e s/\-.*//`;    chop($hostos);
 $hostarch = `uname -m | sed -e s/i.86/x86/`;chop($hostarch);
@@ -26,14 +28,12 @@ if ($?) {
 
 $cross_suffix = "";
 
-if ($ARGV[0] =~ /(.*)(-[.\d]+)/) {
-    if ($1 =~ /(.*-)(.*)/) {
-	$cross_suffix = $1;
-    }
-} else {
-    if ($ARGV[0] =~ /([^\/]*-)([^\/]*$)/) {
-	$cross_suffix = $1;
-    }
+if (dirname($compiler_name) ne ".") {
+    $cross_suffix .= dirname($compiler_name) . "/";
+}
+
+if (basename($compiler_name) =~ /(.*-)(.*)/) {
+    $cross_suffix .= $1;
 }
 
 $compiler = "";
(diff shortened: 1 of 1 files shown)


yesimon:
Thanks for the fix!

tkelman:
This doesn't work correctly if `CC` includes any flags, e.g. `x86_64-w64-mingw32-gcc -march=x86-64 -m64`, results in `CROSS_SUFFIX` getting set to `x86_64-w64-mingw32-gcc -march=x86-64 -` which doesn't work.

buffer51:
@tkelman Hm, tricky combo. But at the same time, `CROSS_SUFFIX` is here to extract the common part to call other executables too (`ar`, ...).

Since the flags apply only to `gcc`, I think the proper way to go would be to pass those as compilation flags. @xianyi can probably point you to the prefered way to do that.

tkelman:
Including them in `CC` has worked for years, this is a regression.

buffer51:
@tkelman Can you try buffer51/OpenBLAS@157ee49? If it doesn't help, can you describe your host environment and the `make` command you use?

tkelman:
This was cross-compiling via mingw-w64 in cygwin (`make -C deps install-openblas` from julia), only one of i686 or x86_64 was causing an issue because somewhere in this or another script it was not treating one of them as cross compiling. Will try to test your proposed change when I get a chance.

tkelman:
Okay, the problem was cross-compiling from 64 bit cygwin to x86_64 mingw-w64, since compiling from 64 bit x86_64 to 32 bit i686 is special-cased and not flagged as cross-compilation in this script.

Anyway, does look like your change in https://github.com/buffer51/OpenBLAS/commit/157ee49 works without needing the workaround I applied on the Julia side.

PR #897, opened 2016-05-31
STRSM optimized for MSA

ksraste:
STRSM kernels further optimized for MSA
Removed USE_GEMM3M for mips target


ksraste:
Hi Xianyi,

I have submitted PR for TRSM. Please review and let me know any issue.

Thanks,
Kaustubh

Issue #718, opened 2015-12-14
catastrophic error: unknown register name %1 in asm statement 


akakcolin:
I got an error :

../kernel/x86_64/cscal_microk_haswell-2.c(233): catastrophic error: unknown register name %1 in asm statement 
 
compilation aborted for ../kernel/x86_64/cscal.c (code 4) 
make[1]: *** [cscal_k.o] Error 4 
../kernel/x86_64/zscal_microk_haswell-2.c(309): catastrophic error: unknown register name %1 in asm statement 
 
compilation aborted for ../kernel/x86_64/zscal.c (code 4) 
make[1]: *** [zscal_k.o] Error 4 
make: *** [libs] Error 1

brada4:
Which linux, which compiler? Compiler flags?

martin-frbg:
Could be that your version of binutils (assembler and related files) is too old

akakcolin:
The linux system is  Centos
>Distributor ID:	CentOS
>Description:	CentOS Linux release 7.1.1503 (Core) 
>Release:	7.1.1503
>Codename:	Core
Linux hsw-01 3.10.0-229.20.1.el7.x86_64

The Package binutils-2.23.52.0.1-55.el7.x86_64  is already installed and is the latest version

Using gcc/gfortran can build  successully.
Using icc/ifort   cannot build and  with the same error 
 
using "make BINARY=64 CC=icc FC=ifort"

martin-frbg:
Hmm. #493 was solved by updating to binutils 2.25 ? (The error message there was different -
but the only reference to your "unknown register name %<number>" I can find was relevant for an icc version from ten years ago that did not yet recognize the gcc-style register naming.).
If you are sharing the same source tree between different machines or different compilers, be sure to do a "make clean" before the "make"

brada4:
It means ICC does not support AVX2 asm in OpenBLAS. You can link gcc-built libraries using icc.

xianyi:
@akakcolin ， could you try gcc? `make CC=gcc FC=ifort`

akakcolin:
@xiaoyi, thx

OpenBLAS build complete. (BLAS CBLAS LAPACK LAPACKE)

  OS               ... Linux             
  Architecture     ... x86_64               
  BINARY           ... 64bit                 
  C compiler       ... GCC  (command line : gcc)
  Fortran compiler ... INTEL  (command line : ifort)
[...]

PR #915, opened 2016-06-24
remove input from clobbered list

mdong:
Hi,

My ASM is bit rusty, please let me know if i am wrong. ^_^ 

Code below would not work for ICC, it would work for GCC or PGI. I assume both compiler do not do much of checking on ASM code.

```
__asm__ __volatile__
(  mov
   :
   : "r" (XXX), ....
   : "cc", "%0", "%1", "%2",
```

In GCC manual, I see following explanation:

> You may not write a clobber description in a way that overlaps with an input or output operand. For example, you may not have an operand describing a register class with one member if you mention that register in the clobber list. Variables declared to live in specific registers (see Explicit Reg Vars), and used as asm input or output operands must have no part mentioned in the clobber description. There is no way for you to specify that an input operand is modified without also specifying it as an output operand. Note that if all the output operands you specify are for this purpose (and hence unused), you then also need to specify volatile for the asm construct, as described below, to prevent GCC from deleting the asm statement as unused.

As far as i understand, in this case since we have `__asm__ __volatile__`. This code will not be removed as long as it is reachable. Since memory is in the clobber registers, the code would not be optimized away. So remove `"%0", "%1", "%2",` should not cause any issue, and it would allow ICC to build OpenBlas. Please correct me if i am mistaken.

Build with flags:

```
DYNAMIC_ARCH=1 USE_THREAD=1 USE_OPENMP=0 NUM_THREADS=128 NO_LAPACK=1 INTERFACE64=1 BINARY=64
```

```
$ icc -V
Intel(R) C Intel(R) 64 Compiler for applications running on Intel(R) 64, Version 16.0.3.210 Build 20160415
Copyright (C) 1985-2016 Intel Corporation.  All rights reserved.
```

Error messages:

```
catastrophic error: unknown register name %2 in asm statement
```

Possible related to issue #718 

Best,
Min Dong

### DIFF
--- a/kernel/x86_64/cscal_microk_bulldozer-2.c
+++ b/kernel/x86_64/cscal_microk_bulldozer-2.c
@@ -120,7 +120,7 @@ static void cscal_kernel_16( BLASLONG n, FLOAT *alpha, FLOAT *x)
 	  "r" (n),  	// 0
           "r" (x),      // 1
           "r" (alpha)   // 2
-	: "cc", "%0", "%1",
+	: "cc", //"%0", "%1",
 	  "%xmm0", "%xmm1", "%xmm2", "%xmm3", 
 	  "%xmm4", "%xmm5", "%xmm6", "%xmm7", 
 	  "%xmm8", "%xmm9", "%xmm10", "%xmm11", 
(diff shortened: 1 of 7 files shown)


mdong:
By the way, with PGI compiler following message will be generated:

```
PGC-W-0351-Unsupported clobber register '%1' ignored in extended ASM 
```

martin-frbg:
Finally merging despite my sore lack of assembler experience as all documentation I can find supports this, and (as expected) it demonstrably fixes the icc compile problem (also encountered in #962 and in investigation of  #1069) without any ill effect. (I note also that the equivalent code for sandybridge does not attempt to put its input operands on the clobber list...)
Thank you very much for this and the detailed explanation.

PR #916, opened 2016-06-24
change old GNU keyword __real__ __imag__ to CREAL, CIMAG macro

mdong:
Hi,

Most file use the macro defined in `common.h`, but i guess `cdot.c` and `zdot.c` missed the memo ^_^ .

Best,
Min Dong


martin-frbg:
Just for the record I do not think now that this will work as is across all platforms, as in some cases the CREAL will expand to a macro or function rather than the keyword - which is illegal on the lefthand side.

martin-frbg:
Thank you for highlighting this problem (and eventually leading me in the right direction, I hope) all the same. It's a pity that your contribution apparently was not looked at for so long.

mdong:
Hi, sorry if the deletion seems to be odd. I just come to understand that
my understanding was not complete, thanks for your explanation.

PR #941, opened 2016-08-08
MIPS n32 ABI support, MSA support detection and rename ARCH, ARCHFLAGS

sva-img:
MIPS n32 ABI support, MSA support detection and rename ARCH, ARCHFLAGS

Signed-off-by: Shivraj Patil <shivraj.patil@imgtec.com>


xianyi:
Hi @sva-img ，

Why do you need to  rename ARCH, ARCHFLAGS?

sva-img:
Hi xianyi,

Existing ARCH and ARCHFLAGS is confusing whether it means architecture flags or archiever flags. 
I think, renaming ARCH to AR and ARCHFLAGS to ARFLAGS makes it more clear that these are archiever only flags.

If this looks good, I will resubmit this change in new PR.

kernel/Makefile
```diff
@@ -4,7 +4,6 @@ endif
 
 TOPDIR	= ..
 include $(TOPDIR)/Makefile.system
-include $(TOPDIR)/Makefile.conf
```
xianyi: @sva-img , why do you delete Makefile.conf?

xianyi:
@sva-img ,  `ARCH` and `ARCHFLAGS` flags come from upstream LAPACK implementation. Thus,   I suggest keep the flags with upstream.

kernel/Makefile
sva-img: @xianyi, This is reverted to original code which was added in previous commit.

sva-img:
Ok. I have reverted rename change to original in last commit.

sva-img:
@xianyi, Please have a look of combined changes from 'all commits' for better clarity.

PR #944, opened 2016-08-17
For gemm multi-threading, simply split M.

subwood:
e.g.
layer 1: A (1600k, 576), B(576, 64)

B is very small. We split M.


martin-frbg:
While it may be the case that B is very small for typical workloads in deep learning, I suspect this unconditional change will not be appropriate for general application in scientific or engineering code bases that rely on BLAS/LAPACK (which would explain why it lives on that optimized_for_deeplearning branch)

xianyi:
@subwood @martin-frbg ,

Actually, we are developing a version,  which is optimized for deep learning workload.

PR #959, opened 2016-09-11
Remove unnecessary const from cblas.h

Rufflewind:
In standard C, top-level `const` in function declaration parameters have no effect on the API at all, nor does it influence the actual implementation.  They can be safely removed to avoid unnecessary clutter in the type signatures.

### DIFF
--- a/cblas.h
+++ b/cblas.h
@@ -52,309 +52,309 @@ typedef enum CBLAS_UPLO      {CblasUpper=121, CblasLower=122} CBLAS_UPLO;
 typedef enum CBLAS_DIAG      {CblasNonUnit=131, CblasUnit=132} CBLAS_DIAG;
 typedef enum CBLAS_SIDE      {CblasLeft=141, CblasRight=142} CBLAS_SIDE;
 
-float  cblas_sdsdot(OPENBLAS_CONST blasint n, OPENBLAS_CONST float alpha, OPENBLAS_CONST float *x, OPENBLAS_CONST blasint incx, OPENBLAS_CONST float *y, OPENBLAS_CONST blasint incy);
-double cblas_dsdot (OPENBLAS_CONST blasint n, OPENBLAS_CONST float *x, OPENBLAS_CONST blasint incx, OPENBLAS_CONST float *y, OPENBLAS_CONST blasint incy);
-float  cblas_sdot(OPENBLAS_CONST blasint n, OPENBLAS_CONST float  *x, OPENBLAS_CONST blasint incx, OPENBLAS_CONST float  *y, OPENBLAS_CONST blasint incy);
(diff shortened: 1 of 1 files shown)


xianyi:
@Rufflewind , thank you for the PR.

I want to keep `const` to maintain `cblas.h` as same as MKL header file.

Rufflewind:
Stripping off the top-level `const` does not change the semantics of the code in any way.  The difference from the MKL header file would be completely superficial.

hfp:
> In standard C, top-level const in function declaration parameters have no effect on the API

It is good practice to be const-correct in C as well. Further according to ANSI, two function arguments may not alias if one of which is const-qualified. The latter can make a difference in generated code (which I doubt for OpenBLAS to be relevant due to not relying too much on compiler's optimizations).

martin-frbg:
Same `const` declarations are used in the cblas.h of the reference implementation ( https://github.com/Reference-LAPACK/lapack/blob/master/CBLAS/include/cblas.h ), so perhaps any (proposed) changes should start there ? (Also, the current mkl_cblas.h still appears to have these `const int` as well ?)

Rufflewind:
> It is good practice to be const-correct in C as well.

This only matters in the *implementation* code. The C standard permits the header and the implementation to differ in top-level `const`-ness:

```c
// foo.h
void foo(int);

// foo.c
void foo(const int x) {
  ...
}
```

See WG14/N1256 §6.7.5.3.15 [emphasis added]:

> (In the determination of type compatibility and of a composite type, each parameter declared with function or array type is taken as having the adjusted type and **each parameter declared with qualified type is taken as having the unqualified version of its declared type**.)

Further elaboration can be found at https://abseil.io/tips/109.

Therefore, the presence of top-level `const` in the parameters of declarations **is only a stylistic preference**. It is discouraged by most style guides (e.g. https://clang.llvm.org/extra/clang-tidy/checks/readability-avoid-const-params-in-decls.html) because the `const` is not meaningful to the caller and can cause needless confusion as it did in this thread. However, I won't press the issue if OpenBLAS prefers to stick to the prevailing style in the BLAS standard.

martin-frbg:
Thank you for the additional references - still I (at least) would prefer to consider the reference implementation as normative, unless there is a demonstrable practical advantage (rather than "just" stylistic purism) in diverging from it.

hfp:
I was wrong, I thought this discussion was about `const double*` vs. `double*`. Indeed, `const double` vs `double` (or other types) as part of a function's signature can be considered superfluous ("blown interface"). It also does not introduce a separate overload in C++, which is a strong argument against such constness. It simply "protects" the implementation of a function from accidentally using ("abusing") a function argument as local variable. Though, the call-side of a such "blown interface" is protected anyway since the arguments in question are passed by value.

Issue #957, opened 2016-09-06
CMake INSTALL project?


PrimarchOfTheSpaceWolves:
Hi all,

I'm using Windows 7 with Visual Studio 2013, and I use CMake a fair amount.  It would be nice if the current CMakeLists.txt file also generated the appropriate INSTALL project.

I've been trying to roll my own with some success; any suggestions would be welcome:

FILE(GLOB INCLUDE_FILES "${CMAKE_CURRENT_SOURCE_DIR}/*.h")
install (TARGETS ${OpenBLAS_LIBNAME} DESTINATION lib)
install (FILES ${INCLUDE_FILES} DESTINATION include)

PrimarchOfTheSpaceWolves:
Followup: this is my patch file so far
[CMakeLists-patch.txt](https://github.com/xianyi/OpenBLAS/files/462355/CMakeLists-patch.txt)

BarrySmith:
Could you turn this into a pull request? It's crazy the cmake stuff doesn't all ready have an install, but I am too incompetent at cmake (and spelling) to change the cmake files myself.

martin-frbg:
While updating the pull request I built from your patch, I noticed that your proposal appears to be adding the _d postfix twice - once by redefining the LIBNAME, and a second time by attaching DEBUG_POSTFIX ?

martin-frbg:
Now merged from PR #988 - please double-check if you can spare the time.

PR #988, opened 2016-10-19
Add CMAKE install target

martin-frbg:
Add CMAKE install target (copied from a patch provided by PrimarchOfTheSpaceWolves in #957)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,9 +31,10 @@ set(NO_LAPACKE 1)
 endif()
 
 if(BUILD_DEBUG)
-set(CMAKE_BUILD_TYPE Debug)
+	set(CMAKE_BUILD_TYPE Debug)
+	set(OpenBLAS_LIBNAME "${OpenBLAS_LIBNAME}_d")
 else()
-set(CMAKE_BUILD_TYPE Release)
+	set(CMAKE_BUILD_TYPE Release)
 endif()
 
 if(BUILD_WITHOUT_CBLAS)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -198,3 +205,22 @@ set_target_properties(${OpenBLAS_LIBNAME} PROPERTIES
...
+FILE(GLOB_RECURSE INCLUDE_FILES "${CMAKE_CURRENT_SOURCE_DIR}/*.h")
+install (FILES ${INCLUDE_FILES} DESTINATION include)
+ 
+if(NOT MSVC)
+	set(DEBUG_POSTFIX "")
+	if(BUILD_DEBUG)
+		set(DEBUG_POSTFIX "_d")
+	endif()
+	install (TARGETS ${OpenBLAS_LIBNAME}_static${DEBUG_POSTFIX} DESTINATION lib)
+endif()
```
xantares: DEBUG_POSTFIX is already defined earlier, the install could be moved there to avoid duplicate code

CMakeLists.txt
```diff
@@ -31,9 +31,10 @@ set(NO_LAPACKE 1)
 endif()
 
 if(BUILD_DEBUG)
-set(CMAKE_BUILD_TYPE Debug)
+	set(CMAKE_BUILD_TYPE Debug)
+	set(OpenBLAS_LIBNAME "${OpenBLAS_LIBNAME}_d")
 else()
-set(CMAKE_BUILD_TYPE Release)
+	set(CMAKE_BUILD_TYPE Release)
 endif()
```
martin-frbg: That is xianyi's code from 7ac7e147d, the patch only changed the indentation here.
I take it this evaluation of DEBUG_BUILD should be removed completely ?

martin-frbg: More to the point, should this block become something like
if( NOT CMAKE_BUILD_TYPE )
  set( CMAKE_BUILD_TYPE Release)
endif()
if (CMAKE_BUILD_TYPE = Debug)
 set (DEBUG_POSTFIX "_d")
 set(OpenBLAS_LIBNAME "${OpenBLAS_LIBNAME}_d")
else()
 set DEBUG_POSTFIX ""
endif()
which would also consolidate the definitions of DEBUG_POSTFIX as in your other comment ?

martin-frbg:
I have reworked the PR now to query BUILD_DEBUG only if CMAKE_BUILD_TYPE is not set, and moved the conditional appending of a debug prefix to use the cmake-provided setting.
(Apologies again for the mess of incremental commits to the same file, I was working via the web interface and do not see a means of squashing the commit history there)

Issue #971, opened 2016-09-30
Update LAPACKE to latest master


vladimir-ch:
Would it be possible to update (at least) lapack-netlib/LAPACKE to the latest master? It contains a fix for the bug https://github.com/Reference-LAPACK/lapack/issues/59 that is causing failures in our tests https://github.com/gonum/lapack/pull/143.

martin-frbg:
Probably quicker to do a pull request with just the fix for lapack#59. (As far as I know, OpenBLAS tends to synchronize to major releases of LAPACK and just cherry-pick patches in between. At least in the past when the "netlib" LAPACK was not yet on github and there was only their user forum to go to, lapack bugs were sometimes fixed here first.)

PR #996, opened 2016-10-26
Lapack 3.6.1

grisuthedragon:
The pull request will import LAPACK 3.6.1. into OpenBLAS. 

4 functions are added by the LAPACK team and many bugs are fixed.


martin-frbg:
While I do not feel competent to review, I am a bit worried how this meshes with previous individual changes to the OpenBLAS lapack-netlib subtree made since 3.6.0 - as a trivial but easy to check example your PR would appear to revert 8d652f1 ? (This is why I suggested to submit only the minimal necessary PR for #971 to get that accepted quickly)

grisuthedragon:
I changed the _nep.in_ file in order to fit the one created by 8d6552f1. Furthermore, I imported directory by directory an checked the makefiles if there are OpenBLAS related things in for examples the special tuned LAPACK routines like dgetrf provided by LAPACK. So the "main" makefile of LAPACK (lapack-netlib/SRC/makefile) only differs in the addition of the 4 new routines that exist in LAPACK 3.6.1.

I hope that helps.

grisuthedragon:
Finally, I checked all changes that were done on LAPACK 3.6.0 since the import into OpenBLAS and applied them, if necessary.  Issue #971 is also fixed because this is not inlcuded in LAPACK 3.6.1.

xianyi:
Thank you, I will review this patch these days.

vladimir-ch:
Based on a comment in #1019 is there anything I can do to speed the review of this PR?

grisuthedragon:
I will check #1019 for integration in my pull request tomorrow.

martin-frbg:
@grisuthedragon I believe #1019 is already part of your PR as the fix is tagged 3.6.1 in the netlib repo.
So the answer to "anything I could do" would seem to be "clone xianyi" unfortunately. (And I still believe the point fix has a greater chance to be approved and committed on the fly than the whole of 3.6.1)

grisuthedragon:
I checked for it and yes it is already in

xianyi:
Looks good for me. I merge this patch at first. Maybe we can test more in future.

PR #1019, opened 2016-11-23
LAPACKE: fix xlascl family of functions

vladimir-ch:
Applies fixes made by upstream in
https://github.com/Reference-LAPACK/lapack/commit/5eca362f3a654c74e1743fea0c9bbc9bb3d7370f

Our tests in https://github.com/gonum/lapack/pull/206 are failing because of this


brada4:
It is a sub-set of #996, probably it is best to review that and say that that one would fix your test cases too.

xianyi:
@vladimir-ch , I just merged LAPACK 3.6.1. Does it already include this patch?

grisuthedragon:
@xianyi The patch is already included in lapack 3.6.1 so this PR is already done by your merge of LAPACK 3.6.1

PR #1034, opened 2016-12-16
fix error: non-void function 'get_feature' should return a value [-Wreturn-type]

liungkejin:


martin-frbg:
This issue is already addressed (in the same way obviously) by the already open PR #1025

liungkejin:
Okay

Issue #923, opened 2016-07-13
caffe + openblas not using cpu to the max


fenrus75:
When using the combination of caffe + openblas, it turns out that all cpus cores except cpu 0 are only used about 50% (AlexNet benchmark), the rest is idle time.

(this is sort of hidden if you're not careful due to the polling nature of the default openblas config, but it shows up if you fix that, or if you use the openmp setup. In the default polling setup you can see it by how many cycles are spent in the polling loop)

profile below:

  63.28%  libopenblas_haswellp-r0.2.16.so  [.] sgemm_kernel
   8.78%  libopenblas_haswellp-r0.2.16.so  [.] sgemm_incopy
   6.10%  libcaffe.so.1.0.0-rc3            [.] caffe::im2col_cpu<float>
   4.55%  libopenblas_haswellp-r0.2.16.so  [.] sgemm_itcopy
   2.59%  libopenblas_haswellp-r0.2.16.so  [.] sgemm_oncopy
   1.32%  libcaffe.so.1.0.0-rc3            [.] caffe::PoolingLayer<float>::Forward_cpu
   1.27%  libgomp.so.1.0.0                 [.] gomp_barrier_wait_end
   1.22%  libm-2.23.so                     [.] __powf

(and then a tail of very small things)

I'm kind of lost in terms of how to diagnose this further; increasing the number of worker threads does nothing; each thread does proportionally less work but the total cpu use does not change significant. 

It feels (subjectively) that something is either not feeding the worker threads enough, or that there is a sequential step going on that is not threaded but that takes about 50% of wallclock time.

Is this a known behavior? Any suggestions on how I can diagnose this further?

martin-frbg:
What is your hardware, and many threads are you letting OpenBLAS create ? If this is on a CPU that sports hyperthreading, using more threads than your number of actual full-featured cores will only hold everything up.

fenrus75:
This is a normal Broadwell system; 2 cores (so 4 threads)

(I'll not agree with your characterization of hyperthreading; the cpu can do 2 vector FMA's per cycle... that's slightly hard to feed if you don't do hyperthreading)

fenrus75:
even with running just 2 threads, the 2nd core is not more than 63% utilized.

brada4:
Somewhere in alexnet description:
```
shape {
      dim: 128
      dim: 3
      dim: 224
      dim: 224
    }
```
Even if you multiply all data sample is quite small...

[...]

Issue #900, opened 2016-05-31
performance in sched_yield (multisocket)


phirus2:
Hi!!!
I'm using mumps with openblas 0.2.18 compiled from sources with this options:
- gcc 5.3.0
- USE_THREAD = 1
- NUM_THREADS = 128
- NO_WARMUP = 1
- #NO_AFFINITY = 1
- #BIGNUMA = 1
-  MAX_STACK_ALLOC = 8128

I'm running my code in some machines multisocket (2 or 4) with 24 until 128 cores

The problem is that most of the time is executed by the system. To be more acurate is doing not I/O is doing Context switch. You can see it in the below pictures
![spin_time](https://cloud.githubusercontent.com/assets/19668974/15677414/788d31e2-274a-11e6-9da3-493261b30eff.png)
![sched_time](https://cloud.githubusercontent.com/assets/19668974/15677415/78e247f4-274a-11e6-898a-f90372069c3a.png)

This doesn't happend in 1 socket machine, but if I force to use just one CPU (taskset -c 0-31 MyAPP) the performance is also poor.

What can I do to give you more information and try to help¿¿¿

brada4:
How big is data sample? What CPU? Maybe all the time is spent ripping tiny data sample between CPU caches or NUMA nodes?

brada4:
sched_yield is that descheduling your flawed automated debugger is recommending.
man sched_yield ?

jeromerobert:
@phirus2 What is are the size of the GEMM ? Could you get a sample of them with adding a trace in `interface/gemm.c` or with a debugger ?

MUMPS is a sparse solver so he is probably working on small blocks on which you can't get any benefits from the OpenBLAS multi-threading. Anyway if blocks are small we should not even enter in `blas_thread_server` so that's odd.

brada4:
There is no 32-core intel CPU being manufactured
Results are fake at best. There is no issue.

phirus2:
```
hwloc-info
depth 0:        1 Machine (type #1)
 depth 1:       4 NUMANode (type #2)
  depth 2:      4 Socket (type #3)
   depth 3:     4 L3Cache (type #4)
    depth 4:    64 L2Cache (type #4)
     depth 5:   64 L1dCache (type #4)
      depth 6:  64 L1iCache (type #4)
       depth 7: 64 Core (type #5)
        depth 8:        128 PU (type #6)
```

```
[...]

PR #1051, opened 2017-01-08
Replace sched_yield with usleep(1) on Linux

martin-frbg:
to avoid the massive overhead of the sched_yield call on Linux kernels since its semantics were changed in early 2003 (late 2.5 series) to include reordering of the thread queue. Ref. #900,#923

### DIFF
--- a/common.h
+++ b/common.h
@@ -313,6 +313,10 @@ typedef int blasint;
 #if defined(OS_SUNOS)
 #define YIELDING	thr_yield()
 #endif
+	
+#if defined(OS_LINUX)
+#define YIELDING	usleep(1)
+#endif
 
 #if defined(OS_WINDOWS)
 #if defined(_MSC_VER) && !defined(__clang__)
--- a/common.h
+++ b/common.h
@@ -315,7 +315,7 @@ typedef int blasint;
 #endif
 	
 #if defined(OS_LINUX)
-#define YIELDING	usleep(1)
+#define YIELDING	__asm__ __volatile__ ("nop;nop;nop;nop;nop;nop;nop;nop;\n");
 #endif
 
 #if defined(OS_WINDOWS)


xianyi:
@martin-frbg Do you measure the overhead of sched_yield or usleep(1) ?

martin-frbg:
Unfortunately I do not have access to any serious HPC system (and I have to agree the two issue threads I refered to are a bit confusing, with attempts to modify the GEMM thresholds intermixed with the changes to YIELDING ). On my lowly dual-core Kaby Lake laptop I do see a significant increase in throughput with the (improved) deig.R benchmark (2k x 2k going from 8000Mflops/28sec to 24000/9sec) though most of that will probably be due to avoiding thermal throttling.
Note also that wernsaar already replaced sched_yield with asm(nop) for a number of AMD cpus in the past while I do not think the underlying issue is in cpu hardware. Should I open a separate issue to discuss ?

brada4:
@xianyi just start idle CPU hog e.g.:
`nice yes > /dev/null &`
Then see how much spare time you get running same sample with different YIELDING options.
It demonstrates more CPU to competing users, or less heat for nothing in absence of such...
I don't get significant boost in wall time spent, but system CPU time consumption drops noticeably, somebody with watt-meter could be better positioned to measure.

martin-frbg:
I'll see if I can retask at least a 4-core Haswell to benchmarking tomorrow.

martin-frbg:
First results (still with deig.R, on a quadcore Haswell with turboboost disabled to ensure repeatability) show a slight _slowdown_ rather than the 10 percent speedup reported for fenrus75' test case. The change however does allow the cores to drop to "idle" intermittently, where with sched_yield the core allocation was split between userspace and system time - this is probably what helps thermal management and would make the calculation run faster with dynamic overclocking enabled.

brada4:
for me YIELDING {} gave best result with 0.1um speedup

martin-frbg:
Interestingly an empty YIELDING is what was used in early versions of libGoto2 up to at least 1.08, while 1.13 had sched_yield.

martin-frbg:
Some preliminary data (more to come next weekend if time and workload permits). deig.R benchmark
modified for matrix size 10240x10240, four threads on otherwise idle quadcore Haswell 3.4GHz w/o turboboost, YIELDING defined as:
sched_yield 62626.0 MFlops 457.1 sec
usleep(1)    56996.8 MFlops 502.2 sec
nothing       62442.8 MFlops 458.4 sec
8xnop         62759.1 MFlops 456.1 sec
So usleep(1) appears to be significantly slower, the other alternatives are basically on par in terms of speed but as mentioned above sched_yield keeps the cores from going idle. Results for 2 and 8 threads are similar to this (though I only have them for small matrix sizes up to 2048x2048)

brada4:
you can cat() whole z: z[1] would be user time (that makes flops) and z[2] system time (that generates heat)
2k*2k*8B=32MB, probably dont try at NUMA...

brada4:
From sched_yield(2) manual page:
       Strategic  calls to sched_yield() can improve performance by giving other threads or processes a chance to run when (heavily) contended resources (e.g., mutexes) have been released by the caller.  **Avoid calling sched_yield() unnecessarily or inappropriately**
       (e.g., when resources needed by other schedulable threads are still held by the caller), since doing so will result in unnecessary context switches, which will degrade system performance.

martin-frbg:
Yeah we can all read manpages - my problem is when would a call to sched_yield clearly be considered unnecessary, inappropriate and indecent, and why would K.Goto and all who came after him decide this was not the case here ?

martin-frbg:
PR updated to current favorite just to prevent accidental merging of known bad idea.

martin-frbg:
Did not get around to much more yet, so no complete picture. Quick comparison of deig.R run with 4 threads and 10240x10240 matrix, i7-4770 with and without turboboost enabled:
sched_yield, fixed freq  62625 MFlops 457 sec   turboboost 64515 MFlops 443 sec
asm(8xnop), fixed freq  62759 MFlops 456 sec   turboboost 65407 MFlops 437 sec

brada4:
Seems that slack turns up for small samples more without impacting overall time.

jeffhammond:
Just remember that `sched_yield` has a significant upside over `nop` when oversubscribing.  It would be a lot better to allow users to select the more forgiving option if they intend to run in a multi-tenant or desktop environment.

martin-frbg:
Thanks for that comment - indeed I am still looking for reasons why sched_yield was preferred so far. If I am not mistaken, a default build of OpenBLAS will have a built-in thread limit of two times the number of kernels on x86 at least and so far the behaviour of a "lightly" overloaded system without sched_yield did not appear to be worse. (And I did see a clear improvement in terms of thermal management with my - apparently poorly designed - early Kaby Lake laptop). So perhaps it would be sufficient to provide a choice of implementations for "YIELDING" in Makefile.rule or couple it to the NUM_THREADS value in relation to physical cores detected ? (Lastly there is always the option to limit the number of threads to some managable quantity for the given system at runtime)

martin-frbg:
Closing this for now as the results so far are a bit inconclusive and I need to get rid of my current fork as doing subsequent PRs from it seems to have led to an unintended revert of #988

PR #1061, opened 2017-01-11
Add new targets for ARM64

ashwinyes:


martin-frbg:
Seems the travis check failed only because the vm could not fetch compiler packages from ubuntu.com

ashwinyes:
May I know what all is tested in the travis check?

xianyi:
@ashwinyes ，It doesn't test ARM target on Travic-CI.

ashwinyes:
Ok. Thanks. In that case these should not cause any failures.

Moreover it is adding new targets for ARM64. So it should not affect any existing ARM64 targets

ashwinyes:
Reopened the pull request to trigger the travis checks again.

ashwinyes:
@xianyi The Travis checks have completely successfully. Request you to merge if the changes looks  OK.

Issue #1077, opened 2017-02-02
DGESVD slow coparing with intel implementation


OtacilioNeto:
Dears, when I run this command on a scilab running on freebsd compiled with openblas I get a time of  37 seconds
A1=rand(1000, 1000); tic();[S, v, D]=svd(A1);toc()
 ans  =
 
    37.276 

The same command on windows with intel library I get  1 second

A1=rand(1000, 1000); tic();[S, v, D]=svd(A1);toc()
 ans  =
 
    1.382  

Could you please give me a hint why is so different?

martin-frbg:
One trivial factor is probably that the OpenBLAS build system did not pass the "-O2" compiler option to the fortran compiler until very recently - as #843 showed this has a noticable impact on calculations at least compared to netlib LAPACK (practically identical code but using -O2 optimization level for the default build). You need to either correct the "override FFLAGS" line in Makefile.system or set the FFLAGS  environment variable accordingly before invoking "make".
Less trivial could be that you are running OpenBLAS multithreaded and with too many threads to be efficient while MKL may be smart enough to use only one or two threads in this case. In that case running with OPENBLAS_NUM_THREADS=1 or 2 - or even building OpenBLAS without thread support may improve the timing.

brada4:
Are you with OpenBLAS?
```
R> x<-matrix(rnorm(1e6),1e3,1e3)
R> system.time(r<-svd(x))
   user  system elapsed 
  2.516   0.824   1.032 
```

brada4:
You are not with OpenBLAS by default
https://svnweb.freebsd.org/ports/head/math/scilab/Makefile?view=markup#l48

You have to rebuild scilab to use OpenBLAS, whose port in turn is ""good enough"" i.e likely half-speed of mkl without what @martin-frbg offered.
https://svnweb.freebsd.org/ports/head/math/openblas/Makefile?view=markup#l43

OtacilioNeto:
I did a rebuild before open this issue. This is my system status:

[ota@nostromo /usr/ports/math/openblas]$ ldd /usr/local/bin/scilab-bin  | grep blas
	libopenblasp.so.0 => /usr/local/lib/libopenblasp.so.0 (0x808a00000)

[ota@nostromo /usr/ports/math/openblas]$ pkg info scilab 
scilab-5.5.2_4
Name           : scilab
Version        : 5.5.2_4
Installed on   : Thu Feb  2 15:28:49 2017 BRT
[...]

PR #1079, opened 2017-02-03
dgesvd_ bechmark

brada4:
from #1077 
Shows some surplus threads at small samples:
_NUM_THREADS=1
```
           128x128 :    1554.22 MFlops   0.009000 sec
           256x256 :    3197.26 MFlops   0.035000 sec
```
_NUM_THREADS=4
```
           128x128 :    1076.00 MFlops   0.013000 sec
           256x256 :    2869.33 MFlops   0.039000 sec
```

### DIFF
--- a/benchmark/scripts/R/dgesvd.R
+++ b/benchmark/scripts/R/dgesvd.R
@@ -1,0 +1,55 @@
+#!/usr/bin/Rscript
+
+argv <- commandArgs(trailingOnly = TRUE)
+
+nfrom <- 128
+nto <- 2048
+nstep <- 128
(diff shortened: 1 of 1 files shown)


grisuthedragon:
Is it not better to have a C benchmark for this. Furthermore, the flopcount/rate for the SVD is a bit dangerous because although it works like a direct method, it involves an iterative process. So depending on the "quality" of the random input data this values may differ on the same machine.

brada4:
You are right. I better make one script that drills through lapack calls that R core wraps closely in its simplest functions.

PR #1085, opened 2017-02-07
LAPACKE: fix incorrect value of lda_t in lapacke_?laswp_work

vladimir-ch:
This is a small subset (https://github.com/Reference-LAPACK/lapack/pull/118/commits/07e1fbd89752bed74d35c48e92287d467646a158) of changes committed recently to the Reference LAPACK (https://github.com/Reference-LAPACK/lapack/pull/118). Since I'm not sure whether there is any policy on when and which changes are merged from upstream to OpenBLAS, I'm submitting only changes that prevent our tests (https://github.com/gonum/lapack/pull/219) to pass.


martin-frbg:
I am not the one to set policy, but I would expect that bugfixes should be merged as soon as possible. In that sense most of Reference-LAPACK/lapack#118 (excepting just the purely cosmetic changes) looks important enough to me. Thanks for flagging this.

PR #1091, opened 2017-02-10
CPUID mappings for Core i5-7600K (Kaby Lake)

staticfloat:
On my new computer, I found that OpenBLAS didn't recognize the CPU.  This was because it was getting an `exfamily` of 9, instead of the `8` that mapped to Kaby Lake in the existing OpenBLAS source.  This patch fixes that and adds in a few new CPUID mappings as best I could from the sandpile table.  Please verify that I interpreted the sandpile webpage correctly, there was some ambiguity with regards to huge pages and I just did my best to guess and choose the most reasonable choices.  With this patch, the tests all pass on my local machine.

### DIFF
--- a/cpuid_x86.c
+++ b/cpuid_x86.c
@@ -636,6 +636,13 @@ int get_cacheinfo(int type, cache_info_t *cacheinfo){
 	LD1.associative = 8;
 	LD1.linesize    = 64;
 	break;
+      case 0x63 :
+  DTB.size        = 2048;
+  DTB.associative = 4;
+  DTB.linesize    = 32;
+  LDTB.size       = 4096;
+  LDTB.associative= 4;
+  LDTB.linesize   = 32;
       case 0x66 :
 	LD1.size        = 8;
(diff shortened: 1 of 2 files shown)


martin-frbg:
Honestly I cannot comment on the validity of the cacheinfo data - I never saw this updated nor was asked to provide updates when I or others added cpuids. What will be needed though is a similar update of the code in driver/others/dynamic.c that does the runtime cpu type detection for the DYNAMIC_ARCH builds. (Exfamily 8 btw is what the first wave of low-power dualcore KabyLakes for notebook identifies as)

staticfloat:
Ah, yes.  Thank you.  I've added that as well now, I'm verifying this works by calling `openblas_get_config()`.  Here's the output, (invoked from Julia) in case you're interested:

```
julia> unsafe_string(ccall((:openblas_get_config64_, Base.libblas_name), Ptr{UInt8}, ()))
"USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell"
```

PR #1102, opened 2017-02-20
Correct Apollo Lake CPUID identification in dynamic_arch builds

brada4:
Detect apollo lake like in cpuid_x86, otherwise it falls through to default ?prescott?
Also reading through there is dead break; remove that.


martin-frbg:
Thanks for catching this typo. I changed the title to make it more obvious what this commit did when browsing the git log.

PR #1121, opened 2017-03-09
Add `csymv` and `zsymv` into `@lapackobjs2` for exporting

staticfloat:
@martin-frbg is this the proper way to do this?


staticfloat:
I got frustrated by the inconsistent formatting of this file and have pushed a second commit that cleans things up a bit.  If you don't want that, just let me know, and we can drop that from this PR.

martin-frbg:
Thanks, you beat me to it while I was asleep. I think the cleanup is OK too as it is in a separate commit so it should be possible to track the actual code change in the history if and when necessary.

PR #1129, opened 2017-03-14
Address unlikely memleak in zimatcopy interface

brada4:
Temp buffer is leaked in (unlikely) fall-through path, now no more.

### DIFF
--- a/interface/zimatcopy.c
+++ b/interface/zimatcopy.c
@@ -251,7 +251,7 @@ void CNAME( enum CBLAS_ORDER CORDER, enum CBLAS_TRANSPOSE CTRANS, blasint crows,
 		}
 
 	}
-	free(b)
+	free(b);
 	return;
 
 }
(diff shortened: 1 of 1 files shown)


martin-frbg:
What is the motivation for this change, which seems to boil down to the addition of a single "free(b)" for the impossible case that "order" or "trans" have an illegal value ? At least I would like to see all the early returns preserved, if only for human readability.

brada4:
Interface can be called by user directly, 
Correct would be to malloc b before it is used, i.e in 2 valid places, and freed in branches. I am not sure about preserving readability then.

brada4:
ie if trans=grabage, then leak as it stands now.

martin-frbg:
Still would prefer to see just the minimal change made (i.e. addition of the "free(b)" before the return of the fall-through path, no removal of the early returns)

brada4:
Can fix that. no probl

Issue #1137, opened 2017-03-28
COPY memory alignment SIGSEGV bug


apotschka:
Minimal example [check_copy.c](https://gist.github.com/apotschka/bbe4475315019dad80f8a598c9d04ffe), compiled via
```
$ cc -g -o check_copy check_copy.c /usr/lib/libopenblas.so.0
```

Error occurs for `offset == 1`:
```
$ valgrind ./check_copy
==16357== Memcheck, a memory error detector
==16357== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==16357== Using Valgrind-3.11.0 and LibVEX; rerun with -h for copyright info
==16357== Command: ./check_copy
==16357==
1.000000 1.000000
0.500000 0.500000
0.333333 0.333333
==16357==
==16357== Process terminating with default action of signal 11 (SIGSEGV)
==16357==  General Protection Fault
==16357==    at 0x64FB980: dcopy_k_HASWELL (in /usr/lib/libopenblasp-r0.2.18.so)
==16357==    by 0x6EEE82F: (below main) (libc-start.c:291)
```

Changing the increment of `offset` to `offset += 8` makes the code run fine.

System:
```
$ uname -a
Linux *** 4.4.0-67-generic #88-Ubuntu SMP Wed Mar 8 16:34:45 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
$ cc --version
cc (Ubuntu 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609
$ zcat /usr/share/doc/libopenblas-base/changelog.Debian.gz | head -1
openblas (0.2.18-1ubuntu1) xenial; urgency=medium
$ cat /proc/cpuinfo | head -30
processor       : 0
vendor_id       : GenuineIntel
cpu family      : 6
model           : 63
model name      : Intel(R) Core(TM) i7-5820K CPU @ 3.30GHz
stepping        : 2
microcode       : 0x36
cpu MHz         : 1201.921
cache size      : 15360 KB
physical id     : 0
siblings        : 12
core id         : 0
cpu cores       : 6
apicid          : 0
initial apicid  : 0
fpu             : yes
fpu_exception   : yes
cpuid level     : 15
wp              : yes
[...]

PR #1138, opened 2017-03-29
Replace SSE With C in Haswell dcopy

brada4:
Should stop crashes #1137 
Tested on one single broadwell.


martin-frbg:
As noted in #1137, the problem is not specific to Haswell and the C fallback kernel appears to perform poorly compared to other options.

PR #1153, opened 2017-04-13
update s390x port

sharkcz:
I have rebased the z13 branch on top of master and added one missing fix allowing openblas to build on Fedora/s390x.


martin-frbg:
FWIW the z13 branch was merged into develop back in february, I suspect it could be useful to have the "one missing fix" as a separate commit so as to be able to apply it there as well if still necessary ?

sharkcz:
Unless I'm missing something then https://github.com/xianyi/OpenBLAS/pull/1153/commits/d4f9ebac7499f76b71a5f4d91dc19e76137a7013 is still missing even in the develop branch. Will update the PR in a moment.

PR #1156, opened 2017-04-15
Use GNUInstallDirs to allow changing target directories

SoapGentoo:
* Multi-lib distributions need to change the libdir
  which is only portably possible with `GNUInstallDirs`.
* Multi-arch distributions such as Debian and Exherbo
  need to be able to change the bindir.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@
 ## Author: Hank Anderson <hank@statease.com>
 ##
 
-cmake_minimum_required(VERSION 2.8.4)
+cmake_minimum_required(VERSION 2.8.5)
 project(OpenBLAS)
 set(OpenBLAS_MAJOR_VERSION 0)
 set(OpenBLAS_MINOR_VERSION 2)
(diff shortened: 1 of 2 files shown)


martin-frbg:
I have no opinion either way, but wouldn't cmake 2.8.5 be sufficient rather than bumping the minimum requirement to 3.0 (granted both are not brand new) ?

SoapGentoo:
@martin-frbg `GNUInstallDirs` was only added with 3.0. Some packages tend to bundle the upstream cmake file, but imo this is less than optimal, as it will lead to bitrot. Debian includes 3.0 and thus can deal with this build system. I know that this is painful for some users, but not having to hack the build system to finally support changing all paths is a bigger win in my book.

martin-frbg:
https://cmake.org/Wiki/CMake_Version_Compatibility_Matrix/StandardCMakeModulesFindL
has it as a feature of 2.8.5 though it could be that early implementations were flawed ?
I do agree that making it easier to change the default locations would be an advantage, at least as long as the defaults are at least as "sane" as the current ones. (Found reference to one project that dropped it claiming that it changed their "lib" default to something like lib/x86_64-linux-gnu on at least ubuntu 14)

SoapGentoo:
@martin-frbg changed it to 2.8.5 after checking that the module was indeed only integrated with 2.8.5. I have pushed `GNUInstallDir` into a number of projects and haven't seen any issues yet.

brada4:
It is very dangerous to install local build outside /usr/local (see FHS)

SoapGentoo:
@brada4 it is not the job of the build system to prevent users from doing stupid things. The defaults will install in a FHS compatible manner to `/usr/local`.

PR #1178, opened 2017-05-05
MIPS threading fixes

jcowgill:
https://bugs.debian.org/861486

This PR fixes some threading issues on MIPS found while trying to build Julia in Debian.

The first commit implements `MB` and `WMB` on MIPS (using GCC builtins, but hopefully that's fine). The second commit removes the broken `blas_lock` implementations so the code falls back to using GCC builtins which do work correctly on MIPS.


sebastien-villemot:
I have tested the two commits on a mips64 machine and I confirm that they fix the issue.

martin-frbg:
Thanks for the PR, and thanks for testing :-)

PR #1179, opened 2017-05-05
Fixes to driver/others/memory.c

jcowgill:
This PR fixes some issues I noticed in `driver/others/memory.c` while debugging the MIPS threading issues. Both appear to have been programming mistakes.

The first commit moves the `position < NUM_BUFFERS` condition to the front of the `&&` so it actually has an effect (it was being optimized away before).

The second commit ensures the case where `position == NUM_BUFFERS` is handled correctly in `blas_memory_alloc`. Having said that, this situation is pretty unlikely since `NUM_BUFFERS` is supposed to be greater than the number of CPUs (according to its definition in `common.h` anyway).

### DIFF
--- a/driver/others/memory.c
+++ b/driver/others/memory.c
@@ -1015,7 +1015,7 @@ void *blas_memory_alloc(int procpos){
   mypos = WhereAmI();
 
   position = mypos;
-  while (position > NUM_BUFFERS) position >>= 1;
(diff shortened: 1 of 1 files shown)


martin-frbg:
Looks good to me, thank you very much. I suspect the NUM_BUFFERS vs number of CPUs situation may be a bit more fragile than originally intended, in particular when build and deployment hardwares differ.

PR #1188, opened 2017-05-25
build: Flang compiler support

pawosm-arm:
With Flang compiler now publicly available ([https://github.com/flang-compiler](https://github.com/flang-compiler)), we can add Flang support to OpenBLAS.

Signed-off-by: Paul Osmialowski <pawel.osmialowski@arm.com>


martin-frbg:
Looks good to me, but shouldn't the unrelated change of allowing OpenMP support with CLANG be mentioned in the commit message (or get its own PR) ?

pawosm-arm:
I updated commit message to announce Clang change.

PR #1189, opened 2017-05-27
build: Flang has the same interface as PGI

pawosm-arm:
Honestly, I didn't spot it at first sight.


martin-frbg:
Good to know, thanks for this.

PR #1190, opened 2017-05-30
Update test to use openblas_make_complex_float and openblas_make_comp…

oviradoi:
Update test to use openblas_make_complex_float and openblas_make_complex_double functions


martin-frbg:
Just out of interest - is this to fix an actual compatibility issue for some platform/compiler, or "just" to improve code quality ?

oviradoi:
I was trying to build OpenBLAS with FORCE_OPENBLAS_COMPLEX_STRUCT turned on.
When I did that, the test_potrs.c file would fail to compile because it would use `float _Complex` and `double _Complex` even though OpenBLAS was using the struct.
This is why I changed this test to use the `openblas_make_complex_float` and `openblas_make_complex_double` functions.

PR #1199, opened 2017-06-08
pthreads: prefer `-pthread` during linking

cconrads-scicomp:
The documentation of [GCC](https://gcc.gnu.org/onlinedocs/gcc-7.1.0/gcc/Preprocessor-Options.html#index-pthread) and [Clang]( https://clang.llvm.org/docs/ClangCommandLineReference.html) state that `-pthread` is to be used during compiling and *linking* for POSIX threads support. Moreover, there is no pthread library in the Android NDK (try to search for `libpthread.*`) meaning there will be linker errors when using `-lpthread` (this flag is given in the pkg-config file, at least for arm-linux-androideabi targets).

Currently, `-pthread` is not used when compiling C code and `-lpthread` is regularly used during linking, possibly leading to errors.

This pull request changes *only* the linker flags for most compilers including the Intel Fortran compiler ([`pthread` documentation](https://software.intel.com/en-us/node/678318)) and the `EXTRALIBS` variable; the linker flags for the PathScale and the Open64 compilers remain unchanged because their manuals ([PathScale manual from 2008](http://www2.physics.uiowa.edu/~ghowes/teach/ihpc10/manuals/PathScaleCompiler.pdf), [Open64 4.5.2 manual](http://amd-dev.wpengine.netdna-cdn.com/wordpress/media/2012/10/open64.pdf)) do not mention this flag.

See the linked Google Groups threads for a discussion on whether `-lpthread` should work during linking:
https://groups.google.com/forum/#!topic/comp.programming.threads/NCEpG0EOCCY

### DIFF
--- a/Makefile.system
+++ b/Makefile.system
@@ -306,7 +306,7 @@ ifneq ($(OSNAME), CYGWIN_NT)
 ifneq ($(OSNAME), Interix)
 ifneq ($(OSNAME), Android)
 ifdef SMP
-EXTRALIB   += -lpthread
+EXTRALIB   += -pthread
 endif
 endif
 endif
--- a/Makefile.x86
+++ b/Makefile.x86
@@ -6,9 +6,9 @@ ARFLAGS		= -m x86
 endif
 
 ifndef SMP
-LIBMKL		= -L$(MKLPATH)/32 -Wl,-rpath,$(MKLPATH)/32 -lmkl_intel -lmkl_sequential -lmkl_core -lguide -lpthread -lm
+LIBMKL		= -L$(MKLPATH)/32 -Wl,-rpath,$(MKLPATH)/32 -lmkl_intel -lmkl_sequential -lmkl_core -lguide -pthread -lm
 else
-LIBMKL		= -L$(MKLPATH)/32 -Wl,-rpath,$(MKLPATH)/32 -lmkl_intel -lmkl_intel_thread -lmkl_core -lguide -lpthread -lm
+LIBMKL		= -L$(MKLPATH)/32 -Wl,-rpath,$(MKLPATH)/32 -lmkl_intel -lmkl_intel_thread -lmkl_core -lguide -pthread -lm
 endif
 
 # LIBMKL		= -L$(MKLPATH)/32 -lmkl_lapack -lmkl_ia32 -lguide -lpthread -lm
@@ -16,7 +16,7 @@ endif
 ifndef SMP
 LIBATLAS	= -L$(ATLAS) -lf77blas -latlas -lg2c -lm
 else
-LIBATLAS	= -L$(ATLAS) -lptf77blas -latlas -lpthread -lg2c -lm
+LIBATLAS	= -L$(ATLAS) -lptf77blas -latlas -pthread -lg2c -lm
 endif
 
 ifeq ($(COMPILER_F77), g77)
@@ -52,5 +52,5 @@ LIBVECLIB	= /System/Library/Frameworks/vecLib.framework/Versions/Current/vecLib
 ifndef SMP
 LIBATLAS	= -L$(ATLASPATH)/32 -lcblas -lf77blas -latlas -lm
 else
-LIBATLAS	= -L$(ATLASPATH)/32 -lptf77blas -lptatlas -lpthread -lm
+LIBATLAS	= -L$(ATLASPATH)/32 -lptf77blas -lptatlas -pthread -lm
 endif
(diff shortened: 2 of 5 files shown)


brada4:
POSIX votes for -lpthread and pulling macros/prototypes via headers. There is also not-so-portable -mthread and -pthreads in some niche compilers.

brada4:
Reiterating around your description - you are trying to work around absence of posix pthreads in Android NDK by making a global change, can you make it with ifdef-s addressing the issue only?

martin-frbg:
Looks good to me in principle (bit of googling found https://github.com/google/protobuf/issues/1373 for instance that confirms the Android case), though making it an exception for Android alone may be safer than changing the default for most platforms (and probably requires changes to fewer files)

brada4:
MKL and ACML link parameters changed are dead code and would better be carefully removed.

cconrads-scicomp:
The Groups thread linked above confirms that POSIX requires `-lpthread` but in practice, the very popular compilers GCC and Clang use `-pthread`. The goal of this patch is to adopt OpenBLAS to the functionality offered by these compilers but POSIX-conformance is definitely desirable.

brada4:
It should be `ifdef android` in Makefile.system and os.cmake (with proper capitalisation), other changes dont change anything for any android build. 
e.g. iOS/XCode has just blank libpthread.so while actual functions are in C library...

xianyi:
Should we merge this patch?

brada4:
No, it is 'sed s' 10x where 2x if(def) should be.

cconrads-scicomp:
@xianyi @martin-frbg Please decide which variant you want and I will modify the patch as necessary.

brada4:
if(ANDROID__NDK) in cmake
ifdef ANDROID in Makefile

And dont touch other stuff that is not for android.

cconrads-scicomp:
@brada4 lol. love your bitchy tone.

martin-frbg:
On re-reading this, it appears to me that the only actual bug to be fixed in this context might be that cmake/os.cmake is much less careful than Makefile.system in avoiding to add -lpthread on platforms that do not support this.
I suspect this is where the erroneous entry in the pkgconfig file comes from, and if that turns out to be the only actual issue it could be fixed by copying the exclusion of  CYGWIN_NT and ANDROID targets
from lines 304,305 of Makefile.system onto line 80 of cmake/os.cmake.
Or is the "-pthread" option actually mandatory for the code to function on Android (and not implied by any other option already used) ? 
Changing an option that "appeared to work just fine so far" over a purely semantic issue may bring the risk of breaking compilation on some system, however exotic or outdated, that worked just fine in the past without gaining anything.

cconrads-scicomp:
> Or is the "-pthread" option actually mandatory for the code to function on Android (and not implied by any other option already used) ?

Google dropped GCC support for Android so I will only discuss building code with Clang. [The Clang command line reference](https://clang.llvm.org/docs/ClangCommandLineReference.html) states the following for `-pthread`:
> Support POSIX threads in generated code
I read this to mean that the `-pthread` flag must be provided at compile _and_ at link time on all platforms.

Yet Android's libc called [Bionic](https://en.wikipedia.org/wiki/Bionic_(software)) contains a POSIX threads implementation (see, for example, all files starting with `pthread_` in the Bionic [`libc/bionic`](https://github.com/android/platform_bionic/tree/master/libc/bionic) source code folder). Hence, you can probably get away without using `-pthread` at link time.

Anyway, I just found out that the root cause of the problem is somewhere else. In [_How to build OpenBLAS for Android_](https://github.com/xianyi/OpenBLAS/wiki/How-to-build-OpenBLAS-for-Android) the installation command is given as
```
make PREFIX=/path/to/install-dir install
```
and this is _wrong_ because Make does not know about your cross-compilation attempt. Thus, the pkg-config file will contain the compiler and linker flags for the host system and not for the target system  (Make may find out that you compile for Android if the `bin` folder of the standalone toolchain comes first in the environment variable `PATH`). If I use the same flags on the Make command line during compilation and installation, the issue goes away. In fact, Paul Mustiere fixed the `-lpthread` linker problem a year ago in commit 7f7d04dcd27b76b4808c99643e84fe81697eda2e.

Another way to highlight this issue is to cross-compile for Android and set `ARM_SOFTFP_ABI=1` for compilation but not for the installation. On my Linux system, the pkg-config linker flags contain `-Wl,-lm_hard` if I use
```
make PREFIX=/path/to/install-dir install
```
instead of
```
make ARM_SOFTFP_ABI=1 PREFIX=/path/to/install-dir install
```

The proper solution for this problem is to update the Android build instructions.

PR #1212, opened 2017-06-23
Add Microsoft Windows 10 UWP build support

neilsh-msft:
This change adds support for building Windows UWP compatible binaries.
- Removed gen_config_h, getarch and getarch_2nd from CMake generation when building for UWP.  When built as UWP applications, these require running in an appcontainer and cannot be invoked during build time.  In order to build OpenBLAS for UWP, first run CMake as regular win32/64 configuration to generate executables.  Then remove CMakeCache.txt and CMakeFiles directory and re-run CMake for UWP configuration and build resulting OpenBLAS.sln.
- UWP doesn't support environment variables.  Changed environment variable functions to no-ops.
- UWP doesn't support TerminateThread(). #ifdef'ed out for UWP compilation.
- config.h still requires #define OS_WINNT 1 to build UWP library.  Explicitly add during CMake phase.
- Update CMakeLists.txt to still build gencblas, genconfig, and genf77blas for ALL_BUILD target.


brada4:
It looks good. It could be perfectioned with lots of typing to distinguish 2 build steps involved, but your description could land in FAQ instead as well.

xianyi:
@neilsh-msft , Could you add a note at https://github.com/xianyi/OpenBLAS/wiki/ ?

neilsh-msft:
I've updated the page https://github.com/xianyi/OpenBLAS/wiki/How-to-use-OpenBLAS-in-Microsoft-Visual-Studio

brada4:
I have added explanatory 3rd party source to avoid misunderstandings. Feel free to fix me with permalink.

neilsh-msft:
Link to Wikipedia article on UWP looks fine.

Issue #788, opened 2016-03-05
add ReLaPACK to OpenBLAS?


carlkl:
I want to draw the attention to a new linear algebra library called [Recursive LAPACK](http://hpac.rwth-aachen.de/~peise/relapack) sitting on the top of BLAS and Lapack, see:

http://hpac.rwth-aachen.de/~peise/relapack
https://github.com/HPAC/ReLAPACK (MIT license)

According to the authors this library even beats optimized Blas/LAPACK implementations in several cases (see the paper from their website). The API seems to be very simple: just use the augumented functions names instead of the original Lapack functions: [relapack.h](https://github.com/HPAC/ReLAPACK/blob/master/inc/relapack.h)

I propose to add this library to OpenBLAS and to switch between Lapack and ReLAPCK functions steered with an environment variable and/or an function. Any hints how to implement this in an easy way?

xianyi:
* You can create a `relapack` folder or use `git submodule`.

* Then, add a flag (e.g. `BUILD_RELAPACK`) at `Makefile.rule`.

* Remove othe redundant objs at `interface/lapack/Makefile` and `lapack-netlib/SRC/Makefile` with `BUILD_RELAPACK=1`

elmar-peise:
I think ReLAPACK would be a great and easy way to speed up LAPACK performance in OpenBLAS: As seen in our [paper on arXiv](http://arxiv.org/abs/1602.06763), with OpenBLAS ReLAPACK is faster than LAPACK for most of its routines. 

In case one wants to select whether to call ReLAPACK or another LAPACK implementation depending on the architecture (or other factors), I see two alternatives:
* Use ReLAPACK's configuration header `config.h` to specify which of its routines to make available under their default LAPACK interfaces.  This is a compile time option.
* Disable the LAPACK interface altogether and selectively call ReLAPACK's own/internal routines (named, e.g., RELAPACK_dgetrf) which have the same interfaces as their LAPACK counterparts.

Either way, I'm more than happy to help with any ReLAPACK-related questions or issues and suggestions and feedback are very welcome!

xianyi:
@elmar-peise , thank you for your comment. I read your paper. Great work.

gaming-hacker:
i think the performance will vary wildly depending upon the hardware and precision needed.  i can't see using it with  mulitprecision libs as the overhead would probably at least be O^2 but it's another tool in the toolbox if someone wants to use.  recursion also has its own limitations of numerical stability.

[...]

PR #1214, opened 2017-06-28
Initial import of ReLAPACK

martin-frbg:
Add Peise's recursive implementation of LAPACK functions from https://github.com/HPAC/ReLAPACK for issue #788 , enabled by default in Makefile.rule (mostly for the automated checks now) . Cmake integration still needs to be done.


xianyi:
It looks like travis-ci failed because of multiple definitions. e.g.

`multiple definition of `slauum_'`

martin-frbg:
Yes, but at the moment I do not see where they are coming from - on Linux both compiling natively and cross-compiling for Windows with mxe the relevant netlib definitions get replaced by their relapack wrappers as they should. Is there any way to access the travis build tree after a failure to see  where things ended up ?

martin-frbg:
Did not notice that it is mxe x86_64-w64 that is giving the trouble, my local test was i686-w64 only. Seems the behaviour of `ar` is different in the 64bit version (though the only hint is a message about
option "-u" being ignored that is only generated by x86_64-w64-mingw32)

martin-frbg:
Got it now - the optimized LAPACK functions (such as slauum) from interface/ are expressly built with varying suffixes on the object file (.o,.obj) depending on the target operating system, while the pure netlib parts are compiled to .o files throughout. This needs to be reflected in the relapack Makefile for the substitution by `ar -r` to work. And I had failed to notice that my local cross-compiles had errored out before even building the shared library as they did not skip the "make test" step (which would have windows binaries incompatible with the Linux host) for some reason. Not sure yet if that is a separate issue or just an oversight on my part.
Now that all tests are passing, the remaining question is whether ReLAPACK should be enabled by default in Makefile.rule (as I needed it to be for the travis/appveyor runs)

xianyi:
@martin-frbg Great work!

xianyi:
@martin-frbg I suggest ReLAPACK is disable by default for 0.2.20 version.

Makefile.rule
```diff
@@ -97,7 +100,7 @@ BUILD_LAPACK_DEPRECATED = 1
 NO_WARMUP = 1
 
 # If you want to disable CPU/Memory affinity on Linux.
-NO_AFFINITY = 1
+#NO_AFFINITY = 1
```
nalimilan, line 103: Did you intend to revert this here? It's been considered as a good default by several projects using OpenBLAS, cf. https://github.com/JuliaLang/julia/pull/9639#issuecomment-68912940.

stevengj: Yes, hopefully this was an inadvertent change, since it seems unrelated to this PR …

Processor affinity is a poor default for running on non-dedicated machines.

martin-frbg:
Ah, sorry, probably not. Good catch. I was working on cpuset support at around the same time so it seems some unwanted cross-pollination may have happened. :(

Makefile.rule
martin-frbg: Fixed by #1495 a minute ago (unless I made another blunder)

PR #1216, opened 2017-06-29
Defaulting not-yet-implemented ARMv7 softfp functions to their naive C implementations

### LINKED ISSUE #1215: Addressing the hard float ABI support removal from Android NDK
Since the [r12](https://github.com/android-ndk/ndk/wiki/Changelog-r12) release of the Android NDK, hard float ABIs are no longer supported. This problem was aggravated starting from the r15 version due to the [removal of the libm_hard](https://stackoverflow.com/questions/43445539/android-ndk-libc-support-for-hard-float) from the development kit.

From what I've gathered, the **develop** branch already addresses this issue by supporting the ARM_SOFTFP_ABI flag, recently incorporating changes for the **arm_soft_fp_abi** branch. However, there are still a number of issues such as #1168 and #1145 which show that defaulting to **-mfloat-abi=hard** flag can provoke a lot of confusion. Also, in #1168 it is highlighted that even though the ARM_SOFTFP_ABI flag is supported, real support is limited to single float operations.

Given this:
- Does it make sense to change ARMV7 builds, making them use naive C loops just as ARMV5 does? #777
- Or should the ARMV7 only use the naive implementations for double precision, while still benefiting from single precision assembly code?
- Would there be any difference in performance when explicitly stating to the compiler that we are targeting an ARMV7 arch instead of an ARMV5 for naive implementations?
- Should the project default the ARM_SOFTFP_ABI to 1?

I'd be very grateful for any comments of people more involved with the project, since I am fairly new to all this cross-compilation stuff.

akira-miasato:
This temporarily resolves #1215 as discussed in the issue. When compiling with ARM_SOFTFP_ABI=1 for ARMv7, non-single precision OPs will default to their pure C implementations. This is done via Makefile conditions with the aforementioned flag.

### DIFF
--- a/kernel/arm/KERNEL.ARMV7
+++ b/kernel/arm/KERNEL.ARMV7
@@ -1,97 +1,112 @@
 
 #################################################################################
+
+############################################################
+# Files already implemented in ARM Assembly for softfp ABI #
+############################################################
 SAMAXKERNEL  = iamax_vfp.S
-DAMAXKERNEL  = amax.c
-CAMAXKERNEL  = zamax.c
-ZAMAXKERNEL  = zamax.c
 
 SAMINKERNEL  = iamax_vfp.S
-DAMINKERNEL  = amin.c
-CAMINKERNEL  = zamin.c
-ZAMINKERNEL  = zamin.c
 
 SMAXKERNEL   = iamax_vfp.S
-DMAXKERNEL   = max.c
 
 SMINKERNEL   = iamax_vfp.S
-DMINKERNEL   = min.c
 
 ISAMAXKERNEL = iamax_vfp.S
-IDAMAXKERNEL = iamax.c
-ICAMAXKERNEL = izamax.c
-IZAMAXKERNEL = izamax.c
 
 ISAMINKERNEL = iamax_vfp.S
-IDAMINKERNEL = iamin.c
-ICAMINKERNEL = izamin.c
-IZAMINKERNEL = izamin.c
 
 ISMAXKERNEL  = iamax_vfp.S
-IDMAXKERNEL  = imax.c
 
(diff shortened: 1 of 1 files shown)


ashwinyes:
This is not a proper fix. Changes are needed in param.h. Otherwise DGEMM will result in error.

Its difficult to explain the changes if you are not familiar with OpenBLAS code.

brada4:
And it unconditionally degrades raspbian experience....

martin-frbg:
@ashwinyes @akira-miasato sorry I honestly did not realize that.
@brada4 would raspbian have  ARM_SOFTFP_ABI defined at all ?

ashwinyes:
@martin-frbg Since there are lot of issues on ARM softfp, I am working on trying to add softfp support on all ARM assembly files. @xianyi had submitted the support for few single precision APIs but as you know it is not complete.

My changes can be found in this branch https://github.com/ashwinyes/OpenBLAS/commits/develop_arm_softfp
This is a work in progress. Current status is almost all APIs fall-back to C implementations if softfp is detected. I have not tested this branch yet.

brada4:
arm debian has 3 copies of libc for each capability, nothing removed yet, it will work all ways.

akira-miasato:
> Current status is almost all APIs fall-back to C implementations if softfp is detected.

And this means that all but the already implemented single-precision functions will default to C? If your branch is a proper fix to the issue, I could test it in my case at least. If this is the case, I may close this PR.

ashwinyes:
You are welcome to test my branch.

Most functions (including some of those for which xianyi had added support) fall back to C implementations in that branch. I will be updating the branch as and when I add support to new functions. I will create a PR when eveything is done. If there are any functions which you need at priority, let me know.

You may close this PR anyways as there are issues with this PR.

PR #1218, opened 2017-06-30
Optimise loads on Power9 LE

m-brow:
This changes the load instruction for some subroutines on powerpc.
On POWER9 LE the lxvd2x instruction performs far better than the current lxvw4x.

This doesn't change any functionality, but improves performance on P9 LE systems.
Note: This still works on P8


martin-frbg:
I take it that "still works on P8" includes "causes no performance penalty" on that platform either, or should these actually be #ifdef'd (and POWER9 be given its own TARGET option) ?

m-brow:
Yes, there's no penalty on the P8 platform. 
Making a P9 target may be useful for future work or work using P9 specific instructions. But as this works on P8 as well I think it's ok to leave it in here.

xianyi:
Thank you for the patch.

If you need some specific P9 instructions, I can create a branch in future.

PR #1220, opened 2017-07-01
arm64: Change mtune/mcpu options for THUNDERX2T99 target

ashwinyes:


ashwinyes:
@martin-frbg @xianyi Request you to merge this at your earliest convenience.

No targets are affected other than THUNDERX2T99.

Issue #1208, opened 2017-06-19
Issue compiling OpenBLAS


mariasmo:
Hi,

I am running into an issue trying to compile the OpenBLAS as part of compiling Kaldi for Android. I followed all the instructions on this blog(http://jcsilva.github.io/2017/03/18/compile-kaldi-android/ ) until the following where I am stuck:

make TARGET=ARMV7 HOSTCC=gcc CC=arm-linux-androideabi-gcc NOFORTRAN=1 NUM_THREADS=32 libs 

This is the error I am getting when trying to run the above command:
-----------------------------------------------------------------
C Compiler (arm-linux-androideabi-gcc) is something wrong.
1 at ./c_check line 28.
make: *** [config.h] Error 2
Makefile.system:165: Makefile.conf: No such file or directory
Makefile.system:912: Makefile.: No such file or directory
make: *** No rule to make target `Makefile.'. Stop.
-----------------------------------------------------------------

I got the latest branch of OpenBLAS.

I did set the PATH to the android tool chain correctly.

I am running out of options.

Any help is greatly appreciated.

Thanks in advance.

martin-frbg:
Still looks as if the PATH to the android toolchain is not set correctly (or the toolchain is incomplete).
At that point in c_check it tries to run the command `arm-linux-androideabi-gcc -E ctest.c` - you could try entering this on a command line to see what error message you get.

brada4:
Unlikely there will be 32-core 32bit android any time soon. What is your NDK vrsion? Is it recent clang-based?

mariasmo:
Hi martin-frbg & brada4: thanks for your responses.

The issue seemed to be that it did not have permissions to the toolchain directory when running as my user. As a test, i ran it as root and it compiled fine. 

My NDK version is:
15.0.4075724 

Thanks again for your help.( I will ping if I run into any more issues::))

brada4:
But close this issue in the meantime, Android NDK is not OpenBLAS component.

mariasmo:
Hi martin-frbg & brada4: 

I ran into the issue described in the following thread:

https://github.com/xianyi/OpenBLAS/issues/1173

Since that thread was closed, I am putting in my questions here:

So after doing the 
make TARGET=ARMV7 HOSTCC=gcc CC=arm-linux-androideabi-gcc NOFORTRAN=1 NUM_THREADS=32 libs

command, I executed the following command:

make install PREFIX=`pwd` /install

[...]

Issue #1206, opened 2017-06-15
segmentation fault when use cblas_sgemm.


ctgushiwei:
cpuinfo: processors:4      ARMv7     rev1      cpu arch:7   Hardware:hi3536
make:TARGET=ARMV7 NOFORTRAN=1 HOSTCC=gcc NO_LAPACK=1 ONLY_CBLAS=1

my openblas is 0.2.19.when i use cblas_sgemm,  backtrace from crash:blas_server.c and the fault is segmentation fault

how can i solve this problem

brada4:
Could you close previous issue on same subject?

- does it crash with single-thereaded OpenBLAS
- can you generate backtrace: run your program in gdb, then when it catches fault run 't a a bt' in gdb, and attach resulting output (do not copy/paste)

martin-frbg:
Can you show a simple example of your cblas_sgemm usage ? It could be that you got some calling parameter wrong, in particular if you used the documentation for the regular blas function sgemm, as array handling differs between c (cblas) and fortran (blas).

ctgushiwei:
@brada4  both single-threaded OpenBLAS or multi-threaded OpenBLAS crash at the same function. blas_sever.c when use exec_blas function..   and my test code is sample,it is ok when test at PC.

ctgushiwei:
@brada4 @martin-frbg  my test code is

float Aa[2*3]={1,2,3,4,5,6};
float Bb[3*4]={1,2,3,4,5,6,7,8,9,10,11,12};
float y1[2*4]={1,2,3,4,5,6,7,8};

cblas_sgemm(CblasRowMajor,CblasNoTrans,CblasNoTrans,2,4,3,1.0,Aa,3,Bb,4,1.0,4);
printf("%f",y1[0]);

when use single-threaded OpenBLAS it crashed at the int exec_blas(BLASLONG num,blas_queue_t*queue) when call (routine)(queue->args,queue->range_m,queue->range_n,queue->sa,queue->sb,0)

when use multi-threaded OpenBLAS ,it can not call the blas_sever.c,i do not know what function it use?because it only generate blas_server.o

how can i  solve this problem

brada4:
Are you compiling/linking OpenBLAS and your program with 'hardfp' ?
Does it stop crashing when using ARMV5 target instead (which is C-only and works with softfp)?

ctgushiwei:
[...]

Issue #1168, opened 2017-04-26
Problem with ARMv7 (cblas_ddot)


misbullah:
Hi,

Recently, I used OpenBLAS with Kaldi Toolkit for Android.
I can use OpenBLAS for Intel ATOM platform and get desired result from Kaldi.

When, I build OpenBLAS for armv7-a, there is no error but I get error when I run the executable file.
The error is shown in the following.

**LOG (dnn_batch_static:void kaldi::IvectorExtractor::ComputeDerivedVars()():ivector/ivector-extractor.cc:183) Computing derived variables for iVector extractor
WARNING (dnn_batch_static:void kaldi::TpMatrix<Real>::Cholesky(const kaldi::SpMatrix<Real>&) [with Real = double]():matrix/tp-matrix.cc:110) Cholesky decomposition failed. Maybe matrix is not positive definite. Throwing error
terminate called after throwing an instance of 'std::runtime_error'
  what():  Cholesky decomposition failed.
Aborted**

I trace matrix/tp-matrix.cc code, then I found that the error was happened from this template:

template<typename Real>
void TpMatrix<Real>::Cholesky(const SpMatrix<Real> &orig) {
  KALDI_ASSERT(orig.NumRows() == this->NumRows());
  MatrixIndexT n = this->NumRows();
  this->SetZero();
  Real *data = this->data_, *jdata = data;  // start of j'th row of matrix.
  const Real *orig_jdata = orig.Data(); // start of j'th row of matrix.
  for (MatrixIndexT j = 0; j < n; j++, jdata += j, orig_jdata += j) {
    Real *kdata = data; // start of k'th row of matrix.
    Real d(0.0);
    for (MatrixIndexT k = 0; k < j; k++, kdata += k) {
      Real s = cblas_Xdot(k, kdata, 1, jdata, 1);
      // (*this)(j, k) = s = (orig(j, k) - s)/(*this)(k, k);
      jdata[k] = s = (orig_jdata[k] - s)/kdata[k];
      d = d + s*s;
    }
    // d = orig(j, j) - d;
    d = orig_jdata[j] - d;
    
    if (d >= 0.0) {
      // (*this)(j, j) = std::sqrt(d);
      jdata[j] = std::sqrt(d);
    } else {
      KALDI_WARN << "Cholesky decomposition failed. Maybe matrix "
          "is not positive definite. Throwing error";
      throw std::runtime_error("Cholesky decomposition failed.");
    }
  }
}

[...]

PR #1221, opened 2017-07-01
arm: add support for softfp in arm vfp assembly files

ashwinyes:


ashwinyes:
This PR adds support for softfp in all arm vfp assembly kernel files.

I have done only basic testing of this PR since I don't have an ARM machine at my disposal. All my testing were done on QEMU emulator.

Request owners of following issues to help test this PR and see whether the issue is resolved or not.

#1088 #1145 #1168 #1206 #1208 #1215 #1145 @misbullah @gangm @odieXin @ctgushiwei  @mariasmo @akira-miasato @gangm

ashwinyes:
@tomaklutfu  Thanks for reviewing.

Yes. Those were wrong. I agree. Had corrected those in my local workspace. But still the GEMM answer was not correct. I didnt explore it much as these files were no longer required after all the changes were made in the assembly files.

Please see that I have deleted those files in the last commit.

xianyi:
@ashwinyes Thank you for the PR.

Do I need to merge #1216 at first?

ashwinyes:
No #1216 is not required. It can closed.

I would prefer these changes to get tested by someone working on android and get a "Tested" stamp on it. I have only tested it on QEMU emulator.

xianyi:
We will test it this week

mariasmo:
ashwinyes & xianyi,

i am trying to understand this request.

so i have been compiling openblas using the following command:

make TARGET=ARMV7 HOSTCC=gcc CC=arm-linux-androideabi-gcc NOFORTRAN=1 NUM_THREADS=32 libs

I believe it defaults to "hardfloat" right?

my understanding is that the PR removes the "hardfloat" by default? and can i use the same above mentioned command or do i need to use the following:
ARM_SOFTFP_ABI=1

and to test, do can i get the latest from develop branch? or should i wait until the PR is merged?

thanks

ashwinyes:
Let me explain.

softfp and hard float ABI differs in the way the arguments are passed on to functions and how values are returned. If your project has full C files only, then compiler will take care of everything.

But issue is when you have hand written assembly code. The code in OpenBLAS was always assuming hard float ABI was used. So the code expects values to be certain registers according hard fp ABI which will not be the case when you use softfp ABI.

Now coming to your question. arm-linux-androideabi-gcc (also arm-linux-gnueabi-gcc) by default is **softfp ABI**. When you use this compiler to compile OpenBLAS code, the issue I mentioned above arises.

The PR doesn't remove anything. It just checks whether you are using a softfp or hard fp compliant and adjusts the arguments/return values accordingly.

So, basically you don't need to give **any** additional make parameters to OpenBLAS with the PR.

To test, you can use the branch https://github.com/ashwinyes/OpenBLAS/tree/develop_arm_softfp until this PR gets merged.

Hope I cleared your doubt.

mariasmo:
ashwinyes

thanks for the explanation.

so prior to your changes OpenBLAS was assuming hard fp even though the default for arm-linux-androideabi-gcc  is softfp. Am i correct in my understanding?

and now with your changes , it will compile for only soffp and any other libraries i compile down the stream i shuould assume softfp and not hardfp, correct?

thanks

ashwinyes:
Yes. You are correct.

mariasmo:
thanks

akira-miasato:
I'll be checking it today regarding #1168 and #1215.

akira-miasato:
Tested and issue #1168 seems solved with Android NDK r15b, with speedups comparable to what I achieved with the dirty adaptation from #1215/#1216.

However, some changes regarding GCC/Clang support will be needed (NDK dropped GCC support). I should open a new issue for those things, since this PR focuses on the softfp/hard ABI issues.

ashwinyes:
"fnmacs s3 , s1, s4" is a valid ARM instruction. If clang doesn't recognize, then its more of a clang issue than OpenBLAS issue. But I will try to see if I can workaround it.

ashwinyes:
Looks like clang is not recognizing pre-UAL mnemonics. We will have to replace with UAL mnemonics.

http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0473c/CIHFAHFG.html

ashwinyes:
@akira-miasato 
Addressed the clang issue in https://github.com/xianyi/OpenBLAS/pull/1226

jxllh123456:
thanks a lot, India guy

jxllh123456:
thanks a lot, sure enough, India's software industry has good superiority.

PR #1226, opened 2017-07-07
arm: Fix clang compilation for ARMv7

ashwinyes:
clang is not recognizing some pre-UAL VFP mnemonics like fnmacs, fnmacd,
fnmuls and fnmuld. Replaced them with equivalent UAL mnemonics which are
vmls.f32, vmls.f64, vnmul.f32 and vnmul.f64 respectively.

### DIFF
--- a/kernel/arm/axpy_vfp.S
+++ b/kernel/arm/axpy_vfp.S
@@ -71,14 +71,14 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #if defined(DOUBLE)
 
 #define	FMAC_R1	fmacd
-#define FMAC_R2 fnmacd
+#define FMAC_R2 vmls.f64
 #define	FMAC_I1	fmacd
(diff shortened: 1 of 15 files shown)


akira-miasato:
I cherry-picked this commit over #1221 and compiled with the following commands in Ubuntu 17.04:
```
export PATH=/path/to/ndk/binaries:$PATH
make TARGET=ARMV7 HOSTCC=clang-4.0 CC=clang NO_SHARED=1 NOFORTRAN=1 USE_THREAD=1 NUM_THREADS=32 ARM_SOFTFP_ABI=1 libs
```
And tested over #1168. Everything worked fine. Thanks a lot.

One thing that I purposely omitted when referencing this problem was that there were issues when linking GCC-compiled OpenBLAS with my clang-compiled Kaldi due to stderr definitions, but only when USE_THREAD was enabled. When compiling OpenBLAS with clang, I expected the error to disappear, and indeed it did. I'll reference this PR in #1215 with these details and close the issue for now.

PR #1230, opened 2017-07-09
Add sched_getcpu implementation for pre-2.6 glibc

### LINKED ISSUE #1210: Cannot compile with thread affinity enabled, undefined reference to `sched_getcpu'
I have tried to compile both the latest release and the latest git code, but both fail to compile if I enable thread affinity.
I am pretty sure I really need to have thread affinity working, because on a 4 socket Barcelona cluster (4*12=48 cores per node) I am getting absolutely abysmal scaling on DGELSY, sometimes even heavy negative scaling. I am fairly sure this is due to having 4 NUMA regions, and thread affinity could help resolve it. But I cannot get openblas to compile with thread affinity on this cluster.
The cluster is using GCC 4.7.2.

Here is the part where failures happen:

> gfortran -O2 -Wall -m64   -o dblat1 dblat1.o ../libopenblas_barcelonap-r0.2.20.dev.a -lm -lpthread -lgfortran -lm -lpthread -lgfortran
> sblat2.f: In function ‘schk1’:
> sblat2.f:646:0: warning: ‘nargs’ may be used uninitialized in this function [-Wmaybe-uninitialized]
> dblat2.f: In function ‘dchk1’:
> dblat2.f:646:0: warning: ‘nargs’ may be used uninitialized in this function [-Wmaybe-uninitialized]
> ../libopenblas_barcelonap-r0.2.20.dev.a(init.o): In function 'gotoblas_affinity_init':
> init.c:(.text+0xd07): undefined reference to 'sched_getcpu'
> collect2: error: ld returned 1 exit status
> make[1]: *** [sblat1] Error 1
> make[1]: *** Waiting for unfinished jobs....

martin-frbg:
Fixes #1210, compilation on RHEL5 with affinity enabled

### DIFF
--- a/driver/others/init.c
+++ b/driver/others/init.c
@@ -354,6 +354,24 @@ static int numa_check(void) {
   return common -> num_nodes;
 }
 
+#if defined(__GLIBC_PREREQ)
+#if !__GLIBC_PREREQ(2, 6)
+int sched_getcpu(void)
+{
+int cpu;
+FILE *fp = NULL;
+if ( (fp = fopen("/proc/self/stat", "r")) == NULL)
+   return -1;
+if ( fscanf( fp, "%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%*s%d", &cpu) != 1) {
+  fclose (fp);
+  return -1;
+  }
+  fclose (fp);
+  return(cpu);
+}
+#endif
+#endif
+
 static void numa_mapping(void) {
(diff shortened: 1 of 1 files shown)


brada4:
It compiles without errors
It uses all cores with sched_autogroup_enabled (default in elrepo.org 3.2 kernel) where build without affinity would not.

martin-frbg:
Err, do I take it that this is a good sign, or not ? Note this does not address the cgroup awareness issue in any way - I have another patch ready for that, but would consider it post-0.2.20 material while this one here is expected to "have no side effects at all" (TM)

brada4:
Firstly it does not break anything, secondly we are able to bind threads to cores. (Tested only on Xeon X5650 aka westmere xeon)

martin-frbg:
Thanks, committing then.

Issue #1155, opened 2017-04-13
cgroup awareness


kcgthb:
Hi!

I understand that OpenBLAS tries to automatically detect the number of CPU cores on a machine at runtime, to determine the number of threads to start when `{OPENBLAS,GOTO,OMP,_NUM_THREADS}` is not set.

It works fine in most cases, but when the process runs in a cgroup context, for instance one where the `cpuset` subsystem is in use, it may result in less-than-optimal behavior.

For instance, on a 16-core machine, if a process runs inside a cgroup where 4 CPUs have been allocated via `cpuset`, OpenBLAS will start 16 threads, which will be pinned on just 4 CPU-cores and which will compete with each other. In the end, the performance will be about 1/4th of what it would have been by just starting 4 threads.

So I'm wondering if any thought has been given about this already, and how OpenBLAS could try to detect if it's running in a constrained context, in order to properly allocate the resources it can use.

Thanks!

brada4:
cgroups do not signal affected process at all. Do you want to check before setaffinity for duplicate processors?

brada4:
You can check around lxc and docker reports and lists where they come to conclusion that it is how kernel works. Unless you have stable code to detect crippled cpu number please close this issue. It is you configuring resource fences, should not be too hard to slip an environment variable between the lines

kcgthb:
With the rise of containers of all kind, I think better detecting what resources are actually usable  would be a sensible thing to tend to. And would also avoid to produce erroneous bug reports regarding performance regression.

Getting CPU affinity could be done with [`sched_getaffinity()`](https://linux.die.net/man/2/sched_getaffinity), and it would also cover the cases where a parent process is called with `numactl` or `taskset`.

brada4:
You mentioned that containers are broken. What workaround you offer?
sched_getaffinity() does not see cpuset.

martin-frbg:
From my very limited understanding (gained by reading the cpuset, sched_setaffinity and CPU_SET manpages) a possible (and probably Linux-only) solution could be to use the number of cpus originally obtained by get_num_procs (in driver/others/init.c) as input to CPU_ALLOC() and CPU_ALLOC_SIZE() to setup a sufficiently large cpuset struct and then query the CPU_COUNT_S() of that cpuset and continue working with this value instead of the global cpu count. 
[...]

PR #1239, opened 2017-07-15
Honor cgroup/cpuset limits when enumerating cpus

martin-frbg:
Fix for #1155


xianyi:
@jirutka , We will revert this PR, then release a new version.

jirutka:
I don't think it's needed to revert entire PR or commit, just fix ifdefs in memory.c.

jirutka:
I don't think it's needed to revert entire PR or commit, just fix ifdefs in memory.c.

martin-frbg:
I have reverted the commit as I cannot fix it immediately. From #630 it could be that init.c also needs changing. Will try try to set up a test environment locally before redoing the PR...

PR #1241, opened 2017-07-17
travis: compile 32 bit binary

xoviat:


xoviat:
I acutally don't have time to finish this, so closing for now.

PR #1249, opened 2017-07-25
Honor cgroup/cpuset limits when enumerating cpus

martin-frbg:
Fix for #1155 redone, with ifdefs reworked to build on non-glibc systems as well

### DIFF
--- a/driver/others/init.c
+++ b/driver/others/init.c
@@ -826,15 +826,58 @@ void gotoblas_affinity_init(void) {
   common -> shmid = pshmid;
 
   if (common -> magic != SH_MAGIC) {
+    cpu_set_t *cpusetp;
+    int nums;
+    int ret;
+
 #ifdef DEBUG
     fprintf(stderr, "Shared Memory Initialization.\n");
 #endif
 
     //returns the number of processors which are currently online
-    common -> num_procs = sysconf(_SC_NPROCESSORS_CONF);;
 
+    nums = sysconf(_SC_NPROCESSORS_CONF);
+
+#if !defined(__GLIBC_PREREQ)
+    common->num_procs = nums;
+#else
+
+#if !__GLIBC_PREREQ(2, 3)
+    common->num_procs = nums;
+#elif __GLIBC_PREREQ(2, 7)
+    cpusetp = CPU_ALLOC(nums);
+    if (cpusetp == NULL) {
+        common->num_procs = nums;
+    } else {
+        size_t size;
+        size = CPU_ALLOC_SIZE(nums);
+        ret = sched_getaffinity(0,size,cpusetp);
+        if (ret!=0)
+            common->num_procs = nums;
+        else
+            common->num_procs = CPU_COUNT_S(size,cpusetp);
+    }
+    CPU_FREE(cpusetp);
+#else
+    ret = sched_getaffinity(0,sizeof(cpu_set_t), cpusetp);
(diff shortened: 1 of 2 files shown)


jirutka:
No, it’s not fixed at all, `memory.c` is the same as before.

jirutka:
~~Hm, I’m not sure now, develop branch passed on Alpine on Travis #1255. Maybe I’ve made some mistake when backporting the fix. I’ll look at it later.~~ (EDIT: No, it’s really broken, see next comment)

martin-frbg:
If you do not even want the cgroups functionality, you can checkout master instead of develop. I simply reverted the original PR there with the understanding that xianyi will release 0.2.21 as soon as possible to resolve the issue.

jirutka:
I was actually right, it’s still broken. However, it fails only with specific flags: `BINARY=64 USE_OPENMP=0 NO_LAPACK=0 NO_AFFINITY=1 TARGET=core2`. I guess that `NO_AFFINITY=1` is the one that reveals this error. See https://travis-ci.org/xianyi/OpenBLAS/jobs/258627266.

jirutka:
I didn’t say that I don’t want cgroups functionality, musl of course supports cgroups.

All I want is to update Alpine’s OpenBLAS package from 0.2.19 to the latest version (currently 0.2.20). I can’t, due to 29fc429d9a34a513e737df287c8a8785c9f7a332 that breaks compatibility with non-glibc systems. I’ve tried to pick v0.2.20 and port changes introduced in this PR that is supposed to rework the previous faulty commit, but it has exactly the same problem as the previous one.

After I removed irrelevant changes in whitespaces, this is what I got:

```
--- a/driver/others/init.c
+++ b/driver/others/init.c
@@ -837,8 +837,12 @@ void gotoblas_affinity_init(void) {
     //returns the number of processors which are currently online

     nums = sysconf(_SC_NPROCESSORS_CONF);
-
-#if !defined(__GLIBC_PREREQ) || !__GLIBC_PREREQ(2, 3)
+
+#if !defined(__GLIBC_PREREQ)
+    common->num_procs = nums;
+#else
+
+#if !__GLIBC_PREREQ(2, 3)
     common->num_procs = nums;
 #elif __GLIBC_PREREQ(2, 7)
     cpusetp = CPU_ALLOC(nums);
@@ -870,8 +874,8 @@ void gotoblas_affinity_init(void) {
     common->num_procs = CPU_COUNT(sizeof(cpu_set_t),cpusetp);
 #endif

-#endif
-
+#endif
+#endif
     if(common -> num_procs > MAX_CPUS) {
       fprintf(stderr, "\nOpenBLAS Warning : The number of CPU/Cores(%d) is beyond the limit(%d). Terminated.\n", common->num_procs, MAX_CPUS);
       exit(1);
```

It does not fix `memory.c`.

I didn’t want to revert 29fc429d9a34a513e737df287c8a8785c9f7a332 completely in the package, because it introduces larger change.

martin-frbg:
If you pull from master instead of develop, you should currently get a 0.2.20 with just the (broken) cgroups patch reverted - that PR was not intended for 0.2.20 anyway as I had done only (too) limited testing. Difference to current develop besides the (apparently still broken) cgroups patch would be the L1 cache detection fix for recent Intel chips. I'll see if I can do (and test!) the fix for memory.c on develop this weekend - past week has been quite stressful for me.

Issue #1252, opened 2017-07-27
Building v0.2.20 fails on non-glibc systems


pbregener:
Building OpenBLAS 0.2.20 fails on non-glibc systems, due to unsafe use of the `__GLIBC_PREREQ` preprocessor macro. Will there be a patch release soon or should downstream maintainers patch this?

brada4:
It is #1247 that reverts offending change

martin-frbg:
Master branch has the problematic change already reverted in anticipation of a quick interim release,
current develop branch has a **partially** fixed version of the intended functionality (**that still breaks when compiled with NO_AFFINITY=1, hopefully fixed for good in a few hours**)

brada4:
I think it is more to @xianyi if he intends 0.2.21 tomorrow....

martin-frbg:
Inadvertently closed with my PR for develop - reopening until 0.2.21 is out.

martin-frbg:
Guess we can skip 0.2.21 now as it would no longer be a "quick fix" after six months have passed, but hopefully 0.3.0 can be released "very soon now".

Issue #1089, opened 2017-02-09
`dtbmv` in `libopenblas_haswellp-r0.2.18.dylib` on macOS Sierra (10.12.3) fails?


hiro4bbh:
`dtbmv` in `libopenblas_haswellp-r0.2.18.dylib` on macOS Sierra (10.12.3) may fail on large scale as the following code.

```c
// clang -L/usr/local/opt/openblas/lib -I/usr/local/opt/openblas/include -lopenblas openblas_dtbmv.c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

#include <cblas.h>

int main(int argc, char **argv)
{
  // if n <= 2097136 then dtbmv work almost surely...
  int n = 2097136 + 1;
  double *p = calloc(n, sizeof(double));
  p[0] = 4.0;
  p[1] = 2.0;
  p[2] = 1.0;
  printf("p@%p = (%g, %g, %g, %g, ...)\n", p, p[0], p[1], p[2], p[3]);
  cblas_dtbmv(CblasColMajor, CblasUpper, CblasNoTrans, CblasNonUnit, n, 0, p, 1, p, 1);
  printf("p@%p = (%g, %g, %g, %g, ...)\n", p, p[0], p[1], p[2], p[3]);
  free(p);
  return 0;
}
```

I have only MacBook running macOS Sierra, so I have no memory debugger like `valgrind`.
However, I quote the console output running with `lldb`.

```
$ lldb ./a.out
(lldb) target create "./a.out"
Current executable set to './a.out' (x86_64).
(lldb) r
Process 48585 launched: './a.out' (x86_64)
p@0x104800000 = (4, 2, 1, 0, ...)
Process 48585 stopped
* thread #2: tid = 0x125f57, 0x00007fffe2027c13 libsystem_platform.dylib`_platform_bzero$VARIANT$Haswell + 115, stop reason = EXC_BAD_ACCESS (code=1, address=0x107800000)
    frame #0: 0x00007fffe2027c13 libsystem_platform.dylib`_platform_bzero$VARIANT$Haswell + 115
libsystem_platform.dylib`_platform_bzero$VARIANT$Haswell:
->  0x7fffe2027c13 <+115>: movq   %rsi, (%rdi,%rdx)
    0x7fffe2027c17 <+119>: subq   $0x8, %rdx
    0x7fffe2027c1b <+123>: jae    0x7fffe2027c13            ; <+115>
[...]

PR #1255, opened 2017-07-28
Travis: Rewrite config, build and test also on Alpine Linux (musl libc)

jirutka:
1. Refactores `.travis.yml` to use Travis Build Stages. This allows to define jobs with very different configuration in more clean way.
2. Replaces `sudo apt-get` with [APT addon](https://docs.travis-ci.com/user/installing-dependencies/#Installing-Packages-with-the-APT-Addon). Since sudo is not needed anymore, it runs on Travis containers-based infrastructure that is much faster than their VMs infrastructure (used when sudo is needed).
3. Adds jobs that build and test OpenBLAS on Alpine Linux v3.6 (inside chroot), i.e. verify that it works with musl libc. I hope that this will help to avoid errors like #1252 in future.

You’ve been still running on Ubuntu Presty builders, but new default is Trusty ([read more](https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming)). Thus I‘ve explicitly set `dist: presty` to let it stay on Presty, to not change build environment by this commit.

Alpine jobs needs sudo (for chroot), so they run on VMs infrastructure. That’s why they are much slower than other jobs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,67 @@
+# XXX: Precise is already deprecated, new default is Trusty.
+# https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming
+dist: precise
+sudo: false
 language: c
+compiler: gcc
+
+jobs:
+  include:
+    - &test-ubuntu
+      stage: test
+      addons:
+        apt:
+          packages:
+            - gfortran
+      script:
+        - set -e
+        - COMMON_FLAGS="DYNAMIC_ARCH=1 TARGET=NEHALEM NUM_THREADS=32"
+        - make QUIET_MAKE=1 $COMMON_FLAGS $BTYPE
+        - make -C test $COMMON_FLAGS $BTYPE
+        - make -C ctest $COMMON_FLAGS $BTYPE
+        - make -C utest $COMMON_FLAGS $BTYPE
+      env:
+        - TARGET_BOX=LINUX64
+        - BTYPE="BINARY=64"
+
+    - <<: *test-ubuntu
+      env:
+        - TARGET_BOX=LINUX64
+        - BTYPE="BINARY=64 USE_OPENMP=1"
+
+    - <<: *test-ubuntu
+      env:
+        - TARGET_BOX=LINUX64
+        - BTYPE="BINARY=64 INTERFACE64=1"
+
+    - <<: *test-ubuntu
+      addons:
+        apt:
+          packages:
+            - gcc-multilib
+            - gfortran-multilib
+      env:
+        - TARGET_BOX=LINUX32
+        - BTYPE="BINARY=32"
+
+    - stage: test
+      addons:
+        apt:
+          packages:
+            - binutils-mingw-w64-x86-64
+            - gcc-mingw-w64-x86-64
+            - gfortran-mingw-w64-x86-64
+      script:
+        - make QUIET_MAKE=1 DYNAMIC_ARCH=1 TARGET=NEHALEM NUM_THREADS=32 $BTYPE
+      env:
+        - TARGET_BOX=WIN64
+        - BTYPE="BINARY=64 HOSTCC=gcc CC=x86_64-w64-mingw32-gcc FC=x86_64-w64-mingw32-gfortran"
+
+# whitelist
+branches:
+  only:
+    - master
+    - develop
 
 notifications:
   webhooks:
@@ -7,32 +70,3 @@ notifications:
     on_success: change  # options: [always|never|change] default: always
     on_failure: always  # options: [always|never|change] default: always
     on_start: never     # options: [always|never|change] default: always
-
-compiler:
-  - gcc
-
-env:
-  - TARGET_BOX=LINUX64 BTYPE="BINARY=64"
-  - TARGET_BOX=LINUX64 BTYPE="BINARY=64 USE_OPENMP=1"
-  - TARGET_BOX=LINUX64 BTYPE="BINARY=64 INTERFACE64=1"
-  - TARGET_BOX=LINUX32 BTYPE="BINARY=32"
-  - TARGET_BOX=WIN64 BTYPE="BINARY=64 HOSTCC=gcc CC=x86_64-w64-mingw32-gcc FC=x86_64-w64-mingw32-gfortran"
-
-before_install:
- - sudo apt-get update -qq
- - sudo apt-get install -qq gfortran
- - if [[ "$TARGET_BOX" == "WIN64" ]]; then sudo apt-get install -qq binutils-mingw-w64-x86-64 gcc-mingw-w64-x86-64 gfortran-mingw-w64-x86-64; fi
- - if [[ "$TARGET_BOX" == "LINUX32" ]]; then sudo apt-get install -qq  gcc-multilib gfortran-multilib; fi
-
-script: 
- - set -e
- - make QUIET_MAKE=1 DYNAMIC_ARCH=1 TARGET=NEHALEM NUM_THREADS=32 $BTYPE
- - if [ "$TARGET_BOX" == "LINUX32" ] || [ "$TARGET_BOX" == "LINUX64" ]; then make -C test DYNAMIC_ARCH=1 TARGET=NEHALEM NUM_THREADS=32 $BTYPE; fi
- - if [ "$TARGET_BOX" == "LINUX32" ] || [ "$TARGET_BOX" == "LINUX64" ]; then make -C ctest DYNAMIC_ARCH=1 TARGET=NEHALEM NUM_THREADS=32 $BTYPE; fi
- - if [ "$TARGET_BOX" == "LINUX32" ] || [ "$TARGET_BOX" == "LINUX64" ]; then make -C utest DYNAMIC_ARCH=1 TARGET=NEHALEM NUM_THREADS=32 $BTYPE; fi
-
-# whitelist
-branches:
-  only:
-    - master
-    - develop
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,7 +87,10 @@ jobs:
         - TARGET_BOX=LINUX64_MUSL
         - BTYPE="BINARY=64"
 
-    - <<: *test-alpine
+    # XXX: This job segfaults in TESTS OF THE COMPLEX LEVEL 3 BLAS,
+    # so it's "allowed to fail" for now (see allow_failures).
+    - &test-alpine-openmp
+      <<: *test-alpine
       env:
         - TARGET_BOX=LINUX64_MUSL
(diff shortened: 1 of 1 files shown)


jirutka:
~~Travis killed one job due to too long output, because gcc generates tons of warnings on your code…~~ (EDIT: I was wrong, it actually segfaulted)

brada4:
How much warnings are too much? Will halving their volume help?

martin-frbg:
Jobs being killed because of overly long output has not been a problem with the existing travis setup as far as I know - is this some consequence of your build environment (different compiler version etc) or did you use additional CFLAGS ? BTW the oversight from #1252 would easily have been caught with a judiciuosly placed #undef __GLIBC_PREREQ, in my opinion the main problem was that a misunderstanding between xianyi and me led to it being merged right before the release when I wanted to have it on hold until later.

jirutka:
> How much warnings are too much? Will halving their volume help?

Eh, any warnings? gcc doesn’t print them just for fun… But about Travis, it has limit 4 MiB for output on stdout+stderr per job.

The right solution is to fix code, possible workaround is to turn off some of these warnings.

> Jobs being killed because of overly long output has not been a problem with the existing travis setup as far as I know - is this some consequence of your build environment (different compiler version etc) or did you use additional CFLAGS ?

When you look at some of your past jobs, for example [this one](https://travis-ci.org/xianyi/OpenBLAS/jobs/257490686), you’re already on the edge of the limit:

> This log is too long to be displayed. Please reduce the verbosity of your build or download the raw log.

Preparing Alpine environment adds some more text into the log ~~and it just flipped over the limit~~ (EDIT: no, it segfaulted) (on Alpine job with `USE_OPENMP=1`). You will run into this issue sooner or later even with your existing jobs.

> BTW the oversight from #1252 would easily have been caught with a judiciuosly placed #undef __GLIBC_PREREQ…

Of course, as all possible errors… It happened and got into the release without notice because you don’t regularly test against musl libc. That’s why I opened this PR to prevent such mistakes in future.

martin-frbg:
From a quick look at the failed job, a big number of them appear to be the recently introduced gcc indentation warnings, complaints about potentially uninitialized variables that would need to be addressed in netlib lapack and a bunch of harmless pointer type issues that I believe may even be platform-specifc. Suggest to turn off at least the indentation warnings for now.

7heo:
"[...] harmless pointer [...]" — @martin-frbg 

Best joke ever. 🤣 Thanks for making my Friday mate.

martin-frbg:
@7heo care to prove otherwise ? These superficially mismatching casts have been there forever but there are real problems elsewhere in the code that need fixing.

7heo:
@martin-frbg memory management **is** tricky and shouldn't be dealth with lightly. the least you can do is cast your stuff explicitely so it reads what it actually does.

jirutka:
I was wrong, the failing job (musl, `USE_OPENMP=1`) hasn’t been killed for too long output, it segfaulted in tests.\*

https://travis-ci.org/xianyi/OpenBLAS/jobs/258570695#L6743:
```
...
 cblas_ctrmm  PASSED THE TESTS OF ERROR-EXITS
 cblas_ctrmm  PASSED THE COLUMN-MAJOR COMPUTATIONAL TESTS (  2592 CALLS)

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x7f53526ef3ef in ???
#1  0x7f53526ee9e5 in ???
#2  0x7f5352a00f85 in ???
Segmentation fault (core dumped)
make[1]: *** [Makefile:79: all3] Error 139
make: *** [Makefile:119: tests] Error 2
```

\* _Travis marks failed job with a cross when some command in script exited with non-zero status or with exclamation mark when some error occurred. Since the log output exceeded size for being displayed on web page (“This log is too long to be displayed.”), I just quickly assumed that it exceeded 4MiB limit._

jirutka:
We set `USE_OPENMP=0` in Alpine package due to gcc bug [60035](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=60035). I don’t know if it’s still relevant though. I’m gonna mark this particular job as allowed to fail, I’m not really keen to investigate what exactly is wrong.

7heo:
I'm gonna quote Natanael Copa's FOSDEM presentation (first slide):

> ## The good things with musl
>
> [...]
> - musl tries to be (annoyingly) correct
> - when patch is needed, code gets better
>
> [...]

jirutka:
Your existing job with `USE_OPENMP=1` _sometimes_ fails even on Ubuntu Precise, but with different error, see https://travis-ci.org/xianyi/OpenBLAS/jobs/258662730. It may be related to the change from VM to container environment (read https://github.com/xianyi/OpenBLAS/pull/1255#issue-246212393); you can switch it back to VM (i.e. exactly the same environment as you currently have) with `sudo: true` on global or job level. However, it IMHO just confirms that there’s surely something wrong in the code…

I’ve also added job with `NO_AFFINITY=1` which reveals bug #1252 and confirms that it’s _not_ fixed in develop branch (#1249).

The job building/testing OpenBLAS with `USE_OPENMP=1` in Alpine environment (i.e. musl libc) is marked as “allowed to fail”, i.e. it’s failure does not affect result of whole build (you can see it [here](https://travis-ci.org/jirutka/OpenBLAS/builds/258648695)).

I consider this PR finished and ready to merge. CI failure on this PR is not a problem of this PR, but existing problems.

martin-frbg:
ctrmm issue may be the same issue as #1089 (worked around in #1254), at least that test triggers the same debug printf (in gbmv_thread.c) in my working copy. Will retest when I am more awake.

brada4:
@jirutka - your finger-pointing does not help. It is your code and you are a developer by modifying source files.
trmm test happens to be last in row, corruption can happen anywhere before. blank backtrace is of no value, even if coming from semantically picky library

jirutka:
@brada4 What finger-pointing? What are you talking about? I'm not OpenBLAS developer, I've only contributed this PR, in a hope to prevent future breakages of compatibility with non-glibc systems.

brada4:
musl does not support complex numbers. Plese be so kind and generate backtrace with function labels or work around musl shortage yourself.

martin-frbg:
@brada4 by all means let's look into that, but as far as I know OpenBLAS will work around lack of support for complex numbers by creating structs etc (which is where I believe the pointer warnings are from). However the ctrmm issue appears to be real and not limited to musl - will be interesting to learn if #1254 fixes it.

brada4:
Slightly worse - http://wiki.musl-libc.org/wiki/Mathematical_Library#Complex - probably needs some ifdef wrapper to force struct complex.

xianyi:
Hi all,

Because I access travis-ci very slow, I cannot see the full log in travis-ci. I see some error case in travis-ci. Is it expected?

martin-frbg:
One is "ld" getting killed with an actual SIGKILL - this is probably what jirutka attributed to travis aborting when the log gets too big (tons of warnings about implicit float/double conversions in LAPACK etc). The next is "my" cgroups patch breakage (musl does not define __GLIBC_PREREQ) so should hopefully go away with #1257. Guess I should take the courage to merge that now, and re-basing this PR afterwards would probably be sufficient to trigger a re-run of the travis checks. (With the xBMV overrun fix already merged, there should be no test failures remaining)

martin-frbg:
Hmm. Still getting segmentation faults in the run you had marked as "allowed to fail" despite #1254, wonder if it would be possible to generate a backtrace for that.
All activity on the gcc/libgomp bug ticket you mentioned seems to have ceased in late 2014 - I wonder if
the pitfalls of the combination of fork and OMP in a gcc-built program should be documented in the wiki somewhere ?

martin-frbg:
@jirutka re-enable the "allow_failures" if you like - I'm not sure if the commit would end up getting credited to me if I was the last developer to meddle with it. It seems I botched at least parts of the xbmv fix anyway, introducing accuracy errors.Perhaps I need to stop coding for a while until things settle down for me. At least the GLIBC_PREREQ problem seems to be gone now.

jirutka:
> Because I access travis-ci very slow, I cannot see the full log in travis-ci.

Travis is _very_ slow in rendering very long build log. However, you can view it also in raw (as plain text), that’s fast. Just click on “Raw log” above the rendered log (on job page).

If you can’t get there due to slow JS, you can also access it directly. On build page, look at link of particular job, e.g. `https://travis-ci.org/xianyi/OpenBLAS/jobs/259694421`. Then download `https://api.travis-ci.org/jobs/<ID>/log.txt?deansi=true` with curl or web browser, e.g. https://api.travis-ci.org/jobs/259694421/log.txt?deansi=true.

jirutka:
> wonder if it would be possible to generate a backtrace for that.

I’ve prepared you a VM with Alpine and prepared setup for building OpenBLAS, but GitHub doesn’t return any SSH keys for your account https://github.com/martin-frbg.keys. I don’t understand why, this should work for all accounts. o.O

> …re-enable the "allow_failures" if you like…

I’ve removed your commit disabling `allow_failures` and rebased this PR against the current develop branch. All jobs except `LINUX64_MUSL USE_OPENMP=1` (the one “allowed to fail”) now pass!

BTW It didn’t make sense to disable it, the point of `allow_failures` is that Travis runs the listed jobs, correctly report the job status, just doesn’t mark whole build as failed when some of `allow_failures` job fail.

martin-frbg:
1) probably because I do not have a local repository linked to my github account, strange as it may sound I still upload changed files in the browser or use the primitive editor on github for changes.
2) thought as much but just for triggering a rerun of travis this change seemed as good as any (and whitespace changes or deleting and restoring a character did not register for good reason)
(3 seems my panic revert of 1254 was overkill, as it looks now it is a single mistake in gbmv_thread.c
that caused the sudden "loss of precision" errors. My plan for now is to merge your PR tomorrow (if nobody complains), then push the fixed 1254PR and see if it makes the "allowed failure" case pass.

martin-frbg:
Key added now, thanks for setting up the VM. 
The "allowed failure" in cblas_ctrmm appears to be unrelated to the gbmv errors, helgrind shows lots of thread races apparently originating from missing locks in blas_server_omp. Unfortunately valgrind stumbles over the inline assembly when building for HASWELL, a NEHALEM build raises no issues (on a 
glibc system) that would explain an eventual segfault. Edit: helgrind is known to be confused by libgomp, so that one seems to be a red herring as well.

jirutka:
Hm, I can’t reproduce the error in `LINUX64_MUSL USE_OPENMP=1` ctest on the VM or my building machine. It seems that it’s somehow related to Travis environment, maybe kernel version or hardware?

brada4:
Was it chroot inside Ubuntu LTS or ???

PR #1265, opened 2017-08-02
arm64: make iamax kernel buildable by clang

pawosm-arm:
Here's the patch we had to make to get OpenBLAS to build with clang (and flang) for the thunderx2 target. Basically, prevent the compiler unrolling this function, which contains inline asm with a label that is assumed "local".

### DIFF
--- a/kernel/arm64/iamax_thunderx2t99.c
+++ b/kernel/arm64/iamax_thunderx2t99.c
@@ -208,7 +208,8 @@ extern int blas_level1_thread_with_return_value(int mode, BLASLONG m, BLASLONG n
 #endif
 
 
-static BLASLONG iamax_compute(BLASLONG n, FLOAT *x, BLASLONG inc_x)
+static BLASLONG __attribute__((noinline)) iamax_compute(BLASLONG n, FLOAT *x,
+								BLASLONG inc_x)
 {
 	BLASLONG index = 0;
 


ashwinyes:
Thanks for pointing this out.

I didn't check clang compilation for thunderx2 while enabling threading for some of these level 1 APIS.

Request you to make similar change in the following files as well and re-submit the PR so that its fixed in all files.
kernel/arm64/casum_thunderx2t99.c
kernel/arm64/copy_thunderx2t99.c
kernel/arm64/dasum_thunderx2t99.c
kernel/arm64/dot_thunderx2t99.c
kernel/arm64/dznrm2_thunderx2t99.c
kernel/arm64/iamax_thunderx2t99.c
kernel/arm64/izamax_thunderx2t99.c
kernel/arm64/sasum_thunderx2t99.c
kernel/arm64/scnrm2_thunderx2t99.c
kernel/arm64/zasum_thunderx2t99.c
kernel/arm64/zdot_thunderx2t99.c

But, I think the correct fix is to change the local named labels in the inline asm to numbered  labels. I will try to do this when I get time.

ashwinyes:
Created the PR https://github.com/xianyi/OpenBLAS/pull/1266 for fixing the labels.

Please check whether this fixes your clang compilation

pawosm-arm:
re #1266: It seems to fix it. We can build it and it passes all of the tests in the OPB Makefile at least.

martin-frbg:
Closing this PR as superseded by 1266, thanks for your effort all the same.

PR #1266, opened 2017-08-02
THUDNERX2T99: Fix clang compilation

ashwinyes:


pawosm-arm:
It seems to fix it. We can build it and it passes all of the tests in the OPB Makefile at least.

martin-frbg:
Committing then, thanks both of you.

PR #1268, opened 2017-08-06
Travis: Add jobs building with clang and disable job `LINUX64_MUSL USE_OPENMP=1`

jirutka:
/cc @martin-frbg


martin-frbg:
#1256 also has some travis strangeness - a SIGKILL to "ld" in the middle of the build, so perhaps something in the travis setup has become unstable.

PR #1277, opened 2017-08-10
Make: fix installation instructions

christoph-conrads:
The installation instructions as shown after successfully compiling
OpenBLAS are wrong because this arguments used during compilation have
to be provided to Make again.


martin-frbg:
I wonder if it is possible to use something like $MAKEFLAGS here to insert the actual values instead of the <args> placeholder plus explanation ?

PR #1281, opened 2017-08-18
fix detection of generic ARMv8 CPUs

sharkcz:
If I can see right on our systems, then all ARMv8 CPUs have "8" as value for "CPU architecture" in /proc/cpuinfo. This change unblocks building OpenBLAS on APM Mustang system (X-Gene CPU).

### DIFF
--- a/cpuid_arm64.c
+++ b/cpuid_arm64.c
@@ -142,7 +142,7 @@ int detect(void)
   	if(p != NULL)
 	{
 
-		if (strstr(p, "AArch64"))
+		if ((strstr(p, "AArch64")) || (strstr(p, "8")))
 		{
 			return CPU_ARMV8;
 


sharkcz:
In theory we could remove the AArch64 string from the check, but rather don't break anything.

brada4:
My android tv has number 8 there, as minimum this improves detection.

martin-frbg:
Bit of googling shows that the transition from "AArch64" to "8" in the arm64 cpuinfo output happened in late 2014/early 2015 so probably too recent to simply drop the old string already. Looks good to me. (Interestingly xianyi comitted the equivalent to cpuid_arm.c a year ago (for armv8 32bit mode) but apparently did not check cpuid_arm64.c , and conversely cpuid_arm.c does not appeat to know how to handle a plain "7" (or "6") yet)

brada4:
Plain 7 exists:
https://www.raspberrypi.org/forums/viewtopic.php?f=63&t=146742

martin-frbg:
Yep, that's where I got the idea from (and "6" exists as well in the new cpuinfo scheme)

PR #1293, opened 2017-09-07
More canonical installation on Cygwin

embray:
* The DLL is named `cygopenblas.dll`, not `libopenblas.dll`--this is in accordance with the convention for all DLLs linked against Cygwin.
* The import lib (still called `libopenblas.dll.a`) is installed in `$(PREFIX)/lib`.  Although this is not strictly necessary on Cygwin (gcc can link directly with the DLL) it is still useful to have in some cases.

See https://trac.sagemath.org/ticket/23797


brada4:
Can you share screenshot of imports/exports from cygwin emulation lib from depends.exe of cygwin-built blas library?
You should be able to link/use binary (mingw-built) libopenblas.dll too

embray:
That's an interesting question, though I'm curious what it's relevant to?

From cygwin1.dll:

```
__assert_func
__cxa_atexit
__getreent
_impure_ptr
atan2
atan2f
atoi
cabs
cabsf
calloc
cexp
cexpf
cos
cosf
csin
csinf
csqrt
csqrtf
cygwin_detach_dll
cygwin_internal
dll_dllcrt0
exit
exp
expf
fputs
free
log
log10
log10f
logf
lround
lroundf
malloc
memcmp
memmove
memset
posix_memalign
pow
powf
printf
puts
realloc
sin
sinf
sqrt
```

Sure, one could still use the non-Cygwin lib. But this is in the context of building SageMath, where it's not uncommon to custom build many of its dependencies, and for the Cygwin port of Sage I'm working on it's better to have full consistency.  In any case, this PR doesn't change anything about how OpenBLAS is built on Cygwin--just how it's installed there.

brada4:
I hoped it was single crt call that could be ifdef-avoided, nope it turns out full libm...

embray:
Yup, the Cygwin DLL provides libm as well.

brada4:
Change is sane, sorry for bothering with curiousity.

embray:
It's fine--it's an interesting question. Thanks for merging.

PR #1316, opened 2017-09-28
Variable thread count for multi-threaded GEMMs

timmoon10:
When performing smallish GEMMs (M\~1000, N\~100, K\~100) on a 36-core system, I've found that I can achieve a significant speedup by reducing the number of OpenMP threads. It turns out that multi-threading is disabled if `(m < nthreads * SWITCH_RATIO) || (n < nthreads * SWITCH_RATIO)`, so we maintain parallel execution with no algorithmic changes if `nthreads` is changed to `min(m / SWITCH_RATIO, n / SWITCH_RATIO, nthreads)`. This pull request makes this change automatically and also clears some unused code that dates back to GotoBLAS.

### DIFF
--- a/benchmark/gemm.c
+++ b/benchmark/gemm.c
@@ -121,13 +121,15 @@ static void *huge_malloc(BLASLONG size){
 int main(int argc, char *argv[]){
 
   FLOAT *a, *b, *c;
-  FLOAT alpha[] = {1.0, 1.0};
+  FLOAT alpha[] = {1.0, 0.0};
   FLOAT beta [] = {0.0, 0.0};
-  char trans='N';
-  blasint m, n, i, j;
+  char transa = 'N';
+  char transb = 'N';
+  blasint m, n, k, i, j, lda, ldb, ldc;
   int loops = 1;
-  int has_param_n=0;
-  int l;
+  int has_param_m = 0;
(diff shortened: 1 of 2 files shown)


martin-frbg:
Looks good to me, but could you please provide some performance figures just to illustrate how significant "significant" is ? (And if I read your patch correctly, it should not cause a slowdown on less well endowed systems ?)

timmoon10:
Here are some scaling studies on a 36-core system (2 x Intel 18-core Xeon E5-2695 v4, gcc 4.9.3, TARGET=HASWELL USE_OPENMP=1 NUM_THREADS=36 INTERFACE64=0):

surface_gpgpu_sgemm

surface_gpgpu_dgemm

surface_gpgpu_cgemm

surface_gpgpu_zgemm

timmoon10:
Here are some experiments on an 8-core system (Intel Core i7, default build for OSX):

osx_sgemm

osx_dgemm

osx_cgemm

osx_zgemm

brada4:
There is another threading threshold in interface/gemm.c
Graph going down probably means move between L3 of different cores.

brada4:
I forgot to mention that you are on right path, fixed threading threshold as currently is great impairment to say OpenCV/SGEMM

Issue #1191, opened 2017-05-30
Race Condition in OpenBLAS on IBM OpenPower 8


grisuthedragon:
This issue is an follow up to #1071 which extends the problem and gives a clearer example. I consider a code computing the TILE-QR decomposition using OpenMP 4 task depend paralleization. The code, including a makefile to reproduce the errors presented here is available at: [https://github.com/grisuthedragon/openblas-tileqr](https://github.com/grisuthedragon/openblas-tileqr). Details about the algorithm can be found [here](http://buttari.perso.enseeiht.fr/docs/crgc_mcore.pdf). 

System Details:
 * CentOS 7.3 litle endian
 * IBM Power8 (SL822C) (20 Cores, 160 Threads) 
 * GCC 5.4.1

For the first experiment I compiled OpenBLAS in single-thread mode using 
```
make USE_OPENMP=0 USE_THREAD=0 NO_CBLAS=1 NO_LAPACKE=1 
``` 
and compiled and executed the example code via 
```
gfortran -o tileqr_serial -fopenmp tileqr.f90 libopenblas_serial.a -lgomp 
OMP_NUM_THREADS=20 ./tileqr_serial 5120 5120 256 32
````
and I obtained: 
```
LAPACK   -- Time:    8.576    Gflops/s:   20.867
Tiled    -- Time:    0.930    Gflops/s:  192.450
 
Max diff between R from tiled   and LAPACK is: 0.2E+09
```
which is the wrong result because the code has the provide the same result as LAPACK. 

Compiling OpenBLAS with 
```
make USE_OPENMP=1 USE_THREAD=1 NO_CBLAS=1 NO_LAPACKE=1 
```
and compiled and executed the example code via 
```
gfortran -o tileqr_openmp -fopenmp tileqr.f90 libopenblas_openmp.a -lgomp 
OMP_NUM_THREADS=20 ./tileqr_openmp 5120 5120 256 32
````
and I obtained: 
```
LAPACK   -- Time:    6.553    Gflops/s:   27.310
Tiled    -- Time:    1.120    Gflops/s:  159.760
 
Max diff between R from tiled   and LAPACK is: 0.1E+04
```
which is also wrong. 

Finally I used the old threading implementation for level-3 operations by compiling OpenBLAS via: 
```
make USE_OPENMP=1 USE_THREAD=1 USE_SIMPLE_THREADED_LEVEL3=1 NO_CBLAS=1 NO_LAPACKE=1
```
and for the example
[...]

PR #1317, opened 2017-09-28
Save and restore VSX registers

### LINKED ISSUE #1263: OpenPower nonvolatile registers not saved/restored
Hello,
We have found a case where the OpenPOWER nonvolatile registers vsr52-vsr63 are not saved and restored across calls to several assembly language functions.
 
According to the latest Power ABI (rev. 1.4 dated May 10, 2017) found at:
 
https://openpowerfoundation.org/?resource_lib=64-bit-elf-v2-abi-specification-power-architecture
 
The vector registers vr20-vr31 (see Table 2.22) are nonvolatile (callee save). The vector registers vr0-vr31 are overlaid onto  Vector Scalar Registers (VSR) vsr32-vsr63 (see Fig 2.17). Thus vsr52-vsr63 are also nonvolatile, and for a routine to use any of these registers requires a save-restore sequence within the assembly code.
 
Within the OpenBLAS-0.2.[19,20]/kernel/power source code there are 12 files that reference these nonvolatile registers (vsr52-vsr63):
 
cgemm_macros_8x4_power8.S
ctrmm_macros_8x4_power8.S
dgemm_macros_16x4_power8.S
dgemm_ncopy_macros_4_power8.S
dgemm_tcopy_macros_16_power8.S
dtrmm_macros_16x4_power8.S
dtrsm_macros_LT_16x4_power8.S
sgemm_macros_16x8_power8.S
strmm_macros_16x8_power8.S
zgemm_macros_8x2_power8.S
zgemm_tcopy_macros_8_power8.S
ztrmm_macros_8x2_power8.S
 
These 12 files are used in a subset of the .S files which include the assembly macros within these files. These built routines do correctly save and restore fpr14-fpr31, the 64-bit registers which overlay the left half of vsr14-vsr31 (and are specified in Table 2.20 of the ABI document as nonvolatile). However, vsr52-vsr63 are NOT saved and restored. This  omission means that if any of vr20-vr31 (or by different names vsr52-vsr63) are defined and then used after a call which contains any of the 12 assembly files listed above, corruption can, and indeed has, occurred.  We have tracked down instances where these calls have overwritten nonvolatile registers; these cases resulted in a segfault in our test program.
 
We believe a solution is to save and restore vsr52-vsr63 (or vr30-vr31) in the same manner, and roughly in the same locations, as is currently done for fpr14-fpr31.
 
Please feel free to contact us for any additional details or further explanation.

Thanks!
Eric

martin-frbg:
Expected to fix #1263 (the remaining issues from #1078)

### DIFF
--- a/kernel/power/cgemm_kernel_8x4_power8.S
+++ b/kernel/power/cgemm_kernel_8x4_power8.S
@@ -82,15 +82,15 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #endif
 
 #ifdef __64BIT__
-#define STACKSIZE 32000
-#define ALPHA_R_SP 296(SP)
-#define ALPHA_I_SP 304(SP)
-#define FZERO	312(SP)
+#define STACKSIZE 32196
+#define ALPHA_R_SP 296+196(SP)
+#define ALPHA_I_SP 304+196(SP)
+#define FZERO	312+196(SP)
 #else
-#define STACKSIZE 256
-#define ALPHA_R_SP 224(SP)
-#define ALPHA_I_SP 232(SP)
-#define FZERO	240(SP)
+#define STACKSIZE 456
+#define ALPHA_R_SP 224+200(SP)
+#define ALPHA_I_SP 232+200(SP)
+#define FZERO	240+200(SP)
 #endif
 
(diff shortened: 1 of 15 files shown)


martin-frbg:
Committing now after seeing that it solves #1191 in addition of passing all built-in tests and BLAS-Tester

Issue #1270, opened 2017-08-07
OpenBLAS has issues solving triangular matrices on win32


ghost:
Using the 32 bit openblas causes timeouts while running the scipy tests:

ghost:
https://ci.appveyor.com/project/scipy/scipy/build/1.0.40/job/7vqlq835677r4xde

martin-frbg:
Do I read that correctly - you are using a binary distribution of 0.2.19 from some third party site (which could be one of the sourceforge mirrors sites, but then again could be something else) ? Could you retest with the current "develop" snapshot which has a bunch of fixes vs. 0.2.19 (or at least build 0.2.19 yourself on your platform to see if it passes the built-in tests) ? In any case I guess a minimal example run on a well-defined system would be preferable to a big unedited dump from an appveyor vm.

ghost:
That was built 12 days ago: https://github.com/scipy/scipy/pull/7616#issuecomment-318169023

ghost:
>In any case I guess a minimal example run on a well-defined system would be preferable to a big unedited dump from an appveyor vm.

Is there any way to turn on debug output?

ghost:
excuse me; the URL was not apparently correct. Retrying with 2.20 now.

martin-frbg:
not sure if 0.2.20 will build in your environment (depending on how and where you build - a cross-compile on linux like the binaries provided on sourceforge will probably go okay. the issue is a botched commit that breaks things on non-glibc systems, waiting for xianyi to release 0.2.21) which is why I suggested taking a snapshot of "develop" (easily done by clicking on the green "Clone or download" button on the "Code" tab here). Normal build from source should run a few tests at the end automatically (though as a Linux guy I am not completely sure it does this on Windows as well.

martin-frbg:
Thanks for the link to the scipy issue. From reading that, it seems your tests were (at least at some point in the history of that PR) already done with 0.2.20 (?) 
Do you still believe the problematic behaviour is with potrf specifically ?

ghost:
> Thanks for the link to the scipy issue. From reading that, it seems your tests were (at least at some point in the history of that PR) already done with 0.2.20 (?)

That's what I thought, but HEAD had 2.19. I updated to 2.20 just now.

> Do you still believe the problematic behaviour is with potrf specifically ?

[...]

PR #1318, opened 2017-09-30
Add trivial smoketest for xpotrf

pv:
Add a trivial test for the *potrf routines to utest.

We've now been burned a few times by "bad" openblas builds (on some platforms only) that had broken cholesky factorizations. It would be useful to catch those already when building openblas itself, so here is a PR that adds a trivial smoketest that should catch those.

### DIFF
--- a/utest/test_potrs.c
+++ b/utest/test_potrs.c
@@ -394,3 +394,150 @@ CTEST(potrf, bug_695){
     CTEST_ERR("%s:%d  got NaN", __FILE__, __LINE__);
   }
 }
+
+
+// Check potrf factorizes a small problem correctly
+CTEST(potrf, smoketest_trivial){
+  float A1s[4] = {2, 0.3, 0.3, 3};
+  double A1d[4] = {2, 0.3, 0.3, 3};
+  openblas_complex_float A1c[4] = {
+    openblas_make_complex_float(2,0),
+    openblas_make_complex_float(0.3,0.1),
+    openblas_make_complex_float(0.3,-0.1),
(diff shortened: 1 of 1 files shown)


matthew-brett:
Guys - would you mind testing this patch on Windows 32-bit?  That is where we are seeing the numpy failure.

pv:
The problematic case does not require just windows 32-bit, but also the specific compiler and maybe compiler flag choice, which I'm not sure is useful for openblas devs to spend time on. I tested myself that this test case correctly fails when using the "bad" build binaries, and it passes for correctly working potrf routines (as indicated by appveyor&travis).

matthew-brett:
At the moment, though, the only situation we know this fails, is on Windows 32-bit, with the given compiler / flags and so on; so if this test passes with a standard mingw build, on Windows 32-bit, then we (numpyers) can try and track down which of our tweaks is causing the problem.

pv:
@matthew-brett: the "standard" mingw-w64/32bit compiler shipped on appveyor produces problem-free binaries (indeed, remember that we already had this discussion vs. scipy). If necessary, let's discuss that elsewhere since it's probably not so relevant for openblas development.

Anyway, even if the failing compiler setup is rare, it would still be useful if the standard openblas test suite would have at least minimal testing of the potrf routines.

matthew-brett:
Ah - no - sorry - I had not registered that https://github.com/xianyi/OpenBLAS/issues/1270 was the same issue.

martin-frbg:
I'm all for detecting errors early, thanks :-)

Issue #1326, opened 2017-10-11
sdot yields wrong results for odd sizes > 2^24 and all sizes > 2^29


kohr-h:
When I compute a dot product of a vector of all ones with itself, the result should be the size of the vector. This starts to no longer hold at size 2^24 + 1 as the following program shows:
```C
// file tmp.c
#include <stdio.h>
#include <stdlib.h>
#include <assert.h>
#include "cblas.h"

int main(int argc, char **argv) {
    int exp = atoi(argv[1]);
    int offset = atoi(argv[2]);
    size_t size = (1 << exp) + offset;
    float *a = (float *) malloc(size * sizeof(float));
    assert(a != NULL);
    for (size_t i=0; i < size; ++i)
        a[i] = 1;

    float dot = cblas_sdot(size, a, 1, a, 1);
    printf("dot = %f\n", dot);
    printf("size = %zu = (1 << %i) + (%i)\n", size, exp, offset);

    return 0;
}
```
I get
```
$ ./tmp 24 0
dot = 16777216.000000
size = 16777216 = (1 << 24) + (0)

$ ./tmp 24 1
dot = 16777216.000000
size = 16777217 = (1 << 24) + (1)

$ ./tmp 24 2
dot = 16777218.000000
size = 16777218 = (1 << 24) + (2)

$ ./tmp 24 3
dot = 16777220.000000
size = 16777219 = (1 << 24) + (3)

$ ./tmp 24 4
dot = 16777220.000000
size = 16777220 = (1 << 24) + (4)

$ ./tmp 24 3
dot = 16777220.000000
size = 16777221 = (1 << 24) + (5)
```
and then, for sizes larger than 2^29 the result just stays constant at 2^29:
```
$ ./tmp 29 1
dot = 536870912.000000
size = 536870913 = (1 << 29) + (1)

$ ./tmp 29 2
dot = 536870912.000000
size = 536870914 = (1 << 29) + (2)

$ ./tmp 29 4
dot = 536870912.000000
size = 536870916 = (1 << 29) + (4)

$ ./tmp 30 -1
dot = 536870912.000000
size = 1073741823 = (1 << 30) + (-1)

$ ./tmp 30 0
dot = 536870912.000000
size = 1073741824 = (1 << 30) + (0)
```

I'm using the latest ~~`master`~~ `develop` version of OpenBLAS, and I statically link the library to not accidentally dynamically link an older version:
```
[...]

PR #1329, opened 2017-10-16
(Trivial) optimized dsdot implementation for HASWELL

martin-frbg:
For #1326

### DIFF
--- a/kernel/x86_64/KERNEL.HASWELL
+++ b/kernel/x86_64/KERNEL.HASWELL
@@ -24,6 +24,8 @@ DDOTKERNEL = ddot.c
 CDOTKERNEL = cdot.c
 ZDOTKERNEL = zdot.c
 
+DSDOTKERNEL = sdot.c
+
 SAXPYKERNEL = saxpy.c
 DAXPYKERNEL = daxpy.c
 CAXPYKERNEL = caxpy.c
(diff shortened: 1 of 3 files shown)


kohr-h:
Would the same change in the microkernel work for other architectures, too? They all look very similar, apart from the number of floats they process at a time.

martin-frbg:
I expect so, but did not want to mess with all of them at once.

kohr-h:
Sure. I'm asking because in the overarching Numpy issue, people won't be happy with Haswell support only. And by "support" I mean "optimized microkernel". 
Although it seems to me that the part in the microkernel that is removed for dsdot really doesn't make any difference in runtime. Does that agree with your experience?

martin-frbg:
The small optimization in the microkernel may well be immesurable, the important part is the declaration in KERNEL.HASWELL that tells the build system to use the modified sdot.c file for dsdot as well instead of falling back to the slow generic C. This should work equally well for all the other x86_64 cpus that have their own sdot microkernels, and only a little bit slower for those that fallback to the C implementation of sdot_kernel_16. And it probably would work also for arm, power etc. once their local implementations of sdot.c are adapted.

kohr-h:
That's what I was guessing, too. Most of the gain for little work (at least for one platform :-)).

PR #1338, opened 2017-10-30
cmake: Allow to build without fortran

xerus:
Hi, I had some issues to compile without fortran. Adding these checks helped.


martin-frbg:
The travis timeout is probably unrelated, but offhand I do not think building the (cblas) ctest should depend on fortran

brada4:
Neither setting fortran options invokes FC

xerus:
both subprojects (test, ctest) uses Fortran, no?
Anyway a similar check is already on the develop branch. I just wanted to build from master.

PR #1345, opened 2017-11-02
WIP: [appveyor] add flang

ghost:


ghost:
I think we may be suffering a case of the second underscore.

isuruf:
@xoviat, I forgot to mention that I'm working on it here, https://github.com/isuruf/OpenBLAS/tree/flang

PR #1346, opened 2017-11-04
Add flang

ghost:


cmake/lapack.cmake
```diff
@@ -132,7 +132,9 @@ set(CLASRC
   DEPRECATED/cgegs.f  DEPRECATED/cgegv.f  cgehd2.f cgehrd.f cgelq2.f cgelqf.f
   cgels.f  cgelsd.f cgelss.f DEPRECATED/cgelsx.f cgelsy.f cgeql2.f cgeqlf.f cgeqp3.f
   DEPRECATED/cgeqpf.f cgeqr2.f cgeqr2p.f cgeqrf.f cgeqrfp.f cgerfs.f
-  cgerq2.f cgerqf.f cgesc2.f cgesdd.f  cgesvd.f
+  cgerq2.f cgerqf.f cgesc2.f cgesdd.f  cgesvd.f cgejsv.f cgesvdx.f cgesvj.f
+  cgetrf2.f cgges3.f cggev3.f cgghd3.f cggsvd3.f cggsvp3.f cpotrf2.f csysv_aa.f
+  csytrs_aa.f
```
isuruf: @martin-frbg, `CLASRC` is outdated. There are more functions missing which leads to symbol errors. Do you update this list by hand or do you use a script?

martin-frbg: This is updated by hand, and cmake is still a second-class citizen here unfortunately. (The Makefile is correct however, so these should not be removed from exports/gensymbol)

ghost:
@isuruf Can you add me as a collaborator?

exports/gensymbol
```diff
@@ -3053,34 +3053,34 @@
     ## @(MATGEN_OBJ) from `lapack-3.4.1/lapacke/src/Makefile`
     ## Not exported: requires LAPACKE_TESTING to be set and depends on libtmg
     ##               (see `lapack-3.4.1/TESTING/MATGEN`).
-    LAPACKE_clatms,
```
isuruf: These `MATGEN` objects are not built by default in `lapack` right?

martin-frbg: The Makefiles do build them by default (while the netlib cmakefiles do have a separate BUILD_TESTING option as far as I recall). The decision to include them appears to date back to May 2014, 219bcb1

ghost:
Could be that `set_target_properties(openblas PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS TRUE)` is needed.

ghost:
Folks, I think we've got a build working here.

ghost:
@martin-frbg Obstacles to merging?

CMakeLists.txt
```diff
@@ -160,6 +160,8 @@ endif()
 # Set output for libopenblas
 set_target_properties( ${OpenBLAS_LIBNAME} PROPERTIES RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/lib)
 set_target_properties( ${OpenBLAS_LIBNAME} PROPERTIES LIBRARY_OUTPUT_NAME_DEBUG "${OpenBLAS_LIBNAME}_d")
+set_target_properties( ${OpenBLAS_LIBNAME} PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS TRUE)
```
isuruf: If this is needed, then that means `exports\gensymbol` is outdated

appveyor.yml
```diff
@@ -24,23 +24,31 @@ skip_commits:
...
   matrix:
+    #- COMPILER: clang-cl
+    #  DYNAMIC_ARCH: ON
+    #  WITH_FORTRAN: no
     - COMPILER: clang-cl
-      DYNAMIC_ARCH: ON
-    - COMPILER: clang-cl
-    - COMPILER: cl
+      WITH_FORTRAN: yes
+    #- COMPILER: cl
```
isuruf: We should probably uncomment this

CMakeLists.txt
ghost: `zpotrs_` was not exported, and I strongly suspect that most of the other routines weren't either. The question is whether this is preferable to `gensymbol`. As it reduces maintenance, I'd say yes.

isuruf: What's the difference between this and line 156?

ghost: Probably the fact that the compiler is not MSVC (it's clang).

isuruf: CMake sets `MSVC` when `clang` is targeting `MSVC`

ghost: I don't know then. Obviously it made a difference though.

martin-frbg: To me the proposed changes to exports/gensymbol still look completely unrelated to the PR topic.

martin-frbg: And I see the question of running gensymbol vs setting WINDOWS_EXPORT_ALL_SYMBOLS was raised in https://github.com/xianyi/OpenBLAS/pull/1279 already, seemed one is not a full replacement for the other.

ghost:
@martin-frbg The changes to `gensym` are required. May I revert the last commit?

martin-frbg:
What worries me is that the changes to gensymbol appear to be required only for the specific task of getting OpenBLAS to build in a Windows/MSVC/cmake/clang environment while for all I know "the" (or perhaps "most") other platforms/builds are just fine without them. (Not that I would think that the functions you intend to disable are actually important, they just happen to have been available in recent versions so some may actually use them). 
Perhaps the actual "problem" - or difference from the better supported Makefile setup - is that lapacke.cmake is enumerating these functions as "set(MATGEN" without adding that list to the final sources ?

insertinterestingnamehere:
You all seem to have beat me to the punch on submitting a PR for this, but I went ahead and put up the changes that got things working for me as https://github.com/xianyi/OpenBLAS/pull/1350. Feel free to draw on those changes in whatever way will be most useful to you.

ghost:
@isuruf Can you add @insertinterestingnamehere to the repo?

isuruf:
I think #1350 is better than this PR. What this PR has more is the appveyor.yml changes which can be added after #1350 is merged.

ghost:
Closing as cleared.

PR #1349, opened 2017-11-06
[travis] update to trusty

ghost:
I understand you had problems with the sudo: false builds but trusty may have improvements.


ghost:
Unfortunately, the random failures still occur.

Issue #1357, opened 2017-11-14
A lot of warnings in -Wunused-but-set-variable and Wunused-variable


luotao1:
I grep our TeamCity log. There are 816 `-Wunused-but-set-variable` and  386 `-Wunused-variable` warnings come from OpenBLAS: https://github.com/PaddlePaddle/Paddle/issues/3281#issuecomment-344138280

Do you have any plans to fix these warnings? 
Looking forward to your reply!

martin-frbg:
While I can only speak for myself, I did not see them as an immediate priority problem so far given the lack of active developers on this (mainly) volunteer project. Also when this topic last came up months ago I did not want to touch the level3 blas code as it seemed another developer was actively working on the affected functions. Fixing the warnings should be straightforward if time consuming, but I would hate to clean up harmless declarations of currently unused things that actually may be hooks already put in place for later developments. (Though most likely the majority will simply be leftovers from cut-and-paste coding.)
Did you experience any problems in connection with these warnings, or is it just that this creates an unfavorable impression about the general quality of the code ?

martin-frbg:
Regarding the warnings arising from the fallback codes in kernel/generic, I guess I could offer the additional excuse that these are historic slips of the pen, written by Kazushige Goto a decade ago for what was then libGoto or GotoBLAS. So they have been there from the beginning of (our) time :-)

luotao1:
Thanks for your quick reply. 

> Did you experience any problems in connection with these warnings, or is it just that this creates an unfavorable impression about the general quality of the code 

We didn't experience problems, but it creates a very long log for our teamcity. 

>  Fixing the warnings should be straightforward if time consuming 

How about add `Wunused-but-set-variable` and `Wunused-variable` in flags such as CMAKE_CXX_FLAGS, which can avoid these warnings?

martin-frbg:
Suppressing the warnings will probably draw criticism from others for sweeping things under the rug...
At least for the "historic" codes in kernel/generic, the messages arise from varying macro expansion as often in OpenBLAS the same source file is compiled into multiple related functions for handling float/double/complex input etc. So "fixing" would mean adding a ton of `#ifdef`s rather than just commenting or removing truely unneeded lines. It is quite likely that this applies to the level3 blas codes as well, while at least the `j` and `at[]` variables in the x86_64 gemv and symv kernels appear to be genuinely useless. (Still time that could be spent on fixing actual bugs that lead to actual errors)

[...]

PR #1361, opened 2017-11-16
Fix zgeadd_k prototype

martin-frbg:
Third argument to zgeadd_k is a double everywhere else, causing a compiler warning as discussed in #1357


martin-frbg:
Travis experienced lockups that I doubt are related to the code.

martin-frbg:
Not sure what to make of the travis failures, will retry with a rebased tree.

PR #1363, opened 2017-11-18
Fix prototype of zgeadd_k

martin-frbg:
Third argument to zgeadd_k is a double everywhere else, causing a compiler warning as discussed in #1357


martin-frbg:
Pretty sure the travis timeouts are unrelated - still a bit worried about the windows build (job 7) as it fails reproducibly while the first three are random.

martin-frbg:
Closing for now

PR #1366, opened 2017-11-23
Update LAPACK to 3.8.0

martin-frbg:


martin-frbg:
Needs testers. ~~Also the relevant files in cmake/ are not yet updated~~.

insertinterestingnamehere:
The new CMake files work for me with flang and gfortran on Windows. Thanks for putting everything together!

PR #1368, opened 2017-11-26
Eliminate warnings

brada4:
One false buffer size warning from cppcheck
Unused const-s spotted by gcc7
Unused increment result spotted by clang4


brada4:
this is about half-way with warnings x86_64 only, there are still _16 kernels todo and all arch-specific left out and I will try at least popular cross targets, one a weekend.

PR #1373, opened 2017-11-29
README: Use the SVG Travis badge

kevinji:
Title is pretty self-explanatory.


martin-frbg:
Not that it really matters :-) But I see the other two icons are svgs already, thx.

PR #1403, opened 2017-12-23
Address few more warnings

brada4:
All "buffer" variables seem like padding for times of predictable memory allocators. Rest looks like small mistakes/ omissions. Still dead end variables lurk around compute kernels.


brada4:
This is last this year. So far I got rid of about half of gcc warnings, and about as much dead stores not seen by gcc. I will continue on until it starts to require brain work.

PR #1417, opened 2018-01-11
CMake: Use the correct library name on windows

ghost:
FindBLAS and FindLAPACK don't correctly detect the OpenBLAS library name on windows. I am unsure why this was originally set this way, but it has caused me some trouble.

Note that vcpkg patches this to remove its effects.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,11 +12,7 @@ set(OpenBLAS_VERSION "${OpenBLAS_MAJOR_VERSION}.${OpenBLAS_MINOR_VERSION}.${Open
 # Adhere to GNU filesystem layout conventions
 include(GNUInstallDirs)
 
-if(MSVC)
-set(OpenBLAS_LIBNAME libopenblas)
-else()
 set(OpenBLAS_LIBNAME openblas)
-endif()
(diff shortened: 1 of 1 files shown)


martin-frbg:
Is this a general problem, or peculiar to vcpkg ? 
I notice that vcpkg has some of its packages named _libpackagename_ , while others including openblas are just _packagename_ , does this perhaps create problems where other vcpkg packages or tools are looking for openblas.dll instead of libopenblas.dll ? Or does the Windows cmake build leads to something really silly like liblibopenblas.dll in certain situations ?

ghost:
The problem is that FindLAPACK and FindBLAS (from CMake) don't work if the library name is 'libopenblas' on Windows. It is not a general problem, but rather a problem that is specific to each Find* module that CMake defines.

brada4:
Indeed almost nobody prepends library name with lib on windows. I wonder if "make" should not be changed this way to generate directly consumable module too.

martin-frbg:
Thanks, I had not realized that FindLAPACK and FindBLAS are CMake modules. I intend to merge this if we do not get a reaction from xianyi soon.

PR #1434, opened 2018-01-26
CMake: Remove unused wall option when FC=flang

ghost:


martin-frbg:
Could you please clarify why "-Wall" is unused in this context ?

ghost:
Flang doesn't support that option.

martin-frbg:
That seemed unlikely to me, and google could not make up its mind if I was looking for flanges or flags on a wall. Do you happen to know if this is more than a temporary incompatibility (if not to say deficit) in flang, and does its presence actually stop the build ?

ghost:
I am a contributor to the flang project. Personally, I am not planning to add this and there are no open issues for this. Of course I cannot rule this out absolutely and say that if someone asks for it for some reason, I would actively block the feature.

ghost:
[Here is the file](https://github.com/flang-compiler/clang/blob/64043d5cec9fb02d1b0fd80c9f2c4e9e4f09cf8f/lib/Driver/ToolChains/Flang.cpp) that contains the options.

martin-frbg:
Not wanting to critisize flang, just wondering as -Wall seems to have become quite common among compilers (or I need to get out more)

ghost:
So it looks like flang has `-Minform=inform`, which enables all warnings. I suppose I could set that if you want?

ghost:
Or, I could ask whether other folks would be OK with adding Wall...

martin-frbg:
Make -Wall an alias for -Minform=inform ? But mostly I was just curious, sorry for dropping out suddenly but my laptop was tired and I was running out of battery...

martin-frbg:
So flang simply follows the PGI conventions for compiler options ? After a closer look at the various compiler-specific sections in Makefile.system it seems no effort has ever been made to provide consistent warning flags across all supported compilers. The only goal appears to have been to ensure that -Wall is set for the GNU compilers, so merging this now looks fine to me.

PR #1438, opened 2018-01-28
CMake: enable interprocedural optimization

ghost:


martin-frbg:
Does this really need bumping the cmake requirement all the way up to 3.9, which may pose an obstacle for users of at least some (enterprise) linux distributions ?

brada4:
You cannot assume Fortran compiler supports any kind of IPO based on cc test alone.

ghost:
You're correct; this won't work as-is.

ghost:
I've talked with the CMake people and the documentation is wrong. It's possible to check IPO for the fortran compiler as well.

brada4:
That will not make clang IPO work with ifort etc.

ghost:
`check_ipo_supported` attempts to compile a source file with IPO enabled. So if the C compiler is clang and the fortran compiler is ifort, `check_ipo_supported` will return false. As far as individually turning on IPO support for a single language at a time: this is not currently possible so it's not a regression.

ghost:
Also, wouldn't you have the intel C compiler as well? In which case this would enable IPO for all sources.

brada4:
Nope. [They are separate packages](https://software.intel.com/en-us/try-buy-tools)
IPO in 2 compilers does not mean it is same IPO, even if you require latest cmake requirement.

ghost:
The IPO support is tested against the linker. So the function does this:

1. Compiles a test C file with IPO and attempts to link it.
2. Compiles a test Fortran file with IPO and attempts to link it.

If both of those succeed, then IPO is enabled.

ghost:
The worst-case scenario is that if you have ifort, you get no IPO at all, which is what currently happens. The best-case scenario is that if you have the intel linker and both intel C and Fortran compilers, IPO is enabled.

brada4:
Counterexample that your test would permit:
`make CC=clang CCOMMON_OPT=-flto FC=-gfortran FCOMON_OPT=-flto`
i.e clang aka linker has no way to understand IPO objects that gfortran kindly placed on disk and `ar` archives.
It should not be default in unlucky case one got recent cmake.

ghost:
You've uncovered a CMake bug that occurs on Linux. I'm going to close this for now and then come back when the bug is fixed.

brada4:
Undocumented feature failing is not a bug.

brada4:
It will be exactly same combination on OSX with Xcode.

brada4:
Short trip your way:
Imported reference components (LAPACK/LAPACKE) throw around 1000 -Wlto-type-mismatch warnings.
You are doing great job improving cmake templates, though OpenBLAS (the reference LAPACK actually) would not be the most grateful 1st target.
What about starting with "experimental" /quickbuild.ipo script using make for BLAS+CBLAS where IPO actually works and extending it the more stuff starts to "work"?

PR #1447, opened 2018-02-01
Generate CHAR_CORENAME for SPARC

### LINKED ISSUE #1446: Fails to compile on SPARC due to missing CHAR_CORENAME
After fixing #1442, compiling on SPARC still fails:
```
openblas_get_config.c:58:3: error: expected ',' or ';' before 'CHAR_CORENAME'
   CHAR_CORENAME
   ^
openblas_get_config.c: In function 'openblas_get_corename':
openblas_get_config.c:81:10: error: 'CHAR_CORENAME' undeclared (first use in this function)
   return CHAR_CORENAME;
          ^
openblas_get_config.c:81:10: note: each undeclared identifier is reported only once for each function it appears in
openblas_get_config.c:85:1: warning: control reaches end of non-void function [-Wreturn-type]
 }
 ^
make[3]: *** [Makefile:113: openblas_get_config.o] Error 1
```

The problem is that the autogenerated `config.h` does not define `CHAR_CORENAME`:
```
#define OS_SUNOS        1
#define ARCH_SPARC      1
#define C_GCC   1
#define __64BIT__       1
#define PTHREAD_CREATE_FUNC     pthread_create
#define BUNDERSCORE     _
#define NEEDBUNDERSCORE 1
#define SPARC
#define V9
#define DTB_DEFAULT_ENTRIES 32
#define GEMM_MULTITHREAD_THRESHOLD      4
```

When passing `TARGET=SPARC`, everything works because a line `#define CHAR_CORENAME "sparc"` is added.

This should be easy to fix analogous to #1443 but I don't know what the right fix is. Should `CHAR_CORENAME` be defined or not?

martin-frbg:
Should fix #1446, getting automatic target detection working (again?) for the Sparc platform

### DIFF
--- a/cpuid_sparc.c
+++ b/cpuid_sparc.c
@@ -57,3 +57,8 @@ void get_cpuconfig(void){
 void get_libname(void){
   printf("v9\n");
 }
+
+void get_corename(void){
+  printf("sparc\n");
+}
+
--- a/cpuid_sparc.c
+++ b/cpuid_sparc.c
@@ -58,7 +58,7 @@ void get_libname(void){
   printf("v9\n");
 }
 
-void get_corename(void){
-  printf("sparc\n");
+char *get_corename(void){
+  return "sparc";
 }
 
--- a/getarch.c
+++ b/getarch.c
@@ -1116,7 +1116,7 @@ int main(int argc, char *argv[]){
 #ifdef FORCE
     printf("CORE=%s\n", CORENAME);
 #else
-#if defined(INTEL_AMD) || defined(POWER) || defined(__mips__) || defined(__arm__) || defined(__aarch64__) || defined(ZARCH)
+#if defined(INTEL_AMD) || defined(POWER) || defined(__mips__) || defined(__arm__) || defined(__aarch64__) || defined(ZARCH) || defined(sparc)
     printf("CORE=%s\n", get_corename());
 #endif
 #endif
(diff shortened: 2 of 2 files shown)


jdemeyer:
I guess you mean
```diff
diff --git a/cpuid_sparc.c b/cpuid_sparc.c
index ab8f773..6902a8f 100644
--- a/cpuid_sparc.c
+++ b/cpuid_sparc.c
@@ -58,7 +58,7 @@ void get_libname(void){
   printf("v9\n");
 }
 
-void get_corename(void){
-  printf("sparc\n");
+char *get_corename(void){
+  return "sparc";
 }
 
```

jdemeyer:
By the way, I think that `CHAR_CORENAME` is required on every arch and this should simply be unconditional:
```C
#if defined(INTEL_AMD) || defined(POWER) || defined(__mips__) || defined(__arm__) || defined(__aarch64__) || defined(ZARCH) || defined(sparc)
    printf("#define CHAR_CORENAME \"%s\"\n", get_corename());
#endif
```

martin-frbg:
Ah yes, copypasting in a hurry is seldom a good idea. Unless I am mistaken again, the only architectures that do not yet have a get_corename function are the - most likely legacy - Alpha and Itanium.

jdemeyer:
This now actually works.

martin-frbg:
Thank you, I'll merge it now.

PR #1448, opened 2018-02-01
retire 3DNOW assembly

brada4:
There is no 3DNOW in modern processors (reciting xianyi on absence of this extension in current CPUs)
Leftover 3DNOW prefetch instructions are not used in OpenBLAS assembly.
Impact - respective athlon(32) gets 20% slower at GEMM. Raspberry pi jumps around it in circles today for difference in electricity expense alone.

### DIFF
--- a/kernel/x86/KERNEL.ATHLON
+++ b/kernel/x86/KERNEL.ATHLON
@@ -1,4 +1,4 @@
-SGEMMKERNEL    =  gemm_kernel_2x4_3dnow.S
+# SGEMMKERNEL    =  gemm_kernel_2x4_3dnow.S
 SGEMMINCOPY    =  ../generic/gemm_ncopy_2.c
 SGEMMITCOPY    =  ../generic/gemm_tcopy_2.c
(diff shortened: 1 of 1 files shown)


martin-frbg:
Not sure what exactly you are trying to retire ? Seems to me you are commenting out something that is used only by old Athlon and Via C3 - if any of these still exist (maybe embedded somewhere), but does not affect anything else.

brada4:
mmx register manipulations by obsolete 3dnow so that -DHAVE_3DNOW would mean presence of prefetch(w), present on nearly all AMD and on Broadwell+ (unsure about celeron though in this case)

martin-frbg:
Sorry, I still don't get it. So you actually want to get rid of all the `#ifdef HAVE_3DNOW... prefetchw ... #endif` in the .S files that will be ignored on reasonably modern cpus anyhow ?

brada4:
I cannot find CPU to run 3Dnow, I have no hate for prefetch/prefetchw.

Issue #1439, opened 2018-01-29
openblas performance on 8xA53


wl3b10s:
test gemm on 8 core  A53.
Setting 
 omp_set_num_threads(4);
 openblas_set_num_threads(4);
is much faster than:
 omp_set_num_threads(8);
 openblas_set_num_threads(8);

i think this may because of multithread doesn't work well on more cores.

my build config : all default  and build native on arm a53 and ubuntu.

my test config : gemm  float, both no transpose
M = 16, N = 76800, K = 98
M = 32, N = 19481, K = 144
M = 128, N = 1271, K = 1728

brada4:
is it aarch32 or aarch64?
... check NUMA layout detected by ubuntu. Traversing NUMA borders is slow and happens in 64MB (intel64) or so blocks being dragged via system bus between CPUs.
Probably there is something in ARM datasheets about stride sizes, unless you make inputs 10x NUMA stride it will not gain for extra multicore  CPU.
You can still issue two _GEMM in parallel.

ashwinyes:
M = 16, N = 76800, K = 98
M = 32, N = 19481, K = 144
M = 128, N = 1271, K = 1728

Can you try with proper square matrices? Like 1024x1024 or 2048x2048 etc. For skewed matrices like "M = 32, N = 19481, K = 144" , the partitioning may not be good at 8 cores.

martin-frbg:
@ashwinyes I wonder what target gets used here - I guess the A53 is compatible with the A57 (?), but does it get autodetected as such or would users end up with the generic C ARMV8 kernels ? (Same for the A72)

wl3b10s:
@martin-frbg i guess it detect as generic armv8 asm kernel.

ashwinyes:
@martin-frbg Thanks for pointing this. Yes it would use the ARMV8 C kernels. I think its time we make the generic ARMV8 target to use the assembly kernels. All ARMV8 machines will support the assembly kernels .

@wl3b10s For the time being, you may also try building with "make TARGET=CORTEXA57". You will see some performance improvement.

martin-frbg:
The quick and dirty alternative would be to treat A53 and A72 as A57 in cpuid_arm64. Am I right in thinking that the hardfp/softfp issue that lead to a re-introduction of C kernels for ARM affects that 32bit platform only ?

ashwinyes:
[...]

PR #1449, opened 2018-02-06
Enable assembly kernels for the generic ARMV8 target and treat CortexA53,A72 as A57

martin-frbg:
as discussed in #1439

### DIFF
--- a/cpuid_arm64.c
+++ b/cpuid_arm64.c
@@ -114,8 +114,9 @@ int detect(void)
 
 	fclose(infile);
 	if(cpu_part != NULL && cpu_implementer != NULL) {
-		if (strstr(cpu_part, "0xd07") && strstr(cpu_implementer, "0x41"))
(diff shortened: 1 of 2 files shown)


martin-frbg:
@ashwinyes I noticed that your ?amax.S all appear to support USE_MIN to double as the corresponding ?amin functions, yet the KERNEL files still point to the generic C implementations for those. Should this be changed ?
PS: similarly dot.S and its thunderx counterparts appear to be prepared to handle DSDOT, but are not declared as DSDOTKERNEL anywhere as far as I can tell.

ashwinyes:
Those are not tested properly as far as I remember. If you could test it and it works fine, you can push it for generic ARMv8. But please don't change thunderx kernels.

martin-frbg:
Thanks for the feedback, I will see what I can do with the available hardware. (Having a smartphone with a thunderx processor would be nice though, hehe)

PR #1450, opened 2018-02-06
Fix issues with forking on Cygwin

embray:
In most places OpenBLAS treats Cygwin as equivalent to Windows, and in most cases this is fine *except* when it comes to forking a process that's using OpenBLAS (since native Windows does not know about Cygwin's fork()).

In particular, the win32 threading module is broken after fork, and should be shut down before fork, so we must support adding a `pthread_atfork` hook for this on Cygwin.  Note, this is still using the native win32 threading module even on Cygwin--in principle Cygwin could use pthreads instead, but the win32 threading actually works just fine and probably involves less overhead.  It just needs the additional care around fork.

What does *not* work around fork, is using the native Windows memory allocation functions on Cygwin, so this patch also changes memory allocation to use mmap on Cygwin instead.

### DIFF
--- a/Makefile.system
+++ b/Makefile.system
@@ -304,6 +304,7 @@ endif
 ifeq ($(OSNAME), CYGWIN_NT)
 NEED_PIC = 0
 NO_EXPRECISION = 1
+OS_CYGWIN_NT = 1
 endif
 
 ifneq ($(OSNAME), WINNT)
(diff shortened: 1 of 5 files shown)


embray:
Thanks!

martin-frbg:
Fork test seems to hang on ARMv8 as well, probably better to enable it only on systems where we know it should pass...

Issue #365, opened 2014-05-01
Drotmg incorrect result


btracey:
In rotmg.c, lines 146 and 179 should say
"else if(dflag == ONE)"
rather than just "else"

The first pass through the code, if the flag is zero, it sets d11, d12, and sets dflag to -1. If it needs to go through the while loop a second time, dflag is now -1, (which is not 1), and so it changes dh21 to -1 and dh12 to -1. This should not happen. ATLAS has this flag check, as does the accelerate framework on OSX.

PR #1454, opened 2018-02-09
Keep the flag handling separate from the scaling loops in rotmg

### LINKED ISSUE #1452: some question about rotmg
Well,when i used rotmg function,the same input had different result between cublas and openblas.
input is as follows:
d1=5.9e-8
d2=5.960464e-8
x1=1.0
y1=150.0
param[5]={0.0, 0.0, 0.0, 0.0, 0.0}

And the output of openblas:
d1=0.999956
d2=0.989812
x1=0.036623
y1=150.0
param[5]={-1.0, 0.000002, -0.000244, 0.000244, 0.000002}

And the output of cublas:
d1=0.999956
d2=0.989812
x1=0.036623
y1=150.0
param[5]={-1.0, 0.000002, -0.000244, 1.0, 0.000002}

I referred some papers in order to know the implementation of the function.And I found that in some situation, this function will have a scaling operation.It seems that cublas does not have this operation? It is a bug of cublas?I think it's a bug of cublas, but i'm not sure.

martin-frbg:
Fixes #1452 and is more in line with how ATLAS does it. The earlier fix from #365 only moved the bug elsewhere, but we will never want the iterative rescaling to change the dflag setting and variable associations with each cycle.

### DIFF
--- a/interface/rotmg.c
+++ b/interface/rotmg.c
@@ -136,18 +136,21 @@ void CNAME(FLOAT *dd1, FLOAT *dd2, FLOAT *dx1, FLOAT dy1, FLOAT *dparam){
 
 		if(*dd1 != ZERO)
 		{
-			while( (*dd1 <= RGAMSQ) || (*dd1 >= GAMSQ) )
+			if( (*dd1 <= RGAMSQ) || (*dd1 >= GAMSQ) )
 			{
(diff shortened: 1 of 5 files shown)


martin-frbg:
I intend to add the resurrected utest for this tomorrow.

PR #1467, opened 2018-02-20
Minor fixes

brada4:
Rewrite one pair of identical branches
Clear few dead trailing codes
Eliminate two unused variables
All from clang checker, strangely missed previously.

### DIFF
--- a/benchmark/gemm.c
+++ b/benchmark/gemm.c
@@ -237,7 +237,7 @@ int main(int argc, char *argv[]){
     timeg = time1/loops;
     fprintf(stderr,
 	    " %10.2f MFlops %10.6f sec\n",
-	    COMPSIZE * COMPSIZE * 2. * (double)m * (double)m * (double)n / timeg * 1.e-6, time1);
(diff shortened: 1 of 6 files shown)


martin-frbg:
Close and reopen should also work to trigger a retest, no need to add otherwise unneeded changes.
Not sure what makes that one appveyor test hit the 1 hr limit every now and again (and finish in under 55 minutes otherwise)

brada4:
Point taken of one subject per PR.

martin-frbg:
I'd still prefer to see the benchmark fixed in a separate PR, and the "take out unreachable branch to retrigger CI" removed as it only makes the code harder to read ? (Also it will probably mess up "git blame" which currently allows to trace the actual "if 0" code change to issue 130 (and I am not entirely sure if that old change was even an appropriate fix for the issue)

brada4:
OK

martin-frbg:
BTW where did the additional "+2" (as in "k+2") come from in your patch for 1474 ?

Issue #1470, opened 2018-02-22
daxpy 10x slower on macOS (Haswell)


simonbyrne:
Calls to `daxpy` seem to be about 10x slower on macOS than when run inside a Linux VM on the same machine (both give the config as "USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell").

This affects a lot of code which depends on it (e.g. `dgebrd`).

downstream: https://github.com/JuliaLang/LinearAlgebra.jl/issues/501

martin-frbg:
Nobody here currently has a Mac I fear. From anecdotal evidence in previous issues (#730, JuliaLang/julia#901) you could try removing the `.align 16` in kernel/x86_64/daxpy_microk_haswell-2.c, or try with the current "develop" tree as that happens to have that microkernel replaced with the non-AVX2 Sandybridge one for unrelated reasons (the respective microkernel is #included at the top of kernel/x86_64/daxpy.c)

brada4:
Can you check if virtual machine has AVX2 inside in /proc/cpuinfo (i.e haswell or sandybridge is used)

simonbyrne:
> Can you check if virtual machine has AVX2 inside in /proc/cpuinfo (i.e haswell or sandybridge is used)

Yes, it does.

simonbyrne:
Using the current `develop` branch (e3a80e6aa8a289228de9ff1c97364bcb7340f7e9) doesn't fix it either.

martin-frbg:
Does killing any and all .align16 in the daxpy microkernel file change anything ?

simonbyrne:
Yes, commenting that out in `daxpy_microk_sandy-2.c` seems to fix it.

ViralBShah:
Is it possible to perhaps have a new release once we fix this, given that we haven't had one for a long time? We're happy (in the Julia community) to try out an RC and give feedback before release.

simonbyrne:
Ah, does it have anything to do with [this commit comment](https://github.com/xianyi/OpenBLAS/commit/6e7be06e072cddc5d34617f28f60a32484b9e910)?

According to [the Mac developer docs](https://developer.apple.com/library/content/documentation/DeveloperTools/Reference/Assembler/040-Assembler_Directives/asm_directives.html#//apple_ref/doc/uid/TP30000823-TPXREF146):

[...]

PR #1473, opened 2018-02-26
Replace .align with .p2aligns in dscal.c and the Nehalem microkernels as well

martin-frbg:
for issue #1470


martin-frbg:
Test failures look like unrelated timeouts, retrying to be on the safe side

PR #1480, opened 2018-03-03
Rewrite ROTMG to address cases not covered by the netlib algorithm

### LINKED ISSUE #1452: some question about rotmg

martin-frbg:
Solution copied from the new implementation in https://github.com/gonum/gonum/commit/58e39c9 by kortschak et al. based on an algorithm proposed by Tim Hopkins in 1998. Fixes #1452 (see reference to the original paper there).

### DIFF
--- a/utest/utest_main2.c
+++ b/utest/utest_main2.c
@@ -70,7 +70,7 @@ CTEST (drotmg,rotmg)
 	  te_param[i]=tr_param[i]=0.0;
 	}
 
-	//reference values as calulated by netlib blas
+	//reference values as calculated by netlib blas
 
         tr_d1= 0.1732048;
         tr_d2= 0.03840234;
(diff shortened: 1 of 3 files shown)


utest/utest_main2.c
```diff
@@ -115,8 +117,8 @@ CTEST (drotmg,rotmg_issue1452){
 	for(i=0; i<5; i++){
 	  te_param[i]=tr_param[i]=0.0;
 	}
-
-	//reference values as calulated by netlib blas
+	te_param[3]=1./4096.;
+	//reference values as calulated by gonum blas with rotmg rewritten to Hopkins' algorithm
```
kortschak: Spelling of "calculated" if you want.

kortschak (approved):
This looks good with very minor preexisting nit.

Issue #875, opened 2016-05-13
build error in linux: memory.c:1348, memory.c:1406


hainanwang:
Dear friends, 

I am building openblas in linux. I got the following error due to function defined in memory.c. Is this a bug or some other issue? Thanks! 
gcc -O2 -DMAX_STACK_ALLOC=2048 -Wall -m64 -DF_INTERFACE_GFORT -fPIC -DSMP_SERVER -DNO_WARMUP -DMAX_CPU_NUMBER=32 -DASMNAME=memory -DASMFNAME=memory_ -DNAME=memory_ -DCNAME=memory -DCHAR_NAME=\"memory_\" -DCHAR_CNAME=\"memory\" -DNO_AFFINITY -I../.. -c memory.c -o memory.o
memory.c:1348: error: wrong number of arguments specified for 'constructor' attribute
memory.c:1406: error: wrong number of arguments specified for 'destructor' attribute
make[1]: **\* [memory.o] Error 1

line 1348: void CONSTRUCTOR gotoblas_init(void)
line 1406: void DESTRUCTOR gotoblas_quit(void)

Regards,

Hainan

martin-frbg:
Which version of OpenBLAS are you trying to build ? This is closed issue #767, assumed to be fixed by adding compiler-specific guards around the use of "modern" constructor/destructor qualifiers.

hainanwang:
Hi Martin,

I am building openblas 0.2.18.  I tried removing (101) following the instruction in #767 and #700. Now it can be built successfully. FYI, Here is my compiler/OS: 
gcc version 4.1.2 20080704 (Red Hat 4.1.2-52)

Thanks, 

Hainan

brada4:
Your compiler is buggy
There are miscompilation fixes from 2012 and 2014 available from red hat.
If you get in retro computing- would you expect windows 7 to run on an i386?

jeromerobert:
@hainanwang From #700 and http://pkgs.fedoraproject.org/cgit/rpms/openblas.git/tree/openblas-0.2.15-constructor.patch I guess you need:

```patch
--- a/driver/others/memory.c
+++ b/driver/others/memory.c
@@ -144,7 +144,7 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #if defined(_MSC_VER) && !defined(__clang__)
 #define CONSTRUCTOR __cdecl
[...]

PR #1504, opened 2018-04-02
Allow building on OpenBSD

ararslan:
With this change, OpenBLAS builds and all tests pass on OpenBSD 6.2 using Clang. Tested on x86-64 only, with and without `DYNAMIC_ARCH=1`.

The only quirk here is that Clang is the default compiler on OpenBSD 6.2, and yet I still had to use `CC=clang` when building, otherwise OpenBLAS uses GCC. The default GCC version on OpenBSD is very old, so we run into #875. Let me know if there's something in the Makefile that needs to be adjusted that I've overlooked.

cc @ibara, who may be interested in this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,11 +91,7 @@ ifeq ($(OSNAME), $(filter $(OSNAME),Linux SunOS Android))
 	@ln -fs $(LIBSONAME) $(LIBPREFIX).so
 	@ln -fs $(LIBSONAME) $(LIBPREFIX).so.$(MAJOR_VERSION)
 endif
-ifeq ($(OSNAME), FreeBSD)
-	@$(MAKE) -C exports so
-	@ln -fs $(LIBSONAME) $(LIBPREFIX).so
-endif
-ifeq ($(OSNAME), NetBSD)
+ifeq ($(OSNAME), $(filter $(OSNAME),FreeBSD OpenBSD NetBSD))
 	@$(MAKE) -C exports so
 	@ln -fs $(LIBSONAME) $(LIBPREFIX).so
 endif
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ ifeq ($(OSNAME), $(filter $(OSNAME),Linux SunOS Android))
 	@ln -fs $(LIBSONAME) $(LIBPREFIX).so
 	@ln -fs $(LIBSONAME) $(LIBPREFIX).so.$(MAJOR_VERSION)
 endif
-ifeq ($(OSNAME), $(filter $(OSNAME),FreeBSD OpenBSD NetBSD))
+ifeq ($(OSNAME), $(filter $(OSNAME),FreeBSD OpenBSD NetBSD DragonFly))
 	@$(MAKE) -C exports so
 	@ln -fs $(LIBSONAME) $(LIBPREFIX).so
 endif
--- a/Makefile.install
+++ b/Makefile.install
@@ -72,12 +72,7 @@ ifeq ($(OSNAME), $(filter $(OSNAME),Linux SunOS Android))
 	ln -fs $(LIBSONAME) $(LIBPREFIX).so ; \
 	ln -fs $(LIBSONAME) $(LIBPREFIX).so.$(MAJOR_VERSION)
 endif
-ifeq ($(OSNAME), FreeBSD)
-	@cp $(LIBSONAME) "$(DESTDIR)$(OPENBLAS_LIBRARY_DIR)"
-	@cd "$(DESTDIR)$(OPENBLAS_LIBRARY_DIR)" ; \
-	ln -fs $(LIBSONAME) $(LIBPREFIX).so
-endif
-ifeq ($(OSNAME), NetBSD)
+ifeq ($(OSNAME), $(filter $(OSNAME),FreeBSD OpenBSD NetBSD))
 	@cp $(LIBSONAME) "$(DESTDIR)$(OPENBLAS_LIBRARY_DIR)"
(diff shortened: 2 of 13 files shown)


martin-frbg:
WIthout wanting to start a "political" discussion, do you expect OpenBSD to deviate significantly from the already supported FreeBSD ? Else I think your PR could be simplified to affect much fewer files...
Also, does OpenBSD install both gcc and clang side by side by default, and still default to clang (how?) ? I have already queued a PR to (finally) get rid of #875, but we would probably want to use a recent clang rather than an old gcc if given the choice.

ararslan:
> do you expect OpenBSD to deviate significantly from the already supported FreeBSD ?

I'm not sure what you mean, they're different systems. Can you elaborate?

> Also, does OpenBSD install both gcc and clang side by side by default, and still default to clang (how?) ?

Yes, OpenBSD 6.2, which I tested this on, ships with both Clang 4.0.0 and GCC 4.2.1. It uses Clang as the default compiler starting with OpenBSD 6.2, i.e. `cc` refers to `clang`, and previous versions defaulted to GCC. Despite Clang being default for my version, somehow without specifying `CC=clang` when invoking Make, OpenBLAS chooses to build with GCC.

martin-frbg:
It just occured to me that many of your changes added OS_OPENBSD where there already is OS_FREEBSD, so _if_ it would be sufficient to define OS_FREEBSD when the actual name is either FreeBSD or OpenBSD fewer files would need to be touched.
Regarding the CC issue, are you building with make or cmake ? I believe the former should pick up $CC from the environment (through Makefile.prebuild which invokes c_check with that setting), I am less sure about the latter.

ararslan:
> so if it would be sufficient to define OS_FREEBSD when the actual name is either FreeBSD or OpenBSD fewer files would need to be touched.

Ah, I see. That may work, but I'm hesitant to say one way or another whether it's completely kosher, as I'm not familiar with all of the differences between the FreeBSD and OpenBSD kernels. It might be best to leave them separate for now and perhaps simplify things later if we find that they can be safely combined in full generality. Does that seem okay to you?

>  are you building with make or cmake ?

GNU Make, I didn't try using CMake.

I just checked the output of `gmake -p` and `CC` is indeed getting set automatically to `cc`, so I'm not sure why it's picking up `gcc`. ¯\\\_(ツ)\_/¯

ibara:
It is probably OK to have OpenBSD defined as `OS_FREEBSD` for building. However, OpenBSD doesn't have a `__FreeBSD__` cpp define, so at least the `__OpenBSD__` cpp define would need to be added where needed, as would the `Makefile` additions be needed. It would be nice if OpenBLAS reported running on OpenBSD when actually running on OpenBSD (if OpenBLAS does in fact report OS at runtime, I've never used it).

Some OpenBSD architectures ship with both gcc-4.2.1 and clang-5.0.1 but some ship with only clang. Additionally, clang is actively tracked and updated on OpenBSD as new releases come out. On systems where both are shipped, `cc(1)` is a symlink to clang, and the `$CC` variable is set to `cc` for official package builds (you may have to set it yourself for manual builds), so it would only ever build with clang in the scenarios we care about, aka official package builds. In all cases, it would either be clang or a more modern gcc (of the 4.9.4 vintage right now, but will be upgraded in the future) from packages that would be building OpenBLAS, so I don't think that #875 would be a problem. Also to note, that eventually, systems that ship with both gcc-4.2.1 and clang will only ship with clang, so this is a non-issue.

ararslan:
Something that may be a little more robust would be to define something like
```c
#if defined(__FreeBSD__) || defined(__OpenBSD__)
OS_BSD_COMMON
#endif
```
then use that in place of all current uses of `OS_FREEBSD`. That makes the code a little clearer as well (at least to me) since you know that what you're reading applies to multiple BSD systems rather than just FreeBSD. (I imagine that DragonFly would fit into all of the same conditionals as well.)

I can also just define `OS_FREEBSD` on OpenBSD, but I'm not sure how to structure the change in `c_check`. I guess in that file OpenBSD would be omitted and reported as FreeBSD?

martin-frbg:
Thank you very much for the detailed explanation. I guess at that point it is not useful to go through extra hoops just to avoid changing one or two of the files. 
The compiler mis-detection is a bit intriguing - what does `$CC -E ctest.c` return on that system (which is what c_check invokes, the `COMPILER_whatever` output should be above the OS_OPENBSD line) ? Maybe that file needs another patch, if "your" clang does not identify itself by defining `__clang__` .

ararslan:
On this branch:
```
$ cc -E ctest.c
# 1 "ctest.c"
# 1 "<built-in>" 1
# 1 "<built-in>" 3
# 317 "<built-in>" 3
# 1 "<command line>" 1
# 1 "<built-in>" 2
# 1 "ctest.c" 2
# 12 "ctest.c"
COMPILER_CLANG
# 44 "ctest.c"
COMPILER_GNU
# 64 "ctest.c"
OS_OPENBSD
# 105 "ctest.c"
ARCH_X86_64
# 137 "ctest.c"
BINARY_64
```

ararslan:
Ah, the Makefile is setting the compiler to `gcc` because `$(origin CC)` is `default`. See https://github.com/xianyi/OpenBLAS/blob/develop/Makefile.system#L14-L20.

ararslan:
Since (nearly) the exact same edits are needed for supporting DragonFly BSD, I've added a separate commit that includes the appropriate changes for DragonFly. I tested it with the default compiler, GCC, on its only supported architecture, x86-64.

ibara:
OpenBSD clang defines `__clang__`

ararslan:
Thank you both, @ibara and @martin-frbg!

Issue #231, opened 2013-06-10
should default CC=clang on mac


cartazio:
The system assembly on Mac doesn't understand AVX instructions, which is probably an ongoing source of confusion when people try to build open blas on a mac, 

the work around is using CC=clang, because the clang assembler goes via llvm.

this problem is dealt with when using mac homebrew to install on a mac, but would be a good idea to make clang the default for doing assembling on mac (and compiling too i suppose)

PR #1505, opened 2018-04-03
Compile with cc rather than gcc whenever possible

ararslan:
See discussion in #1504 regarding misdetection of the appropriate compiler. This will require some verification, especially from someone familiar with Windows, that the approach taken here is legit.

### DIFF
--- a/Makefile.system
+++ b/Makefile.system
@@ -17,15 +17,19 @@ NETLIB_LAPACK_DIR = $(TOPDIR)/lapack-netlib
 #   http://stackoverflow.com/questions/4029274/mingw-and-make-variables
 # - Default value is 'cc' which is not always a valid command (e.g. MinGW).
 ifeq ($(origin CC),default)
+
+# Check if $(CC) refers to a valid command and set the value to gcc if not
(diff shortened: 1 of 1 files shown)


ararslan:
The failed Windows build appears to have just timed out.

martin-frbg:
Yes, one of the appveyor configurations does this occasionally, probably taking too long to download and install the build environment. Note however that we have no CI tests set up for OSX.
If I read #231 correctly, forcing CC to clang on OSX (which your patch seems to undo) was done to pick up the builtin assembler from llvm rather than the inferior system assembler that gcc would use.
Maybe this could also be achieved by other means, but perhaps it would be safer to move the  UNAME_S  line up and make the whole "if (origin CC)" thing conditional on "none of the BSDs" ?  
Or at least leave the Darwin case in place, as it seems to serve a different purpose...

ararslan:
Sure, I can certainly reinstate the macOS-specific logic. I assumed it was outdated, as for quite a while now `cc` on macOS has been Clang rather than GCC. I probably shouldn't be making such assumptions. :slightly_smiling_face:

PR #1510, opened 2018-04-07
Add macOS to the Travis testing matrix

ararslan:
Let's see what happens...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ language: c
 jobs:
   include:
     - &test-ubuntu
+      os: linux
       stage: test
       compiler: gcc
       addons:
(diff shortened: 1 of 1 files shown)


martin-frbg:
"Could not parse .travis.yml" unfortunately...

martin-frbg:
I _think_ I got it working now, simply by moving the osx build to the top of the include list.

ararslan:
> "Could not parse .travis.yml" unfortunately...

Ah yes, I had meant to validate the YAML using Travis' tool before pushing. Probably why the Travis webhook didn't show up at all.

> I think I got it working now

Where were you able to do that? I don't see macOS in the Travis config, since the commit here was reverted without replacement.

I think I know what the issue was, I can try again in another PR.

ararslan:
See #1512

martin-frbg:
I have experimented with travis in my fork, after adding the travis connection to it. (Which I assume you did in your fork now as well ?)

Issue #1492, opened 2018-03-16
Test failures for 32-bit macOS compile


matthew-brett:
I have just noticed that, when updating to latest `develop` branch, I am getting test failures when compiling for 32-bit macOS only.  `make test` seems to fail (but runs to completion on 64-bit):

https://travis-ci.org/MacPython/openblas-libs/jobs/354493385

```
Note: The following floating-point exceptions are signalling: IEEE_DIVIDE_BY_ZERO
Note: The following floating-point exceptions are signalling: IEEE_DIVIDE_BY_ZERO
Note: The following floating-point exceptions are signalling: IEEE_DIVIDE_BY_ZERO
Note: The following floating-point exceptions are signalling: IEEE_DIVIDE_BY_ZERO
Note: The following floating-point exceptions are signalling: IEEE_DENORMAL
Note: The following floating-point exceptions are signalling: IEEE_DENORMAL
make[1]: *** [run_test] Error 1
make: *** [tests] Error 2
```

martin-frbg:
Which (emulated) hardware is this trying to build on ? (I think I see you are using the current develop branch, which should be good, but I cannot figure out much more. From the long list of what seem to be mobile devices at the start, could it be armv7 which received patches for softfp support relatively recently ?) Any chance to find out _which_ of the tests are failing ? And which if any version used to pass these tests before ?

matthew-brett:
The building machinery is here:

https://github.com/MacPython/openblas-libs

The build that is failing is the 32-bit build of OpenBLAS, on 64-bit macOS - so no emulation, just a 32-bit build.  This just corresponds to:

```
make DYNAMIC_ARCH=1 USE_OPENMP=0 NUM_THREADS=64 BINARY=32
```

I don't know what test is failing - but I am running the same thing now on my desktop.

matthew-brett:
Ok - here is the test output on my desktop:

```
./openblas_utest
TEST 1/22 amax:samax [OK]
TEST 2/22 drotmg:rotmg [OK]
TEST 3/22 drotmg:rotmg_issue1452 [OK]
TEST 4/22 drotmg:rotmg_D1eqD2_X1eqX2 [OK]
TEST 5/22 drotmg:drotmg_D1_big_D2_big_flag_zero [OK]
TEST 6/22 axpy:daxpy_inc_0 [OK]
TEST 7/22 axpy:zaxpy_inc_0 [OK]
[...]

PR #1512, opened 2018-04-07
Add macOS to the Travis testing matrix: Take 2!

ararslan:
Don't merge until we're sure that a Travis macOS build made it through.

I'm pretty sure this should fix the invalid YAML syntax issue. The `travis-lint` Ruby gem appears not to have been updated to work with build stages so I haven't been able to verify correctness locally.

Edit: Appears to be working!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ language: c
 jobs:
   include:
     - &test-ubuntu
+      os: linux
       stage: test
       compiler: gcc
       addons:
@@ -57,7 +58,8 @@ jobs:
         - TARGET_BOX=LINUX32
         - BTYPE="BINARY=32"
 
-    - stage: test
+    - os: linux
+      stage: test
       compiler: gcc
       addons:
         apt:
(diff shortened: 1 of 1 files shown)


ararslan:
As an aside, you might be able to cut down on the overall CI time here by moving some of the Travis Linux builds to CircleCI. In particular the Alpine Linux builds might make more sense there, as Circle is mostly Docker-based, so you can just load an Alpine Docker image rather than going through the steps currently done in the Travis config.

martin-frbg:
Yes, looks good now (and better than my kludge of adding it before the existing tests), will merge one the checks have completed.

martin-frbg:
Maybe add a BINARY=32 versions as well for the likes of #1492 ?

ararslan:
I've experimented a bit on my fork with using CircleCI instead of Travis for the Alpine testing. See https://circleci.com/gh/ararslan/OpenBLAS/30 for example. It might be worthwhile to take some of the load off of Travis, especially since this PR is adding two more builds to the already long queue.

Circle is running an official Alpine Docker image rather than installing it in a chroot in Ubtuntu as is currently being done on Travis. On Travis, the Alpine builds seem to be taking on average 5-10 minutes longer than the other builds (aside from MinGW), whereas the Alpine builds on my fork took a total of 6-7 minutes from VM startup to tests finishing.

martin-frbg:
@xianyi what do you think about adding CircleCI ?

Issue #1525, opened 2018-04-16
RISC-V Architecture support


jerryz123:
RISCV is an open source ISA gathering a lot of support in the hardware community. There's a lot of interest in HPC on RISCV, and future extensions to the ISA look to support vectorized operations. 

Would it be possible to add RISC-V to the list of supported architectures? I have a working fork that compiles for RISC-V, and I'm willing to work on integrating the changes.

xianyi:
@jerryz123 Great work!

We like to merge your RISC-V PR.

PR #1526, opened 2018-04-16
Add support for RISC-V

jerryz123:
#1525


martin-frbg:
At first glance the majority of the kernel files added by your PR are copies of the respective "generic" implementation ? (If that is the case, having KERNEL define e.g. SAMAXKERNEL=../generic/amax.c should work just as well and result in a much smaller changeset...)

jerryz123:
Most of these were actually taken from the ARM folder, which oddly has a lot of the generic implementations. I thought it might be weird to point to the ARM kernels from here, so I opted to copy the files instead.

Also, the generic folder only contains a handful of generic kernels, for some reason?

martin-frbg:
Point taken. "Historic reasons" most likely - probably "generic" was just a small number of files for generic x86 targets in the original libGoto, and the ARM port had to create its own pool of generic C implementations to get started.

xianyi:
Hi All,

I just created a risc-v branch for merging PR.

davidlt:
Any plans to have this in a release?
Btw, shouldn't there be an update to `TargetList.txt` to mention RISCV64?

martin-frbg:
Sometime after 0.3.1 would be my guess - so far this is basically just a generic C target, and #1528 hints at fundamental problems due to different optimization strategies needed for that platform.

davidlt:
I am attempting to apply the current patch on top of 0.3.0 (applies as-is) for Fedora RISC-V. Let's see if it builds.

davidlt:
I hit this: https://github.com/xianyi/OpenBLAS/blob/develop/getarch.c#L1066

```
BUILDSTDERR: getarch.c:1085:2: error: #error "This arch/CPU is not supported by OpenBLAS."
BUILDSTDERR:  #error "This arch/CPU is not supported by OpenBLAS."
BUILDSTDERR:   ^~~~~
BUILDSTDERR: make: *** [Makefile.prebuild:62: getarch] Error 1
BUILDSTDERR: Makefile.system:1005: Makefile.: No such file or directory
BUILDSTDERR: make: *** No rule to make target 'Makefile.'.  Stop.
```
I guess, we need to add something like this:
```
 907 #ifdef FORCE_ARMV8                                                                                                                                                                                                             
 908 #define FORCE                                                                   
 909 #define ARCHITECTURE    "ARM64"                                                 
 910 #define SUBARCHITECTURE "ARMV8"                                                 
 911 #define SUBDIRNAME      "arm64"                                                 
 912 #define ARCHCONFIG   "-DARMV8 " \                                               
 913        "-DL1_DATA_SIZE=32768 -DL1_DATA_LINESIZE=64 " \                          
 914        "-DL2_SIZE=262144 -DL2_LINESIZE=64 " \                                   
 915        "-DDTB_DEFAULT_ENTRIES=64 -DDTB_SIZE=4096 -DL2_ASSOCIATIVE=32 "          
 916 #define LIBNAME   "armv8"                                                       
 917 #define CORENAME  "ARMV8"                                                       
 918 #endif   
```

RPM: http://fedora-riscv.tranquillity.se:3000/rpms/openblas/src/branch/master-riscv64
Log: http://fedora-riscv.tranquillity.se/kojifiles/work/tasks/9574/19574/build.log

Note, that `TARGET` was not set for riscv64.

davidlt:
I guess, I could set it to `GENERIC`.

jerryz123:
What commands did you use? I run "make ARCH=riscv64 TARGET=RISCV64
HOSTCC=gcc CC=riscv-gcc"

As for the kernels, I'm not sure right now which implementations to target
for optimizations.

jerryz123:
Note building with generic doesn't pass all the tests for me.

davidlt:
I have added your suggested flags and soon will push another build.

At this point we don't run tests for packages (same for other distros).

davidlt:
Failed again for me.
```
[..]
riscv64 -ru ../../libRblas_riscv64-r0.3.0.dev.a sbdsvdx.o spotrf2.o [..]
[..]
BUILDSTDERR: make[2]: *** [Makefile:556: ../../libRblas_riscv64-r0.3.0.dev.a] Error 127
make[2]: Leaving directory '/builddir/build/BUILD/openblas-0.3.0/Rblas/lapack-netlib/SRC'
make[1]: Leaving directory '/builddir/build/BUILD/openblas-0.3.0/Rblas/lapack-netlib'
BUILDSTDERR: make[1]: *** [Makefile:21: lapacklib] Error 2
BUILDSTDERR: make: *** [Makefile:211: netlib] Error 2
[..]
```
`ar` command becomes `riscv64`. Right here: https://github.com/xianyi/OpenBLAS/blob/develop/lapack-netlib/SRC/Makefile#L556

Looking into lapack-netlib:
```
[..]
INSTALL/make.inc.gfortran:ARCH      = ar
INSTALL/make.inc.gfortran:ARCHFLAGS = cr
INSTALL/make.inc.gfortran_debug:ARCH      = ar
INSTALL/make.inc.gfortran_debug:ARCHFLAGS = cr
[..]
```
We override them somewhere.

ctest.c
```diff
@@ -149,3 +149,7 @@ ARCH_ARM
 ARCH_ARM64
 #endif
 
+#if defined(__riscv)
```
davidlt, line 152: Shouldn't you check for `__riscv_xlen` (32, 64, 128) or better for pointer size?

martin-frbg:
the make.inc in lapack-netlib should have been generated by the toplevel Makefile (and yes, the double meaning of  "ARCH" as both "architecture" (OpenBLAS) and "archiver" (lapack-netlib) is unfortunate.)

jerryz123:
Sorry, I just ran it again. Correct command is 

`make TARGET=RISCV64 CC=riscv64-unknown-linux-gnu-gcc HOSTCC=gcc FC=riscv64-unknown-linux-gnu-gfortran all`

davidlt:
This is the used in Fedora:
```
+ make -C Rblas TARGET=RISCV64 DYNAMIC_ARCH=0 USE_THREAD=0 USEOPENMP=0 FC=gfortran CC=gcc COMMON_OPT= 'FCOMMON_OPT=-O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -fasynchronous-unwind-tables -fstack-clash-protection -fPIC -frecursive' NUM_THREADS=128 LIBPREFIX=libRblas LIBSONAME=libRblas.so INTERFACE64=0
```
and it does compile files using `gfortran`.

My current guess would be that somehow `f_check` is confused/silently fails (?) resulting in `NOFORTRAN` and thus `ARCH` not being overwritten correctly.

I will have to investigate tomorrow.

martin-frbg:
With f_check failing and setting NOFORTRAN, the lapack-netlib tree should not be accessed (as it is the - only - part that requires gfortran).

dloghin:
I successfully cross-compiled using this command: 
`make ARCH=riscv64 TARGET=RISCV64 BINARY=64 CROSS=1 CROSS_SUFFIX=riscv64-unknown-linux-gnu- CC=riscv64-unknown-linux-gnu-gcc FC=riscv64-unknown-linux-gnu-gfortran HOSTCC=gcc USE_THREAD=0 USEOPENMP=0`

However, there are two (minor) issues:
1. c_check fails withe the error:
```
1 at ./c_check line 169.
```

This can be avoided by adding:
```
    $defined = 1;
}
```
in c_check

2. The issue with `ARCH` in `lapack-netlib` mentioned above. This can be solved by replacing ARCH with AR, for example.

PR #1751 uses these two approaches to solve the issues.

davidlt:
Here is what worked for me in Fedora/RISC-V: http://fedora-riscv.tranquillity.se:3000/rpms/openblas/commits/branch/master-riscv64

StefanBruens:
Is there any progress on integrating it into the release branch?

martin-frbg:
While I am not adverse to merging it into develop, unfortunately it seems that my observations from https://github.com/xianyi/OpenBLAS/pull/1526#issuecomment-394161654 still stand - this is still a "generic" port without any target-specific optimizations, and indications were that the blocking and packing algorithms at the core of OpenBLAS (GotoBLAS) may not even be appropriate for the RISCV architecture.

jerryz123:
A good BLAS port should contain implementation-specific optimizations. Since currently the space of available RISC-V cores is very diverse, and since there is not a large demand for an optimized BLAS for any specific implementation, it does not make sense to optimize an implementation specific version at the moment.

We hope high-performance RISC-V cores will eventually converge to using variable-length vector instructions to exploit parallelism. I experimented with a vector-optimized BLAS built with OpenBLAS, and found that the existing blocking and packing algorithms are suboptimal for these architectures.

If anybody wants to optimize a BLAS for their own RISC-V implementation, this branch is a good place to start.

Issue #1469, opened 2018-02-21
utest failures on various arches


sharkcz:
After commit e7366a41616 we see test failures on various non-x86 arches

s390x
`````
TEST 11/22 dsdot:dsdot_n_1 [FAIL]
  ERR: test_dsdot.c:47  expected -2.393e-03, got -2.393e-03 (diff 1.035e-10, tol 1.000e-13)
````

aarch64
`````
TEST 11/22 dsdot:dsdot_n_1 [FAIL]
  ERR: test_dsdot.c:47  expected -2.393e-03, got -2.393e-03 (diff 1.035e-10, tol 1.000e-13)
`````

ppc64
`````
TEST 16/21 rot:csrot_inc_0 [FAIL]
  ERR: test_rot.c:109  expected 3.125e-01, got -2.148e-01 (diff 5.273e-01, tol 1.000e-04)
TEST 18/21 rot:zdrot_inc_0 [FAIL]
  ERR: test_rot.c:71  expected 3.125e-01, got -2.148e-01 (diff 5.273e-01, tol 1.000e-13)
`````
and even more on armv7
````
./openblas_utest
TEST 1/22 amax:samax [OK]
TEST 2/22 drotmg:rotmg_D1eqD2_X1eqX2 [OK]
TEST 3/22 drotmg:rotmg_issue1452 [OK]
TEST 4/22 drotmg:rotmg [OK]
TEST 5/22 axpy:caxpy_inc_0 [FAIL]
  ERR: test_axpy.c:109  expected 2.000e+00, got -3.000e+00 (diff 5.000e+00, tol 1.000e-13)
TEST 6/22 axpy:saxpy_inc_0 [FAIL]
  ERR: test_axpy.c:90  expected 2.000e+00, got 4.000e+00 (diff -2.000e+00, tol 1.000e-13)
TEST 7/22 axpy:zaxpy_inc_0 [FAIL]
  ERR: test_axpy.c:71  expected 2.000e+00, got -3.000e+00 (diff 5.000e+00, tol 1.000e-13)
TEST 8/22 axpy:daxpy_inc_0 [FAIL]
  ERR: test_axpy.c:52  expected 2.000e+00, got 4.000e+00 (diff -2.000e+00, tol 1.000e-13)
TEST 9/22 zdotu:zdotu_offset_1 [OK]
TEST 10/22 zdotu:zdotu_n_1 [OK]
TEST 11/22 dsdot:dsdot_n_1 [FAIL]
[...]

PR #1535, opened 2018-04-23
Use generic zrot.c on ppc64/POWER6 to work around utest failure from …

martin-frbg:
…#1469


sharkcz:
zrot.c is not available in "generic" but in "arm", and with this change the utests are passing and build for POWER6 is successful. With the proposed change I got "no rule to make ../kernel/power/../generic/zrot.c"

Issue #1529, opened 2018-04-19
Can a cmake build system determine what options OpenBLAS was built with?


brianborchers:
I'm interested in setting up cmake for my software package, which makes use of BLAS and LAPACK routines.  My code is written using OpenMP and I would prefer to link with an OpenMP version of OpenBLAS so that I can manage the number of threads explicitly.  I'd like to warn the user about potential problems if the library is built with pthreads, or warn of poor performance if the library is built single threaded.  

I'm not aware of any way to see whether or not the version of OpenBLAS that I'm linking to was compiled with OpenMP support, single threaded, etc.  Is there some way (e.g. by symbols provided or not in the library) that cmake or other build systems could determine these aspects of the OpenBLAS library? 

One very painful option would be to distribute OpenBLAS and build it along with my code- that's obviously not a desirable way to package things.

martin-frbg:
Hmm. That came up in #1416 already, but I am not sure what the best solution would be. Neither the maximum number of threads nor the type of multithreading used are written to openblas_config.h as far as I know (and I am not entirely sure if it would help if they were). Using ldd (or equivalent) on the library should show its dependencies but I am not sure if there is a cmake function for this. If you want to check symbols, it should be a case of either blas_server() or blas_server_omp() being compiled into the library.
There is also openblas_get_num_threads() which could be called from a small test program - though it was changed in #760 to return the number of currently active threads, not the maximum the library was built to handle.

brada4:
Not wildly documented, but you can load/open library file and try:
https://github.com/xianyi/OpenBLAS/wiki/OpenBLAS-Extensions
... helps if you mention platform ....
I would suggest for inspiration to take idea from R - they put generic BLAS/LAPACK libs in a directory and you are absolutely free to replace it with OpenBLAS, cuBLAS or MKL as you see fit.

martin-frbg:
openblas_get_parallel() would return 2 for OpenMP, 1 for non-OpenMP, 0 for single-threaded. Maybe this should be put in the FAQ to make it more visible (although I am not sure if it is more convenient than a standard cmake check_library_exists test for blas_server_omp). And it might make sense to add the NUM_THREADS setting to the output of openblas_get_config()...

martin-frbg:
[...]

PR #1557, opened 2018-05-12
Add threading and OpenMP information to output

martin-frbg:
For #1416 and #1529, more information about the options OpenBLAS was built with is needed. Additionally we may want to add this data to the openblas.pc file (but not all projects use pkgconfig, and as far as I am aware the cmake module for accessing it does not make such "private" declarations available)


martin-frbg:
The pkg-config topic is addressed in #1559, and I found a quick hack to access arbitrary pkg-config variables from cmake described in https://bloerg.net/2015/03/06/pkg-config-variables-in-cmake.html

PR #1559, opened 2018-05-13
Add build-time configuration options to pkgconfig file

martin-frbg:


martin-frbg:
This adds an openblas_config variable to openblas.pc, that will look like:
USE_64BITINT= NO_CBLAS= NO_LAPACK= NO_LAPACKE= DYNAMIC_ARCH=1 NO_AFFINITY=1 USE_OPENMP= HASWELL MAX_THREADS=4
and can be queried by `pkg-config --variable openblas_config openblas` (the idea is taken from the qt_config line found in some qt-related pc files)

PR #1581, opened 2018-05-31
Fix paths to LIN and EIG tests

martin-frbg:
should fix 1574


martin-frbg:
This is not optimal as it makes the OpenBLAS version of lapack_testing.py different from netlib - on the other hand, its current usage in the netlib build system is different already.

Issue #1593, opened 2018-06-04
_Atomic definition


diorcety:
Here https://github.com/xianyi/OpenBLAS/blob/develop/common.h#L653
ifndef is used to check if _Atomoc is defined. It doesn't seems to work with clang (6.0 for me).
Removing the 3 lines of this ifndef make the compilation succeed
https://releases.llvm.org/3.1/tools/clang/docs/LanguageExtensions.html#c_atomic

martin-frbg:
Hmm. Somehow that block looks reversed, maybe it would make more sense to write
```
#if (__STDC_VERSION__ >= 201112L)
#include <stdatomic.h>
#else
#ifndef _Atomic
#define _Atomic volatile
#endif
#endif

diorcety:
Indeed :)

dzyong:
Hi @martin-frbg 

The old code can work well with gcc (but the new code couldn't, because of https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65467), if it has some problem with clang, maybe we should distinguish compiler here.

martin-frbg:
Sorry, I had completely forgotten about the gcc issue that made this odd code necessary. As far as I can tell, this should have been fixed with gcc 7, so making it conditional on `#if defined(C_GCC) && (__GNUC__ < 7) ` should be sufficient ?

PR #1598, opened 2018-06-06
Restore _Atomic define before stdatomic.h for old gcc

martin-frbg:
see #1593

### DIFF
--- a/common.h
+++ b/common.h
@@ -663,6 +663,7 @@ __declspec(dllimport) int __cdecl omp_get_num_procs(void);
 #ifndef _Atomic
 #define _Atomic volatile
 #endif
+#endif
 
 #else
 #ifdef __ELF__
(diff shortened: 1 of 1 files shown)


dzyong:
I haven't found `C_GCC` and `__GNU_C__`, where do you get that?
I only find `__GNUC__`, it's used like this:

`#if __GNUC__ > 7`

martin-frbg:
C_GCC is defined by gcc I think - it is already used in several places in the file. `__GNU_C__` was just a misspelling in my issue comment (fixed now).

dzyong:
I compile this code on ubuntu 16.04 with gcc:
```
#include <stdio.h>

void main(){
#ifdef C_GCC
    printf("C_GCC defined\n");
#else
    printf("C_GCC not defined\n");
#endif
}
```
It print `C_GCC not defined`.

martin-frbg:
Ah, sorry. C_GCC is actually set by c_check (written to config.h)

PR #1604, opened 2018-06-07
Improve performance of GEMM for small matrices when SMP is defined.

oon3m0oo:
Always checking num_cpu_avail() regardless of whether threading will actually
be used adds noticeable overhead for small matrices when SMP is defined.
Most other uses of num_cpu_avail() do so only if threading will be used, so do 
the same here.


oon3m0oo:
Whoops, found a bug here... let me try again.

PR #1609, opened 2018-06-10
Create OpenBLASConfig.cmake in cmake builds as well

martin-frbg:
Patch provided by loriab in #1529

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,7 @@ option(BUILD_WITHOUT_LAPACK "Without LAPACK and LAPACKE (Only BLAS or CBLAS)" ON
 endif()
 option(BUILD_WITHOUT_CBLAS "Without CBLAS" OFF)
 option(DYNAMIC_ARCH "Build with DYNAMIC_ARCH" OFF)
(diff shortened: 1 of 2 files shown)


xsacha:
Thanks,
Will be good to see this in the next version.
I did the same thing here:
https://github.com/hunter-packages/OpenBLAS/commit/cfbd3563dbd1fa8f41f1547d6f642cee33e533e9
One thing I noticed after doing this is that the install path was not pushed to the Config
I needed to add this in a separate commit here:
https://github.com/hunter-packages/OpenBLAS/commit/fd4832ba3d74e67ad2adbf004e9603bea27c8892

martin-frbg:
Thanks. Unfortunately I am not sure how to address this, as your version seems to be quite different from loriab's and I am not sure what to look for in the resulting config file. His version of the cmake.in seems to be much more elaborate than yours, so perhaps this supplies the required information ?

xsacha:
Yes, this version is fine. It adds some things I didn't such as versioning.
Include path usually doesn't matter if you install to system.

martin-frbg:
Thanks again for the review.

PR #1611, opened 2018-06-10
Return a somewhat sane default value for L2 cache size if cpuid retur…

### LINKED ISSUE #1610: OpenBLAS should be resilient to reported zero-size L2 cache lines
On current versions of the "crostini" virtual machine on Google's Pixelbooks, the hypervisor erroneously reports all 0s for CPUID 0x80000006, including a zero-size L2 cache line. This causes openblas to spin forever on most operations:
https://bugs.chromium.org/p/chromium/issues/detail?id=848488

The bug here is clearly in the hypervisor, but I'd like to request that OpenBLAS could still be better-behaved here, and either die with an error or assume a default cacheline size.

martin-frbg:
…ned something unexpected

Fixes #1610, the KVM hypervisor on Google Chromebooks returning zero for CPUID  0x80000006, causing DYNAMIC_ARCH builds of OpenBLAS to hang

### DIFF
--- a/kernel/setparam-ref.c
+++ b/kernel/setparam-ref.c
@@ -647,7 +647,9 @@ static int get_l2_size_old(void){
       return 6144;
     }
   }
-  return 0;
+//  return 0;
+fprintf (stderr,"OpenBLAS WARNING - could not determine the L2 cache size on this system, assuming 256k\n");
+return 256;
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


nelhage:
I've confirmed that https://github.com/nelhage/openblas/tree/issue1610 resolves the issue on my chromebook, but that the patch on this PR so far is not enough. Thanks for the quick turnaround!

PR #1612, opened 2018-06-11
Fixed a few more unnecessary calls to num_cpu_avail.

oon3m0oo:
I don't have as many benchmarks for these as for gemm, but it should still
make a difference for small matrices.

### DIFF
--- a/interface/axpy.c
+++ b/interface/axpy.c
@@ -40,11 +40,11 @@
 #include "common.h"
 #ifdef FUNCTION_PROFILE
 #include "functable.h"
-#endif 
+#endif
 #if  defined(Z13)
 #define MULTI_THREAD_MINIMAL  200000
 #else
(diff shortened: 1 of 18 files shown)


martin-frbg:
@ashwinyes as this touches thunderx code...

oon3m0oo:
@ashwinyes gentle ping on this.  It shouldn't be too terribly controversial since the calls are essentially moot when threading isn't selected.  Please do check the logic, though!

martin-frbg:
Perhaps you could leave the arm64 parts of the PR until later if ashwinyes is busy ? While the changes do look harmless to me, I do not want to be responsible for breaking his Cavium server platform _another time_

ashwinyes:
Sorry for the delayed response.

I dont have any issue with this PR. 

You may go ahead with the commit.

martin-frbg:
Thanks for the review, merging now.

Issue #103, opened 2012-05-02
increase default threading cutoff


JeffBezanson:
With OPENBLAS_NUM_THREADS=1:

julia> a=rand(10,10); @time for i=1:100000; a*a; end
elapsed time: 0.11344099044799805 seconds

With OPENBLAS_NUM_THREADS=2:

julia> a=rand(10,10); @time for i=1:100000; a*a; end
elapsed time: 0.23067402839660645 seconds

This is openblas v0.1.1 on a core i5 650. There seems to be a slowdown for small matrices with threads enabled. In the same setup I see nice speedups for big matrices, e.g. over 100x100.

Can anything be done about this? Setting OPENBLAS_NUM_THREADS is easy, but we'd rather get the best of both worlds if possible :)

ViralBShah:
In Julia, we have been using 50 as the threshold which works well. Given that there is a way to address this in the build system, this could be closed.

martin-frbg:
I find it worrying that your threshold of 50 is so far from the current default, and if anything the threading behaviour has become more complicated with individual magic numbers used as threshold
values in specific BLAS functions. So I am hesitating to close this - at the very least, the description
of this parameter in Makefile.rule should be amended to make it clear that the sane range of values
extends to (or it seems some may think, begins around) 50. 
On reviewing this thread, there was also a comment by jiahao https://github.com/xianyi/OpenBLAS/issues/103#issuecomment-38914601 about improvements seen with choosing a different OpenMP loop scheduler in blas_server_omp, which sadly elicited no response. Thoughts ?

ViralBShah:
I'm not familiar with the openmp work, but one issue is that openmp is not available on all the platforms/OS combinations.

martin-frbg:
> one issue is that openmp is not available on all the platforms

I am certainly aware of that. I see OpenMP in this context as a subtopic that got buried quickly the first time it came up, and I would like to get some comment on it before it gets buried again. A speed gain
from switching away from the "static" scheduler would appear to suggest an unbalanced workload where some threads take longer than others  ?(Though the example given in https://github.com/JuliaLang/julia/issues/3239#issuecomment-38355077 was a bit vague in that it picked the `auto` scheduler without determining what the compiler chose).

martin-frbg:
My own experiments with the OMP scheduler were a bit inconclusive - which scheduler works best seems to depend very much on the functions and workload, so #1620 makes the choice available at
[...]

PR #1620, opened 2018-06-14
Add build-time option for OMP scheduler; document MULTITHREAD_THRESHOLD range

martin-frbg:
to address topics mentioned in #103

### DIFF
--- a/Makefile.rule
+++ b/Makefile.rule
@@ -60,6 +60,14 @@ VERSION = 0.3.1.dev
 # This flag is always set for POWER8. Don't modify the flag 
 # USE_OPENMP = 1
 
+# The OpenMP scheduler to use - by default this is "static" and you
+# will normally not want to change this unless you know that your main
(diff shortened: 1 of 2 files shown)


Makefile.rule
```diff
@@ -157,7 +165,8 @@ NO_AFFINITY = 1
 
 # If any gemm arguement m, n or k is less or equal this threshold, gemm will be execute
 # with single thread. You can use this flag to avoid the overhead of multi-threading
```
sandwichmaker: This documentation is not quite correct, since this threshold is now used a multiplier to a per function constant and has nothing to do directly with particular matrix dimensions.

martin-frbg: That's right, though as far as I tried to understand the transition the idea was that it was closer to the originally intended effect. Not sure how to phrase this, unless we want to call it a controversial magic number with dubious real-world equivalent :grinning:

PR #1625, opened 2018-06-18
Further improvements to memory.c.

oon3m0oo:
- Compiler TLS is now used only used when the compiler supports it
- If compiler TLS is unsupported, we use platform-specific TLS
- Only one variable (an index) is now in TLS
- We only access TLS once per alloc, and never when freeing
- Allocation / release info is now stored within the allocation itself, by
  over-allocating; this saves having external structures do the bookkeeping, and
  reduces some of the redundant data that was being stored (such as addresses)
- We never hit the alloc lock when not using SMP or when using OpenMP (that was
  my fault)
- Now that there are fewer tracking structures I think this is a bit easier to
  read than before

PR #1626, opened 2018-06-19
Use BLAS rather than CBLAS in test_fork.c

oon3m0oo:
This is handy for people not lapack.

### DIFF
--- a/utest/CMakeLists.txt
+++ b/utest/CMakeLists.txt
@@ -25,14 +25,12 @@ endif ()
 
 # known to hang with the native Windows and Android threads
 # FIXME needs checking if this works on any of the other platforms
-if (NOT NO_CBLAS)
 if (NOT USE_OPENMP)
 if (OS_CYGWIN_NT OR OS_LINUX)
 set(OpenBLAS_utest_src
   ${OpenBLAS_utest_src}
   test_fork.c
   )
-endif()
 endif()
 endif()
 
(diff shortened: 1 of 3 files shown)


sandwichmaker:
err @oon3m0oo meant people not building the cblas API.

martin-frbg:
Rerun the tests once 1625 is in, I suspect right now people not fork...

oon3m0oo:
It's still useful and caught a real bug that would have occurred if people do fork.  I've been running my testing with #1625 applied, but no hurry.

martin-frbg:
My point is just that CI for this PR will be using the state of the develop branch without 1625 applied, which is likely the reason for the fork utest failing on some platforms

martin-frbg:
Fails all the iNTERFACE64=1 tests it seems (the xcode one passes, but is not configured to run utest at all). Probably needs to be `BLASINT n` now rather than `int n` ?

oon3m0oo:
Ahh yep, you're probably right.

martin-frbg:
Travis seems to think otherwise, but has been acting up a bit lately.

martin-frbg:
It would probably make sense to remove the `not (NO_CBLAS)` restriction from the utest Makefile and CMakeLists.txt with the same PR ?

oon3m0oo:
Yep, done.

PR #1634, opened 2018-06-21
Fix data races reported by TSAN.

oon3m0oo:


oon3m0oo:
@fenrus75 you might want to take a look at this.  TSAN was unhappy with just volatile on the working array, even with the barriers.

oon3m0oo:
Given the comments in #1622 I've restricted this to the real positive in memory.c.  We can revisit level3_thread.c later.

Issue #1595, opened 2018-06-05
Large number of micro-architectures in DYNAMIC_ARCH=1


sandwichmaker:
I would like to reduce the number of micro-architectures being built when DYNAMIC_ARCH=1, is there a way to do that? My naive attempts at editing the makefiles led to linking failures.

martin-frbg:
Unless you want to remove the more modern architectures (which could be done by setting NO_AVX2=1
or even NO_AVX=1), there is currently no clean way to do this. You would need to hack driver/others/dynamic.c directly to either remove all traces of the architectures that you removed from the DYNAMIC_CORE list of Makefile.system, or `#define` their gotoblas_KERNEL function pointer to point to one of the remaining architectures.

sandwichmaker:
Thanks @martin-frbg that is what I gathered too, that this would require some kind of dynamic.c hacking. Having a fallback mechanism would be nice.

martin-frbg:
related to #791 - it has been noted before that the number of dynamic targets is getting out of hand (and it is causing trouble for the continuous integration tests as well). Which architectures would you like to remove (or see removed from the default list) ?

sandwichmaker:
@martin-frbg I only care about newer CPUs.

roughly Zen, Nehalem, sandybridge and haswell.

martin-frbg:
Hmmm. #791 had people asking to keep even Prescott, guess you can't please everybody. 

It should be easy to override the DYNAMIC_CORE list in Makefile.system with a user-defined selection from Makefile.rule (or the environment), but I do not see how to parse this list on the preprocessor level and add redirections for all "standard" types _not_ on this list in dynamic.c.
Perhaps creating a handful of predefined subsets would work, say DYNAMIC_NO_ANCIENT (
only nehalem and newer), DYNAMIC_NO_AMD and DYNAMIC_NO_INTEL ? At least that would
make it easy to add blocks of `#define gotoblas_SOMEKERNEL gotoblas_SOMEFALLBACK` to
dynamic.c

sandwichmaker:
@martin-frbg a few extra architectures are not a problem. So if you had a group that had Prescott, I would be okay with it. 

I am displaying my lack of knowledge about these architectures, but could we introduce an ordering of these microarchitectures which could be used for reasoning about fallback?

martin-frbg:
[...]

PR #1639, opened 2018-06-23
Add DYNAMIC_LIST option for user-defined list of dynamic targets

martin-frbg:
Hackish proof-of-concept for #1595. Just for discussion, not for immediate merging (seems the local fork I developed this with is out-of-sync as well)


driver/others/dynamic.c
```diff
@@ -338,6 +460,23 @@ static gotoblas_t *get_coretype(void){
 	  return &gotoblas_NEHALEM;
 	}	
 	return NULL;
+      case 6:
```
sandwichmaker: this should go in independent of this PR no?

Makefile.system
```diff
@@ -497,6 +497,14 @@ endif
 endif
 endif
 
+ifdef DYNAMIC_LIST
+override DYNAMIC_CORE = PRESCOTT $(DYNAMIC_LIST)
+XCCOMMON_OPT = -DDYNAMIC_LIST -DDYN_PRESCOTT
```
sandwichmaker, line 502: This uses PRESCOTT as the fallback for everything. 

I think there are two things to reason about here.

1. Should we expect the user to know what the lowest common denominator is when using the DYNAMIC_LIST option, instead of specifying it for them? I think the answer is yes. 

Then we have the error handling questin, , what to do in the case a CPU is detected at runtime which is worse in its capabilities than what the user specified as the minimum CPU. 

One thing to do would be what you are doing here, which is the safe thing, have everything fallback to PRESCOTT, the other would be to actually crash with an error which says that the library does not have support for the architecture detected.

I would argue that we should do the latter, since we should not paper over the user's poorly chosen list of arches. You should only use DYNAMIC_LIST if you know what you are doing.

2. Given the list of dynamic architectures given by the user, should we make an attempt to match each architecture that is NOT present in the list to fall back to its "nearest" archiecture in the DYNAMIC_LIST or should they all fallback to the lowest common denominator?

Doing this correctly would require specifying for each architecture a descending chain and building the logic to intersect this chain with DYNAMIC_LIST and picking the top element.

sandwichmaker:
Thanks for taking a whack at this @martin-frbg.

Makefile.system
martin-frbg: Crashing is not an option for a library in my opinion, as the user affected by it may be someone several steps remote from whoever made the initial decision. Crawling along at the speed of a Prescott cpu may be bad enough in practice. On the other hand, actually interpreting the list of architectures (with the limited means provided by the C preprocessor) to identify appropriate fallback targets for each and every possible combination seems disproportionate for a case that is not supposed to happen.

driver/others/dynamic.c
martin-frbg: That is why I mentioned the local tree being out of sync and this PR being just for discussion. That `case 6` for Cannon Lake has already been merged into the develop branch earlier.

sandwichmaker: ah nm me then.

Makefile.system
sandwichmaker: Fair point about crashing and the user maybe far removed from the decision maker.  I was taking a narrower view of this option, as an expert only thing.  

Would you be open to making the default fallback also configurable instead of hardcoding it to be PRESCOTT?

martin-frbg: I could add a DYNAMIC_FALLBACK option, but I am not really convinced yet that it would be useful. After all this would be needed only for cpus not on the list, and if we want to avoid crashing it would have to be suitable for any old system that OpenBLAS supports - which basically leaves only Prescott I think. Setting up optimum fallback for each cpu would mean long strings of ifndef(THISCPU) ... ifdef(PREVIOUS_GENERATION)...elseif(NEXT_OLDER)....elseif...elseif...

sandwichmaker: Fair enough. what other kind of feedback are you looking for before refining/finalizing this PR?

martin-frbg: Well, do you think it would even address your wish for a short, user-defined list (the fallback issue nonwithstanding, perhaps I could try to fallback to one or two intermediate types, e.g. Sandybridge and Nehalem if available before defaulting to Prescott) ?

sandwichmaker: Yes I think it will. 
Your argument for PRESCOTT as the fallback is cogent and I agree with it. I was trying to over-optimize. 
Let me try this patch on my internal build and get back to you later today.

martin-frbg: Revised PR will now fall back to Sandybridge or Nehalem if available, with Prescott only being used as a last resort (except for old cpus where it remains the only option). Also updated from develop to get rid of the spurious unrelated changes.

PR #1642, opened 2018-06-25
Rewrite &= -> = and simplify the initial blocking phase.

oon3m0oo:
@fenrus75 it looked to me like the initial read spin could be combined in the loop with the write, since it was actually checking more values than it needed to, but correct me if my logic is wrong.

### DIFF
--- a/driver/level3/level3_thread.c
+++ b/driver/level3/level3_thread.c
@@ -344,12 +344,6 @@ static int inner_thread(blas_arg_t *args, BLASLONG *range_m, BLASLONG *range_n, 
     div_n = (n_to - n_from + DIVIDE_RATE - 1) / DIVIDE_RATE;
     for (js = n_from, bufferside = 0; js < n_to; js += div_n, bufferside ++) {
 
-      /* Make sure if no one is using workspace */
-      START_RPCC();
-      for (i = 0; i < args -> nthreads; i++)
-	while (job[mypos].working[i][CACHE_LINE_SIZE * bufferside]) {YIELDING;MB;};
-      STOP_RPCC(waiting1);
-
 #if defined(FUSED_GEMM) && !defined(TIMING)
 
       /* Fused operation to copy region of B into workspace and apply kernel */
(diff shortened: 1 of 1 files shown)


driver/level3/level3_thread.c
```diff
@@ -462,14 +485,16 @@ static int inner_thread(blas_arg_t *args, BLASLONG *range_m, BLASLONG *range_n,
...
 #ifdef TIMING
           ops += 2 * min_i * MIN(range_n[current + 1]  - js, div_n) * min_l;
 #endif
-          
+
           /* Clear synchronization flag if this thread is done with region of B */
           if (is + min_i >= m_to) {
+            ANNOTATE_IGNORE_WRITES_BEGIN();
             job[current].working[mypos][CACHE_LINE_SIZE * bufferside] &= 0;
+            ANNOTATE_IGNORE_WRITES_END();
```
fenrus75: the &= 0 in the line above is odd btw, it makes the write LESS atomic if notihing else.
should just really be an assignment

fenrus75 (commented):
If you use LLVM and know it enough... would you mind putting the appropriate compiler barriers for MB/WMB in place? without those I can see the analysis getting double confused

martin-frbg:
Seems you would need to include <valgrind/drd.h> (or its contents) for this to work ? Not sure if that would be worth it, just to suppress a warning from one specific tool.

oon3m0oo:
@martin-frbg it's up to you.  Worst case we can have a local patch on our side that adds the annotations.

@fenrus75 The &= 0 has been there for some time, I actually assumed you had put it in there knowing some secret Intel sauce!  If the right thing to do here is not have the annotations but still move the loop and eliminate the &= 0 I can do that here, instead.

martin-frbg:
That &=0 line seems to come from the original GotoBLAS code...
(Undecided about the annotations, but in their current form they fail all CI tests

fenrus75:
&= 0 is basically equal to = 0

but because of the atomics /volatile I think it'll make the compiler
generate worse, less atomic code.

so I did not touch it before, but I think it should be = 0 so that we don't
have read-modifiy-write just write.

oon3m0oo:
Alright this is now the bare minimum.  I'll set up a local process for adding the annotations.

PR #1654, opened 2018-06-30
Add compiler option to avx512 test and hide test output

martin-frbg:


brianborchers:
If I try to build the development version (downloaded 5 minutes ago, so I believe it should include this change) using the default gcc 5.4 on my Ubuntu 16.04 Skylake system, the build fails with:
 
cc -O2 -DMAX_STACK_ALLOC=2048 -fopenmp -Wall -m64 -DF_INTERFACE_GFORT -fPIC -DSM
P_SERVER -DUSE_OPENMP -DNO_WARMUP -DMAX_CPU_NUMBER=8 -DMAX_PARALLEL_NUMBER=1 -ma
rch=skylake-avx512 -DASMNAME=sasum -DASMFNAME=sasum_ -DNAME=sasum_ -DCNAME=sasum
 -DCHAR_NAME=\"sasum_\" -DCHAR_CNAME=\"sasum\" -DNO_AFFINITY -I.. -I. -UDOUBLE  
-UCOMPLEX -c asum.c -o sasum.o
dot.c:1:0: error: bad value (skylake-avx512) for -march= switch
 /*********************************************************************/
 ^
axpy.c:1:0: error: bad value (skylake-avx512) for -march= switch
 /*********************************************************************/
 ^
.....

martin-frbg:
This change is not yet in develop (as the PR has not been merged), but I believe your report confirms that it is needed to avoid a build failure with older gcc or binutils. (Noticed this when I did a DYNAMIC_ARCH build to reproduce sandwichmaker's problem with static linking - this problem would not have arisen earlier as automatic addition of skylake-avx512 option for skylake is rather new)

PR #1661, opened 2018-07-03
Ensure that the gotoblas lookup table is always initialized.

oon3m0oo:
It is possible to build a program that calls a non-GEMM OpenBLAS routine from
a static initializer. Since the order of initialization is undefined, and even
less defined when using __attribute__((constructor)) in one TU and a C++ static
initializer in another TU, it can happen (and does, unfortunately) that
gotoblas_init is not called before the first BLAS routine. This results in a
segfault when trying to index into the gotoblas table.

The solution I've put here is indirection: rather than directly using the table
use an inlined function to first check if it's been initialized. Since it will
only not have been done once, the branch prediction should keep things fast.

### DIFF
--- a/common.h
+++ b/common.h
@@ -641,6 +641,18 @@ void gotoblas_dynamic_quit(void);
 void gotoblas_profile_init(void);
 void gotoblas_profile_quit(void);
 
+#ifdef DYNAMIC_ARCH
+/* Do not reference gotoblas directly, instead use get_gotoblas(), below. This
+is to prevent initialization order issues where an initializer attempts to call
+into OpenBLAS before gotoblas_dynamic_init has been called. */
+extern gotoblas_t *gotoblas;
+static __inline__ gotoblas_t *get_gotoblas() {
+  if (gotoblas) return gotoblas;
+  gotoblas_dynamic_init();
+  return gotoblas;
+}
+#endif
+
 #ifdef USE_OPENMP
 
 #ifndef C_MSVC
@@ -652,7 +664,7 @@ __declspec(dllimport) int __cdecl omp_get_num_procs(void);
 #endif
 
 #if (__STDC_VERSION__ >= 201112L)
-#if defined(C_GCC) && ( __GNUC__ < 7) 
+#if defined(C_GCC) && ( __GNUC__ < 7)
 // workaround for GCC bug 65467
 #ifndef _Atomic
 #define _Atomic volatile
--- a/common_c.h
+++ b/common_c.h
@@ -229,49 +229,49 @@
 #define CIMATCOPY_K_CTC         cimatcopy_k_ctc
 #define CIMATCOPY_K_RTC         cimatcopy_k_rtc
 
-#define CGEADD_K                cgeadd_k 
+#define CGEADD_K                cgeadd_k
 
 #else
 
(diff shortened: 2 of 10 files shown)


martin-frbg:
Looks like the definition of LIKELY_ONE in memory.c needs to be done unconditionally (the failing CI tasks are those that do not have DYNAMIC_ARCH set)...

martin-frbg:
Oh, and please mention issue 1652 in your comment so that it gets linked to the PR and vice versa.

oon3m0oo:
I've removed LIKELY_ONE. It seems like it's more trouble than it's worth:
https://godbolt.org/g/5G9PZ6

It's generating more instructions, and not seeming to do anything particularly intelligent about the branch.

oon3m0oo:
@martin-frbg or @fenrus75 any thoughts here?

martin-frbg:
No time to look at this right now, and frankly I am hoping for some less invasive feature to put into 0.3.2 and release that first, so that people have a somewhat usable release to work with again.

oon3m0oo:
No worries, I might want to find a simpler way to do this, anyway.  Let's hold off this commit.

oon3m0oo:
I've found an internal solution for this, but it's worth recording the problem for posterity.

We have a situation where a dynamic library loaded by a C++ program has a static initializer that calls a BLAS function. This gets complicated.

The normal order of operations when a linux program starts is this:
1) the loader loads the executable, and starts loading dynamic libraries it depends on
2) the loader loads each library and calls its initializers
3) the loader calls all initailizers in the main executable, including gotoblas_init()
4) main() is executed

In our situation, we have a dynamic library with an initializer that uses BLAS:
1) the loader loads the executable, and starts loading dynamic libraries it depends on
2) the loader loads the library with the initializer and starts executing it
3) the initializer calls into OpenBLAS
4) gotoblas is NULL since it has not yet been initialized

Digging around a bit, it seems that GCC/clang do not guarantee the order of constructors across translation units.  Within the same TU, you can specify the priority (as memory.c does), but there is no global ordering, and certainly not across libraries.  In general, this is what you want, since you want your libraries to initialized themselves.  Things get complicated with shared libraries, though, and how they can depend on code elsewhere that isn't necessarily in another shared library.

The solution for us was to ensure that in this situation we also use OpenBLAS as a shared library, so that it gets loaded first (and thus gotoblas_init is called) before any initializer that depends on it in another module.  For fully static binaries, we still use a static library.

Who knew dynamic linking could be so hard? =)

martin-frbg:
Maybe this should be somewhere in the wiki... do other compilers guarantee the order of constructors across translation units, or is this perhaps even a c-something standards issue ?

oon3m0oo:
As far as I know no compilers do a global ordering guarantee.  Certainly gcc and clang don't.

jktomer:
No, the order of static initializers across translation units is *expressly* unspecified, and it is undefined behavior to depend on any given order. Even with a given compiler the order may vary from one build to the next.

The only way to ensure that some variable of static storage duration is never accessed before it is initialized, is not to expose it outside the single translation unit where it is initialized. (Unfortunately that would prevent inlining calls to get_gotoblas() without link-time optimization, which may be unacceptable.)

Issue #1677, opened 2018-07-12
windows build failed. snprintf not supported in vs2013


yplypl:
snprintf was called in openblas_get_config.c:81 of function:CNAME(), but not supported in windows, should change to _snprintf instead.

martin-frbg:
Thanks for the report, should be fixed by PR #1678. I assume you are using a version of Visual Studio earlier than about 2014, or are there other compilers on Windows that have this limitation ?

yplypl:
yes, u are right, i am using vs2013, i just tried vs2015, snprintf is supported success.

yplypl:
i notice that, snprintf have been defined as _snprintf_s, but in vs2013, those two function does not have the same parameters.
![_snprintf_s](https://user-images.githubusercontent.com/6971032/42622476-0a6baf5e-85f3-11e8-9f59-a43c3038e9d9.png)
![_snprintf](https://user-images.githubusercontent.com/6971032/42622478-0aa0a74a-85f3-11e8-9a56-67eeb871239b.png)

martin-frbg:
Interesting. thanks. I just reused a declaration from utest/ctest.h, but it seems that one never actually gets compiled at all (one call is inside `#ifdef APPLE', the other inside `#ifdef CTEST_SEGFAULT` which is expressly unset for MINGW and MSVC). So the mismatched arguments were never flagged. I'll fix both with a new PR.

PR #1678, opened 2018-07-12
Define snprintf for older versions of MSVC

martin-frbg:
for #1677


yplypl:
i notice that, snprintf have been defined as _snprintf_s, but in vs2013, those two function does not have the same parameters. 
_snprintf_s
_snprintf

Issue #1672, opened 2018-07-07
Test failures on mips64 with OpenBLAS 0.3.1


svillemot:
Hi,

The Debian package for OpenBLAS 0.3.1 fails some of the "u"-tests on mips64, see:
https://buildd.debian.org/status/fetch.php?pkg=openblas&arch=mips64el&ver=0.3.1%2Bds-1&stamp=1530702292&raw=0

Note that the tests were passing with 0.2.20, so this is a regression.

Best,

Sébastien

svillemot:
Note that the Debian mips64el port uses the N64 ABI with hardware floating point and the MIPS64R2 ISA.

The OpenBLAS package is currently built with TARGET=SICORTEX (I am not 100% sure whether this is the right target for the generic package though).

martin-frbg:
Thanks for the report. I suspect the 0.2.20 utests passing was mostly (or entirely) due to there being fewer of them - most were reactivated in 0.3.0. (Which makes this ticket basically a continuation of #1469, unfortunately the Fedora folks do not build-test mips and I have not acquired a mips64 system yet.) Hopefully my fixes for mips32 will apply, in which case there should be a PR for you to test later today...

martin-frbg:
Not an easy fix for me unfortunately. I note mips64 SAXPY/DAXPY got fixed in #891 but I am not sure I can transfer that fix to CAXPY/ZAXPY with my sketchy knowledge of assembly language. DSDOT is just another case were it was assumed that converting the DDOT result to double precision would be enough, completely disregarding the loss of precision in the earlier summations. Again I am not sure yet how to translate that knowledge into a fix for the assembly implementation. ROT and SWAP both have no easy to spot early exits for the zero increment case like the ARMV7 implementation had, but 
maybe rearranging the final statements in zswap.S after the example of the working swap.S will be sufficient. (Note though that the utests are testing corner cases, so "most" regular code would still work despite these failures)

martin-frbg:
DSDOT is the only one I managed to fix, the others I plan to replace with their mips C counterparts that are already used in the I6400, I6500 and P6600 targets. Unfortunately the person who contributed the fix  for axpy.S that I failed to adapt for zaxpy.S no longer has an account on github.

PR #1683, opened 2018-07-15
Fix utest failures with the default MIPS64 target

martin-frbg:
fixes (or works around) #1672 by retiring most of the affected (original GotoBLAS) kernels in favor of the generic C implementations that are already used by the I6400, I6500 and P6600 targets. Adds argument
promotion to dot.S in DSDOT mode to avoid loss of precision.


martin-frbg:
Meh. Misplaced the updated kernel file.

PR #1684, opened 2018-07-15
Work around utest failures in the MIPS64 SICORTEX target

### LINKED ISSUE #1672: Test failures on mips64 with OpenBLAS 0.3.1

martin-frbg:
by switching from the original GotoBLAS2 kernels to the mips C implementations already used by I6400, I6500, P6600. Promote input arguments of dot to double in the DSDOT case. Fixes #1672.

### DIFF
--- a/kernel/mips64/KERNEL
+++ b/kernel/mips64/KERNEL
@@ -1,3 +1,12 @@
+CAXPYKERNEL = ../mips/zaxpy.c
+ZAXPYKERNEL = ../mips/zaxpy.c
+SROTKERNEL = ../mips/rot.c
+DROTKERNEL = ../mips/rot.c
+CROTKERNEL = ../mips/zrot.c
+ZROTKERNEL = ../mips/zrot.c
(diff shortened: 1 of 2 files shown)


jiangtao9999:
+ZSWAPKERNEL = ../mips/zwap.c
Is not "zswap.c" ?

martin-frbg:
Yes, thanks, that is why I did not yet commit this. The CI setup cannot catch this as it does not build for all architectures, and mips64 emulation via qemu is slow on my machine.

svillemot:
@martin-frbg Thanks!

PR #1703, opened 2018-07-30
Set EXPORT_NAME to match OpenBLASConfig.cmake

wsttiger:
This fixes error in find_package(OpenBLAS):

```
CMake Error at prereqs/share/cmake/OpenBLAS/OpenBLASConfig.cmake:93 (get_property):
  get_property could not find TARGET OpenBLAS::OpenBLAS.  Perhaps it has not
  yet been created.
Call Stack (most recent call first):
  src/targets/cpu/CMakeLists.txt:1 (find_package)
```

This is because the CMake target is lowercase `openblas` and it should be `OpenBLAS`. Rather than changing the CMake target name, I changed the export name to `OpenBLAS`.


brada4:
Probably calling side should be changed? Or at least mentioned?

martin-frbg:
Looks good to me, but I am not that familar with cmake - in fact I probably introduced this problem in #1609 with an incomplete adaption of  @loriab 's suggestion from #1529 @xsacha

xsacha:
Probably this is needed then:
https://github.com/hunter-packages/OpenBLAS/commit/877fb38a15150fd0ed3b42ea05288710b6f0d294

I have just been applying that to the top of Head to get my config working.

martin-frbg:
Probably a very silly question, but is the created library still named libopenblas.so (like it is in gmake builds), with your change to OpenBLAS_LIBNAME ?

pfultz2:
> Probably a very silly question, but is the created library still named libopenblas.so (like it is in gmake builds), with your change to OpenBLAS_LIBNAME ?

Yes, it is. Only the exported cmake target is called `OpenBLAS::OpenBLAS`

xsacha:
Yeah that's why I didn't do a pull request. The reason I changed it was the same but just not the right spot for consistency.

wsttiger:
The library is still named `libopenblas.so`.

xsacha:
On the gmake version you get a library named by the processor it targets (and symlinked). Is it worthwhile replicating this?

Thanks for these changes. Means I can use OpenBLAS entirely unmodified now.

martin-frbg:
Thanks to all involved. I am not sure if replicating the target-specific naming would still be worthwile, but it was certainly convenient when there was no easy way to query an installed version for its properties.

wsttiger:
This PR allows me on integrate OpenBLAS using the following cmake commands:

`find_package(OpenBLAS)`
`target_link_libraries(myexe OpenBLAS::OpenBLAS)` where the library name remains `libopenblas.so`.

PR #1707, opened 2018-08-02
Haiku supporting patches

extrowerk:


extrowerk:
Btw; it is from the HaikuPorts repo.
Tested, all test passed.
Thx for the merge.

PR #1712, opened 2018-08-04
replace several assembly implementations with an intrinsics based implementation and add AVX512

fenrus75:
This patch series does, for a range of algorithms the following

Step 1: Replace "evil" gcc inline asm syntax with %0/%1 etc with named variables (%[name[) to make the assembly more readable
Step 2: Replace the now-readable assembly construct with the same algorithm written using C intrinsics
Step 3: Add AVX512 support to the now easier-to-change C intrinsics version

C intrinsics compile to pretty much the same code as the assembly version, but it lets the compiler do register allocation, but more importantly, it makes the non-math part of the code much easier to understand in terms of what is going on and is less fragile in general.

### DIFF
--- a/kernel/x86_64/daxpy_microk_haswell-2.c
+++ b/kernel/x86_64/daxpy_microk_haswell-2.c
@@ -25,54 +25,49 @@ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 *****************************************************************************/
 
+
+
+#ifndef __AVX512CD__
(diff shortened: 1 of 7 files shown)


fenrus75:
Hi,

before I keep adding more of these... could you give an opinion if going down the intrinsics path is ok with you?

martin-frbg:
No opinion on the asm vs intrinsics topic - you are the current expert so use what suits you best. But can you please prepare separate PRs for files that are not directly related, rather than one huge "replace x86_64 openblas with something even weirder" code dump ?

martin-frbg:
Also not all of the CI failures are necessarily spurious... if you just keep piling on other changed files, chances are the entire commit will only work with the particular compiler and operating system you used to develop it.

fenrus75:
yeah I'll split it up no problem

Issue #1710, opened 2018-08-03
incorrect usages of int for blasint with 64-bit API


stevengj:
When I compile openblas with the 64-bit API, I get a lot of warnings about code which incorrectly assumes that `blasint` is the same as `int`.

There are dozens of warnings like:
```c
gemv.c:202:48: warning: absolute value function 'abs' given an argument of type
      'blasint' (aka 'long') but has parameter of type 'int' which may cause
      truncation of value [-Wabsolute-value]
  if (beta != ONE) SCAL_K(leny, 0, 0, beta, y, abs(incy), NULL, 0, NULL, 0);
```
Maybe you need to define a `BIABS(i)` macro that expands to `abs` for the 32-bit API and `labs` (or `llabs` if needed) for the 64-bit API?

I also get a few other miscellaneous cases that use the wrong integer type:
```c
lapack/laswp.c:100:8: warning: incompatible function pointer types passing 'int
      (*)(BLASLONG, BLASLONG, BLASLONG, float, float *, BLASLONG, float *,
      BLASLONG, blasint *, BLASLONG)' (aka 'int (*)(long, long, long, float,
      float *, long, float *, long, long *, long)') to parameter of type
      'int (*)()' [-Wincompatible-function-pointer-types]
                     laswp[flag], nthreads);
                     ^~~~~~~~~~~
../common_thread.h:180:39: note: passing argument to parameter 'function' here
                       void *c, BLASLONG ldc, int (*function)(), int threads);

lapack/zlaswp.c:99:80: warning: incompatible function pointer types passing 'int
      (*)(BLASLONG, BLASLONG, BLASLONG, float, float, float *, BLASLONG, float
      *, BLASLONG, blasint *, BLASLONG)' (aka 'int (*)(long, long, long, float,
      float, float *, long, float *, long, long *, long)') to parameter of type
      'int (*)()' [-Wincompatible-function-pointer-types]
  ...k1, k2, dummyalpha, a, lda, NULL, 0, ipiv, incx, laswp[flag], nthreads);
                                                      ^~~~~~~~~~~
../common_thread.h:180:39: note: passing argument to parameter 'function' here
                       void *c, BLASLONG ldc, int (*function)(), int threads);

1 warning generated.
xerbla.c:16:69: warning: format specifies type 'int' but the argument has type
      'blasint' (aka 'long') [-Wformat]
      fprintf(stderr, "Parameter %d to routine %s was incorrect\n", p, rout);
```

[...]

PR #1713, opened 2018-08-04
Introduce blasabs macro and use it to switch between abs and labs for INTERFACE64

martin-frbg:
fixes (most of) #1710

### DIFF
--- a/common.h
+++ b/common.h
@@ -253,8 +253,10 @@ typedef unsigned long BLASULONG;
 
 #ifdef USE64BITINT
 typedef BLASLONG blasint;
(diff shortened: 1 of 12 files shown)


common.h
```diff
@@ -253,8 +253,10 @@ typedef unsigned long BLASULONG;
 
 #ifdef USE64BITINT
 typedef BLASLONG blasint;
+#define blasabs(x) labs(x)
```
stevengj, line 256: This doesn't seem like it will work on Windows, which is an LLP64 architecture where `long == int` and you have to use `llabs`.

PR #1718, opened 2018-08-07
create a C intrinsics version for DDOT

fenrus75:
This creates a C+intrinsics version for the Haswell DDOT function; this is easier to read/modify, especially for the non-math portion of the code.
It also allows the compiler to perform more optimal register allocation and even inlining into the caller.

(avx512 version to follow once the CI passes)

### DIFF
--- a/kernel/x86_64/ddot_microk_haswell-2.c
+++ b/kernel/x86_64/ddot_microk_haswell-2.c
@@ -43,18 +43,18 @@ static void ddot_kernel_8( BLASLONG n, FLOAT *x, FLOAT *y, FLOAT *dot)
 
 	".p2align 4				             \n\t"
 	"1:				             \n\t"
-        "vmovups                  (%2,%0,8), %%ymm12         \n\t"  // 2 * x
-        "vmovups                32(%2,%0,8), %%ymm13         \n\t"  // 2 * x
-        "vmovups                64(%2,%0,8), %%ymm14         \n\t"  // 2 * x
-        "vmovups                96(%2,%0,8), %%ymm15         \n\t"  // 2 * x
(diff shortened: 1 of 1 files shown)


martin-frbg:
Seems gcc 4.6.3 does not understand "pragma GCC target(avx2)" ?

fenrus75:
time to do archaeology to see what it does support

martin-frbg:
Does it actually need the pragma to enable parsing of immintrin.h ? IIRC you would not end up in that file when avx2 is not available (NO_AVX2 is set) anyway.

fenrus75:
it does not need it if the compiler flags have AVX either explicit or in
the -march= option.

but afaics, with the HASWELL openblas target we don't pass -march=haswell
nor anything close to that

fenrus75:
one of the options is to make a skylakex version of this file and use a
newer compiler for that but that is just more, not less, duplication of the
same base kernel

I'm trying one more thing, and will ask some of our gcc guys if they still
remember 4.6.3 later today, but if all else fails I'll go the skylakex route

martin-frbg:
You probably wanted to write `#if (defined(__GNUC__) && __GNUC__ < 6)` ?

martin-frbg:
> #if )defined

err... wrong phase of moon ?

fenrus75:
that teaches me to not do things before coffee on the train

martin-frbg:
No luck, now mingw hits an ICE that could be related to http://gcc.gnu.org/bugzilla/show_bug.cgi?id=57848

fenrus75:
I think I'm going to go with plan B

1) Only do this for SkylakeX
2) In the skylakeX file, check for gcc "new enough" and if not, #include
the unmodified haswell

for some value of new enough

PR #1721, opened 2018-08-08
Add an AVX512 enabled DDOT function

fenrus75:
written in C intrinsics for best readability.
(the same C code works for Haswell as well)

For logistical reasons the code falls back to the existing
haswell AVX2 implementation if the GCC compiler is not new enough

### DIFF
--- a/kernel/x86_64/ddot.c
+++ b/kernel/x86_64/ddot.c
@@ -37,8 +37,10 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "ddot_microk_piledriver-2.c"
 #elif defined(NEHALEM)
 #include "ddot_microk_nehalem-2.c"
-#elif defined(HASWELL) || defined(ZEN) || defined (SKYLAKEX)
+#elif defined(HASWELL) || defined(ZEN)
 #include "ddot_microk_haswell-2.c"
+#elif defined (SKYLAKEX)
+#include "ddot_microk_skylakex-2.c"
 #elif defined(SANDYBRIDGE)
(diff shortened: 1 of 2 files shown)


fenrus75:
Lets see how the CI / gcc interactions go on this approach

martin-frbg:
Looks good so far, but I think your gcc version check would exclude llvm (where clang disguises itself as `__GNUC__` with major version 5 as far as I know). Or is that intentional ?

fenrus75:
Now that we have at least one passing pattern I'll try to widen this a bit; adding LLVM

if this is passing all CI combos I'll go poke at the other vector functions.

I'm conservative in only picking up recent compilers given the CI ICE's etc

fenrus75:
ok I think this will be it as pattern of compilers that work

martin-frbg:
Great. So ready to merge this PR, or do you plan to add more dots to it ?

fenrus75:
at this point I think I will do separate PRs for each of the functions to separate the CI and make your merging decision easier.... so this one is done

brianborchers:
It would be helpful if documentation somewhere mentioned the required versions of compilers to get the AVX512 support.

martin-frbg:
That could be put in README.md I think (and I guess it is a good opportunity to replace the faq entry for "what are the plans for `<list of outdated cpus>`" that has been looking silly for a while). Perhaps the compiler requirements can be relaxed later e.g. the build failure seen earlier with 4.6.3 appears to have been fixed post gcc 4.9.0. (And apparently it could be worked around even in earlier versions by providing a suitable -march option that switched on sse4 support - it might make sense to add appropriate -march flags for individual target-specific codes automatically instead of building without any by default)

Issue #1720, opened 2018-08-07
GIMP hang / deadlock in get_memory_table / blas_thread_shutdown


amurzeau:
Hi,

There is a bug open in Debian related to gimp 2.10.2 and openblas 3.2: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=903514.

Depending on the machine and environment used, gimp can deadlock at startup because of a deadlock inside glibc.
I'm forwarding what I wrote for the Debian bug tracker:

Using gdb to find where it hung ([gimp-gdb.txt](https://gist.github.com/amurzeau/c0c1bae139388cb6656e471c7b90ac88)) gives threads waiting on
a lock while doing thread-local related stuff and the main thread is in
the process of dl_close-ing openblas waiting the threads to exit using
pthread_join.

It seems that the lock used in `tls_get_addr_tail` [0] is the same as
the one locked by `_dl_close` [1].
A recursive lock is used but here it does not help as the thread calling
`tls_get_addr_tail` and `_dl_close` are not the same.

This deadlock may not happen everytime, in my case, the openblas threads
are still initializing while `_dl_close` is called.

Given this, I think the offending commit in openblas is bf40f806 [2]
which add TLS variables to avoid locking. But many change were done
since then.

One of related bug report is [3] which seems to indicate that the locks
handling is not easy inside glibc.

There were an attempt to fix deadlocks between `tls_get_addr` and a
`_dl_close` of a module whose finalizer joins with that thread [4].

So I see these possibles solutions:
 * Add a breaks between gimp and openblas
 * Disable TLS in openblas build (if possible, but this would cause a
performance loss for users that use openblas without gimp)
 * Patch glibc to not deadlock (but this seems not easy to do at all)

[0] https://github.com/bminor/glibc/blob/glibc-2.27/elf/dl-tls.c#L761
[1] https://github.com/bminor/glibc/blob/glibc-2.27/elf/dl-close.c#L812

[2]
https://github.com/xianyi/OpenBLAS/commit/bf40f806efa55c7a7c7ec57535919598eaeb569d#diff-31f8d4e8863583d95bf2f9529f83844e
[4] https://sourceware.org/ml/libc-alpha/2015-06/msg00062.html

amurzeau:
As I see the implementation of glibc, it seems impossible to guarantee that no deadlock will happen when dl_closing openblas.

I don't think disabling TLS in openblas is a good solution as it would cause a performance hit even for users that don't use gimp.

[...]

PR #1726, opened 2018-08-10
Allow overriding USE_COMPILER_TLS (formerly HAS_COMPILER_TLS).

oon3m0oo:
This is useful, for example, for targets where glibc has deadlock issues with its built-in TLS implementation, as described in #1720.

### DIFF
--- a/driver/others/memory.c
+++ b/driver/others/memory.c
@@ -471,26 +471,27 @@ struct alloc_t {
    for an auxiliary tracking structure. */
 static const int allocation_block_size = BUFFER_SIZE + sizeof(struct alloc_t);
 
+#if !defined(USE_COMPILER_TLS)
 /* Clang supports TLS from version 2.8 */
 #if defined(__clang__) && __clang_major__ > 2 || \
     (__clang_minor__ == 2 || __clang_minor__ == 8)
-#define HAS_COMPILER_TLS
+#define USE_COMPILER_TLS 1
 #endif
 
 /* GCC supports TLS from version 4.1 */
 #if !defined(__clang__) && defined(__GNUC__) && \
     (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 1))
-#define HAS_COMPILER_TLS
+#define USE_COMPILER_TLS 1
 #endif
 
 /* MSVC supports TLS from version 2005 */
 #if defined(_MSC_VER) && _MSC_VER >= 1400
-#define HAS_COMPILER_TLS
+#define USE_COMPILER_TLS 1
 #endif
 
 /* Versions of XCode before 8 did not properly support TLS */
 #if defined(__apple_build_version__) && __apple_build_version__ < 8000042
-#undef HAS_COMPILER_TLS
+#define USE_COMPILER_TLS 0
 #endif
 
 /* Android NDK's before version 12b did not support TLS */
@@ -501,9 +502,10 @@ static const int allocation_block_size = BUFFER_SIZE + sizeof(struct alloc_t);
 #if defined(__ANDROID__) && defined(__clang__) && defined(__NDK_MAJOR__) && \
     defined(__NDK_MINOR__) &&                                               \
     ((__NDK_MAJOR__ < 12) || ((__NDK_MAJOR__ == 12) && (__NDK_MINOR__ < 1)))
-#undef HAS_COMPILER_TLS
+#define USE_COMPILER_TLS 0
 #endif
 #endif
+#endif /* !USE_COMPILER_TLS */
 
 /* Holds pointers to allocated memory */
 #if defined(SMP) && !defined(USE_OPENMP_UNUSED)
@@ -511,7 +513,7 @@ static const int allocation_block_size = BUFFER_SIZE + sizeof(struct alloc_t);
    server plus the number of threads in the thread pool */
 #  define MAX_ALLOCATING_THREADS MAX_CPU_NUMBER * 2 * MAX_PARALLEL_NUMBER * 2
 static int next_memory_table_pos = 0;
-#  if defined(HAS_COMPILER_TLS)
+#  if USE_COMPILER_TLS
 /* Use compiler generated thread-local-storage */
 static int THREAD_LOCAL local_memory_table_pos = 0;
 #  else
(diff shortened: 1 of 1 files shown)


oon3m0oo:
Travis failures seem to get apt-get install issues...?

martin-frbg:
Travis has been a bit unstable lately, restarting the failed jobs usually works.

martin-frbg:
Not sure if that define gets passed through the Makefiles automatically, and perhaps USE_COMPILER_TLS should default to 0 on glibc systems ? (Is there any major advantage in using compiler vs pthreads tls where both are available ? I think so far we simply do not know if the lockups are limited to glibc, or if musl etc are similarly affected)

oon3m0oo:
Actually, this whole thing is a bit fishy, I'm going to respond on the other thread.

driver/others/memory.c
```diff
@@ -501,17 +502,18 @@ static const int allocation_block_size = BUFFER_SIZE + sizeof(struct alloc_t);
 #if defined(__ANDROID__) && defined(__clang__) && defined(__NDK_MAJOR__) && \
     defined(__NDK_MINOR__) &&                                               \
     ((__NDK_MAJOR__ < 12) || ((__NDK_MAJOR__ == 12) && (__NDK_MINOR__ < 1)))
-#undef HAS_COMPILER_TLS
+#define USE_COMPILER_TLS 0
```
amurzeau, line 505: Same here: possible redefinition of `USE_COMPILER_TLS` as `__clang__` is checked and will cause that define to be set to 1 at line 478.

driver/others/memory.c
```diff
@@ -471,26 +471,27 @@ struct alloc_t {
...
 /* MSVC supports TLS from version 2005 */
 #if defined(_MSC_VER) && _MSC_VER >= 1400
-#define HAS_COMPILER_TLS
+#define USE_COMPILER_TLS 1
 #endif
 
 /* Versions of XCode before 8 did not properly support TLS */
 #if defined(__apple_build_version__) && __apple_build_version__ < 8000042
-#undef HAS_COMPILER_TLS
+#define USE_COMPILER_TLS 0
```
amurzeau, line 494: Is it possible to have this condition match while also have a previous one match ?
This would mean a previous line (like 478) does `#define USE_COMPILER_TLS 1` and then this one redefine it to 0 causing a compiler warning or error.
Maybe just add a `#undef USE_COMPILER_TLS` before this line.

amurzeau (commented):
I've added 2 comments about possible redefinition of USE_COMPILER_TLS in case of android and macos targets.

amurzeau:
I've compiled this patch and indeed, the build system need a modification to pass around `USE_COMPILER_TLS`.
I've applied this:
```diff
--- a/Makefile.system
+++ b/Makefile.system
@@ -1016,6 +1016,10 @@ ifdef USE_SIMPLE_THREADED_LEVEL3
 CCOMMON_OPT   += -DUSE_SIMPLE_THREADED_LEVEL3
 endif
 
+ifdef USE_COMPILER_TLS
+CCOMMON_OPT     += -DUSE_COMPILER_TLS=$(USE_COMPILER_TLS)
+endif
+
 ifndef SYMBOLPREFIX
 SYMBOLPREFIX =
 endif

```

Also, to try harder if the deadlock might occurs or not, I've made a small C test program that does a bunch of dlopen and dlclose and check in gdb if it runs or not (easy to see as gdb will show whenever a thread is created):
```c
#include <dlfcn.h>
#include <stdio.h>

int main(int argc, char* argv[]) {
        while(1) {
                void* handle = dlopen(argv[1], RTLD_NOW);
                if(handle == 0) {
                        printf("Can't open openblas: %s\n", dlerror());
                        return 1;
                }
                dlclose(handle);
        }
        return 0;
}
```

And this patch + the additional change on the build system to support USE_COMPILER_TLS fix the deadlock. The full patch is here: https://salsa.debian.org/science-team/openblas/blob/e51b3b26641901a5a93247ef9bb9dbdddc82cfae/debian/patches/0008-Allow-overriding-USE_COMPILER_TLS-formerly-HAS_COMPI.patch
As an additional note, when compiling I got these warnings:
```
cc -Wdate-time -D_FORTIFY_SOURCE=2 -g -O2 -fdebug-prefix-map=/<<PKGBUILDDIR>>=. -fstack-protector-strong -Wformat -Werror=format-security -march=native -mtune=native  -DMAX_STACK_ALLOC=2048 -Wall -m64 -DF_INTERFACE_GFORT -fPIC -DNO_LAPACKE -DSMP_SERVER -DNO_WARMUP -DMAX_CPU_NUMBER=4 -DMAX_PARALLEL_NUMBER=1 -DUSE_COMPILER_TLS=0 -DASMNAME=z_abs -DASMFNAME=z_abs_ -DNAME=z_abs_ -DCNAME=z_abs -DCHAR_NAME=\"z_abs_\" -DCHAR_CNAME=\"z_abs\" -DNO_AFFINITY -I../.. -c -DDOUBLE abs.c -o z_abs.o
memory.c: In function 'get_num_procs':
memory.c:191:7: warning: unused variable 'n' [-Wunused-variable]
 int i,n;
       ^
memory.c:191:5: warning: unused variable 'i' [-Wunused-variable]
 int i,n;
     ^
memory.c: In function 'get_memory_table':
memory.c:542:32: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
   int local_memory_table_pos = (int)pthread_getspecific(local_storage_key);
                                ^
memory.c:555:44: warning: cast to pointer from integer of different size [-Wint-to-pointer-cast]
     pthread_setspecific(local_storage_key, (void*)local_memory_table_pos);
```

oon3m0oo:
We should probably pass this over in favor if #1739.

rgommers:
> We should probably pass this over in favor if #1739.

That, and the issue referenced in the PR description was closed 6 years ago. I'm not even sure that a build config option to change TLS handling is desirable - it's one more variation to test, and it's better to properly fix TLS issues if there are any left in 2024 (many projects use thread-local storage, and it's a hard necessity usually - just turning it off isn't really an option).

Probably best to close this PR?

martin-frbg:
Both these PRs have been cherry-picked for improvements ages ago, I had only kept them open for reference. If you think they should be closed to make the PR list look nicer, this can be done.

rgommers:
Ah if they're merged, I would indeed close them.

PR #1731, opened 2018-08-11
add short blurb about avx512 and needed compiler to README

fenrus75:
based on a request in the original sdot PR


martin-frbg:
typo "skyalke"...

Issue #992, opened 2016-10-24
Optimize OpenBLAS for LinuxOne s390x zEC13 SIMD: BLAS 1/2 single precision


edelsohn:

papaaloabob:
Aloha edelsohn,
Is this still open?  I see that quickwritereader did a lot of related work.  Is he not interested in this one?

Thank you,
bob

quickwritereader:
As there is not direct hardware simd I postponed it. as for some functions I believed using simd will not improve performance much.  So I decided to take other issue before returning to this one. 
I'm thinking to try it in near future. 
But If you want to try go ahead. After all solving this issue will benefit community.

edelsohn:
There is Power VSX SIMD support for single precision, so there should be opportunities for optimization and tuning. It may be best to implement in C with vector types than directly using intrinsics. One needs to experiment.

papaaloabob:
Thank you very much for your quick and thoughtful reply.  I'll look closer.

Bob Holmes
808-640-1799

mtk-i:
The processors in [IBM LinuxONE](https://developer.ibm.com/linuxone) can't handle SIMD single precision instructions, I have successfully assembled SIMD single precision instructions but I got it `Illegal instruction` when I run it. It's good to point me to a capable Z14 VM since I ported many functions to single precision so I can test it. Thanks in advance.

edelsohn:
Unfortunately, there are no publicly accessible z14 systems. The best that I can do is test on a z14 system internal to IBM and provide feedback.

mtk-i:
Some functions need many tests to get optimal performance, However I will do my best to make it optimized as much as possible and send it to you, Maybe I will provide many implementations for some functions, Do you have problem in that?

mtk-i:
hmm I have an idea, since I can compile the instructions, I will write separated unit tests that compatible with  functions implementations and just send the execution files to you for test.

edelsohn:
I don't know if QEMU for s390x is up to date yet for these instructions. That's another option.

quickwritereader:
If these additions are for z13 I could share the pc given to me.

mtk-i:
@edelsohn Not sure if QEMU give native performance, Although I do not think so. I think providing source code to you "not execution files since can't be trusted by IBM" for test should work. Let's see how we can test it later.
@quickwritereader unfortunately it's z14 specific, Z13 can't handle that.

mtk-i:
Done
[...]

PR #1738, opened 2018-08-14
detect z14 arch on s390x

sharkcz:
This patch add detection of the z14 architecture in s390x/zarch, but falls back to the z13 implementation. Without this change z14 would be detected as generic only.


martin-frbg:
Thanks. I expect we will get z14 support soon from #992 (z14 support fork linked from there also has the detection code already)

Issue #1735, opened 2018-08-13
OpenBLAS : Program will terminate because you tried to start too many threads.


logidelic:
I recently built OpenBLAS 0.3.2 on Ubuntu 16 and am running into this error.

`
OpenBLAS : Program will terminate because you tried to start too many threads.
`

My program is using a library that allocates many threads for various reasons. Most of the threads are sleeping most of the time, so there are no performance issues, but I guess that the large number of threads is too much...

Does linking to OpenBLAS really limit the max number of threads that the program creates, even if those threads are not making blas calls? Is there a way around this?

FWIW, I have done a lot of searching and reading on the topic, but still can't quite figure out how to solve this... Some notes:

 - I was not getting this error with a previous version of OpenBLAS (whatever is in he Ubuntu repo)
 - I also tried building with `USE_OPENMP=1` but that doesn't seem to fix the problem.
 - I tried with `OPENBLAS_NUM_THREADS=1` environment variable which seemed to make no difference
- I tried calling `openblas_set_num_threads(1)` at runtime which seemd to make *some* difference, but eventually I got the same error
- I tried building with `USE_THREAD=0`. I think this prevents the error but I haven't found clear documentation as to what the implications are. Can I still call blas functions in multiple threads safely (on unrelated data of course)? Does blas do everything in a single thread in this case or in whatever thread it's called from?

Thanks much.

brada4:
If you call openblas from multiple threads you need omp or single thread version. Check Makefile.rule for more options

brada4:
If you call multi-threaded blas from many threads at once  you lose cache efficiency

martin-frbg:
This appears to be a (still poorly understood) bug introduced with the rewrite (and speedup) of the thread initialization code in 0.3.1 - see #1704 and #1641. A workaround is simply to increase the value of MAX_ALLOCATING_THREADS, but to achieve some kind of final solution it would be very useful to get an idea how many (blas-calling) threads an affected program is/was trying to start.

brada4:
#1704 sort of ended in resource leak where calling thread exits and new comes in place all the time?
@logidelic can you add some cout() where your code creates / exits a thread?

martin-frbg:
Not sure I'd call that a resource leak, it just so happens that there is a fixed size array for thread pointers in the new code where old entries apparently never get reused. #1641 already had musings on making this thing dynamically allocated.

susilehtola:
The same issue has been reported in Fedora
[...]

PR #1739, opened 2018-08-16
Support arbitrary numbers of threads for memory allocation.

oon3m0oo:
When I originally refactored memory.c to reduce locking, I made the (incorrect) assumption that all threads were managed by OpenBLAS.  The recent Issues we've seen (#1735) show that really, any caller can make its own threads and call into OpenBLAS; they don't all come from blas_server.  Thus we have to be able to support an arbitrary number of threads that can come in any time.  The original implementation (before my changes) dealt with this by having a single allocation table, and everyone had to lock to get access to and update it, which was expensive.  Moving to thread-local allocation tables was much faster, but now we have to deal with the fact that thread local storage might not be cleaned up.

This change gives each thread its own local allocation table, and completely does away with the global table.  We cleanup allocations using pthreads' key destructor and Win32's DLL_THREAD_DETACH.

This change also removes compiler TLS, which in the end, wasn't really worth it given the issues with the glibc implementation (#1720).  The overall performance impact was < 1%, anyway.  Removing it also simplifies the code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ jobs:
         - COMMON_FLAGS="DYNAMIC_ARCH=1 TARGET=NEHALEM NUM_THREADS=32"
       script:
         - set -e
-        - travis_wait 45 make QUIET_MAKE=1 $COMMON_FLAGS $BTYPE
+        - make -s $COMMON_FLAGS $BTYPE
         - make -C test $COMMON_FLAGS $BTYPE
         - make -C ctest $COMMON_FLAGS $BTYPE
         - make -C utest $COMMON_FLAGS $BTYPE
@@ -69,7 +69,7 @@ jobs:
             - gfortran-mingw-w64-x86-64
       before_script: *common-before
       script:
-        - travis_wait 45 make QUIET_MAKE=1 $COMMON_FLAGS $BTYPE
+        - make -s $COMMON_FLAGS $BTYPE
       env:
         - TARGET_BOX=WIN64
         - BTYPE="BINARY=64 HOSTCC=gcc CC=x86_64-w64-mingw32-gcc FC=x86_64-w64-mingw32-gfortran"
@@ -94,7 +94,7 @@ jobs:
       script:
         - set -e
         # XXX: Disable some warnings for now to avoid exceeding Travis limit for log size.
-        - alpine make QUIET_MAKE=1 $COMMON_FLAGS $BTYPE
+        - alpine make -s $COMMON_FLAGS $BTYPE
               CFLAGS="-Wno-misleading-indentation -Wno-sign-conversion -Wno-incompatible-pointer-types"
         - alpine make -C test $COMMON_FLAGS $BTYPE
         - alpine make -C ctest $COMMON_FLAGS $BTYPE
@@ -160,7 +160,7 @@ jobs:
         - brew update
         - brew install gcc # for gfortran
       script:
-        - travis_wait 45 make QUIET_MAKE=1 $COMMON_FLAGS $BTYPE
+        - make -s $COMMON_FLAGS $BTYPE
       env:
         - BTYPE="BINARY=64 INTERFACE64=1"
 
--- a/driver/others/memory.c
+++ b/driver/others/memory.c
@@ -143,14 +143,6 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define FIXED_PAGESIZE 4096
 #endif
 
-#ifndef BUFFERS_PER_THREAD
-#ifdef USE_OPENMP_UNUSED
-#define BUFFERS_PER_THREAD (MAX_CPU_NUMBER * 2 * MAX_PARALLEL_NUMBER)
-#else
-#define BUFFERS_PER_THREAD NUM_BUFFERS
-#endif
-#endif
-
 #define BITMASK(a, b, c) ((((a) >> (b)) & (c)))
 
 #if defined(_MSC_VER) && !defined(__clang__)
@@ -338,7 +330,8 @@ int  goto_get_num_procs  (void) {
   return blas_cpu_number;
 }
 
-static void blas_memory_init();
+static void blas_memory_init_from_fork();
+static void blas_tls_init();
 
 void openblas_fork_handler()
 {
@@ -351,7 +344,7 @@ void openblas_fork_handler()
   // implementation of OpenMP.
 #if !((defined(OS_WINDOWS) && !defined(OS_CYGWIN_NT)) || defined(OS_ANDROID)) && defined(SMP_SERVER)
   int err;
-  err = pthread_atfork ((void (*)(void)) BLASFUNC(blas_thread_shutdown), NULL, blas_memory_init);
+  err = pthread_atfork ((void (*)(void)) BLASFUNC(blas_thread_shutdown), NULL, blas_memory_init_from_fork);
   if(err != 0)
     openblas_warning(0, "OpenBLAS Warning ... cannot install fork handler. You may meet hang after fork.\n");
 #endif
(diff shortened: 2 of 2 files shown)


oon3m0oo:
Hmm... I'm not sure why Windows suddenly can't find ::Tls*.

martin-frbg:
Checking if this is a problem with the CI - local Linux builds show no errors so far.

martin-frbg:
I wonder if the Windows build issue comes simply from using C++ idiom in a plain C file (which would also imply the problem was in your earlier code as well, just masked by the default use of compiler TLS).

oon3m0oo:
Aha, you're probably right.  I've removed the ::; let's see what happens.

In the meantime I can reproduce the test failures and am actively debugging.

oon3m0oo:
Yep, that was it.  I've fixed the tests, too.  The failures are make timeouts (45 minutes). Is there a way I can restart them?

martin-frbg:
Not sure if you as the submitter already have the right to restart individual jobs, or perhaps would need to set up a free account with travis. (you should see a symbol with two concetric circles to the right of the time column in the travis jobs table if you can restart jobs by clicking this). I have now restarted the three jobs in question, but the issue with the timeouts could only be resolved by changing the limit in the .travis.yml file. However, these three jobs used to take around twenty minutes like the others, so "make timeout" here probably means a lockup in the tests. (We will see...)

oon3m0oo:
I've tried all three locally and everything was fine... not sure what's happening with travis.

martin-frbg:
I cannot reproduce this either. Travis status blog has a few recent entries about jobs that require "sudo: true" intermittently failing or hanging. It is just worrying that these three jobs should fail repeatedly...

oon3m0oo:
Yeah, it is.  Is there any way to get more verbose output of those jobs?

martin-frbg:
I could remove the settings for "quiet make" from the travis.yml, but ironically travis used to abort jobs for creating too much output without it.

martin-frbg:
Trying some things in my fork now. But without "QUIET", travis tends to kill the job as it exceeds 4MB of output (mostly from warnings in the LAPACK part)...

brada4:
Travis also kills the absolutely silent jobs...
`make -s` ?
I think it barfs out less than a megabyte now.

martin-frbg:
Turns out it is the fork safety utest that is hanging.

brada4:
Linux timeout command can do it, no posix equivalent, but particular test is only linux...

martin-frbg:
@brada4 this does not answer the question why the fork test fails with this change.

brada4:
Just to fix up CI

martin-frbg:
The hang is reproducible with (a vm running) the same version of Ubuntu locally, so probably related to either glibc 2.15 or gcc 4.6.3 . gdb shows one thread in __libc_thread_freeres() called from clone() via start_thread, with the other thread in pthread_join called from blas_thread_shutdown at line 982 of blas_server.c. Not sure yet why this would happen only after this change, when the only relevant difference appears to be in blas_memory_cleanup that happens _after_ the call to blas_thread_shutdown.

martin-frbg:
helgrind sees a race between calls to blas_memory_alloc from memory.c:1071 (reading with no lock held, called from dgemm, gemm.c:403) and memory.c:1097 (writing with a lock held), and another between get_memory_table at memory.c:489 and pthread_key_create called from memory.c:1031

martin-frbg:
Seems access to the newly introduced local_storage_key variable is racey, but adding another lock to protect it does not help.

martin-frbg:
I believe I fixed the race conditions  in the "mem" branch of my fork (at the expense of another mutex lock), but the issue with travis looks like a glibc < 2.23 bug. If I move the CI to Ubuntu xenial, the utest passes. 
The question now becomes how to handle these older systems:  An `#if !defined(__GLIBC_PREREQ) || __GLIBC_PREREQ(2,23)` for the "new" code is easy enough, but what to fall back to in the `#else` branch - pre-0.3.1 memory.c, or the 0.3.2 memory.c without compiler TLS and with an increased MAX_ALLOCATING_THREADS ?

oon3m0oo:
I've added a tiny bit of locking that ought to resolve things... we'll see.  I suppose it depends on what the issue with glibc is.  The change in logic here might be enough to avoid the issue altogether (crosses fingers).

martin-frbg:
My impression was that it was somehow related to using malloc in the threads, but I do not have the bug ids handy now. We'll see what travis says now, and I'll also give it a spin with the ubuntu vm later when I have time.

martin-frbg:
Does not look good unfortunately, I see local segfaults (at memory.c:1128) even with relatively current systems as well. I must admit I am beginning to think we should revert to the 0.3.0 memory.c for 0.3.3 to give users a stable release to work with, and reintroduce the rewrite when it is stabilized and tested.

oon3m0oo:
@martin-frbg what tests are you running?

martin-frbg:
Normal build (on opensuse leap 42.2, gcc-4.8.5, glibc 2.22) was enough to trigger a segfault in the sblat2 test with your latest version...

oon3m0oo:
All I have is 2.24 but I can't reproduce any crashes.  Everything seems to run fine, which makes all of this more complicated.  I'm trying to see if there's a good way for me to get something were I can reproduce this, but if you have any ideas, I'd be grateful.

martin-frbg:
Virtualbox with a bunch of vm's from osboxes.org or similar ? Also running `valgrind --tool=helgrind` (or just the default memcheck mode valgrind) on the blas tests and/or the utest binary is often helpful.

martin-frbg:
BTW the sblat2 test crashes with alloc_table=0x0 (making alloc_table[0] an illegal access) in line 1128 of memory.c

martin-frbg:
Seems to me the crucial problem is that your revised code is missing a call to blas_tls_init() at the start of blas_memory_alloc() (or the `static int tls_initialize` there is not working as intended for whatever reason). Net effect seems to be that get_memory_table gets invoked with an undefined local_storage_key.

oon3m0oo:
Actually it gets called from blas_memory_init, from blas_memory_alloc. And somehow that line is gone.  I'll add it back in.

martin-frbg:
Seems the timeouts/lock ups in the "fork" utest are still there on older systems. My quick test (with the Ubuntu CI setup updated to avoid the perceived glibc "fork test" problem) passed everything except the mingw cross-build. (Not sure yet what happened there, could be same glibc problem). Still not quite happy with putting this in 0.3.3, at least the previous version did not lock up on older systems.

oon3m0oo:
Hmm... I have to admit it is rather frustrating that I can run something locally and have everything pass and run fine (no seg faults), and end up with failures on Travis.  I don't envy the huge portability aspect in pulling this project together.  I do wish that the logs at least would have more details in the case of failure (it seems like only one job gives verbose output).

martin-frbg:
I have hacked my .travis.yml into your branch now, but the problem with the old Ubuntu "Precise" version will probably need to be debugged in a local installation. I believe the earlier problems with segfaults were a direct consequence of invoking undefined behaviour (by omitting tls key initialization in that one code path), so did not affect all Linux flavors int the CI and may not have occured on whatever you use for local testing.

martin-frbg:
Checking "my" version locally, I "only" get the fork hang with the old Ubuntu systems, but not the segfaults of "your" version. Note the version in my fork has added safeguards against alloc_table itself being NULL, where your original only checked alloc_table[position]...

martin-frbg:
I have now put my modified version of the code into #1742, which would again make the original memory.c the default unless compiled with -DUSE_TLS (and on a system with a sufficiently capable libc).
I think this should provide a reliable solution for 0.3.3, although it puts the burden of choosing on the end user (or distribution packager) for now. Once the remaining issues are resolved, we can just change the default setting of USE_TLS .

PR #1751, opened 2018-09-05
replace ARCH with AR in lapack-netlib

dloghin:
This PR solves two issues when cross-compiling OpenBLAS for RISCV64.

1. c_check fails withe the error:
```
1 at ./c_check line 169.
```

I have added this in c_check:
```
    $defined = 1;
}
```

2. compiling lapack-netlib fails with:
```
riscv64 -ru ../../libRblas_riscv64-r0.3.0.dev.a sbdsvdx.o spotrf2.o
...
...
```

I have replaced ARCH and ARCHFLAGS with AR and ARFLAGS.


martin-frbg:
The problem with the netlib patch is that it is too invasive IMHO - the lapack-netlib tree is imported from https://github.com/Reference-LAPACK/lapack and needs to be updated whenever a new netlib LAPACK is released. I suspect setting AR=/path/to/riscv/ar on the `make` command line would get around the original problem.

dloghin:
Fair point. Actually, there is amuch simpler way to make it work: using override keyword on ARCH in lapack-netlib/make.inc which is generated by the main Makefile. Hence, lapack doesn't need to be modified. I will create a new PR.

PR #1753, opened 2018-09-06
Override ARCH (archiver) in lapack-netlib/make.inc to avoid conflict with ARCH (architecture)

dloghin:
This PR solves two issues when cross-compiling OpenBLAS for RISCV64.

1. compiling lapack-netlib fails with:

```
riscv64 -ru ../../libRblas_riscv64-r0.3.0.dev.a sbdsvdx.o spotrf2.o
...
...
```

ARCH needs an override keyword in lapack-netlib/make.inc because it is taken as ARCH (architecture) from the main Makefile (hence what should be riscv64-unknown-linux-gnu-ar becomes riscv64). I have added the override when make.inc is generated from the main Makefile.

2. c_check fails withe the error:

```
1 at ./c_check line 169.

```
I have added this in c_check:

```
    $defined = 1;
}
```


brada4:
Actually it should read "support riscv64"
I am unsure if setting ARCH helps netlib lapack, to my search & look-around understanding it sets ARCH to 'ar' in absence of one in top config file.

dloghin:
Well, if ARCH is set from the command line, then the ARCH = $(AR) in netlib is ignored. That's why we should use `override`.

martin-frbg:
~~I must admit it is stlll not clear to me where the "wrong" setting of the ARCH entry in make.inc originates. While the OpenBLAS Makefile does use this keyword internally to denote architecture, I do not see how this would be able to override the AR entry that is used in writing make.inc. Does the shell environment already contain an entry named ARCH by default ?~~

martin-frbg:
Never mind, I think I understand now (again) that the issue is caused by _having_  to specify the ARCH option on the `make` command line. Normally c_check should already set this automatically, perhaps the real issue is that there is another round of setting $architecture from parsing $data around line 200 of c_check that seems to have been overlooked when the riscv64 patch/branch
was created. I wonder what ends up as ARCH in your Makefile.conf when you do not specify it yourself ?

dloghin:
Indeed, if you don't specify ARCH on the command line, it gets detected in c_check based on the compiler (CC) and written in Makefile.conf. In this case, top level ARCH is ignored in lapack-netlib. However, when someone wants to specify it on the command line, it gets priority (details here https://www.gnu.org/software/make/manual/html_node/Overriding.html).

In my opinion, using ARCH to denote both architecture and archive in the same project is an issue that should be addressed.

martin-frbg:
The naming conflict only arose from combining the two separate projects. Perhaps the original GotoBLAS2 would have been better off using something like CPUARCH throughout to avoid confusion, but I am not sure if the required changes in most Makefiles would be worthwile now. Perhaps Makefile.prebuild should simply override any ARCH that was specified on the command line if it serves no useful purpose ?

PR #1757, opened 2018-09-09
fix small typo in strmm_ LN

brada4:
Fix small old typo that affects SANDYBRIDGE BULLDOZER PILEDRIVER STREAMROLLER EXCAVATOR HASWELL ZEN SKYLAKEX strmm_ and via this [some lapack](http://www.netlib.org/lapack/explore-html/db/dc9/group__single__blas__level3_ga5a8ec25aba550c224cf6941fca7a2c98_ga5a8ec25aba550c224cf6941fca7a2c98_icgraph.svg)


brada4:
... from original gotoblas codes ....

martin-frbg:
Dating back to the very first (GotoBLAS-1.00) release in 2006 even, I just dug up my old copy to check. (Given that track record, the practical impact of that typo must have been very small - good catch all the same.)

brada4:
It just used to copy one single value from previous row near tip of L triangular matrix.

PR #1760, opened 2018-09-11
remove unused 'levelness' parameter from num_cpu_avail()

brada4:
and remove it from all callers, probably moderating the thread limit for small sets could go to interface/?
i have some doubts about extra roll with same num_cpu_avail() in kernel/ ... shouldnt it be all in one place?

### DIFF
--- a/common_thread.h
+++ b/common_thread.h
@@ -129,7 +129,7 @@ typedef struct blas_queue {
 
 extern int blas_server_avail;
 
-static __inline int num_cpu_avail(int level) {
+static __inline int num_cpu_avail(void) {
 
 #ifdef USE_OPENMP
 	int openmp_nthreads=0;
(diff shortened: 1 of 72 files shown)


martin-frbg:
While this parameter has been unused for ages, I am not sure if it is worth changing so many files just to get rid of it ? (Seems in GotoBLAS2-1.08 it was used to disable multithreading for some callers, but that code was already removed in 1.13)

martin-frbg:
PS callers in kernel/* are a handful of cases where parallelization is implemented in cpu-specific code.

brada4:
Rationale that in this place integer parameter does not have any future use or so. interface/ indeed can be perfectly calibrated per function x coretype

brada4:
would it help reinstating old prototype with warning?

PR #1764, opened 2018-09-16
Allow to install the 'interface64' version concurrently with the regular version

yurivict:


timokau:
Is it necessary to use cmake to achieve this? The README still recommends using plain GNU make.

brada4:
You could always pass arbitrary SUFFIX to `make`, this just adds feature to cmake.

martin-frbg:
No, suffixing has already been possible with the GNU make build since october 2015 (issue 646 as mentioned in the linked issue 1763)

Issue #1748, opened 2018-09-03
trmv! gives warning for moderately large matrices


dlfivefifty:
```julia
julia> using LinearAlgebra

julia> BLAS.vendor()
:openblas64

julia> n = 10000; A = randn(n,n); b = randn(n); BLAS.trmv!('U', 'N', 'N', A, b);

WARNING unrolling of the trmv_U loop may give wrong results
```

The offending line seems to be https://github.com/xianyi/OpenBLAS/blob/e11126b26ada8d97b4a522e461ca92311653bfc6/driver/level2/trmv_U.c#L66

See discussion https://discourse.julialang.org/t/why-lapack-trtrs-not-blas-trsv/14239/3

martin-frbg:
This is related to #1332. Is it the same hardware as in #1698 ? I did not expect that loop to be entered on "common" hardware after the hack, but clearly your definition of "moderately large" is different from mine. Probably best to fully disable the suspicious code block until the actual issue is found and fixed. (Too bad that these issues with original codes from GotoBLAS come up only now.)

martin-frbg:
Actually BLAS-Tester results (both with the smallish default problem sizes and larger ones like `./xdl2blastst -R trmv -U 1 U -n 15000`) suggest that the warning is unnecessary as the results appear to match the reference values even with unrolling of the gemv loop (as long as trmv is not allowed to be parallelized again - this conclusion from #1332 remains valid).

PR #1768, opened 2018-09-20
Try harder at disabling gemv_n unrolling

martin-frbg:
for #1748


martin-frbg:
No longer convinced this is even necessary... there seems to be a small lack of precision caused by the Haswell implementation of gemv_n in dgemv_n_4.c compared to the plain C arm/gemv_n.c but this is hardly the fault of the trmv_U.c caller and shows up only in the (fixed) test case from #1332 as a MAXERR of .2842D-13 instead of .1421D-13 achievable with the plain C version for N=507. By comparison, allowing multithreading in interface/trmv.c increases the error to around .15D-12

PR #1781, opened 2018-09-30
use blas lock on LOONGSON3A

fengrl:
Time wasted on pthread_lock/unlock on LOONGSON3A platform.  Blas_lock/unlock can increase speed about 50%.

### DIFF
--- a/common.h
+++ b/common.h
@@ -208,6 +208,13 @@ extern "C" {
 #undef	USE_PTHREAD_SPINLOCK
 #endif
 
+#ifdef OS_LINUX
+#ifdef LOONGSON3A 
+#undef  USE_PTHREAD_LOCK
+#undef  USE_PTHREAD_SPINLOCK
+#endif
+#endif
+
 #if   defined(USE_PTHREAD_LOCK)
 #define   LOCK_COMMAND(x)   pthread_mutex_lock(x)
 #define UNLOCK_COMMAND(x)   pthread_mutex_unlock(x)
--- a/common_mips64.h
+++ b/common_mips64.h
@@ -94,7 +94,7 @@ static inline unsigned int rpcc(void){
 #define RPCC_DEFINED
 
 #ifndef NO_AFFINITY
-#define WHEREAMI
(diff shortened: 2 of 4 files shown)


martin-frbg:
Blas_lock on mips64 is just the YIELDING macro that maps to sched_yield() I think. Is that sufficient and appropriate ? (Seems on some platforms sched_yield is about the worst one can do from a performance standpoint, so YIELDING is often mapped to asm(nop) or the like in common.h). Historically most parts of OpenBLAS were not thread-safe, but I must admit that most of the locks that I put in were based on valgrind/helgrind reports rather than qualified analysis of the code. So there may have been some false positives, but on the other hand some race conditions tended to show up only on fast computers with many cores. (Not sure how big and fast Loongson-based hardware can get ?)

fengrl:
While compile with:
make CC=gcc FC=gfortran HOSTCC=gcc TARGET=LOONGSON3A CROSS=1 NO_LAPACKE=1 BINARY=64

in common.h, SMP was defined.
#ifdef SMP
#define USE_PTHREAD_LOCK
#undef  USE_PTHREAD_SPINLOCK
#endif
So, USE_PTHREAD_LOCK will be used.

Like what we do with WINDOWS, blas_lock same code added on LOONGSON3A.

martin-frbg:
Could you create a separate PR for the sgemm_kernel bug please ? 
And sorry for sitting on this PR for so long, but I have been very busy elsewhere. Both the common_mips64.h change and the one to memory.c have already been made through separate PRs in the meantime I think - your #1800 and brada4's #1785. Regarding the original locking issue, I think the difference to windows is that the x86 and x86_64 platforms define their own implementation of blas_lock in common_x86.h and common_x86_64.h respectively, where the generic sched_yield() is followed by  an InterlockedExchange (MSVC) or xchgl instruction to - I believe - ensure thread safety. The common_mips64.h does not provide anything like this (compare lack of BLAS_LOCK_DEFINED)  so it would end up with the pure sched_yield() from the fallback code which I fear would be unsafe.

PR #1784, opened 2018-10-03
Create a AVX512 enabled version of DGEMM

fenrus75:
This patch adds dgemm_kernel_4x8_skylakex.c which is
* dgemm_kernel_4x8_haswell.s converted to C + intrinsics
* 8x8 support added
* 8x8 kernel implemented using AVX512

Performance is a work in progress, but already shows a 10% - 20%
increase for a wide range of matrix sizes.

### DIFF
--- a/kernel/x86_64/KERNEL.SKYLAKEX
+++ b/kernel/x86_64/KERNEL.SKYLAKEX
@@ -2,18 +2,12 @@ include $(KERNELDIR)/KERNEL.HASWELL
 
 SGEMMKERNEL    =  sgemm_kernel_16x4_skylakex.S
 
+DGEMMKERNEL    =  dgemm_kernel_4x8_skylakex.c
 
-#DTRMMKERNEL    =  ../generic/trmmkernel_16x2.c
-#DGEMMKERNEL    =  dgemm_kernel_16x2_skylakex.S
-#DGEMMINCOPY    =  ../generic/gemm_ncopy_16.c
-#DGEMMITCOPY    =  ../generic/gemm_tcopy_16.c
-#DGEMMONCOPY    =  ../generic/gemm_ncopy_2.c
-#DGEMMOTCOPY    =  ../generic/gemm_tcopy_2.c
-#DGEMMINCOPYOBJ =  dgemm_incopy$(TSUFFIX).$(SUFFIX)
-#DGEMMITCOPYOBJ =  dgemm_itcopy$(TSUFFIX).$(SUFFIX)
-#DGEMMONCOPYOBJ =  dgemm_oncopy$(TSUFFIX).$(SUFFIX)
-#DGEMMOTCOPYOBJ =  dgemm_otcopy$(TSUFFIX).$(SUFFIX)
-
+DGEMMINCOPY    =  ../generic/gemm_ncopy_8.c
+DGEMMITCOPY    =  ../generic/gemm_tcopy_8.c
+DGEMMONCOPY    =  ../generic/gemm_ncopy_8.c
+DGEMMOTCOPY    =  ../generic/gemm_tcopy_8.c
 
 SGEMM_BETA = ../generic/gemm_beta.c
 DGEMM_BETA = ../generic/gemm_beta.c
--- a/kernel/x86_64/dgemm_kernel_4x8_skylakex.c
+++ b/kernel/x86_64/dgemm_kernel_4x8_skylakex.c
@@ -1,0 +1,1288 @@
+/*********************************************************************************
+Copyright (c) 2015, The OpenBLAS Project
+All rights reserved.
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+1. Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+2. Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in
+the documentation and/or other materials provided with the
+distribution.
+3. Neither the name of the OpenBLAS project nor the names of
+its contributors may be used to endorse or promote products
+derived from this software without specific prior written permission.
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ARE DISCLAIMED. IN NO EVENT SHALL THE OPENBLAS PROJECT OR CONTRIBUTORS BE
+LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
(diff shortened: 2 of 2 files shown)


martin-frbg:
Travis (musl-libc builds) failure is probably the compiler version problem again, guess the new dgemm kernel needs the same "if CC version too old, include haswell kernel instead" lines added before the `#include <immintrin.h>` as in your other recent contributions.

fenrus75:
that's.... more complicated since the haswell code needs the "4" version for incopy while skylake needs the "8" version; one cannot just use the .S file as is.
(or .c file; I made a .c version for the haswell algorithm first)

martin-frbg:
I guess the simpler alternative is to add the `#include <immintrin.h>` to the AVX512 compatibility test in the build system.

brada4:
Upgrading alpine chroot could help too

martin-frbg:
Upgrading the alpine environment might fix the CI test, but not the behaviour on such systems elsewhere... (But adding the include to the compile test did not deter it from trying to build for AVX512, 
unfortunately the log does not contain any hints. Removing QUIET_MAKE completely makes it hit the outpuf file size limit shortly after the build enters "kernel", I am now trying with a modified Makefile.)

fenrus75:
is there a way to run the container locally for debug? I might need to add this one to my testing

martin-frbg:
The way I read .travis.yml, it is basically doing a 'wget` to download a specially crafted installation of Alpine from their github account into a stock ubuntu installation and then chroots to it.

martin-frbg:
Did the chroot installation locally now, and can reproduce the problem. For some reason, the SkylakeX
dgemm kernel lacks its _SKYLAKEX suffix in the DYNAMIC_ARCH build. This is probably a build system issue, perhaps the KERNEL.SKYLAKEX needs to specify DGEMMKERNELOBJ with its TSUFFIX explicitly to ensure that it gets compiled to the expected name. (SImilarly to what is already in place for the DGEMMINCOPY etc. C files)

chriselrod:
AVX-512 doubles the number of registers from 16 to 32, as well as doubling their size.
Therefore loading an 8x8 block from matrix C only uses 8 registers, leaving 24.

Assuming column major, C += A * B...
A much larger kernel, eg 40x5 or 16x14 is optimal.

16x14 dedicates 28 registers to the block from C, 2 to a column of A, and 2 to broadcasts from B.
24x9 dedicates 27 registers to the block from C, 3 to a column of A, and 2 to broadcasts from B.
32x6 dedicates 24 registers to the block from C, 4 to a column of A, and 4 to broadcasts from B.
40x5 dedicates 25 registers to the block from matrix C, leaving 5 to a column of A, and 2 broadcasts at a time from B.

While SIMD intrinsics + LLVM is not totally optimal (there is 1 unnecessary vmovapd), I can easily test a variety of kernels in Julia by using generated functions:
```julia
using CpuId, Base.Cartesian, SIMD, StaticArrays, BenchmarkTools
const REGISTER_SIZE = simdbytes()

@generated function generic_kernel!(D::MMatrix{M,P,T},
                        A::MMatrix{M,N,T},
                        X::MMatrix{N,P,T}) where {M,P,N,T}
    MSoT = M*sizeof(T)
    L = REGISTER_SIZE ÷ sizeof(T)
    Q, r = divrem(M, L) #Assuming M is a multiple of L
    r == 0 || throw("Number of rows $M not a multiple of register size: $REGISTER_SIZE.")
    V = Vec{L,T}
    quote
        pD, pA = pointer(D), pointer(A)

        @nexprs $P p -> @nexprs $Q q -> Dx_p_q = vload($V, pD + $REGISTER_SIZE*(q-1) + $MSoT*(p-1))
        @inbounds for n ∈ 1:N
            @nexprs $P p -> begin
                vX = $V(X[n,p])
                @nexprs $Q q -> begin
                    vA_q = vload($V, pA + (n-1)*$MSoT + $REGISTER_SIZE*(q-1))
                    Dx_p_q = fma(vA_q, vX, Dx_p_q)
                end
            end
        end
        @nexprs $P p -> @nexprs $Q q -> vstore(Dx_p_q, pD + $REGISTER_SIZE*(q-1) + $MSoT*(p-1))
        D
    end
end

const D16_14 = @MMatrix zeros(16,14)
const A16_50 = @MMatrix randn(16,50)
const X50_14 = @MMatrix randn(50,14)

const D40_5 = @MMatrix zeros(40,5)
const A40_56 = @MMatrix randn(40,56)
const X56_5 = @MMatrix randn(56,5)

const D8_8 = @MMatrix zeros(8,8)
const A8_175 = @MMatrix randn(8,175)
const X175_8 = @MMatrix randn(175,8)
```
```julia
julia> @benchmark generic_kernel!(D16_14, A16_50, X50_14)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     192.995 ns (0.00% GC)
  median time:      194.078 ns (0.00% GC)
  mean time:        201.990 ns (0.00% GC)
  maximum time:     275.496 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     631

julia> @benchmark generic_kernel!(D40_5, A40_56, X56_5)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     190.127 ns (0.00% GC)
  median time:      190.885 ns (0.00% GC)
  mean time:        200.710 ns (0.00% GC)
  maximum time:     296.633 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     646

julia> @benchmark generic_kernel!(D8_8, A8_175, X175_8)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     220.581 ns (0.00% GC)
  median time:      220.994 ns (0.00% GC)
  mean time:        230.962 ns (0.00% GC)
  maximum time:     354.935 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     496

julia> 8 * 175 * 8
11200

julia> 40 * 56 * 5
11200

julia> 16 * 50 * 14
11200
```
Bigger kernels (using almost all registers) are faster than smaller (8x8).

I also have code for a non-packing implementation that I can benchmark with, but it loses performance quickly once matrices no longer fit in the L2 cache, which is 200x200 on my Skylake-X.
Still, that would be a more reasonable test, because of movement of the need to move data between the L1 and L2 caches, where square kernels ought to do better.

I'd be happy to contribute C with intrinsics. Julia doesn't really support linking to Intel MKL: libraries like Arpack, which are in the dependency list of many others, only support OpenBLAS.
But OpenBLAS lags painfully far behind MKL on SkylakeX dgemm.

martin-frbg:
Meh. Needs to be `CNAME` rather than `dgemm_kernel` in line 828 so that the build system can supply whatever name it needs...

martin-frbg:
Fixed the naming for you, will leave the algorithm discussion to the adults with SkylakeX access. Just feel free to keep the AVX512 kernels coming ...

fenrus75:
yes the kernels will evolve. I know the dgemm kernel is not nearly great
yet (I have a bunch of things in prototype), but this is the infrastructure
to put in place to be able to do further optimizations....
(small steps kind of thing)

fenrus75:
(my threshold was "be already better than existing, and get the code into a
shape that makes it easier to hack on".. I rather work on integration
issues on the most simple minimum version rather than make it 5x bigger and
then realize some fundamental thing needs redone)

PR #1787, opened 2018-10-04
Fix unknown type name __WAIT_STATUS on RHEL5

### LINKED ISSUE #1519: test_fork utest fails to build with GCC 4.7.2
So there is this HPC cluster where the newest GCC available is 4.7.2, and I have tried to build a fresh dev version of OpenBLAS. If threading is disabled it builds just fine, but with threading enabled the build of one of the utests fail. I suspect this is due to the ancient compilers/libraries/kernel present on the system. Older versions of OpenBLAS did not have this issue.
Log:
```
<snip>
gcc -O2 -DUTEST_CHECK -DSANITY_CHECK -DREFNAME=test_forkf_ -DMAX_STACK_ALLOC=2048 -Wall -m64 -DF_INTERFACE_GFORT -fPIC -DSMP_SERVER -DNO_WARMUP -DMAX_CPU_NUMBER=24 -DASMNAME=test_fork -DASMFNAME=test_fork_ -DNAME=test_fork_ -DCNAME=test_fork -DCHAR_NAME=\"test_fork_\" -DCHAR_CNAME=\"test_fork\" -I..   -c -o test_fork.o test_fork.c
In file included from test_fork.c:35:0:
/usr/include/sys/wait.h:116:22: error: unknown type name ‘__WAIT_STATUS’
In file included from test_fork.c:35:0:
/usr/include/sys/wait.h:169:23: error: unknown type name ‘__WAIT_STATUS’
/usr/include/sys/wait.h:179:38: error: unknown type name ‘__WAIT_STATUS’
test_fork.c: In function ‘__ctest_fork_safety_run’:
test_fork.c:109:13: warning: implicit declaration of function ‘wait’ [-Wimplicit-function-declaration]
test_fork.c:111:13: warning: implicit declaration of function ‘WEXITSTATUS’ [-Wimplicit-function-declaration]
make[1]: *** [test_fork.o] Error 1
<snip>
```

jeromerobert:
With glibc 2.5 one must have #define _XOPEN_SOURCE >= 500 to use wait.
But reading glibc code this is actually needed only if stdlib.h was
included before sys/wait.h. This was the case here through
openblas_utest.h. So changing include fix compilation on RHEL5 and
should ne hurt with more recent distro.

* Problem found when using with gcc 5.5 and 4.7.2 on RHEL5/CENTOS5
* Fix #1519

### DIFF
--- a/utest/test_fork.c
+++ b/utest/test_fork.c
@@ -31,10 +31,10 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 **********************************************************************************/
 
-#include "openblas_utest.h"
 #include <sys/types.h>
 #include <sys/wait.h>
 #include <cblas.h>
+#include "openblas_utest.h"
 
 void* xmalloc(size_t n)
 {


martin-frbg:
Thanks for fixing this, looks as if I lost track of that issue at some point.

PR #1789, opened 2018-10-05
update travis alpine chroot with avx512 intrinsics headers

brada4:
v3.7 should suffice, no harm in later one
https://github.com/alpinelinux/alpine-chroot-install/commits/master/alpine-chroot-install
code fragment is copied verbatim from 
https://github.com/alpinelinux/alpine-chroot-install

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -85,8 +85,8 @@ jobs:
       sudo: true
       language: minimal
       before_install:
-        - "wget 'https://raw.githubusercontent.com/alpinelinux/alpine-chroot-install/v0.6.0/alpine-chroot-install' \
-              && echo 'a827a4ba3d0817e7c88bae17fe34e50204983d1e  alpine-chroot-install' | sha1sum -c || exit 1"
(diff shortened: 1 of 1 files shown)


brada4:
Original has no more quotes???

martin-frbg:
Probably would not hurt to update Alpine, but I am not sure I see the point ? 
The recent build failure with  #1784 was actually not caused by missing headers as I first thought, but by a genuine problem in the code. Due to lack of actual AVX512 hardware in the CI hosts this would only show up in DYNAMIC_ARCH builds that actually included the new SkylakeX code rather than its non-AVX512 Haswell fallback.
BTW anybody trying that Alpine chroot installation locally should be aware that it appears to create hard links to directories on the host, so a quick decision to just `rm -r` its toplevel directory when no longer needed is likely to end in tears.

brada4:
Later than previously used alpine 3.6 (installer 0.6.0) supplies those headers.
alpine binutils are almost very latest at all times, no problem with assemblies.
on `rm -r` note -  it also does make double mounts (not bind type mounts), no matter how hard hardlinks destroy the system you cannot delete the structure....

PR #1800, opened 2018-10-09
Update common_mips64.h for the 1st loop of blas_memory_alloc

fengrl:
The 1st loop work on mips64 and x86 32 only, sync mips64 with other platforms.


martin-frbg:
The history of this function on MIPS is unclear, seems it was copied over from the original libGoto implementation for x86 when the MIPS port was done. With 0.2.12, "experimentally remove expensive function calls" disabled it for x86 so apparently the cost of determining core locality first was greater than any benefits.

Issue #1797, opened 2018-10-08
Compilation failure with gcc-6 ('_mm256_set1_pd': target specific option mismatch)


ayounes-nviso:
Platform: Linux-Intel standard 16.04 (gcc-6)
OpenBLAS HEAD of develop: 6e2c4945567ad8fb5f11be7ac7c0b10c6b4f42cb

Going back to this commit fixes the compilation error:
02ef20a1 Merge pull request #1786 from martin-frbg/immintrin

Full cmake build error:

In file included from /home/ayounes/devel/nv3dfi-build/linux_host/src/openblas_project-build/kernel/CMakeFiles/dgemm_kernel_SKYLAKEX.c:8:0:
/home/ayounes/devel/nv3dfi/ext/pkg-openblas/src/kernel/x86_64/dgemm_kernel_4x8_skylakex.c: In function 'dgemm_kernel_SKYLAKEX':
/home/ayounes/devel/nv3dfi/ext/pkg-openblas/src/kernel/x86_64/dgemm_kernel_4x8_skylakex.c:48:7: warning: AVX vector return without AVX enabled changes the ABI [-Wpsabi]
  ymm4 = _mm256_setzero_pd();  \
  ~~~~~^~~~~~~~~~~~~~~~~~~~~
/home/ayounes/devel/nv3dfi/ext/pkg-openblas/src/kernel/x86_64/dgemm_kernel_4x8_skylakex.c:1307:4: note: in expansion of macro 'INIT4x8'
    INIT4x8()
    ^~~~~~~
/home/ayounes/devel/nv3dfi/ext/pkg-openblas/src/kernel/x86_64/dgemm_kernel_4x8_skylakex.c:736:7: warning: AVX512F vector return without AVX512F enabled changes the ABI [-Wpsabi]
  zmm4 = _mm512_setzero_pd();  \
  ~~~~~^~~~~~~~~~~~~~~~~~~~~
/home/ayounes/devel/nv3dfi/ext/pkg-openblas/src/kernel/x86_64/dgemm_kernel_4x8_skylakex.c:1565:4: note: in expansion of macro 'INIT8x1'
    INIT8x1()
    ^~~~~~~
In file included from /usr/lib/gcc/x86_64-linux-gnu/6/include/immintrin.h:41:0,
                 from /home/ayounes/devel/nv3dfi/ext/pkg-openblas/src/kernel/x86_64/dgemm_kernel_4x8_skylakex.c:36,
[...]

PR #1812, opened 2018-10-11
Use KERNEL_DEFINITIONS rather than COMMON_OPTS to pass -march=skylake…

martin-frbg:
…-avx512


martin-frbg:
Actually kernel/CMakeLists.txt may be a better place for this...

martin-frbg:
confirmed to fix #1797/#1806

Issue #1566, opened 2018-05-17
Can't locate File/Temp.pm in @INC (you may need to install the File::Temp module)


rayjs:
I am using a custom tool chain to compile OpenBLAS for the Target below:
Target is armv7a (32 bit) (EmbeddedOS)
Host is x86_64 Ubuntu 16

Please note that I have previously compiled successfully for armv7a for android.

I get the following error:
```
Can't locate File/Temp.pm in @INC (you may need to install the File::Temp module) (@INC contains: /opt/imx-toolchain/sysroots/x86_64-pokysdk-linux//usr/lib/perl/site_perl/5.22.1 /opt/imx-toolchain/sysroots/x86_64-pokysdk-linux//usr/lib/perl/vendor_perl/5.22.1 /opt/imx-toolchain/sysroots/x86_64-pokysdk-linux//usr/lib/perl/5.22.1 /opt/poky/2.1.3/sysroots/x86_64-pokysdk-linux/usr/lib/perl/site_perl/5.22.1/ /opt/poky/2.1.3/sysroots/x86_64-pokysdk-linux/usr/lib/perl/site_perl/5.22.1 /opt/poky/2.1.3/sysroots/x86_64-pokysdk-linux/usr/lib/perl/vendor_perl/5.22.1/ /opt/poky/2.1.3/sysroots/x86_64-pokysdk-linux/usr/lib/perl/vendor_perl/5.22.1 /opt/poky/2.1.3/sysroots/x86_64-pokysdk-linux/usr/lib/perl/5.22.1/ /opt/poky/2.1.3/sysroots/x86_64-pokysdk-linux/usr/lib/perl/5.22.1 /opt/poky/2.1.3/sysroots/x86_64-pokysdk-linux/usr/lib/perl/5.22.1 .) at ./c_check line 4.
BEGIN failed--compilation aborted at ./c_check line 4.
make: *** [config.h] Error 2
Makefile.system:171: Makefile.conf: No such file or directory
make: *** No rule to make target 'Makefile.conf'.  Stop.
[...]

PR #1813, opened 2018-10-11
Do not fix on /usr/bin/perl, use plain perl instead.

brada4:
Prefer invoking perl command from path, e.g. yocto sdk - called poky - would set PERL5LIB to own perl library path in case local perl is installed,  breaking /usr/bin/perl

### DIFF
--- a/c_check
+++ b/c_check
@@ -1,4 +1,4 @@
-#!/usr/bin/perl
+#!/usr/bin/env perl
 
 use File::Basename;
 use File::Temp qw(tempfile);
--- a/exports/gensymbol
+++ b/exports/gensymbol
@@ -1,4 +1,4 @@
-#!/usr/bin/perl
+#!/usr/bin/env perl
 
 # Changelog
 # 2017/09/03 staticfloat
--- a/f_check
+++ b/f_check
@@ -1,4 +1,4 @@
-#!/usr/bin/perl
+#!/usr/bin/env perl
 
 $hostos   = `uname -s | sed -e s/\-.*//`;    chop($hostos);
 


martin-frbg:
Background for this was #1566, but if the only actual problem was with the yocto toolkit and File::Temp as used in c_check, I would prefer to fix only that rather than risk introducing new incompatibilities.

brada4:
It is universally available in all POSIX systems, just Linux prepends it to many scripts ?in hope somebody does not complain about broken tools?

brada4:
e.g 
`#!/usr/bin/env python3`
has good chance to be fixed up by user, while
'#!/usr/bin/python3.6`
Makes them rely on well behaved root user.

martin-frbg:
Counter arguments are typically about security, e.g. https://pagure.io/packaging-committee/issue/700 and I am not sure if we could rely on "env" itself being universally available in /usr/bin. Again if it is just the File::Temp that is causing problems I would prefer to fix only that. Or does the yocto PERL5LIB setting cause other conflicts ?

brada4:
It could be `#!perl` 
I am sad issue report did not get feedback from OPs

martin-frbg:
Definitely not, that would only work if you had perl in the current directory. From your comment in #1566 I had assumed you had done a build with yocto, poky or whatever.

brada4:
Yes, I installed a toolkit, once you install build tools like gcc and make, and re-run `oe-init-build-env` , /usr/bin/perl appears broken. 
I dont know - sanitize script to permit web mode without parsing variables (`perl -Tt`) 
Or re-execute with environment cleaned (env could help here too - we sort of need basic basic perl)

Issue #1563, opened 2018-05-17
Issue with Nehalen dtrmm


andreasnoack:
From Julia linked to the 0.3.0 prerelease, I'm getting
```julia
julia> BLAS.trmm!('L', 'U', 'N', 'N', Float64(1), ones(Float64,3,3), ones(Float64,3,3))
3×3 Array{Float64,2}:
 3.0           3.0           3.0
 2.0           2.0           2.0
 6.92492e-310  6.92492e-310  6.92492e-310
```
It gives the correct result for `s`, `c`, and `z` or when setting `OPENBLAS_CORETYPE=penryn`.

martin-frbg:
Would you happen to know from previous Julia tests if this is a regression ? The only difference between 0.2.20 and 0.3.0 that I am aware off is that I switched KERNEL.NEHALEM to use the generic implementation of dswap (which I think does not affect trmm). I did notice trmm problems (lapack test failures) on some 32bit platforms including mips but my impression so far is that they are not new.

brada4:
Does CONSISTENT_FPCSR=1 build parameter change things? (number looks like in denormals to the ankles)

martin-frbg:
Bisecting now as I see it worked with ab87ee6b from late october.

martin-frbg:
git bisect says e5cc3d7 is the first bad commit - this is "core.IdenticalExpr clang501 checker" from the "code cleanup" PR #1419 of january 19 which brought the simplification of a lengthy block of conditionals in the generic trmm_ltcopy and trmm_utcopy routines. Sorry (again) for not noticing the problems with this PR when I merged it, I will return the affected files to their earlier state now.

PR #1817, opened 2018-10-13
Clean tailing dead assignments

brada4:
Staying away from rewriting any code sequences.


martin-frbg:
So this is a retry of the parts of #1419 that needed reverting following #1563 ?

brada4:
But withiut introducing rewritten faulty  logic

Issue #1820, opened 2018-10-16
OpenBLAS becomes single-threaded


tjoli:
Hi everybody,

I am using standard lapack routines to diagonalize moderately large matrices with openBLAS.
The box is a xeon with 6 cores and hyperthreading. At runtime execution starts with 12 threads
by default and after some reproducible time goes down to one thread with miserable performance
but execution is without errors. I have observed the same behavior with OMP_NUM_THREADS=2,4,6
and using also OPENBLAS_NUM_THREADS. I tried to set OMP_DYNAMIC=FALSE with no change of behavior. It is a locally compiled version of openBLAS 0.3.3 with gcc/gfortran 7.3.0

any hints would be appreciated !

martin-frbg:
Weird. Does matrix size change by the time you notice the number of active threads going down ? Many BLAS drivers now fall back to using a single thread when the overhead of multithreading is expected to exceed the gain from parallelizing, but this is always local to the individual function. Also most LAPACK functions are copied unmodified from the netlib reference implementation, so the LAPACK itself is not multithreaded. 
Offhand I can only think of "something in your code" tying all threads to the same core via sched_setaffinity. Some minimal code sample that reproduces the problem would be great, but it would probably already help to know if you see this same behaviour with older versions (e.g 0.3.0, 0.2.20)

tjoli:
In fact it happens during treatment of a single matrix. I noticed the same behavior with 0.2.20 hence I upgraded. The inference of single-thread behavior comes from looking at output of "top"
also I used repeatedly the same code on a local computer center and it gives full speed with expected performance.

added:
I checked that a standard openMP code runs all the time with the correct number of threads, by standard I mean without calling any openBLAS.

martin-frbg:
So what are the dimensions of that particular matrix, and which function is it that appears to execute single-threaded ? (Could be a missed or misguided optimization in OpenBLAS, but if you already saw this with 0.2.20 at least it is not a recent regression.) I presume your "standard openMP code" will be using some other implementation of BLAS (like MKL or ATLAS), or is it avoiding BLAS completely ?

tjoli:
I checked matrices of sizes 1800 and 3800. 

The problem appears with lapack routine ZHEEV:
if I want all eigenvectors (so ZHEEV('V'....)) then it slows down to one thread while if I want eigenvalues only ZHEEV now with argument
[...]

PR #1825, opened 2018-10-20
Delay _hemv threading in attempt to address #1820

brada4:
Use more than one thread once there is improvement as illustrated in a graph.
Since there is still a heap of uncertainty and room fro improvement left take following shortcuts:
* use same pre-computed threshold for c_ and z_ versions (see graph in issue - z_ gains from multiple threads, c_ uses second when its gain gets stable)
* dont read global CPU number if we decide one is enough

### DIFF
--- a/interface/zhemv.c
+++ b/interface/zhemv.c
@@ -195,7 +195,12 @@ void CNAME(enum CBLAS_ORDER order, enum CBLAS_UPLO Uplo, blasint n, void *VALPHA
   buffer = (FLOAT *)blas_memory_alloc(1);
 
 #ifdef SMP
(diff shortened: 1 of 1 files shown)


martin-frbg:
Suggest to `#define MULTI_THREAD_MINIMAL` at the top of the file to make it easier to adjust threshold per architecture like it was recently done in axpy.c/zaxpy.c ?

brada4:
In a minute

brada4:
I am certain fixed threshold is essentially wrong even for x86_64 alone, but the idea of holding back threading saves most lapack cases where it churns down to smaller and smaller BLAS operations which would otherwise get heavy damage from being threaded.

PR #1827, opened 2018-10-21
[WIP] Add more documentation regarding parallelism

TiborGY:


brada4:
Aint it better to improve Makefile.rule where the mystery options are set instead?

TiborGY:
> 
> 
> Aint it better to improve Makefile.rule where the mystery options are set instead?

Maybe, but I am working on a more extensive piece of documentation on threading. Too long to be reasonably included in Makefile.rule.

brada4:
Are you sure it is necessary
* there is no complete OMP hierachy support, but will be hopefully
* new code trigerred by USE_TLS=1 in Makefile.rule will make most MM parallel without extra locking

martin-frbg:
No offence meant, but I wonder if it would be more appropriate if you developed this piece of documentation in the wiki of your fork first ? (BTW OpenBLAS is continuing the original GotoBLAS work of K.Goto, a statement like  "the project aims to develop a BLAS package" makes it look as if it was incomplete or its API unstable)

TiborGY:
> 
> 
> No offence meant, but I wonder if it would be more appropriate if you developed this piece of documentation in the wiki of your fork first ? (BTW OpenBLAS is continuing the original GotoBLAS work of K.Goto, a statement like "the project aims to develop a BLAS package" makes it look as if it was incomplete or its API unstable)

Oh OK, the fact that even the wiki is forked on GitHub was never in my thoughts, good idea.
The language used may need adjustment/rephrasing. All suggestions are welcome.

TiborGY:
Given the currently open threading issues, I will probably wait with doing this until things have settled down, one way or another.

brada4:
Check https://github.com/xianyi/OpenBLAS/issues/1882#issuecomment-441563356 with some input for what happens in case where you try to mix OMP implementations.

rgommers:
This PR is only a few sentences and at this point seems abandoned. So I think it can be closed.

Note that in the docs there are a number of places where parallelism is discussed, e.g.:

- http://www.openmathlib.org/OpenBLAS/docs/faq/#how-can-i-use-openblas-in-multi-threaded-applications
- http://www.openmathlib.org/OpenBLAS/docs/distributing/#threading-related-options
- http://www.openmathlib.org/OpenBLAS/docs/faq/#using-openblas-with-openmp

TiborGY:
sure, I never got around to progress with this

PR #1831, opened 2018-10-22
disable threading in C/ZSWAP copying from S/DSWAP

brada4:
SMP is about 10x slower than single thread up until reaches same speed as one thread at 5-10x outer cache sized arguments
 \+ add adjusted arm threading threshold under belief it is sort of data size.

### DIFF
--- a/interface/zswap.c
+++ b/interface/zswap.c
@@ -42,6 +42,14 @@
 #include "functable.h"
 #endif
 
+#if defined(THUNDERX2T99) || defined(VULCAN) || defined(ARMV8)
+// Multithreaded swap gives performance benefits in ThunderX2T99
+#else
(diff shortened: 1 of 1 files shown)


brada4:
I'll check all interfaces with unconditional threading and place some viable threshold 
What I have is 2MB/core cache of 2x haswell and 4x piledriver - same as ZEN and quite close to 2.5MB in modern Xeons.
I'll rename PR when finished, but keep this branch only with stable ideas in the meantime.

martin-frbg:
Separate PRs for different interfaces will probably make more sense, especially if parts should ever need to be reverted ? (and cc: @ashwinyes as this change here affects his thunderx platform)

brada4:
Point taken, will cc involved too.

PR #1833, opened 2018-10-26
register should be push/pull with sdc1/ldc1 on mips64

fengrl:
64 bit register command should be used to push/pop from stack. Otherwise, data will be lost.
Bug had been found about this.

### DIFF
--- a/common.h
+++ b/common.h
@@ -208,6 +208,13 @@ extern "C" {
 #undef	USE_PTHREAD_SPINLOCK
 #endif
 
+#ifdef OS_LINUX
+#ifdef LOONGSON3A 
+#undef  USE_PTHREAD_LOCK
+#undef  USE_PTHREAD_SPINLOCK
+#endif
+#endif
+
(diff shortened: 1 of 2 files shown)


brada4:
In principle you can do that in ABI-neutral way via assembly clobber specification
https://www.ibiblio.org/gferg/ldp/GCC-Inline-Assembly-HOWTO.html#ss5.3
i.e write down registers used in asm section, and compiler saving/restoring them behind the scenes, additionally giving compiler optimizer a chance to avoid them and complete save/restore pair in surrounding code.
It is not done properly and completely elsewhere , and say x86_64 would generate broken code with -O3.
EDIT: present patch does what it has to, though improved one may be much more long-lived.

martin-frbg:
As I said before I know very little about MIPS64 assembly, but isn't LD/ST `#define`d to be the appropriate ldc1/stc1 instruction in common_mips64.h  already ?

PR #1834, opened 2018-10-26
register push/pop command change

fengrl:
64bit push/pop register command should be used. Otherwise, data will lost.

### DIFF
--- a/kernel/mips64/KERNEL
+++ b/kernel/mips64/KERNEL
@@ -6,6 +6,11 @@ CROTKERNEL = ../mips/zrot.c
 ZROTKERNEL = ../mips/zrot.c
 CSWAPKERNEL = ../mips/zswap.c
 ZSWAPKERNEL = ../mips/zswap.c
+SDOTKERNEL   = ../mips/dot.c                                                                                        
+DDOTKERNEL   = ../mips/dot.c                                                                                        
+CDOTKERNEL   = ../mips/zdot.c                                                                                       
+ZDOTKERNEL   = ../mips/zdot.c    
+
 
 ifndef SNRM2KERNEL
 SNRM2KERNEL = snrm2.S
--- a/kernel/mips64/sgemm_kernel_8x4_ps.S
+++ b/kernel/mips64/sgemm_kernel_8x4_ps.S
@@ -146,11 +146,11 @@
 	sd	$21,  40($sp)
 	sd	$22,  48($sp)
 
-	ST	$f24, 56($sp)
-	ST	$f25, 64($sp)
-	ST	$f26, 72($sp)
-	ST	$f27, 80($sp)
-	ST	$f28, 88($sp)
+	sdc1	$f24, 56($sp)
+	sdc1	$f25, 64($sp)
+	sdc1	$f26, 72($sp)
+	sdc1	$f27, 80($sp)
+	sdc1	$f28, 88($sp)
 
 #if defined(TRMMKERNEL)
 	sd	$23,  96($sp)
@@ -161,10 +161,10 @@
 #endif
 
 #ifndef __64BIT__
-	ST	$f20,120($sp)
-	ST	$f21,128($sp)
-	ST	$f22,136($sp)
-	ST	$f23,144($sp)
+	sdc1	$f20,120($sp)
+	sdc1	$f21,128($sp)
+	sdc1	$f22,136($sp)
+	sdc1	$f23,144($sp)
 #endif
 
 	.align	4
@@ -7766,11 +7766,11 @@
 	ld	$21,  40($sp)
 	ld	$22,  48($sp)
 
-	LD	$f24, 56($sp)
-	LD	$f25, 64($sp)
-	LD	$f26, 72($sp)
-	LD	$f27, 80($sp)
-	LD	$f28, 88($sp)
+	ldc1	$f24, 56($sp)
+	ldc1	$f25, 64($sp)
+	ldc1	$f26, 72($sp)
+	ldc1	$f27, 80($sp)
+	ldc1	$f28, 88($sp)
 
 #if defined(TRMMKERNEL)
 	ld	$23,  96($sp)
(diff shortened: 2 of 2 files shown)


martin-frbg:
As commented on the closed PR, LD/ST seems to be mapped to ldc1/stc1 by common_mips64.h already - or is this mapping not done correctly (in which case other mips64 kernels would be affected as well) ?

fengrl:
ldc1/sdc1 was defined by DOUBLE.
And as my understand, while 64 BIT defined, ldc1/sdc1 should be used. But more test should be done and more code should be reviewed if DOUBLE was change to 64 BIT.

martin-frbg:
`__64BIT__` is set when the operating system is 64bit, I believe this only affects the handling of integers (related to `INTERFACE64` option in Makefile.rule). `DOUBLE` is set when the function is expected to handle double precision float values - as this is not the case for sgemm, LD/ST map to lwc1/swc1 here. Do you see an actual miscalculation or even crash on your platform, or did the code only look suspicious to you ?

fengrl:
For 64 bit platform, FP registers are 64 bit. So while push/pop to/from stack, 64bit comand should be used. If 32 bit command was used, maybe only 32 bit data protected. 
Yes, I see actual problem, that other 32 bit lost, code will get a wrong result.

martin-frbg:
Do you see actual problem running the tests, or running your own code ? After all it is only the 32bits of a float argument that are moved to and from the stack here. While I have no mips64 hardware to test, I would expect a problem of this magnitude would have come up at some point in the _seven_ years since this code was written.

fengrl:
grep sdc1/ldc1 under kernel/mips64, you can see that they were used frequently while push/pop to/from the stack. As I understand, under this mips64 directory, sdc1/ldc1 should be used at all places instead of ST/LD.
Yes, I have mips64 hardware test about this.
There is a problem about this operation.

fengrl:
@martin-frbg  Currently, there are only few engineers working on mips64 platform. I guess it is the reason that this problem not come up in the past seven years. And on the other side, this problem is not always, it exists sometimes.
In general, review code of this part can see that it is a problem.
If you have other suggests about this change, please let me know.

martin-frbg:
Maybe this is why I cannot reproduce the problem (with either the provided tests or xianyi's BLAS-Tester that is derived from ATLAS). And I only have qemu for testing, which may not reproduce the actual behaviour in full. What I would like to be sure of is that what you are facing is really an assembly problem and not one of the thread safety issues that seem to keep popping up despite all efforts.

martin-frbg:
qemu is unable to show any different behaviour with or without this specific PR, but replacing all LD/ST with ldc1/stc1 results in many test failures.

FengRillian:
It is not easy to estimate replacing all LD/ST with ldc1/stc1. Related modify can be made after new problems found.

martin-frbg:
Could you comment on what the "dot problem" is please ?

fengrl:
dot problem is Openblas/test,  
      Test of subprogram number 13            SDSDOT
      Test of subprogram number 13            DSDOT

martin-frbg:
Thanks. So if this is "only" about DSDOT, it could be sufficient to add entries redirecting DSDOT and SDSDOT to the mips dot.c (and I guess also remove my apparently misguided attempts to "correct" DSDOT precision from dot.S) ?

fengrl:
My result is that kernel/mips64/../mips/dot.c works while I add SDOTKERNEL/DDOTKERNEL to kernel/mips64/KERNEL

martin-frbg:
I have now obtained a small MIPS64 system for testing (quad-core Cavium Octeon III) and do not see any test failures for DOT there. Perhaps your change could be limited to KERNEL.LOONGSON3A rather than the global KERNEL (and if it is only SDSDOT and DSDOT that fail, add only SDSDOTKERNEL=../mips/dot.c and DSDOTKERNEL=../mips.dot.c there) ?

fengrl:
OPENBLAS_NUM_THREADS=1 OMP_NUM_THREADS=1 ./sblat1
 Real BLAS Test Program Results


 Test of subprogram number  1             SDOT 
                                    ----- PASS -----

 Test of subprogram number  2            SAXPY 
                                    ----- PASS -----

 Test of subprogram number  3            SROTG 
                                    ----- PASS -----

 Test of subprogram number  4             SROT 
                                    ----- PASS -----

 Test of subprogram number  5            SCOPY 
                                    ----- PASS -----

 Test of subprogram number  6            SSWAP 
                                    ----- PASS -----

 Test of subprogram number  7            SNRM2 
                                    ----- PASS -----

 Test of subprogram number  8            SASUM 
                                    ----- PASS -----

 Test of subprogram number  9            SSCAL 
                                    ----- PASS -----

 Test of subprogram number 10            ISAMAX
                                    ----- PASS -----

 Test of subprogram number 11            SROTMG
                                    ----- PASS -----

 Test of subprogram number 12            SROTM 
                                    ----- PASS -----

 Test of subprogram number 13            SDSDOT
                                       FAIL

 CASE  N INCX INCY  I                             COMP(I)                             TRUE(I)  DIFFERENCE     SIZE(I)

   13  2    2   -2  1                     -0.15000001E+00                      0.29999999E-01 -0.1800E+00  0.1700E+01
   13  4    2   -2  1                     -0.10000000E+00                      0.94999999E+00 -0.1050E+01  0.3300E+01
   13  2   -2    1  1                     -0.44000000E+00                     -0.69000000E+00  0.2500E+00  0.1700E+01
   13  4   -2    1  1                      0.52000004E+00                     -0.63999999E+00  0.1160E+01  0.3300E+01
   13  2   -1   -2  1                      0.40000001E+00                      0.43000001E+00 -0.3000E-01  0.1700E+01
   13  4   -1   -2  1                      0.40000001E+00                      0.13700000E+01 -0.9700E+00  0.3300E+01

fengrl:
in kernel/mips64/KERNEL, KERNEL.LOONGSON3A, there is no SDSDOTKERNEL setting.

martin-frbg:
Thank you for the log. Weird that it fails with such a large error, while it passes in qemu and on my cheap new router.
When there is no specific setting for a DSDOTKERNEL, the build system  will use the SDOTKERNEL and pass -DDSDOT to it. (All that does in dot.S is convert the float arguments to double and perform the addition with madd.d instead of madd.s - maybe that hack is too simple to work on serious hardware like Loongson3A). 
As SDOT (and DDOT) apparently works correctly, there should be no need to replace the assembly code for dot in general with the (probably slower) generic C implementation. (And I made a mistake in my earlier message - a separate entry for SDSDOTKERNEL is not needed, just the single line `DSDOTKERNEL=../mips/dot.c` needs to be added )

martin-frbg:
Merging this to get the fix for the initial sgemm kernel issue , but I intend to follow up with a commit that restores the mips64 DOT microkernels and adds a DSDOTKERNEL entry to KERNEL.LOONGSON3A instead

PR #1837, opened 2018-10-28
Ensure that blas_thread_init has been called in openblas_set_num_threads

embray:
Not doing so can lead to bugs if the process happens to have been forked at some point.  For example:

```c
pid = fork();
if (pid) {
    openblas_set_num_thread(8);
    // <call some multi-threaded blas function (dgemm, etc.)>
}
```

This leads to segfaults because currently `openblas_set_num_threads` assumes `blas_thread_init()` has already been called and `blas_server_avail == 1`.  But after a fork, due to the `pthread_atfork` handler which calls `blas_thread_shutdown`, this is no longer the case.  But then when you get into `exec_blas_async` it does call `blas_thread_init()` even though there are already threads running, creating a likelihood that one or more of those threads segfault.

### DIFF
--- a/driver/others/blas_server.c
+++ b/driver/others/blas_server.c
@@ -850,6 +850,11 @@ void goto_set_num_threads(int num_threads) {
 
   long i;
 
+#ifdef SMP_SERVER
+  // Handle lazy re-init of the thread-pool after a POSIX fork
+  if (unlikely(blas_server_avail == 0)) blas_thread_init();
+#endif
+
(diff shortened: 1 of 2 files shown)


brada4:
Should work fine.
It is one-time code in general, "unlikely" tagging just impairs readability.

embray:
Sure. It was literally just copy/pasted from elsewhere. I see no harm in including unlikely() but I also agree that in most cases it is not significant.

PR #1839, opened 2018-10-30
remove certain duplicate #include-s

brada4:
Only pertaining stdio math stdlib string headers.
Should any problem arise it can be compensated from common.h


martin-frbg:
Probably needs rearranging the other `#include`s so that they come after common.h in case any of them depend on one of the headers this removes (on any of the supported platforms). 
Not sure this is worth it anyhow - changing over 500 files just to make them prettier ?

brada4:
I knew you will be looking at 
```
bench.../....c:

#ifdef __CYGWIN32__
#include <sys/time.h>
```

That can be removed to as sys/time.h is included unconditionally in common.h 
Ths is just braindead grunt work part. There is some intelligence involved in cleaning other.

brada4:
In principle it should become
```
#include "common.h"
#include <extra_system.h>
#include "uncomon.h"
```

brada4:
I will split the patch into parts that have just single surplus file (like 400) then pay more attn to multiple , carefully sorting with common.h firss firstly.

PR #1846, opened 2018-11-01
gemm/dgemm: add a way for an arch kernel to specify preferred sizes

fenrus75:
sgemm/dgemm: add a way for an arch kernel to specify preferred sizes

    The current gemm threading code can make very unfortunate choices, for
    example on my 10 core system a 1024x1024x1024 matrix multiply ends up
    chunking into blocks of 102... which is not a vector friendly size
    and performance ends up horrible..

    this patch adds a helper define where an architecture can specify
    a preference for size multiples.
    This is different from existing defines that are minimum sizes and such.

    The performance increase with this patch for the 1024x1024x1024 sgemm
    is 2.3x (!!)


martin-frbg:
Interesting to say the least... but I expect the net effect is smaller on more mundane AVX2 hardware ?

fenrus75:
only sort of. lets say 10 core (20 thread) system, 1024x1024x1024 matrix.
makes blocks of 51 wide...

so that is a 32 wide stride, then a 16 (both can still be performant) then a 2 and a 1
the 2 and 1 each do a full "k loop" down and have a large number of memory accesses, but no real parallel math to speak of so the 2 and 1 are same cost as the 16 .. even on AVX2.

by doing similar rounding up to nice multiples, this same thread does 2 "K loops" of 32 each instead of 4 "K loops"

zerothi:
A small nit-pick. It should be *preferred* no?
I can make a PR if desired?

Issue #1844, opened 2018-10-31
OpenBLAS threadsafety issues for downstream libraries (NumPy)


seberg:
Since OpenBLAS is one of the most commonly used BLAS implementations with NumPy, there are currently two points being discussed within NumPy. One thing is the spawning of too many threads when using multi processing/python threads (although is probably something we need to find a solution for, but pointers are welcome!).

The main issue is the thread safety of OpenBLAS when multiple threads are used. This may not be problematic for many projects explicitly using OpenBLAS, but NumPy is a bit different:
  * NumPy does not actually know which BLAS is being used.
  * Thus, NumPy has no control on the number of threads (the system default is used).

Together this means NumPy will often use multi threaded OpenBLAS. Now if someone uses a numpy function within threads they might get silently incorrect results (https://github.com/numpy/numpy/issues/11046) and this seems true for OpenBLAS versions 0.2.18, 0.2.20, 0.3.0 (probably generally).

The possibility of silently wrong results deeply troubles me. Downstream users of NumPy are unlikely to be aware of such issues, or might not even notice that they are using BLAS at all. The problem is that there seems to be no easy solution.

We could use locking to disable threading, but that probably is incorrect/unnecessary often as well.
I have been talking with @mattip and others at NumPy, and hoped we can start a small discussion here. Maybe there are fixes/workarounds we can implement in NumPy, but maybe we can also provide help/resources to make OpenBLAS thread safe?

brada4:
Spawning too many threads is not exactly thread safety issue, it is a configuration choice made by user.

The accuracy issue indeed reproduces even with OpenBLAS threading disabled (but not with ATLAS or reference BLAS)

btw numpy perfectly knows which blas is used, and can set threads as needed (at first sight - could be only problem is no support for OMP nesting in OpenBLAS, or that MKL gets some special treatment, or both)
```
python -c "import numpy; print(numpy.__config__.show())"
lapack_opt_info:
    libraries = ['openblas', 'openblas']
    library_dirs = ['/usr/lib64']
    define_macros = [('HAVE_CBLAS', None)]
    language = c
blas_opt_info:
    libraries = ['openblas', 'openblas']
    library_dirs = ['/usr/lib64']
    define_macros = [('HAVE_CBLAS', None)]
    language = c
openblas_info:
    libraries = ['openblas', 'openblas']
    library_dirs = ['/usr/lib64']
    define_macros = [('HAVE_CBLAS', None)]
    language = c
[...]

Issue #532, opened 2015-04-06
OpenBLAS 6 times slower than MKL on DGEMV()


hiccup7:
Small vector scenario.
26.7 seconds for OpenBLAS in Julia:

```
blas_set_num_threads(CPU_CORES)
const trans = 'N'
const a = ones((201, 150))
const x = ones(150)
@time for k=1:1000000; s = BLAS.gemv(trans, a, x); end
```

4.6 seconds for MKL in Python:

```
import numpy as np
from scipy.linalg.blas import dgemv
from timeit import default_timer as timer
alpha = 1.0
a = np.ones((201, 150), order='F')
x = np.ones(150)
start = timer()
for k in range(1000000):
    s = dgemv(alpha, a, x)
exec_time=(timer() - start)
print
print("Execution took", str(round(exec_time, 3)), "seconds")
```

Large vector scenario.
15.7 seconds for OpenBLAS in Julia:

```
blas_set_num_threads(CPU_CORES)
const trans = 'N'
const a = ones((4, 100000))
const x = ones(100000)
@time for k=1:100000; s = BLAS.gemv(trans, a, x); end
```

7.9 seconds for MKL in Python:

```
import numpy as np
from scipy.linalg.blas import dgemv
from timeit import default_timer as timer
alpha = 1.0
a = np.ones((4, 100000), order='F')
x = np.ones(100000)
start = timer()
for k in range(100000):
    s = dgemv(alpha, a, x)
exec_time=(timer() - start)
print
print("Execution took", str(round(exec_time, 3)), "seconds")
```

Tested environment is WinPython-64bit-3.4.3.2FlavorJulia at http://sourceforge.net/projects/winpython/files/WinPython_3.4/3.4.3.2/flavors/
The same Python time was measured in 64-bit Anaconda3 v2.1.0.

From versioninfo(true) in Julia:

```
Julia Version 0.3.7
System: Windows (x86_64-w64-mingw32)
CPU: Intel(R) Core(TM) i7-4700HQ CPU @ 2.40GHz
WORD_SIZE: 64
BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell)
```

[...]

Issue #1847, opened 2018-11-01
Segmentation fault using serial OpenBLAS with OpenMP on Windows


hjbreg:
Updated test results (OpenBLAS 0.3.3)

### Windows 64-bit

gcc version 4.9.3 20150626 (Fedora MinGW 4.9.3-1.el7)

- USE_THREAD=0 NUM_THREADS=1
**segmentation fault**

- USE_THREAD=0 NUM_THREADS=16
**incorrect result**

- USE_THREAD=1 NUM_THREADS=2 USE_OPENMP=1 NUM_PARALEL=16
**incorrect result**

- USE_THREAD=1 NUM_THREADS=2 openblas_set_num_threads(1)
**incorrect result**

### Linux 64-bit

gcc version 4.4.7 20120313 (Red Hat 4.4.7-4)

- USE_THREAD=0 NUM_THREADS=1
**too many memory regions**

- USE_THREAD=0 NUM_THREADS=16
**OK**

----

Segmentation fault if OMP_NUM_THREADS > 4, but linking to netlib-lapack is ok, so I think this may be OpenBLAS side problem. I also tested it on Linux, no segmentation fault.

**GCC: x86_64-w64-mingw32-gcc 4.9.3**

**Single threaded OpenBLAS is built**

USE_THREAD=0 DYNAMIC_ARCH=1 DYNAMIC_OLDER=0 NO_CBLAS=1 NO_LAPACKE=1 NO_SHARED=1

**Test code: segfault-win.cpp**

```
#include <cstdlib>
#include <vector>
#include <iostream>

// g++ segfault-win.cpp -lopenblas -lgfortran -lquadmath -fopenmp

#define bint int

extern "C"
void dgels_(char *trans, bint *m, bint *n, bint *nrhs, double *a, bint *lda, double *b, bint *ldb,
            double *work, bint *lwork, bint *info);

extern "C" void openblas_set_num_threads(int);
extern "C" int openblas_get_num_threads();
extern "C" int omp_get_max_threads();

int C_dgels(char trans, bint m, bint n, bint nrhs, double *a, bint lda, double *b, bint ldb)
{
    bint info = 0;
    double wkopt;
    bint lwork = -1;
    dgels_(&trans, &m, &n, &nrhs, a, &lda, b, &ldb, &wkopt, &lwork, &info);
    if (info == 0) {
        lwork = static_cast<bint>( wkopt );
        double *work = new double[lwork];
[...]

PR #1852, opened 2018-11-03
Revert change from #532 due to unsafe use of static buffer

martin-frbg:
see #1844

### DIFF
--- a/driver/level2/gemv_thread.c
+++ b/driver/level2/gemv_thread.c
@@ -62,11 +62,6 @@
 #endif
 #endif
 
-#ifndef TRANSA
-#define Y_DUMMY_NUM 1024
-static FLOAT y_dummy[Y_DUMMY_NUM];
-#endif
-
 static int gemv_kernel(blas_arg_t *args, BLASLONG *range_m, BLASLONG *range_n, FLOAT *dummy1, FLOAT *buffer, BLASLONG pos){
 
   FLOAT *a, *x, *y;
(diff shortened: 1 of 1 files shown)


hjbreg:
This does not work for #1847, results are still incorrect.

martin-frbg:
#1847 seems unrelated now (possibly mingw/Windows-only and specific to Sandybridge).
And it seems it may be possible to keep the optimization from #532 by declaring either `#pragma omp threadprivate(y_dummy)` (OpenMP) or `static __thread FLOAT y_dummy` (pthreads) rather than just `static FLOAT y_dummy`.

seberg:
I can confirm that this (at least for pthreads I think), fixes the specific issue from gh-1844/https://github.com/numpy/numpy/issues/11046

A bit curious, is there a way to add tests for such issues?

martin-frbg:
"this" being the revert as per this PR, or my belated suggestion to make the y_dummy array thread-local ? 
I guess we could add a test case for this to the "utest" directory to keep it from reappearing, but (a) it would need to be a self-contained C program (or function) rather than a python/numpy script and (b) it would still be a bit outlandish compared to the other, simple checks. I guess in an ideal world the test would have been someone commenting "stop, you cannot do that" on the PR, but this project happens to have very few contributors compared to the number of (often unwitting indirect) users it managed to acquire over the years.

seberg:
Oh, sorry with "this" I mean the PR. I have been hesitant to dive in the code.

Many users and few contributors is a general issue and I am sure it is much worse for OpenBLAS compared to numpy.  Maybe we will add the "test" to numpy, sometimes downstream is a good "test suit" ;).
I guess ideally the PR changing such things has the tests to make everyone confident it works – even without inspecting the code perfectly. But, testing threading stuff is hard I guess :(.

martin-frbg:
I think I am going to rework this to keep the problematic optimization, while making sure that it only gets built when conditions allow thread-safe execution. (That is, with the y_dummy array either  threadprivate in conjunction with USE_OPENMP or annotated with __thread where the compiler supports it)

martin-frbg:
Closing as superseded by my PR #1865

PR #1856, opened 2018-11-05
[Arm64) Revert A53 detection as A57

### LINKED ISSUE #1855: Revert detecting A53 as A57
On OpenHPC, I have proposed we default to building for A57, but that's because OpenHPC is an HPC oriented build, not a generic one. 

While defaulting A53 to build like A57 has shown some improvements on the hardware I have access to, it may not be true for other, more mobile oriented, hardware.

I recommend we revert that change, ie. make A53 build ARMv8 by default and, in the future, improve the ARMv8 build with use of NEON by default. I will create a pull request for this issue, just wanted to give some background.

rengolin:
This patch reverts the decision of treating A53 like A57, which was
based on an analysis done on server class hardware and is not
representative of all A53s out there.

Fixes #1855.

### DIFF
--- a/cpuid_arm64.c
+++ b/cpuid_arm64.c
@@ -115,8 +115,8 @@ int detect(void)
 	fclose(infile);
 	if(cpu_part != NULL && cpu_implementer != NULL) {
 		if (strstr(cpu_implementer, "0x41") && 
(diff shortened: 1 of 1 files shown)


martin-frbg:
Just for reference, this was PR#1449 based on issue #1439 (basically at a time when ARMV8 was still a generic C implementation)

Issue #1690, opened 2018-07-19
Threading and thread safety documentation


TiborGY:
The documentation of multithreading/thread safety capabilities are currently incomplete.

Case in point, I want to build a Fortran library that makes some BLAS/LAPACK calls, then use this library in a C++ program, that uses C++11 parallelism to call a subrutine from said Fortran library in parallel. If I want the BLAS/LAPACK operations to run single threaded, do I still need to set `USE_OPENMP = 1` and `NUM_PARALLEL = N`, or is it irrelevant if I set `USE_THREAD = 0`?
What if I wanted to use OpenMP in my C++ code, instead of C++11 parallelism?

So a decision tree of possible use cases or a flowchart would be nice to help choose the proper values for `USE_THREAD  USE_OPENMP  NUM_THREADS NUM_PARALLEL` and when to use `FCOMMON_OPT = -frecursive`

IMO, the number of threading related build time variables, and their relationships to each other are getting more and more numerous, without proper documentation.

martin-frbg:
No need to dramatize this, the only recent addition is NUM_PARALLEL which is more of an experimental workaround to allow several instances of OpenBLAS, invoked from different parts of an OpenMP main program, to coexist without trashing thread memory. So you want to set/change this only if you build with USE_OPENMP=1, if at all. When you build with USE_THREAD=0 for a single-threaded OpenBLAS, there is no need to set USE_OPENMP=1 - in fact you will get a warning message from Makefile.system if you try. (It _may_ still be necessary to set NUM_THREADS to a number greater than 1 in some cases just to ensure OpenBLAS allocates a sufficiently sized memory buffer internally).
 When you want to use OpenMP in your C++ code _and_ want a parallel OpenBLAS, it is mandatory to build OpenBLAS with USE_OPENMP=1 to avoid locking conflicts.
FCOMMON_OPT=-frecursive is a workaround for a thread safety problem inherited from netlib LAPACK, see http://icl.cs.utk.edu/lapack-forum/viewtopic.php?t=1930, so you would normally want to enable this in multithreaded builds.

TiborGY:
OK, thanks for the clarification. I am probably going to expand the wiki with these threading recommendations.
 I have two more questions though:
- Is there a performance difference between using a single threaded build, and a parallel build set to 1 thread at runtime?
[...]

PR #1857, opened 2018-11-06
Add gfortran -frecursive option from upstream and #1847

brada4:
Most people assume thread-safety from single-threaded calls. This is not the case by default with gfortran. It is hinted already by LAPACK makefile.inc


brada4:
Should fix #1847 thread safety issue

martin-frbg:
I have been meaning to look up the reasoning behind the "only on 64bit Linux" note, for what looks more like a gfortran topic. (And I seem to remember some attempt to eliminate the need for it in Reference-LAPACK a while back).
There probably needs to be a similar change in cmake (fc.cmake ?) as well.

brada4:
It is still there in reference lapack. Probably needs to be looked at while importing...

hjbreg:
also discussed in #1690 
also mentioned in lapack bug list and errata
http://www.netlib.org/lapack/bug_list.html
http://www.netlib.org/lapack/errata_from_342_to_350.html
and
http://wwwf.imperial.ac.uk/~mab201/20120814.html
https://web.stanford.edu/group/fan/S4/dev_info.html#lapack

brada4:
@hjbreg OpenBLAS does not have alternative implementation of LAPACK, it just includes a copy of reference implementation. They use -frecursive with gfortran, and now so will we.

martin-frbg:
Also discussed in http://icl.cs.utk.edu/lapack-forum/viewtopic.php?t=1930 - the one attempt to get rid of -frecursive in netlib was apparently based on a misunderstanding of the scope of the problem.

brada4:
I think discussion there goes about gaming compiler allocating (bigger fixed size private) temporary arrays as global structures. Since there is no deep recursion in LAPACK they get allocated on the stack and we are fine.

martin-frbg:
Yes, but the entire point of `-frecursive` is to keep gfortran from creating static storage for temporary arrays that it "thinks" are too big to put on the stack. (Which is probably why this is a gfortran-specific issue - just not limited to multithreaded builds and/or 64bit linux as the comment in Makefile.rule claims)

brada4:
Exactly.... And in this issue we hit a problem that single-threaded build is presumed thread-safe, when it is not courtesy GCC.

hjbreg:
Do not know why `-frecursive` was disabled by the commit https://github.com/xianyi/OpenBLAS/commit/9d6f2b594e1555c03b3da7434399e3cc810fcbd9 from #390

martin-frbg:
Either he was using a different fortran compiler at that time (and got a warning about an unknown option) or he misjudged the consequences. Putting it in the middle of an unrelated changeset certainly did not help.

Issue #1141, opened 2017-04-04
single thread cause core dump


qiankunhuanying:
I make the source code to generate the libopenblas*.a, and use the APIs in my.a. Then I make a new.so  including both libopenblas*.a and my.a,  and make an executable file that reference this new.so to test my code. When I use the multi-thread libopenblas*.a(just use "make ONLY_CBLAS=1"), the main() function works normally. but when I use the single-thread libopenblas*.a(use "make USE_THREAD=0"), the main() causes core dump--"segmentation fault(core dump)".
    I tried version 0.2.18 and 0.2.20, but both has the core dump, even I just called functions that donot call openblas APIs.
    My platform is allwinner-R16, CPU is ARMV7, and I made the libopenblas*.a with cross compiler.

martin-frbg:
1. Did you do a "make clean" before the second "make" with different options (the "make USE_THREAD=0") ?
2. Can you use a debugger (such as gdb) on that platform to find where the core dump occurs ?

qiankunhuanying:
@martin-frbg  thank you very much.
1. yes,  I make clean  first.
2. as i used a *.so, the gdb on my device noticed that "warning: unable to find dynamic linker breakpoint function. GDB will be unable to debug shared library initializers and track explicitly loaded dynamic code."
so it didnot tell the function having core dump. I will change *.so to just call *.c file, and use GDB to have a try. then I will update the results.

ashwinyes:
@qiankunhuanying  From your first post, looks like you have used the following steps to compile.
Compiling for Multithread - "**make ONLY_CBLAS=1**"
Compiling for single thread - "**make USE_THREAD=0**" 

If the above are the steps that you followed, then they are not consistent with each other.
Could you try compiling single thread with "**make ONLY_CBLAS=1 USE_THREAD=0**" just to confirm that it is not related to using the make param. "ONLY_CBLAS=1"

brada4:
Such crash is usually sign of calling convention mismatch, i.e ld.so calls library initialiser function before main()
probably best start is to run 'file' command against libc and init on the target system and your built .so files

qiankunhuanying:
@ashwinyes 
I didnot describe concretely in my first post, 
1. compiling for Multithread, I used
"make clean
[...]

Issue #938, opened 2016-08-02
'BLAS : Program is Terminated...' Even When Not Cross-Compiled


u55:
Hi OpenBLAS developers,

I recently came across a test failure in python's scipy package, scipy/scipy#6422, when scipy is compiled with openblas using the configuration:

```
"FC=gfortran USE_OPENMP=0 USE_THREAD=1 MAJOR_VERSION=3 NO_LAPACK=0 BUILD_LAPACK_DEPRECATED=1"
```

The scipy test runs several matrix math operations in 20 python threads (within a single process controlled by python's Global Interpreter Lock) and produces the error message 
`'BLAS : Program is Terminated. Because you tried to allocate too many memory regions'` multiple times and then segfaults. 

Having read the [FAQ](https://github.com/xianyi/OpenBLAS/wiki/faq#allocmorebuffers) for openblas, I recompiled openblas with the additional configuration option `'NUM_THREADS=64'`, which seems to fix the problem. However, this seems to me to be an openblas bug, based on a comment from @jeromerobert in issue #889, who said this about `'NUM_THREADS'`:

> ... It should be automatically detected by the build system. The only reason to manually set NUM_THREADS is to build OpenBLAS for an other machine which have more physical cores than the current machine.

Since I am compiling openblas myself on the same machine that is using scipy, this suggests to me that openblas is not detecting the right value of `NUM_THREADS` automatically for my CPU during the build process. However, I find

```
...
CORE=SANDYBRIDGE
LIBCORE=sandybridge
NUM_CORES=8
...
```

written in Makefile.conf, which suggests that openblas is able to detect the correct number of logical cores of my CPU. So I don't understand why the `'NUM_THREADS=64'` configuration is necessary. Can you please confirm if this is an openblas bug, or if it is a "feature"?

Thanks.

martin-frbg:
As I read #889, jeromerobert (and theoractice later in the thread) was suggesting that running such a big number of threads is not expected to bring any performance benefits, hence OpenBLAS' automatic setup is for a much smaller limit. 
(I do note that the #define NUM_BUFFERS (MAX_CPU_NUMBER * 2) is unchanged from Kazushige Goto's work on the original GotoBLAS of ~ 10 years ago, so maybe doing some benchmarks with varying thread numbers _may_ provide arguments for changing this default on current hardware. On the other hand, Sandybridge is not brand-new either and I suspect one of the developers may have done just that for such architecture already.)

brada4:
[...]

PR #1858, opened 2018-11-06
Add minimum threshold for number of buffers

brada4:
And dont break expectation that single-thread version can work in many threads.
#1847


martin-frbg:
Also related are #938, #1141 - though it could be that is it ultimately the GEMM buffer handling that needs looking into.

brada4:
Yes, it is wiping dirt under doormat type of solution.

martin-frbg:
Not sure if we want to go all the way to 64 as at least with the non-tls implementation this will actually have an impact on memory requirements ?

brada4:
Just in memory.c
```
#endif /* defined(SMP) */
    local_memory_table = (struct alloc_t **)malloc(sizeof(struct alloc_t *) * NUM_BUFFERS);
    memset(local_memory_table, 0, sizeof(struct alloc_t *) * NUM_BUFFERS);
#if defined(SMP)
```
i.e more 62 64-bit pointers (496 bytes) get pre-allocated for 1-CPU version

martin-frbg:
Yes, but that's the experimental TLS version you are looking at - pretty sure that in the other/older/default implementation it corresponds to an actual allocation.

brada4:
The "old" version
(8+8+8+8+48)*62=4960
```
static volatile struct {
  BLASULONG lock;
  void *addr;
#if defined(WHEREAMI) && !defined(USE_OPENMP)
  int   pos;
#endif
  int used;
#ifndef __64BIT__
  char dummy[48];
#else
  char dummy[40];
#endif
```

martin-frbg:
Just looking for reasons why nobody tried to uncouple NUM_BUFFERS from NUM_CPU (NUM_THREADS) before, although this problem has come up in some form or other every few months in the past years.

brada4:
I think it is buffer (me) or two (formula there) per thread, the big issue is that single-threaded version depends on this structure at all.

Issue #1851, opened 2018-11-03
Thread safety issue with dgemm and >2 cores


bbbbbbbbba:
On my test computer with 8 cores, the following test snippet fails as long as `OPENBLAS_NUM_THREADS` is set to something >= 3 (or left unset):

```
import numpy as np
import threading

def do_test(idx):
    for i in range(100 * idx, 100 * idx + 100):
        a = np.ones((1 << 20, 2)) * i
        b = np.eye(2)
        mat_0 = a.dot(b).reshape(-1)
        mat_1 = a.dot(b).reshape(-1)
        if np.any(mat_0 != mat_1):
            indices = np.nonzero(mat_0 != mat_1)[0]
            if len(indices) > 100: print('Error %d %d' % (i, len(indices)))
            else: print('Error %d %s' % (i, [(j, mat_0[j], mat_1[j]) for j in indices]))

for idx in range(8):
    threading.Thread(target=do_test, args=(idx,)).start()
```

The version of OpenBLAS I am using is `libopenblasp-r0-8dca6697.3.0.dev.so` (bundled with numpy 1.15.3).

_Originally posted by @bbbbbbbbba in https://github.com/xianyi/OpenBLAS/issues/1844#issuecomment-435145042_

bbbbbbbbba:
After changing `np.ones((1 << 20, 2))` to `np.ones((100000, 2))`, the bug only happens with `OPENBLAS_NUM_THREADS` set to 3, 4, 5, or 6. Debugging seems to suggest that for my installation `GEMM_R` is set to 15856, which would mean the bug only happens when [this loop](https://github.com/xianyi/OpenBLAS/blob/939452ea9dcb57abdcc3f1278c6db668a4690465/driver/level3/level3_thread.c#L639) is executed more than once.

bbbbbbbbba:
OK, I think I get the story of this bug.

With `OPENBLAS_NUM_THREADS` set to `nthreads`, all operations large enough for OpenBLAS multithreading is split into tasks, which comes in batches of `nthreads` _(or smaller, but that doesn't come up here)_. The main thread dispatches `nthreads - 1` of those tasks to worker threads _(there are `nthreads - 1` of them, and they are shared between all main threads; the main thread holds a global `exec_queue_lock` to prevent anyone else from messing with task assignment)_, and leaves one task for itself. Then it waits for results from each worker thread before going on. 

[...]

PR #1865, opened 2018-11-10
Optimize gemv for small M, large N only if it can be done in a threadsafe manner

### LINKED ISSUE #1844: OpenBLAS threadsafety issues for downstream libraries (NumPy)

martin-frbg:
commit 8e5a108 tried to improve the performance of #532 but introduced a static array in the process, breaking programs like numpy that call into OpenBLAS from several concurrent threads. Supersedes the simple revert from #1852 with the intention to fix #1844 without sacrificing performance where possible.


brada4:
Looks like includes the observations from mentioned issue.

martin-frbg:
Sorry, I did not get the comment ? (Unfortunately I cannot test this on my hexacore system this weekend, but I had tested the general idea outlined in the comments on #1852 earlier)

brada4:
It looks that it enables gemv optimisation for OMP only. 
Does not fail with Android UserLAnd (sort of debian chroot, atom CPU) in OMP or pthread builds, though I cannot tell if it ever failed before in this combination.
EDIT: it fails in pre=patches pthread version and ubuntu numpy, though multiprocessorness of this "debian" is not always guaranteed, you somehow need to spin many bg apps to make cpus wake from park. Does not seem possible from "app" I'd wait for more reliable tests...

martin-frbg:
The intention is that it enables the optimization only when the static buffer can be made thread safe, that is either under OMP or when the compiler is capable of assigning C11-like TLS.

brada4:
Patches do the trick most likely just that android tablet chroot is not always multiprocessor like a normal computer....

martin-frbg:
Merged after successful local tests

mattip:
Thanks. Since this is a critical bug for NumPy, will it be part of a bugfix release?

martin-frbg:
Guess 0.3.4 is overdue in any case, will see if I can do it next weekend or at least by the end of this month. Need to try to get a decent handle on the equally ugly #1851 as well.

Issue #1803, opened 2018-10-09
AIX 7.1 with GCC/GFortran - Issue needing resolution


Teej42:
I did review [Issue 463](https://github.com/xianyi/OpenBLAS/issues/463).

I set up everything using the packages provided [here](https://www.ibm.com/developerworks/aix/library/aix-toolbox/alpha.html).  This includes the makefile, GCC and GFortran for AIX 7.1, Python 2.7, and so on.  My objective is to get OpenBLAS set up to get SciPy module working (due to Lapack/BLAS dependency) for Python 2.7.

I updated Makefile.system to comment out the -m32/-m64 as noted in Issue 463:

#ifndef BINARY_DEFINED
#ifdef BINARY64
#CCOMMON_OPT += -m64
#else
#CCOMMON_OPT += -m32
#endif
#endif

I ran this:

`/opt/freeware/bin/make FC=gfortran TARGET=POWER8 >result.txt 2>error.txt`

But in the error.txt, I get this:

Assembler:
/home/fp/OpenBLAS-0.2.20/tmp/cc7zjwrP.s: line 71: Error In Syntax 
/home/fp/OpenBLAS-0.2.20/tmp/cc7zjwrP.s: line 72: invalid opcode or pseudo-op
/home/fp/OpenBLAS-0.2.20/tmp/cc7zjwrP.s: line 73: Error In Syntax 
/home/fp/OpenBLAS-0.2.20/tmp/cc7zjwrP.s: line 100: Error In Syntax 
/home/fp/OpenBLAS-0.2.20/tmp/cc7zjwrP.s: line 101: Error In Syntax 
make[1]: *** [Makefile.L1:622: sasum_k.o] Error 1
make[1]: *** Waiting for unfinished jobs....
Assembler:
/home/fp/OpenBLAS-0.2.20/tmp/ccVmE6OQ.s: line 63: Error In Syntax 
/home/fp/OpenBLAS-0.2.20/tmp/ccVmE6OQ.s: line 64: invalid opcode or pseudo-op
/home/fp/OpenBLAS-0.2.20/tmp/ccVmE6OQ.s: line 65: Error In Syntax 
/home/fp/OpenBLAS-0.2.20/tmp/ccVmE6OQ.s: line 85: Error In Syntax 
/home/fp/OpenBLAS-0.2.20/tmp/ccVmE6OQ.s: line 86: Error In Syntax 
make[1]: *** [Makefile.L1:667: scopy_k.o] Error 1
[...]

PR #1868, opened 2018-11-12
Use prtconf to determine CPU type on AIX

martin-frbg:
for #1803

### DIFF
--- a/cpuid_power.c
+++ b/cpuid_power.c
@@ -56,6 +56,7 @@
 #define CPUTYPE_CELL       6
 #define CPUTYPE_PPCG4	   7
 #define CPUTYPE_POWER8     8
+#define CPUTYPE_POWER9     9
 
 char *cpuname[] = {
   "UNKNOWN",
@@ -66,7 +67,8 @@ char *cpuname[] = {
   "POWER6",
   "CELL",
   "PPCG4",
-  "POWER8"
+  "POWER8",
+  "POWER9"
 };
 
 char *lowercpuname[] = {
@@ -78,7 +80,8 @@ char *lowercpuname[] = {
   "power6",
   "cell",
   "ppcg4",
-  "power8"
+  "power8",
+  "power9"	
 };
 
 char *corename[] = {
@@ -90,7 +93,8 @@ char *corename[] = {
   "POWER6",
   "CELL",
   "PPCG4",
-  "POWER8"
+  "POWER8",
+  "POWER8"   	
 };
 
 int detect(void){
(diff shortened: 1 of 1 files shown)


ayappanec:
Power 9 released recently. We can include that as well.

martin-frbg:
Good point, thanks. Mind you I have no way to test this.

cpuid_power.c
```diff
@@ -120,13 +124,41 @@ int detect(void){
...
+  pclose(infile);
+
+  if (!strncasecmp(p, "POWER3", 6)) return CPUTYPE_POWER3;
+  if (!strncasecmp(p, "POWER4", 6)) return CPUTYPE_POWER4;
+  if (!strncasecmp(p, "PPC970", 6)) return CPUTYPE_PPC970;
+  if (!strncasecmp(p, "POWER5", 6)) return CPUTYPE_POWER5;
+  if (!strncasecmp(p, "POWER6", 6)) return CPUTYPE_POWER6;
+  if (!strncasecmp(p, "POWER7", 6)) return CPUTYPE_POWER6;
+  if (!strncasecmp(p, "POWER8", 6)) return CPUTYPE_POWER8;
+  if (!strncasecmp(p, "POWER8", 6)) return CPUTYPE_POWER8;
```
sbutcher: should this say 
`   if (!strncasecmp(p, "POWER9", 6)) return CPUTYPE_POWER8; `
instead?

sbutcher (commented):
I'm not familiar with the code, but we have a POWER9 and it built successfully with these changes, _however_ i noticed a couple of things i'd like to query. 
Also, not in this PR, but perhaps should be: is line 183 also wrong by having 2 `returns`?
`return return CPUTYPE_POWER8;`

cpuid_power.c
martin-frbg: Err, yes... and I guess the double "return" in 183 proves that I was never good at copypasting.

sbutcher:
Thanks. Can confirm this builds for me on POWER9, and I was able to compile `test_cblas_dgemm.c` against this build.

martin-frbg:
Thanks - and just to confirm, your resulting library has "power9" in its name now rather than the earlier "power5" (and the build used the POWER8 target) ?

sbutcher:
We built using easybuild, and it produces
```
libopenblas_power8p-r0.3.4.dev.a
libopenblas_power8p-r0.3.4.dev.so
```
(no longer requires manually adding POWER8 target by hand)

Checkout from git and running `make` without options produces
```penBLAS build complete. (BLAS CBLAS LAPACK LAPACKE)

  OS               ... Linux             
  Architecture     ... power               
  BINARY           ... 64bit                 
  C compiler       ... GCC  (command line : /usr/bin/gcc)
  Fortran compiler ... GFORTRAN  (command line : /usr/bin/gfortran)
  Library Name     ... libopenblas_power8p-r0.3.4.dev.a (Multi threaded; Max num-threads is 160)
```

martin-frbg:
Great, thanks again.

PR #1875, opened 2018-11-19
Serialize accesses to parallelized level3 functions from multiple cal…

martin-frbg:
…lers

for #1851


brada4:
It is very well placed, still permitting estimated 1-threaded (small) calls to proceed via other parallelism mechanisms.

PR #1877, opened 2018-11-20
Initialize complex _dot result variable consistently at declaration time

brada4:
Fix one bug and wrap similar fragments to avoid confusing static checker that found it (`cppcheck-1.82-lp150.1.1.x86_64`)

@sharkcz should be minor change and for good?

### DIFF
--- a/kernel/arm/zdot.c
+++ b/kernel/arm/zdot.c
@@ -42,15 +42,12 @@ OPENBLAS_COMPLEX_FLOAT CNAME(BLASLONG n, FLOAT *x, BLASLONG inc_x, FLOAT *y, BLA
 	BLASLONG i=0;
 	BLASLONG ix=0,iy=0;
 	FLOAT dot[2];
-	OPENBLAS_COMPLEX_FLOAT result;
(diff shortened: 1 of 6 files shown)


martin-frbg:
Not sure I see the bug here that this shuffling around of initializers is supposed to fix ?

brada4:
You can apply via edit zarch only https://github.com/xianyi/OpenBLAS/pull/1877/commits/ae895f8648f7af07add892565edf847d1b988adf where they go uninitialized past the quick exit as only place initializing them.

martin-frbg:
How so ? The assignment happens at the end of the function, and I do not see any open escape routes.

brada4:
There is missing initalisation in the most likely route after that quick exit i.e one will start adding values to uninitialised memory acquired at declaration time.

brada4:
Take it as a 2 step patch:
1/ add missing initialisation in the further path
2/ figure out it can actually be initialized at declaration
Others are just silencing checker, sort of does not matter much if it s one or two assignments per call initialising result.

martin-frbg:
Sorry, I still do not see anything missing. Either the quick exit is taken after setting the result to zero, or the calculation uses the dot array and stores the result of the final addition in "result" before exiting.

brada4:
Indeed

Issue #1578, opened 2018-05-27
Problem with OpenBLAS 0.3.0, PGI compilers, and OpenMP


brianborchers:
I've been able to build OpenBLAS 0.3.0 using the PGI C and Fortran compilers with 
 
`make CC=pgcc FC=pgfortran DEBUG=1 DYNAMIC_ARCH=1 USE_OPENMP=1
`

The code compiles and it appears to pass its tests.  

I then compiled and linked a simple test with the library using

```
pgcc -c -m64 -mp testdpotrf.c
pgfortran -Mnomain -m64 -mp testdpotrf.o -L/home/brian/openblas/pgiopenblas/OpenBLAS-0.3.0 -lopenblas -o testdpotrfpgi

```

If I set OMP_NUM_THREADS to 1, the test runs and returns correct results.  However, If I set OMP_NUM_THREADS to 2 or higher,  the result is incorrect.  See below:

>setenv OMP_NUM_THREADS 1
>./testdpotrfpgi 10000
> 10000   8.89    37.5  7.04e-13
>setenv OMP_NUM_THREADS 2
>./testdpotrfpgi 10000
> dpotrf info=33
> 10000   0.00    2164568.2  9.80e-01

Here, the output indicates that for a 10,000 by 10,000 test matrix, it took 8.89 seconds to compute the Cholesky factor, for a speed of 37.5 GFLOPS, and the relative error in a solution to the test system of equations was 7.04e-13.  With OMP_NUM_THREADS set to 2, the DPOTRF returns an error indicating that the matrix isn't positive definite (we know that the matrix is PD though.)  
 
In comparison, using OpenBLAS 0.3.0 built with the same options except for CC=gcc and FC=gfortran, the test returns the expected results.  

Am I doing something obviously wrong here, or is this a bug?  
[testdpotrf.zip](https://github.com/xianyi/OpenBLAS/files/2042855/testdpotrf.zip)

brianborchers:
Oops- it turns out that I was getting my gcc built copy of libopenblas.so linked with the pgcc compiled test code.  After fixing the linkage, this works correctly now.

PR #1878, opened 2018-11-21
Correct link flags for PGI compiler.

kiwifb:
I have recently been building openblas with the PGI fortran compiler (18.4) and bumped into a couple of issues. The code from `f_check` for the PGI compiler is dating from the initial import from GOTOblas according from git blame, but even so I am still wondering what the original motivation was.

* f_check currently replace `lib` by `libso` for the path of libraries, specifically for the PGI compiler without explanations. This is also wrong with a recent version of the compiler. Those path are extracted from a verbose call to the compiler and there is no reasons for them to have ever been wrong (that I can fathom).

* Still from the initial import from GOTOblas some libraries are arbitrarily removed from the list of libraries. However in recent version of pgfortran `-lnuma` is needed for openmp support so it shouldn't be filtered.

### DIFF
--- a/f_check
+++ b/f_check
@@ -292,9 +292,6 @@ if ($link ne "") {
 	    && ($flags !~ /^-LIST:/)
 	    && ($flags !~ /^-LANG:/)
 	    ) {
-	    if ($vendor eq "PGI") {
-		$flags =~ s/lib$/libso/;
-	    }
 	    $linker_L .= $flags . " ";
 	}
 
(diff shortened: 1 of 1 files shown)


martin-frbg:
Possibly this affects only very recent versions of PGI, at least it did not come up in e.g. the relatively recent #1578 (No idea what the (intended) consequences of using libso for the library path are or were.)

kiwifb:
It is possible to avoid those effects by setting your environment with the right variables. Certainly on a regular HPC system I would set it up in a way where the effects of `lib` -> `libso` would go unnoticed. Also because I was using gcc+pgfortran (because of some peculiarities of what I was doing) I needed `-lnuma` in the flags. If I had used pgcc for linking it is quite likely that it wouldn't have needed it.

martin-frbg:
I see. Can we be sure that the  non-filtered -lnuma does not cause problems on other operating systems, or with older versions of PGI (with or without using OpenMP ?

kiwifb:
The only possible conflict would be between a system library and a compiler library of the same soname (or equivalent). I just installed PGI 18.10 and it didn't install its own libnuma, it relied on the system provided one from the package numactl - actually that may be because this time I decided not to install PGI's own build of openmpi.
In any case it appears that it is compatible with something that would be provided by the system and not something PGI specific that accidentally has the same name. So I think it is safe.

The rules in `f_check` about PGI and libnuma are more than 8 years old and it is difficult to find a version old enough to figure out why they were put in place.

brada4:
I think in all practicality most PGI users are with fairly recent version, and very older ones are being slowly surpassed by free options.

Issue #1883, opened 2018-11-23
BLAS isn't multi-core


MiXaiLL76:
When trying to work with openblas and dlib, multi-core processing does not work.
Previously (on another system, everything worked)

https://github.com/davisking/dlib/issues/1004#issuecomment-350565532
As I understand it, dlib uses multi-core openblas for multithreading.

OS: Ubuntu 18.04
Arch: arm64
Board: Orange pi 
OpenBLAS version: git clone 

Openblass build:
make CC=aarch64-linux-gnu-gcc FC=aarch64-linux-gnu-gfortran HOSTCC=gcc TARGET=ARMV8 -j8
Config:
```
openblas_get_num_threads: 4
openblas_get_num_procs: 4
openblas_get_parallel: 1
openblas_get_config: NO_AFFINITY ARMV8 MAX_THREADS=4
openblas_get_corename: ARMV8
```

Dlib build:
cmake -DCMAKE_C_FLAGS="-O3 -fprofile-use " -DDLIB_USE_CUDA=NO -DCMAKE_TOOLCHAIN_FILE=/mnt/c/Users/StepanOFF/Desktop/face_detect/min_core_aarch/aarch64.cmake –build –config Release ..

Both compilations have no errors and all libraries are present.

The picture shows that at startup only 1 core works.
![in work](https://user-images.githubusercontent.com/31501945/48930209-11a71080-ef00-11e8-81fe-f6bee4107c44.png)

No processing
![no in work](https://user-images.githubusercontent.com/31501945/48930233-3602ed00-ef00-11e8-86e0-aca1f08f5ea2.png)

martin-frbg:
What is your typical problem size ? OpenBLAS will (should) not use multiple threads for a very small matrix where the administrative overhead would exceed any possible gain from working in parallel.

MiXaiLL76:
> What is your typical problem size ? OpenBLAS will (should) not use multiple threads for a very small matrix where the administrative overhead would exceed any possible gain from working in parallel.

The execution time of the script in 1 thread takes about 5-6 seconds, when multithreading is used, the execution time is 0.8 - 1 second.

It may be a problem in the wrong configuration of dlib, but in theory it should not be so.

brada4:
Reference BLAS is not multicore, that is for certain.

We will need a firm test case or at least what comes out form "perf record ./detect ; perf report"

Another thing - I suspect it is threads there in the pictures?
[...]

PR #1884, opened 2018-11-23
Re-calibrate _AXPY functions

brada4:
Sadly did not stand a chance to to view this in scope of #1883 
* Sweet spot to switch on 2nd thread would be between 30-40kB input and adds 20-30% performance after (with 20x regression when input measures 2kB, below existing threshold), current threshold is close to optimal and keeps pessimal range out for good.
* adding 3rd and 4th core to work gives lower than 1-core performance up until about 20-25MB there, reaching 2-core speed, but using all CPU cores fully, as seen from system side.
* (presumably memory controller saturates, if we have two they have separate memories attached, unlikely to get set up/scheduled gainfully for this case)
* speculatively accommodate X86, as even preserved, real X86 would not have much cores, so the X86 coDe is more likely to use real X86_64.
* we just let system park/reuse other cores, so we get smaller power budget & hopefully 1usec faster result.

### DIFF
--- a/interface/axpy.c
+++ b/interface/axpy.c
@@ -46,6 +46,10 @@
 #else
 #define MULTI_THREAD_MINIMAL  10000
 #endif
+#if defined(ARCH_X86) || defined(ARCH_X86_64)
+#define MULTI_THREAD_MAX_NUM 2
+#endif
+
 #ifndef CBLAS
 
 void NAME(blasint *N, FLOAT *ALPHA, FLOAT *x, blasint *INCX, FLOAT *y, blasint *INCY){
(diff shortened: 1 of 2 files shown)


martin-frbg:
Not sure I understand your comment about "20x regression", but wouldn't it make more sense to make this a MULTI_THREAD_MAX_TWO threshold at "n <= (n corresponding to 20..25 MB value)"  and leave nthreads unlimited (except by available hardware) when going beyond ? That is assuming there is a measurable speed gain from the nthreads=2 option here.

brada4:
It is about +20/30% over single thread. I will make some picture better....

brada4:
Something went wrong, i will try to make it right

Issue #1886, opened 2018-11-26
Functions  are not guarded against early threading


brada4:
Not exhaustive list, just those used in practice, sort of measurement takes some time

Discovered in #1882 - `dgetrf` `zherk` `dsyrk`
From #1883 - `axpy`
From #1897 (mxnet) `cblas_?syrk` `_potri_`
From https://github.com/JuliaRobotics/RigidBodyDynamics.jl/issues/500 `trsm`

brada4:
Probably worth making into task list, sorting by absent guard and suspected wrong guard.

martin-frbg:
Based on preliminary benchmarks it may make sense to double the multithreading thresholds in interface/trsm.c for the two COMPLEX cases and raise it to four or five times the current limit (of 2*GEMM_MULTITHREAD_THRESHOLD) for float and double.

brada4:
https://github.com/xianyi/OpenBLAS/issues/1989#issuecomment-460371387 has test case vs `dsymv` via lapack, there is no performance regression, but no thread threshold either.

martin-frbg:
#1115 is relevant for dsyrk (and suggests that level3 unrolling is an important factor besides the multithreading thresholds or lack thereof in the interface)

drhpc:
Any progress on this? I am trying to get a grip on https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/1  to decide if it is worth it packaging Octave with sparse matrix computations via SuiteSparse and OpenBLAS in pkgsrc. It's quite a bummer to have the prominent notice not to use OpenBLAS for SuiteSparse.

For now I do not even know how real the threat of abysmal performance is (factor 1000?!). Uncertainty, doubt. It's an unfortunate situation that doesn't seem to see progress.

martin-frbg:
I'm just now experimenting with a few things (#3252 that is a bit broken on arm64 right now)

brada4:
@drhpc  the initial idea is fairly simple to draw the line in the sand where computing switches from one to all cpus, it might be slightlt dated with more and more cores per CPU appear.

cholmod manages threads via (G|I)OMP, and calls these:
```
    liblapack.so.3 => zpotrf_,dpotrf_
    libopenblas_pthreads.so.0 => dtrsm_,ztrsm_,dgemm_,dtrsv_,ztrsv_,zgemv_,zherk_,dgemv_,dsyrk_,zgemm_
```

drhpc:
[...]

PR #1887, opened 2018-11-26
Recalibrate _AXPY

brada4:
#1886 #1883

### DIFF
--- a/interface/axpy.c
+++ b/interface/axpy.c
@@ -44,7 +44,11 @@
 #if  defined(Z13)
 #define MULTI_THREAD_MINIMAL  200000
 #else
-#define MULTI_THREAD_MINIMAL  10000
+#ifdef DOUBLE
+#define MULTI_THREAD_MINIMAL  262144
+#else
+#define MULTI_THREAD_MINIMAL  524288
+#endif
 #endif
 #ifndef CBLAS
 
--- a/interface/zaxpy.c
+++ b/interface/zaxpy.c
@@ -44,7 +44,7 @@
 #if  defined(Z13)
 #define MULTI_THREAD_MINIMAL  200000
 #else
-#define MULTI_THREAD_MINIMAL  10000
+#define MULTI_THREAD_MINIMAL  131072
 #endif
 #ifndef CBLAS
 


brada4:
Justification and input.
I basically rigged loop from +=step to <<1
It is approximate, by magnitude bigger to bypass low area for 3rd thread (it just looks more impressive with 4, with 3 it is still < 1/2 of first thread)

[_axpy.zip](https://github.com/xianyi/OpenBLAS/files/2616588/_axpy.zip)

martin-frbg:
I have not yet looked at this in detail, but I noticed that (a) the proposed default for everything including "consumer" hardware exceeds the current limit for an IBM mainframe (b) your thresholds look as if you are counting bytes rather than vector elements - obviously the two are related, but I find it surprising that a vector of 512k floats would be processed faster in a single thread.

brada4:
N is in elements for axpy benchmark threshold is 2MB (cache I have) for S D Z and 1MB for C

The gain from threading at all is not so big, the damage from all 4 threads is quite big.  sort of 20% slowdown because of no threading is nothing in place where it becomes eventual 10x problem.

Z13 wikipedia says has 2MB per cpu cache too, so pretty close to my laptop CPU, maybe it can be/should be unified acknowledging memory effect?

512k*4B perfectly fits in cache (just once, not 3x)

brada4:
Remove all threading gueards in interface/axpy.c
Rig benchmark increment loop into left shift in benchmark/axpy.c
Measurement says better safe (1thr > 2.1MB) than sorry (4thr <1MB)
```
#!/bin/sh
for a in s d c z ; do
        for b in 1 2 4 ; do
                OPENBLAS_LOOPS=97 OPENBLAS_NUM_THREADS=${b} ./${a}axpy.goto 1 35000000 1 2>&1 | tee ${a}axpy${b}.out
        done
done
```

brada4:
Something weird, atom hyperthreads improve speed, when they should get worse in practice.
Could it be l1_thread work splitting should align base for subsequent threads? I will try around idea, do not merge this.

brada4:
I could not invent much better:
 4bytes goes 16 times in maximum practical of today Dcache line, so I round up elements to divisible by 16 , which yields min per-cpu chunk of 64 bytes for S 128 bytes for D C and 256 for Z (and something not so round for X)
It is better than nothing at first sight, presumes arguments aligned. In future can be used to fix alignment and use exact cache lines (at present those are not known at the point, so not so easy)
I will spend some time doing properly.

brada4:
Lets hold back for upcoming release.

PR #1890, opened 2018-11-28
Include version number in openblas_get_config output

### LINKED ISSUE #1889: ENH: add an extension to get the library version
I could not find an interface to query the version number (i.e. 0.3.3, 0.3.0) of OpenBLAS.

In the process I found the [wiki page](https://github.com/xianyi/OpenBLAS/wiki/OpenBLAS-Extensions) for the extensions, but there is not one for the version. Even using `strings *.so` I could not find it.

martin-frbg:
fixes #1889


brada4:
What about starting with OPENBLAS like in:
http://www.netlib.org/lapack/explore-html/d8/d8b/_l_a_p_a_c_k__version_8f_source.html

martin-frbg:
No preference for either option, except "OpenBLAS" is one byte more expensive than "VERSION" :smile: and one would not expect openblas_get_config() to return anything other than OpenBLAS info I guess ? (But I have never been good at marketing, so your suggestion is probably valid)

brada4:
API would be If you scan for OpenBLAS it is the sort of "new API" that has version number, otherwise assume past.

brada4:
If we trip the thin border between marketing and industrial espionage - ATLAS just prints a rich banner page with version

PR #1894, opened 2018-11-30
Use correct ARCH name on BSD powerpc64

pkubaj:
FreeBSD uses powerpc64 name for POWER architecture. Use correct Makefile for this platform.

### DIFF
--- a/Makefile.system
+++ b/Makefile.system
@@ -11,7 +11,11 @@ endif
 
 # Catch conflicting usage of ARCH in some BSD environments
 ifeq ($(ARCH), amd64)
-override ARCH=x86_64
+override ARCH=x86_64 
+else ifeq ($(ARCH), powerpc64)
+override ARCH=power
+endif
+
 endif
 
 NETLIB_LAPACK_DIR = $(TOPDIR)/lapack-netlib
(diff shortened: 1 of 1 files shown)


brada4:
Please `#ifdef` freebsd the specific section.

martin-frbg:
Is this sufficient for building ? ARCH is used in several places (see definition of KERNELDIR a few lines above your change), so it will be better to remap its definition (either `ifeq ($(ARCH),powerpc64) override ARCH=power` or better yet find where this originates.
(And no need to `ifdef` this for freebsd... )

ararslan:
What Martin has suggested is actually already being done to catch amd64 in place of x86-64. So an appropriate approach for this may be simply
```diff
--- a/Makefile.system
+++ b/Makefile.system
@@ -12,6 +12,8 @@ endif
 # Catch conflicting usage of ARCH in some BSD environments
 ifeq ($(ARCH), amd64)
 override ARCH=x86_64
+else ifeq ($(ARCH), powerpc64)
+override ARCH=power
 endif
 
 NETLIB_LAPACK_DIR = $(TOPDIR)/lapack-netlib
```

brada4:
And also aarch64 into arm64

brada4:
This particular patch breaks even FreeBSD x86/i386 port , would be nice to close it without action, and next time test them against own ports CI system before submitting upstream.

pkubaj:
This patch is ok, can you merge it?

PR #1907, opened 2018-12-06
Attempt to maintain cache line alignment in BLAS L1 work splitter.

brada4:
Ugly one-liner to take out easily because of being incomplete as follows:
It aligns to 16 elements which makes 64bytes for S and multiples thereof for others. Still devlivers promise in subject.

... measuring axpy 3rd thread (unconstrained at interface) makes result always few percent better than 2 first only, no matter arg size or thread oddness, they were before ups and downs.
... Threading threshold can be much lower than 2MB estimated before for axpy etc (more PRs to follow), there is small improvement in all odd-sized cases even with current thresholds.

### DIFF
--- a/driver/others/blas_l1_thread.c
+++ b/driver/others/blas_l1_thread.c
@@ -65,7 +65,7 @@ int blas_level1_thread(int mode, BLASLONG m, BLASLONG n, BLASLONG k, void *alpha
     /* Adjust Parameters */
     width  = blas_quickdivide(i + nthreads - num_cpu - 1,
 			      nthreads - num_cpu);
-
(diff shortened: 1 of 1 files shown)


brada4:
@martin-frbg I think this approach is a bit too simplistic
 c/zaxpy double throughput  from multiple threads with this patch already at 64 elements, but no change for s/d. I think there is a need to pass extra parameter down from interface with alignment preference. If the effect is cache-line based we can calculate kilobytes before parallel call too.
I will make more bigger one with adjusting all callers with extra zero parameter, i.e. no code flow change sort of initially, then polish what can be polished.

PR #1931, opened 2018-12-23
Add -mavx2 to TARGET=HASWELL builds

martin-frbg:
to leverage GEMM improvements from PR#1921


brada4:
CentOS 6 gcc 4.4.7 will support the flag, though system binutils `as` will barf out early unless tricked with software collections via FAQ.
Ubuntu 14 would have no problem.

martin-frbg:
I do not think it is worthwhile to add a maze of conditionals to support any distro-specific backports of gcc 4.7 features especially if it would mean checking separately that binutils were upgraded to match. The new gemm_beta code will fall back to the earlier implementation in the absence of -mavx2.

brada4:
I will rewrite FAQ piece , lonely beta is not really worth more ifdefs.

PR #1939, opened 2018-12-27
Fix typo in UNKNOWN core name

TiborGY:
Should be of no consequence, right?


martin-frbg:
Historic typo, but you'd need to fix the cpuid_ codes for the other platforms and the main Makefile as well. (And methinks there is little if any point in fixing this, the define might as well be ABCDEF for what it does.)

TiborGY:
Might as well finish it. Should be ready to go.

Issue #1867, opened 2018-11-12
ARM Travis tests running on Emulator == SLOW


rengolin:
Just opening this ticket to keep a reminder, I'll look into getting a cloud instance to enable native tests on Arm hardware. I don't have access to ARMv6 or ARMv7 hardware but I can provide ARMv8 tests, which can emulate ARMv7 via AArch32 natively on the machines that support AArch32.

With the last meaningful ARMv6 machine being RaspberryPi 1 and with that all but long gone, I think we should be fine just testing v7 and v8.

What are the requirements for the client? Do we need to install a slave in the machine, like Jenkins, or is that done via SSH keys access?

martin-frbg:
No idea actually, the current CI tests use whatever API is provided by github. I should be able to run ARMv7 (Asus Tinkerboard) and ARMv8 (Sony Xperia) tests locally when actually required so the CI addition from #1864 is mostly for convenience. One problem with integration could be that I do not have administrative control of the project and it is unclear if/when xianyi is available again.

rengolin:
I see, that's not optimal.

I could use my cloud instance to keep a local loop and somehow exposing the results (webpage? webservice?) until we get travis to fire at it. Let me see what I can do on my side, and I'll ping you when I have something running, we can decide later.

Worst case scenario, we can try a CI on our personal repos, which we do have control. But would be good to have someone active in the community (like you) with admin access. :)

martin-frbg:
Apparently there is a company (and service) named shippable that provides free CI on ARMV8a hardware http://blog.shippable.com/shippable-arm-packet-deliver-native-ci-cd-for-arm-architecture

rengolin:
Interesting, would that work with Travis without need for admin access on the github project?

martin-frbg:
As far as I can tell it is a Travis-like service that would need the same admin access to set up, I just came across it in another context and wanted to put that information somewhere.

rengolin:
Got an instance running, @martin-frbg send me your public ssh keys so I can give you access to it. My email is name.suname@linaro.org.

rengolin:
Given we don't have admin rights in the project, I'm thinking in using [GitHub's API](https://developer.github.com/v3/) to track commits and pull requests, build a branch or a PR and then make the result available somehow.

[...]

PR #1941, opened 2018-12-29
Validate user supplied TARGET

TiborGY:
This has a tiny side effect, the error gets triggered even during `make clean`, but otherwise it works as intended.

### DIFF
--- a/Makefile.system
+++ b/Makefile.system
@@ -65,6 +65,7 @@ endif
 
 ifdef TARGET
 GETARCH_FLAGS := -DFORCE_$(TARGET)
+GETARCH_FLAGS += -DUSER_TARGET
 endif
 
 # Force fallbacks for 32bit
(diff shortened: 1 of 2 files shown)


TiborGY:
About the Travis failures:
-The ARM builds have timed out, no idea why, maybe the Travis servers are too busy
-There is one legitimate failure. I do not really understand how the Travis builds work, but the build parameters seem to be conflicting. First there is 

> export BTYPE="BINARY=64 NO_AFFINITY=1 USE_OPENMP=0 NO_LAPACK=0 TARGET=core2"

After that,

> $ COMMON_FLAGS="DYNAMIC_ARCH=1 TARGET=NEHALEM NUM_THREADS=32"

Is this supposed to to be a CORE2 or a NEHALEM build? Also, why set DYNAMIC_ARCH=1, if you want a specific target?

martin-frbg:
ARM tests are always timing out as the CI job runs on virtual hardware (qemu) - the idea is to use them for now to catch any glaring compile bugs, and set up automatic integration with a dedicated system later. (#1867).
The failing build is a good demonstration, setting TARGET in addition to DYNAMIC_ARCH can be useful when the build host is not autodetected, or to ensure that the common parts of the code (BLAS driver routines and the entire LAPACK) do not rely on "modern" cpu features present in the build system.

Issue #1947, opened 2019-01-05
Possible incorrect AVX512 detection


yuyichao:
Ref https://github.com/JuliaLang/julia/issues/29652#issuecomment-430202222, see [my version of the code in julia](https://github.com/JuliaLang/julia/blob/5290bba46d79b908fb1a4c0aadc94215df5a9905/src/processor_x86.cpp#L548) (which is mostly copied and updated from LLVM).

Disclaimer: I do not have the hardware to test this so I cannot tell if this is already fixed or not (by looking at the code I assume not, unless it's handled in a very different way than avx support). That's why I wished that the two people that have actually experienced this issue to report them instead. I'm reporting it now since I just noticed that none of them seem to have done that already. Feel free to close if I'm just missing the issue in my search somehow. I won't be able to help testing any patch but I'm happy to explain the code that does the testing in julia (I feel like it's fairly simple and very similar to the avx one though....).

yuyichao:
Oh, just realize that I didn't put a description of the issue here so just to summarize the issue so that ppl don't have to read through the julia issue thread.

The OpenBLAS avx512 detection code seems to be purely based on processor part number. While this is probably necessary for figuring out the performance model, it's not enough to guarantee the support of the AVX512 feature. Similar to AVX, both the CPUID feature bit and the zmm state bit in XCR0 needs to be checked to make sure both the processor and the OS supports AVX512.

If I'm reading my code correctly, the bits should be the 0x10000 and 0x20000 bits of EBX for CPUID(EAX=7, ECX=0) and the 0xe0 bit of xcr0.

mk:
I happy to test this on my machine with avx512.

martin-frbg:
If I am reading this correctly, it would affect "only" VM setups where the virtual hardware claims to have the host CPU but does not provide all its features ? At least as far as I know, there are no physical processors around that identify as SkylakeX and do not support AVX512 (?) 
"Historically" neither VMs nor prepackaged OpenBLAS distributed with Julia or similar played much of a role so past reactions to similiar issues (typically over AVX2 I think) seems to have been to tell to "go fix your VM configuration". On the other hand, the runtime detection code in dynamic.c already has an
active check for AVX that probably dates back to GotoBLAS, most likely none of the later contributors (me included) realized the need add the equivalent for later features.

yuyichao:
> active check for AVX that probably dates back to GotoBLAS

[...]

PR #1949, opened 2019-01-05
Query AVX2 and AVX512VL support when selecting x86 kernels

martin-frbg:
For #1947 (and corresponding earlier cases involving Haswell) where VM or OS limitations made otherwise standard features of the correctly detected CPU model unavailable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -153,7 +153,7 @@ matrix:
       before_script:
         - COMMON_FLAGS="DYNAMIC_ARCH=1 TARGET=NEHALEM NUM_THREADS=32"
         - brew update
-        - brew install gcc # for gfortran
+        - travis_wait 30 brew install gcc # for gfortran
       script:
         - travis_wait 45 make QUIET_MAKE=1 $COMMON_FLAGS $BTYPE
       env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -149,11 +149,11 @@ matrix:
 
     - &test-macos
       os: osx
-      osx_image: xcode8
+      osx_image: xcode8.3
       before_script:
         - COMMON_FLAGS="DYNAMIC_ARCH=1 TARGET=NEHALEM NUM_THREADS=32"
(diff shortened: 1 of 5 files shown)


yuyichao:
I believe the test here are testing hardware support. (from CPUID).

OS support bit are in XCR0, same as AVX. See the bit assignment in https://en.wikipedia.org/wiki/Control_register#XCR0. In julia (and I believe LLVM) we conservatively check all bits from 5 to 7 (i.e. `0xe0`). You can reduce that if you don't need all features I assume....

martin-frbg:
Indeed, thanks. Hope I got it right now, as I do not have any AVX512 hardware to test this on.

yuyichao:
Didn't check all the bits but it looks right to me as well....

brianborchers:
> Indeed, thanks. Hope I got it right now, as I do not have any AVX512 hardware to test this on.

I do have a Xeon Skylake system with AVX512 running Linux that I could test on.  However, it's not clear that I could provide a complete test, since the main problem here is with OS's that disallow the use of AVX512 instructions on hardware that is otherwise capable.  
 
On my machine, the DYNAMIC_ARCH build should figure out that it has AVX512 and use those kernels.  How could I test that this is working correctly?

mk:
@brianborchers the original issue occurred for me using docker which seems to disable avx512.

I can test this when I’m back at my iMac next week.

martin-frbg:
If it does not detect a specific capability, the DYNAMIC_ARCH build is supposed to issue a warning about falling back to a less capable cpu. Unfortunately i seem to have missed this in the SkylakeX case, although I preprared the message. Will update this PR in a minute. 
Similarly a quick build test should show that the modified getarch writes both HAVE_AVX2=1 and HAVE_AVX512VL=1 to Makefile.conf.

martin-frbg:
Travis CI seems to be timing out while still installing dependencies on OSX...

PR #1960, opened 2019-01-16
Add support for Hygon Dhyana

cnjsdfcy:


martin-frbg:
Thank you very much. Hopefully OpenBLAS support for Zen architecture will improve a bit in the near future, currently it is treated as Haswell but without the recent tweaks for that cpu.

PR #1962, opened 2019-01-16
Modrenize R benchmarks slightly

brada4:
Disable NaN checks in R>=3.4.0  - we generate matrix() without NaNs, and after that do not care about numeric outcome (1)
Compact vertically - whitespace and let output format exceed 80 chars
Horizontally - simplify formulas and minimize matrix parameter count
 (RNG) runif is about 2x faster than rnorm - i have some dount if (1) is absolutely true though

### DIFF
--- a/benchmark/scripts/R/deig.R
+++ b/benchmark/scripts/R/deig.R
@@ -33,7 +33,7 @@ cat(sprintf("      SIZE             Flops                   Time\n"))
 
 n <- nfrom
 while (n <= nto) {
-  A <- matrix(rnorm(n * n), ncol = n, nrow = n)
+  A <- matrix(rnorm(n * n), nrow = n)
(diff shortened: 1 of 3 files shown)


brada4:
doc is here:
https://stat.ethz.ch/R-manual/R-devel/library/base/html/options.html

martin-frbg:
Not sure I understand the benefit of these changes, is current R throwing warnings for the old code, or do you want to make the benchmarks run faster ? (And I suspect the mflops formula that you shortened may have been written that way to document how the number is calculated)

brada4:
The novelty is checking NaNs in input matrices in 3.4, this makes 3.3 and 3.4 and numpy (!!!) results comparable, though the damage is just few percent from the new check.

PR #1971, opened 2019-01-18
Shift transition to multithreading towards larger matrix sizes

martin-frbg:
See #1886 and JuliaRobotics issue 500. trsm benchmarks on Haswell and Zen showed that with these values performance is roughly doubled for matrix sizes between 8x8 and 14x14, and still 10 to 20 percent better near the new cutoff at 32x32.

### DIFF
--- a/interface/trsm.c
+++ b/interface/trsm.c
@@ -82,9 +82,9 @@
 #endif
 
 #ifndef COMPLEX
-#define SMP_FACTOR 8
+#define SMP_FACTOR 256
 #else
-#define SMP_FACTOR 4
+#define SMP_FACTOR 128
 #endif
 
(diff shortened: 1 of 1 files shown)


martin-frbg:
Interestingly the impact on very thin matrices with only one of M and N below the threshold (e.g. 500x2 or 2x500) appears to be fairly small.

martin-frbg:
Perhaps not surprisingly, this actually has huge drawbacks for matrix dimensions like 500x20... will see if basing the decision on args.m*args.n like it is done in other interface routines works, or perhaps a combination of both approaches.

martin-frbg:
Meh. appveyor is starting to time out on the dynamic_arch build again.

PR #1978, opened 2019-01-21
Better support for MSVC/Windows in CMake (v0.3.x)

danielgindi:
#1975 #1977


martin-frbg:
Seems the new static build option keeps looking for the openblas.def file (and/or the cmake version check is borked, thinking 3.13 <3.4 - "LESS" should probably be "VERSION_LESS" in a few places but then why would it fail only now ?)

martin-frbg:
Now it fails over what appears to be conflicting settings regarding linking against msvcrt in openblas_utest.

danielgindi:
@martin-frbg Actually you see several different problems there, but these are more or less expected:

1. Least important, the warnings. These are due to compiling by default with the dll-based CRT (msvcrt) from Microsoft. The CI script that compiles everything must be compiling most with static CRT, and CMake takes the msvcrt by default for openblas lib. So the script should be modified to pass `MSVC_STATIC_CRT` preprocessor definition to cmake. That is the reason I've added that flag. You get the same shit when working directly with MSVC, or with CMake, or Conan...

2. `Error copying file "C:/projects/OpenBLAS/build/lib/./openblas.dll" to "C:/projects/OpenBLAS/build/utest/.".` That's a different issue, just trying to copy a DLL that's not there. Due to the fact the you were always expecting a DLL before, and now it actually supports static compilation, it compiles only a lib, and no DLL should be copied when the configuration specifies a static compilation.

I do not know a lot about those CI scripts, but these should be very easy adjustments :-)

martin-frbg:
Indeed I misread the log, thanks. (Still not quite sure if it would do the right thing w.r.t openblas.def on cmake <3.4 as I only fixed the version test)

danielgindi:
Man your CI machines are refusing to wake up 😂

martin-frbg:
Both are free services...

danielgindi:
@martin-frbg Seems to be fine, and also to prove that it works as the test env compiles a static library :-)

PR #1981, opened 2019-01-23
Fix include directory of exported targets

edisongustavo:
# Problem

I was having problems compiling a simple program on Windows from the package on conda-forge:

```
fatal error C1083: Cannot open include file: 'cblas.h': No such file or directory
```

The program can be as simple as:

`CMakeLists.txt`

```cmake
cmake_minimum_required(VERSION 3.0)
project(Guga CXX)

find_package(OpenBLAS CONFIG REQUIRED)

add_executable(foo main.cpp)
target_link_libraries(foo OpenBLAS::OpenBLAS)
```

`main.cpp`

```c++
#include <cblas.h>

int main() {
    return 0;
}
```

So I tried on linux building from OpenBLAS source tree, giving the same problem.

## Cause

The exported target is setting the `INTERFACE_INCLUDE_DIRECTORIES` as `${_IMPORT_PREFIX}/include`.

This is different from the directory being used at the `install()` targets:

```
set(CMAKE_INSTALL_INCLUDEDIR ${CMAKE_INSTALL_INCLUDEDIR}/openblas${SUFFIX64})
```

link: https://github.com/edisongustavo/OpenBLAS/blob/develop/CMakeLists.txt#L295

# Solution

Use the correct directory on the install target.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,7 +157,7 @@ endif()
 
 # add objects to the openblas lib
 add_library(${OpenBLAS_LIBNAME} ${LA_SOURCES} ${LAPACKE_SOURCES} ${RELA_SOURCES} ${TARGET_OBJS} ${OpenBLAS_DEF_FILE})
-target_include_directories(${OpenBLAS_LIBNAME} INTERFACE $<INSTALL_INTERFACE:include>)
+target_include_directories(${OpenBLAS_LIBNAME} INTERFACE $<INSTALL_INTERFACE:include/openblas${SUFFIX64}>)
 
 # Android needs to explicitly link against libm
 if(ANDROID)


brada4:
Yours sort of breaks on Linux against gmake built library.

edisongustavo:
Ok, I will take a look for that. Would you have a log?

martin-frbg:
@brada4 not sure I got your point, but I suspect any inconsistency in naming will have had its roots in #1764 and/or my later #1792

edisongustavo:
@brada4 I'm having trouble in understanding what do you mean. Would you have something to guide me so I can investigate it more?

@martin-frbg You are correct that the inconsistency was created here: #1764 
As you can see, this change is what made all include headers be installed into a different directory:

```cmake
set(CMAKE_INSTALL_INCLUDEDIR ${CMAKE_INSTALL_INCLUDEDIR}/openblas${SUFFIX64})
```

This behavior is inconsistent with the `make` behavior, but maybe this is fine and expected. However, the installed `OpenBLASConfig*.cmake` files must have the correct directory set, which this PR fixes.

edisongustavo:
Hello, is there anything else I can do to get this merged?

PR #1988, opened 2019-01-27
Reword/expand comments in Makefile.rule

TiborGY:
Lots of small changes in the wording of the comments, plus an expansion of the NUM_THREADS and NO_AFFINITY sections.

### DIFF
--- a/Makefile.rule
+++ b/Makefile.rule
@@ -72,10 +72,16 @@ VERSION = 0.3.6.dev
 
 # You can define the maximum number of threads. Basically it should be less
 # than or equal to the number of CPU threads. If you don't specify one, it's
-# automatically detected by the the script.
+# automatically detected by the the build system.
 # If SMT (aka. HT) is enabled on the system, it may or may not be beneficial to 
 # restrict NUM_THREADS to the number of physical cores. By default, the automatic 
 # detection includes logical CPUs, thus allowing the use of SMT.
+# Users may opt at runtime to use less than NUM_THREADS threads.
+#
+# Note for package maintainers: you can build OpenBLAS with a large NUM_THREADS
+# value (eg. 32-256) if you expect your users to use that many threads. Due to the way
+# some internal structures are allocated, using a large NUM_THREADS value has a RAM
+# footprint penalty, even if users reduce the actual number of threads at runtime.
 # NUM_THREADS = 24
 
 # If you have enabled USE_OPENMP and your application would call
@@ -138,6 +144,7 @@ NO_WARMUP = 1
 # to the same core(s) as OpenBLAS, possibly binding all threads to a single core.
 # For this reason, affinity handling is disabled by default. Can be safely enabled if nothing
 # else modifies affinity settings.
+# Note: enabling affinity has been known to cause problems with NumPy and R
 NO_AFFINITY = 1
 
 # If you are compiling for Linux and you have more than 16 numa nodes or more than 256 cpus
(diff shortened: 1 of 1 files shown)


martin-frbg:
I think the part about not wanting to _install_ the static library is technically correct, as the gmake  build proceeds via libopenblas.a in any case.
(And if you want to expand the NO_AFFINITY section it might make sense to spell out what the typical incompatibility is - any subsequent threads of the "other" program end up locked to the same core(s) as OpenBLAS )

brada4 (changes requested):
Your num_threads description states all redistributed packages configured with 32/64/128 threads are wrong.
Why generic fortran option os atributed to ifort? 
Affinity is off by default and only affines threads to cores, nothing with memory. One certain application is numpy.

TiborGY:
> Your num_threads description states all redistributed packages configured with 32/64/128 threads are wrong.

Well, the old description was even worse, suggesting less threads than the number of cores. But I can add a note for package maintainers.
> Why generic fortran option os atributed to ifort?

GCC does not recognize "-i8" at all, the option that turns on 64-bit integers is "-fdefault-integer-8".
> Affinity is off by default and only affines threads to cores, nothing with memory. One certain application is numpy.

OK, I will fix this, I was following the old description, and that stated "CPU/memory affinity". I assumed the old description was not wrong.

martin-frbg:
I suspect the original wording as "Basically should be less than actual number of cores"  was meant in the sense of "you will want to define this only if you want less than what the autodetection would find anyway".
(This and most other comments are unchanged from GotoBLAS by Kazushige Goto. Apart from the slight language issues he probably did not foresee people building his library for redistribution rather than direct compilation on the target system. So allowing more threads than cpus _seen on the build host_ was probably not a plausible scenario back then.)

I'm perfectly OK with labeling the "-i8" option in the comment as ifort dialect for clarity.

Lastly calling it "CPU/memory" affinity probably makes sense  where memory access is not quite uniform e.g. between cores on separate dies. (Again the wording is Goto's, and I think he would have known)

brada4:
Probably worth saying that NUM_THREADS is maximum and can be reduced at runtime if needed (with some examples)
And the proposed edit said that affinity is on by default. It is actually off to play safe with others.

TiborGY:
> And the proposed edit said that affinity is on by default. It is actually off to play safe with others.

"For this reason, affinity handling is disabled by default." I cannot see how I could have made it clearer that it is **disabled by default**

TiborGY:
I think if there are no further objections, this is ready to go.

brada4:
The thing is num_threads sizes memory buffers, and larger than CPUs value is desirable if using threading outside OpenBLAS. Accurate reference to OPENBLAS_NUM_THREADS would help there.

TiborGY:
> 
> 
> The thing is num_threads sizes memory buffers, and larger than CPUs value is desirable if using threading outside OpenBLAS.

AFAIK the current codebase always creates enough memory buffers for at least 50 threads.
> 
> 
> Accurate reference to OPENBLAS_NUM_THREADS would help there.

I refrained from mentioning that environment variable on purpose. There are many ways to set the actual number of used threads at runtime, and that is already documented elsewhere. I feel like mentioning only one of the possible ways could be misleading.

brada4:
NUM_THREADS - Better to make a shortcut, current documentation could wish better structuring.

martin-frbg:
Actually I think the new wording is quite adequate as it is now (minor nit: you retained a "the the" from the previous version in line 75). We may eventually want to move the lengthier descriptions to some new document named "OPTIONS" or the like, or rework the original `GotoBLAS_*` files.

PR #1995, opened 2019-02-03
Disable _TRMV parallelism at file top

brada4:
Restore old SMP code path
And use old UP path via single define

### DIFF
--- a/interface/trmv.c
+++ b/interface/trmv.c
@@ -43,6 +43,9 @@
 #include "functable.h"
 #endif
 
+/*  trmv_thread was found to be broken, see issue 1332 */
+#undef SMP
+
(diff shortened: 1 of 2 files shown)


martin-frbg:
This removes a FIXME without fixing anything except cosmetics ?

brada4:
Sorry, lost in clipboard. Only intent is to avoid unnecessary thread calculation in complex case while FIXME hangs over it.

martin-frbg:
Actually on revisiting that FIXME I am now certain that the proper fix is the same as #1389 - and it is totally unclear to me why I came to a different conclusion in #1332. So this is going to be superseded by #2026...

brada4:
Perfect.

PR #1996, opened 2019-02-04
NBMAX=4096 for gemvn, added sgemvn 8x8 for future

quickwritereader:


martin-frbg:
Given the confusion about the development status of the existing but unused kernels left behind by earlier developers, perhaps it would make sense add a small comment at the top of the unused 8x8 kernel ? Just a short note that it is known to work but does not perform as well as expected on what is now current hardware ?

quickwritereader:
@martin-frbg I added note.

martin-frbg:
Thank you, merging now as the CI does not check POWER anyway

PR #2008, opened 2019-02-07
Fix potential memory leak in cpu enumeration on Linux

### LINKED ISSUE #2003: memory leak in get_num_procs()
Seen in version 0.3.5+ds-1 as packaged by the Debian project for Debian Sid, on an AMD Threadripper 2950X system.

Some of my test programs are tripping GCC's memory leak detection built into their implementation of asan.  It looks like the error handling paths coming out of get_num_procs() are getting triggered, and they leak memory.

Hand-walking through get_num_procs():

 - sysconf(_SC_NPROCESSORS_CONF) returns 32
 - CPU_ALLOC_SIZE(32) returns 8.  However, sizeof(cpu_set_t) is 128.
 - Calling sched_getaffinity(0, CPU_ALLOC_SIZE(32), <a previously allocated cpu_set_t *>) returns -1: Illegal argument.  However, sched_getaffinity(0, sizeof(cpu_set_t), <valid memory>) succeeds.

Its not clear to me just by reading the manpages whether or not it is glibc in error or openblas.  It certainly seems reasonable to pass CPU_ALLOC_SIZE(nprocs) instead of sizeof(cpu_set_t) as the second parameter.

That said, once ret is nonzero, get_num_procs() returns without CPU_FREE()'ing the previously CPU_ALLOC()'ed `cpusetp`, and memory is leaked.

martin-frbg:
An early return after a failed call to sched_getaffinity would leak the previously allocated cpu_set_t. Wrong calculation of the size argument in that call increased the likelyhood of that failure. Fixes #2003

### DIFF
--- a/driver/others/memory.c
+++ b/driver/others/memory.c
@@ -201,10 +201,13 @@ int get_num_procs(void) {
   cpu_set_t cpuset,*cpusetp;
   size_t size;
   int ret;
+
+#if defined(__GLIBC_PREREQ)
 #if !__GLIBC_PREREQ(2, 7)
   int i;
 #if !__GLIBC_PREREQ(2, 6)
   int n;
+#endif
 #endif
 #endif
 
@@ -1725,10 +1728,13 @@ int get_num_procs(void) {
   cpu_set_t cpuset,*cpusetp;
   size_t size;
   int ret;
+
+#if defined(__GLIBC_PREREQ)
 #if !__GLIBC_PREREQ(2, 7)
   int i;
 #if !__GLIBC_PREREQ(2, 6)
   int n;
+#endif
 #endif
 #endif
 
--- a/driver/others/memory.c
+++ b/driver/others/memory.c
@@ -216,7 +216,7 @@ int i,n;
  #endif
 
  #if !__GLIBC_PREREQ(2, 7)
-  ret = sched_getaffinity(0,sizeof(cpu_set_t), cpusetp);
+  ret = sched_getaffinity(0,sizeof(cpusetp), cpusetp);
   if (ret!=0) return nums;
   n=0;
   #if !__GLIBC_PREREQ(2, 6)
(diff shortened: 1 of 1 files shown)


jbrandmeyer (changes requested):
Beware that cpusetp has a pointer type.  I think you really want sizeof(*cpusetp)

martin-frbg:
> Beware that cpusetp has a pointer type. I think you really want sizeof(*cpusetp)

I must admit that I am a bit confused about this. Size of the type evaluates to 512 on a mere dualcore laptop and using that it bombs out with a memory management error in an unrelated free() later.

jbrandmeyer:
At this point, I'm confused, too.  I spent some time reading through the glibc headers and sources.

The cpuset macros are clearly operating by allocating a number of unsigned long.  CPU_ALLOC(n) does not necessarily return a whole cpu_set_t, despite the return type being a `cpu_set_t *`.

sched_getaffinity() is just calling into the kernel.  From sysdeps/unix/sysv/linux/sched_getaffinity.c, it looks like the kernel is returning the number of bytes written into the cpuset output argument.  Then, the syscall wrapper clears any additional memory, according to the size passed in by the caller (ie, openblas).

That suggests that the right argument to pass to sched_getaffinity is the result of calling CPU_ALLOC_SIZE(), and not sizeof(cpu_set_t).  It just so happens that CPU_ALLOC_SIZE(n) for n <= 64 is also sizeof(intptr_t), so everything appears to work on commodity hardware.

My suggestion is to just add the CPU_FREE in the error-handling path for the moment.  I suspect that its actually the kernel that's buggy in my case when the size is 8.  I'm going source-diving to see if I can figure it out in their end.

jbrandmeyer:
FWIW, tools/testing/selftests/rseq/param_test.c from the kernel sources always allocates a whole cpu_set_t on the stack, rather than use dynamic allocation.

martin-frbg:
The actual error appears to be in my declaration of cpusetp as a cpu_set_t`*` which is appropriate for dynamic allocation but pushed an uninitalized struct onto sched_getaffinity otherwise. Not sure yet why CPU_ALLOC returned a wrong value for you, but perhaps this should not be called when the number of processors is well below 1024 (although some documentation suggests otherwise)

martin-frbg:
@jbrandmeyer could you try the lastest iteration please, just to see if sched_getaffinity is still failing with the cpu set returned by CPU_ALLOC ?

martin-frbg:
Committing after testing on Ryzen2-2700 with and without taskset.

Issue #2009, opened 2019-02-12
segfault in dblat2 in dgemv_HASWELL in OpenBLAS 0.3.5


susilehtola:
Rebuilding OpenBLAS 0.3.5 on Fedora rawhide (to-be Fedora 30) results in a segfault in the tests

```
OPENBLAS_NUM_THREADS=1 OMP_NUM_THREADS=1 ./dblat2 < ./dblat2.dat
BUILDSTDERR: Program received signal SIGSEGV: Segmentation fault - invalid memory reference.
BUILDSTDERR: Backtrace for this error:
BUILDSTDERR: #0  0x7f21b3836ca1 in ???
BUILDSTDERR: #1  0x7f21b3835e65 in ???
BUILDSTDERR: #2  0x7f21b368717f in ???
BUILDSTDERR: #3  0x5578496125f2 in dgemv_kernel_4x1
BUILDSTDERR: 	at ../kernel/x86_64/dgemv_n_4.c:143
BUILDSTDERR: #4  0x557849612958 in dgemv_n_HASWELL
BUILDSTDERR: 	at ../kernel/x86_64/dgemv_n_4.c:318
BUILDSTDERR: #5  0x5578487bc333 in dgemv_
BUILDSTDERR: 	at /builddir/build/BUILD/openblas-0.3.5/Rblas/interface/gemv.c:231
BUILDSTDERR: #6  0x5578487b6e8e in dchk1_
BUILDSTDERR: 	at /builddir/build/BUILD/openblas-0.3.5/Rblas/test/dblat2.f:582
BUILDSTDERR: #7  0x5578487bbb02 in dblat2
BUILDSTDERR: 	at /builddir/build/BUILD/openblas-0.3.5/Rblas/test/dblat2.f:305
BUILDSTDERR: #8  0x5578487ad052 in main
BUILDSTDERR: 	at /builddir/build/BUILD/openblas-0.3.5/Rblas/test/dblat2.f:390
BUILDSTDERR: /bin/sh: line 1: 20348 Segmentation fault      (core dumped) OPENBLAS_NUM_THREADS=1 OMP_NUM_THREADS=1 ./dblat2 < ./dblat2.dat
BUILDSTDERR: make[1]: *** [Makefile:32: level2] Error 139
BUILDSTDERR: make[1]: *** Waiting for unfinished jobs....
make[1]: Leaving directory '/builddir/build/BUILD/openblas-0.3.5/Rblas/test'
[...]

PR #2018, opened 2019-02-14
dgemv_kernel_4x4(Haswell): add missing clobbers for xmm0,xmm1,xmm2,xmm3

bartoldeman:
This fixes a crash in dblat2 when OpenBLAS is compiled using
-march=znver1 -ftree-vectorize -O2

See also:
https://github.com/easybuilders/easybuild-easyconfigs/issues/7180

### DIFF
--- a/kernel/x86_64/dgemv_n_microk_haswell-4.c
+++ b/kernel/x86_64/dgemv_n_microk_haswell-4.c
@@ -104,6 +104,7 @@ static void dgemv_kernel_4x4( BLASLONG n, FLOAT **ap, FLOAT *x, FLOAT *y, FLOAT 
           "r" (ap[3]),  // 7
           "r" (alpha)   // 8
 	: "cc", 
+	  "%xmm0", "%xmm1", "%xmm2", "%xmm3",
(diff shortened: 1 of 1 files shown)


martin-frbg:
Thanks. Yet another error pattern to look for ... but at first glance it appears to be a unique oversight in #561. gcc >=8 is really bringing out the worst traits of the code it seems...

martin-frbg:
FYI compiling with -ftree-vectorize -O2 (on Kaby Lake, i.e. Haswell kernels) leads to a "less than half accurate" warning for SGEMV so there appears to be more such niceness. (Possibly related to my misgivings about input arguments beyond %0,%1 getting modified #2009)

martin-frbg:
The SGEMV mishandling does indeed appear to come from abusing one of the input parameters (%8 a.k.a. "lda4") of sgemv_kernel_4x8 in sgemv_n_microk_haswell-4.c . I did not manage to come up with a proper constraint, so my current workaround is to "movq" the value from %8 into the otherwise unused %xmm10 at the start and restore it from there - not sure if that is a sane solution...

PR #2019, opened 2019-02-14
Fix unannounced modification of input operand 8 (lda4) in Haswell GEMVN microkernel

martin-frbg:
Fixes miscompilation with gcc9 -ftree-vectorize (related to issue #2009 and PR #2018)

### DIFF
--- a/kernel/x86_64/sgemv_n_microk_haswell-4.c
+++ b/kernel/x86_64/sgemv_n_microk_haswell-4.c
@@ -37,43 +37,41 @@ static void sgemv_kernel_4x8( BLASLONG n, FLOAT **ap, FLOAT *x, FLOAT *y, BLASLO
 	__asm__  __volatile__
 	(
 	"vzeroupper			 \n\t"
-	"vbroadcastss    (%2), %%ymm12	 \n\t"	// x0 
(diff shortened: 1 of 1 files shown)


bartoldeman:
Another way to do this is to rename. Swapping %2 (x) with %8 and then adding the + for 2 gives a minimal diff, but you could also rename 2 to 3, 3 to 4, .. 7 to 8 and then 8 to 2.

bartoldeman:
That looks good to me. An alternative that avoids this awkward reordering is to use "asmSymbolicName"s, as documented here: https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html
Very old GCCs did not have those: I was digging a bit and found those were added in GCC 3.1 (from 2002).

bartoldeman:
In any case this becomes a moot point once it's rewritten to use intrinsics like the skylake-x code, so for now I understand it's just about fixing bugs.

martin-frbg:
Thanks again for the review and additional pointers. I'll probably go with the reordering for now, although it is a bit uglier with the next file (the dtrsm_kernel_RN_haswell mentioned in #2009). 
Similar bugs exist in SGEMVN for Nehalem and Sandybridge, and in some GEMVN and TRSM
kernels for Bulldozer and Piledriver, I'll see if I can get them fixed over the weekend.

PR #2025, opened 2019-02-17
Thread safety tester using C++11 threading

TiborGY:
As far as I could tell, there was no test that really tested the thread safety of the library, with many concurrent BLAS/LAPACK calls.
This small C++ program aims to remedy that. It uses C++11 threading to launch a configurably large number of simultaneous calculations on random matrices. All threads are given a copy of the same inputs, and after all threads are done, the results are compared and an error is shown if any of the threads return a different result. This process is repeated a configurably large number of times to try and catch quasi-random failures.

WIP, no LAPACK calls and no makefiles yet. Probably going to hook this up as an optional test, in case folks have no C++(11) compiler.

The git history is a bit messy, but I think it should clean itself up if a squash merge is done.

- [ ] Ready for review


martin-frbg:
You are right that there is no specific test for this in OpenBLAS, any thread safety checks so far have been performed by running valgrind or thread sanitizer on either test cases from the benchmark directory or reproducers for individual issues. If your test is not particularly time consuming, we could hook it up to the existing utest framework.

TiborGY:
The DGEMM test is pretty much done, as far as the test itself is concerned. When using a single threaded build of OpenBLAS the test reliably passes with 50 concurrent threads, and reliably fails with 51 threads, exactly as expected, since by default 50 buffers are allocated in OpenBLAS.
Memory usage of the test is quite high, but using 512*512 matrices was not enough to reliably trigger thread safety problems.
I am going to create a LAPACK test as well, and then try to hook up the tests into the build system.

TiborGY:
DGEMM and DGEMV tests are done. I will probably continue working on this branch, and then make a new PR with a cleaner history. I hope to get this merge-ready soon-ish, so I will probably forgo the LAPACK tests for now.

TiborGY:
#2148 supersedes this

PR #2026, opened 2019-02-19
Correct range limiting in trmv_thread and re-enable TRMV multithreading

martin-frbg:
reverts a workaround introduced for #1332 by fixing the problem introduced with #1262. Somehow I had managed to miss trmv_thread.c in/after the earlier PR#1388 that took care of the same problem in related files, although I had suggested a link to this issue there.


martin-frbg:
Pesky appveyor timeout... rerunning after adding DYNAMIC_LIST in the appveyor config

PR #2028, opened 2019-02-24
Move one of clobber fixes to right place

brada4:
ditto


martin-frbg:
Thanks, just noticed this myself.

brada4:
Welcome, just updated my copy and spotted a mistake

Issue #2030, opened 2019-02-27
openblas crashes firefox


beew:
OS Ubuntu 16.04 64 bit
openblas 0.3.4 and 0.3.5 compiled from source with 
```
make  NO_AFFINITY=1 DYNAMIC_ARCH=1 USE_OPENMP=1
```
I use update-alternatives to switch between openblas and intel-mkl, update-alternatives creates a symlink of libblas to /usr/lib and I sue LD_PRELOAD to preload libblas (whichever that is the default in update-alternatives)

If openblas is used, firefox crashes. Trying to start Firefox from terminal after it crashes got these outputs and it crashes immediately again (FF is up to date 65.01)
```
ExceptionHandler::GenerateDump cloned child 13834
ExceptionHandler::SendContinueSignalToChild sent continue signal to child
ExceptionHandler::WaitForContinueSignal waiting for continue signal...

```
This doesn't happen with openblas 0.3.3 and before when compiled and preloaded the same way.

martin-frbg:
What is your hardware ? And is there anything in the firefox output that suggests which component of firefox has a dependency on OpenBLAS ? Also does it work if you do not use LD_PRELOAD (there was a recent issue with sage where it appeared that preloading openblas led to a problem with a signal handler in the python/cython code)

beew:
hardware is 

Intel® Core™ i7-7820HK CPU @ 2.90GHz × 8 
GeForce GTX 1070/PCIe/SSE2

I don't know how to get additional FF crash info other than the exceptions posted.  No it doesn't crash without LD_PRELOAD or LD_PRELOAD older versions of openblas (or intel-mkl)

brada4:
How do you use BLAS in Firefox? Firefox does not even load native libraries nowadays like java etc.
It is quite aparent that OpenBLAS and firefox uses platform-native threading while MKL uses own homebrew equivalents.
It is very hard to debug with LD_PRELOAD, and that is typically used to override malloc with debug versions etc.
Could you work out firefox with gdb so that we see if firefox actually signalled foreign, openblas thread.
EDIT: it boils down to starting firefox in gdb, then pressing r(un), once it crashes typing `t(hread) a(pply) a(ll) bt`
In principle neither firefox nor openblas should lose track of own threads.

martin-frbg:
Must be some third-party plugin for firefox, either for some computationally expensive operation - why would one need BLAS or LAPACk to view websites - or something that creates an indirect dependency, 
[...]

PR #2031, opened 2019-02-27
Do not play with thread affinity under OpenMP

brada4:
Inspired by example in #2030
OpenMP has its own sophisticated affinity handling, If we try to outsmart it we can only make things worse (we do set affinity mask to one core, they may try to land bunch of threads in same CPU socket where automatic process migration is cheap enough to ignore)

### DIFF
--- a/Makefile.system
+++ b/Makefile.system
@@ -222,6 +222,10 @@ override USE_THREAD = 0
 override USE_OPENMP = 0
 endif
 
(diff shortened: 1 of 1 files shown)


martin-frbg:
Not sure if it is necessary to forcibly prevent a non-default setting here ?

brada4:
Maybe it should show the warning to mark the spot where to patch it out if really needed?

PR #2032, opened 2019-02-27
Cleanup useless files in top level

brada4:
All are years behind current OpenBLAS, functional replacement is "just type `make` in top dir of source tree.


martin-frbg (changes requested):
This does not appear to be true - common_linux.h actually still has the call to SYS_mbind() that this patch tries to replace with a "return 0" , from the git comments the patch appears to be needed on old linux kernels (preceding 2.6.34, e.g. RHEL6, SLES11) only.

martin-frbg:
Not sure I see the point of removing the tiny quickbuild scripts, even if they do little more than invoke `make` (for some inexperienced user) ?

brada4:
That inexperienced user would not want other bit-ness library on their system. GCC is not the ultimate build tool for windows, something usable of clang-cl/icl/ifort kind is better fit. Maybe cross-building paragraph could be enhanced with cross-building openblas.dll on linux, or then again Makefile.rule or both.

brada4:
My preference would be to add to both documentation files.

TiborGY:
I think quickbuild.32bit and quickbuild.64bit can be deleted, they are just clutter. Telling an user to run them is no easier than to tell them to run the single line command in those files. Also why does quickbuild.32bit have "-j 2" in it, when quickbuild.64bit does not have that? Seems inconsistent and pointless.

quickbuild.win32/64 could probably stay, as they have a link to the wiki in them.

brada4:
That -j2 is plainly overridden (see Makefile.rule around MAKE_NB_JOBS)
Might be it hints to 2 threads for 1-core CPUs typical at times of gotoblas?

brada4:
Patch is not used downstream https://src.fedoraproject.org/cgit/rpms/openblas.git/tree/openblas.spec
Random segfaults should be here often if it was missing.

martin-frbg:
As noted above the patch is only needed on old Linux kernels from around ten years ago. It is linked in the FAQ and I think I saw it mentioned in issue tickets of at least one other project (cannot check now as my query "triggered an anti-abuse measure" on github for some reason).

brada4:
Probably better off to some directory next to scavenged downstream patches one day.

PR #2039, opened 2019-03-03
Address warning in memory.c

brada4:
Complete #1814 lock cleanup, what was missed that excess locking code actually initialized the var as a side effect.


martin-frbg:
Are you sure that setting this variable to zero is appropriate here - if I am not mistaken the code you removed through #1814 would have initialized it from a call to WhereAmI()

brada4:
Whereami varies array index , like n-th thread tries at n-th spot onwards.
If that used to fail then it started at zero anyway.

PR #2042, opened 2019-03-04
add TARGET support for HiSilicon tsv110 CPUs

maomao194313:
HiSilicon tsv110 was independently designed by Huawei based on ARMv8 architecture license and applied to Huawei Kunpeng 920 server. It significantly improves processor performance by optimizing branch prediction algorithms, increasing the number of execution units, and improving the memory subsystem architecture and so on. The  compilation based on TSV110 requires gcc9.0 support, otherwise the compiler will report an error.


kernel/arm64/KERNEL.TSV110
```diff
@@ -0,0 +1,175 @@
+SAMINKERNEL  = ../arm/amin.c
```
brada4, line 1: The file could be greatly shortened by including ARMV8 kernel and leaving just optimized overrides here. What do you think?

brada4:
there is (quite new) ./cpuid_arm64.c and ./drivers/other/dynamic.c , first is for build time detection 2nd to make DYNAMIC_ARCH=1 package work on your target too.

brada4:
Also if I google right it should be march=armv8.4-a as from linux kernel patches adding support for this CPU (but there should be fallback for few years old compilers that distributions use)

kernel/arm64/KERNEL.TSV110
martin-frbg: Seems to me the recent rewrites by @rengolin did away with the "include KERNEL.ARMV8" , so the current redundancy would be acceptable. (In particular if this is to be seen as initial cpu support to be improved upon later).
Perhaps some rainy afternoon someday could be spent rearranging all the arm64 KERNEL files so that they all list the kernels in the same order for easier comparison :slightly_smiling_face:

maomao194313:
> Also if I google right it should be march=armv8.4-a as from linux kernel patches adding support for this CPU (but there should be fallback for few years old compilers that distributions use

TSV110 implements all the required features of arm v8.2, but only partially supports the features of armv8.4

brada4:
You know better for sure.

kernel/arm64/KERNEL.TSV110
brada4: Yes, it is good, CPUID would be handy, otherwise complete enough

maomao194313: The current file is just adding new platform support. This file is not the final version and will be continuously optimized and modified.

maomao194313:
> there is (quite new) ./cpuid_arm64.c and ./drivers/other/dynamic.c , first is for build time detection 2nd to make DYNAMIC_ARCH=1 package work on your target too.


Sorry, I didn't pay attention to this feature, I will add it later.

martin-frbg:
According to the gcc aarch64-cores.def, the relevant data for autodetection would appear to be implementer = 0x48, cpu_part = 0xd01

maomao194313:
> According to the gcc aarch64-cores.def, the relevant data for autodetection would appear to be implementer = 0x48, cpu_part = 0xd01

Yes, you are right

martin-frbg:
This modified version of cpuid_arm64.c should take care of the compile-time autodetection then:
[cpuid_arm64.txt](https://github.com/xianyi/OpenBLAS/files/2948641/cpuid_arm64.txt)

Issue #2033, opened 2019-02-28
Compile fails on 32-bit


opoplawski:
See https://kojipkgs.fedoraproject.org//work/tasks/5955/33095955/build.log 
```
gfortran -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m32 -march=i686 -mtune=generic -msse2 -mfpmath=sse -mstackrealign -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection  -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m32 -march=i686 -mtune=generic -msse2 -mfpmath=sse -mstackrealign -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection  -frecursive -Wl,-z,relro -Wl,--as-needed  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -o dblat2 dblat2.o ../libRblas-r0.3.6.dev.a -lm -lgfortran -lm -lgfortran -L/usr/lib/gcc/i686-redhat-linux/9 -L/usr/lib/gcc/i686-redhat-linux/9/../../..  -lc  
BUILDSTDERR: /usr/bin/ld: ../libRblas-r0.3.6.dev.a(dynamic.o): in function `support_avx512':
BUILDSTDERR: /builddir/build/BUILD/openblas-0.3.6/Rblas/driver/others/dynamic.c:336: undefined reference to `xgetbv'
BUILDSTDERR: collect2: error: ld returned 1 exit status
BUILDSTDERR: make[1]: *** [Makefile:137: dblat1] Error 1
BUILDSTDERR: make[1]: *** Waiting for unfinished jobs....
[...]

PR #2045, opened 2019-03-05
Do not compile in AVX512 check if AVX support is disabled

martin-frbg:
Availability of helper function xgetbv depends on NO_AVX being undefined - we could change that too, but that combo is unlikely to work anyway. For #2033


brada4:
Built for me (on tumbleweed) with gcc 7 8 clang 5 6 7, I would tend to imply that rawhide is no different.

PR #2046, opened 2019-03-07
ctest.c : add __POWERPC__ for PowerMac

kencu:
this extra definition satisfies the gcc compiler on PowerPC Apple computers

### DIFF
--- a/ctest.c
+++ b/ctest.c
@@ -113,7 +113,7 @@ ARCH_X86
 ARCH_X86_64
 #endif
 
-#if defined(__powerpc___) || defined(__PPC__) || defined(_POWER)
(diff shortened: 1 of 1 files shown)


martin-frbg:
Thanks. I suspect PowerMac support has not been looked at in a while

kencu:
FYI for followup -- autodetecting the POWER compiler flavour on PPC Darwin may not working, but forcing a PPCG4 target is successful on a PowerMac running Leopard, and on a PowerMac running PPC Tiger (2007 -- not bad!) if a newer version "make" is installed.

Forcing the PPC970 target even on a PowerMac with a 970 fails at present -- so perhaps the autodetect is finding the PPC970 but that is not working as a target. Have to check that...

Issue #2047, opened 2019-03-07
Segfault in sgemm_kernel on old x86_64


Celelibi:
The single precision gemm kernel consistently crash on an old x86_64 CPU. But only with large enough matrices. Like 1000*1000.

Here is the simplest program I could write that shows it.
```C
#include <stdlib.h>
#include <cblas.h>

#define SIZE 1000

int main(void) {
	float A[SIZE * SIZE];
	float C[SIZE * SIZE];
	cblas_sgemm(CblasRowMajor, CblasNoTrans, CblasNoTrans, SIZE, SIZE, SIZE, 1, A, SIZE, A, SIZE, 0, C, SIZE);
	return 0;
}
```

Weirdly enough, it doesn't crash for a single thread.
```
$ gcc -o sgemm_test sgemm_test.c -lopenblas
$ ./sgemm_test
zsh: segmentation fault  ./sgemm_test
$ OPENBLAS_NUM_THREADS=1 ./sgemm_test
$ OPENBLAS_NUM_THREADS=2 ./sgemm_test
zsh: segmentation fault  OPENBLAS_NUM_THREADS=2 ./sgemm_test
```

And here is the cpuinfo.
```
$ cat /proc/cpuinfo 
processor       : 0
vendor_id       : AuthenticAMD
cpu family      : 15
model           : 104
model name      : AMD Athlon(tm) 64 X2 Dual-Core Processor TK-55
stepping        : 1
cpu MHz         : 1800.000
cache size      : 256 KB
physical id     : 0
siblings        : 2
core id         : 0
cpu cores       : 2
apicid          : 0
initial apicid  : 0
fpu             : yes
fpu_exception   : yes
cpuid level     : 1
wp              : yes
flags           : fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ht syscall nx mmxext fxsr_opt rdtscp lm 3dnowext 3dnow rep_good nopl cpuid extd_apicid pni cx16 lahf_lm cmp_legacy svm extapic cr8_legacy 3dnowprefetch vmmcall lbrv
bugs            : apic_c1e fxsave_leak sysret_ss_attrs null_seg swapgs_fence amd_e400 spectre_v1 spectre_v2
bogomips        : 3591.07
TLB size        : 1024 4K pages
clflush size    : 64
cache_alignment : 64
address sizes   : 40 bits physical, 48 bits virtual
[...]

PR #2049, opened 2019-03-07
Fix crash in sgemm SSE/nano kernel on x86_64

Celelibi:
Fix bug #2047 

I managed to test the nano kernel. Make it crash and test with my fix. Although it might benefit from further testing, I doubt this fix could introduce new bugs.


martin-frbg:
Thanks.

PR #2050, opened 2019-03-07
PowerMac 970 fixes

kencu:
With these two additional changes, OpenBLAS builds through to completion on a DualG5 PowerMac PPC970 system running either MacOSX Leopard or Tiger, and shows good performance on the test suite. (Tiger needs to use a newer `make` than the system default.)

In `common_power.h`,  the gcc assembler fails with parameter errors if `DCBT_ARG = 8`. I'm not sure the added PPC970 enabler needs to be guarded with a Darwin test as well, but it seemed this approach had least chance of unexpected wreckage. I don't have a PPC970 system set up with Linux at present to see if it's generally needed, or if it's Darwin specific. I can remove the extra Darwin test if you prefer. 

Likewise, in param.h I'm not sure why the extra PPC970 defines have been guarded with a Linux test -- possibly for the same reason of being untested on other systems -- but this approach keeps it confined to tested systems. I can remove the system tests in that block completely if you prefer.

I have not tested 64 bit PPC970 Darwin.


martin-frbg:
Looks good to me with the added guards - the previous code appears to have been unchanged from
GotoBLAS so there is no telling if K.Goto actually tried anything other than Linux (BSD? AIX?) on PPC970.

kencu:
OK -- works on PowerMac G5 970 and PowerMac G4 7450, Tiger and Leopard, so it's good on this end.

Issue #2063, opened 2019-03-18
Memory leak on Cygwin when forking


embray:
In #1450 I made a fix on Cygwin to ensure that the thread pool is properly shut down before forking  process.

However, I've observed now that there's a memory leak associated with this somehow: Repeatedly shutting down the thread pool, then starting it up again to do some multi-threaded operation, then forking again and so on quickly leaks memory.  I was able to reproduce this simply using Python with a Numpy linked against OpenBLAS (where `np.linalg.inv` happens to use dgesv):

```python
>>> import numpy as np
>>> import subprocess as sp
>>> mat = np.random.random(9).reshape((3, 3))
>>> while True:
...     _ = np.linalg.inv(mat)
...     _ = sp.call(['true'])
...
```

While letting this loop run the memory usage of the original process quickly blows up to several gigabytes.

I will investigate further.

embray:
Additionally, I do not see the same behavior on Linux.

brada4:
It is called fork bomb....
Does same happen with python.org product?

embray:
It's not a question of fork bombs.  E.g. if I put a sleep in the loop it still leaks in the original process.

I don't believe plain-Windows Python is relevant here since it doesn't have any notion of fork and basically never shuts down the thread pool except when unloading the DLL, but it would be interesting to check.

embray:
One thing I noticed is that `CloseHandle` is never called on the threads, so there is a leak in thread handles.  However, fixing that doesn't fix the overall problem.

embray:
Yeah, each time through the loop there's a memory allocation of just a smidge over 32 MB, which looks to be the same as 

```
/* The number of bytes that will be allocated for each buffer. When allocating
   memory, we store an alloc_t followed by the actual buffer memory. This means
   that each allocation always has its associated alloc_t, without the need
   for an auxiliary tracking structure. */
static const int allocation_block_size = BUFFER_SIZE + sizeof(struct alloc_t);
```

from driver/others/memory.c,  so some allocation isn't been freed.  Just need to nail down which.  It's not immediately obvious, and I wouldn't necessarily rule out a bug in Cygwin either.

embray:
I haven't had a chance to test yet, but I think it is possibly as a consequence of compiling with the `USE_TLS=1` feature.

[...]

PR #2064, opened 2019-03-19
Fix for #2063

embray:
The `DllMain` used in Cygwin did not run the thread memory pool cleanup upon `DLL_THREAD_DETACH` which is needed when compiled with `USE_TLS=1`.

There is also a `DllMain` implemented in memory.c for some reason, which is not used on Cygwin (just plain Windows), and also has some special bits for when OpenBLAS is used as a static lib.  If anyone who understands this better thinks it would be good to use this on Cygwin instead, and has a good idea how it would be nice to get some input on this.

But otherwise I don't see a cleaner way to do this.  It seems to work perfectly though to fix the memory leak reported in #2063.

### DIFF
--- a/driver/others/memory.c
+++ b/driver/others/memory.c
@@ -1313,6 +1313,13 @@ void blas_memory_free_nolock(void * map_address) {
   free(map_address);
 }
 
+#ifdef SMP
+void blas_thread_memory_cleanup(void) {
+    blas_memory_cleanup((void*)get_memory_table());
+}
+#endif
+
+
 void blas_shutdown(void){
 #ifdef SMP
   BLASFUNC(blas_thread_shutdown)();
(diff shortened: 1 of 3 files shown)


driver/others/memory.c
```diff
@@ -1313,6 +1313,13 @@ void blas_memory_free_nolock(void * map_address) {
   free(map_address);
 }
 
+#ifdef SMP
```
embray, line 1316: I'll note that this `ifdef` is completely superfluous, as this entire section is, I believe, technically under an `#ifdef SMP`.  Nevertheless, it us used in several other places in the surrounding code, so I'm just using it here for consistency's sake.  Later it will be good to clean this up when refactoring this file.

Issue #1997, opened 2019-02-04
Compilation with target POWER8 fails in AIX 7.2


ayappanec:
I am trying to compile OpenBLAS (develop branch) in AIX 7.2 with TARGET=POWER8.
GCC compiler used is 8.1.0 version.

#oslevel -s
7200-03-00-0000

#prtconf
System Model: IBM,8286-42A
Machine Serial Number: 1005CEV
Processor Type: PowerPC_POWER8
Processor Implementation Mode: POWER 8

#make BINARY=64 CC="gcc -maix64" FC="gfortran -maix64" CXX="g++ -maix64" TARGET=POWER8
......
......
gcc -maix64 -c -Ofast -mcpu=power8 -mtune=power8 -mvsx -malign-power -DUSE_OPENMP -fno-fast-math -fopenmp -DMAX_STACK_ALLOC=2048 -fopenmp -Wall -DF_INTERFACE_GFORT
 -fPIC -DSMP_SERVER -DUSE_OPENMP -DNO_WARMUP -DMAX_CPU_NUMBER=2 -DMAX_PARALLEL_NUMBER=1 -DVERSION=\"0.3.6.dev\" -mpowerpc64 -maix64 -DASMNAME=idamax_k -DASMFNAME=i
damax_k_ -DNAME=idamax_k_ -DCNAME=idamax_k -DCHAR_NAME=\"idamax_k_\" -DCHAR_CNAME=\"idamax_k\" -DNO_AFFINITY -I.. -DDOUBLE  -UCOMPLEX -UCOMPLEX -DDOUBLE -DUSE_ABS
 -UUSE_MIN ../kernel/power/idamax.c -o idamax_k.o
gcc -maix64 -c -Ofast -mcpu=power8 -mtune=power8 -mvsx -malign-power -DUSE_OPENMP -fno-fast-math -fopenmp -DMAX_STACK_ALLOC=2048 -fopenmp -Wall -DF_INTERFACE_GFORT
 -fPIC -DSMP_SERVER -DUSE_OPENMP -DNO_WARMUP -DMAX_CPU_NUMBER=2 -DMAX_PARALLEL_NUMBER=1 -DVERSION=\"0.3.6.dev\" -mpowerpc64 -maix64 -DASMNAME=idamin_k -DASMFNAME=i
[...]

PR #2069, opened 2019-03-25
AIX asm syntax changes needed for shared object creation

ayappanec:
As mentioned in the issue #1997 , for shared object creation in AIX from archives, one has to use export file to export the symbols. There are two standard scripts out there to create the export file from an archive, CreateExportList (comes from XLC fileset ) & GNU libtool (m4/libtool.m4). Both the scripts ignores the symbols in the object files that starts with "." 
So the current AIX specific asm syntax needs change to avoid missing out necessary symbols while creating shared object from the final archive.

### DIFF
--- a/common_power.h
+++ b/common_power.h
@@ -598,9 +598,14 @@ REALNAME:;\
 #ifndef __64BIT__
 #define PROLOGUE \
 	.machine "any";\
+	.toc;\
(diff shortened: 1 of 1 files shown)


ayappanec:
Thanks @martin-frbg

Issue #2016, opened 2019-02-14
Power9 VSX BLAS 3 double precision real


edelsohn:
Optimize BLAS 3 double precision real (DGEMM, DTRMM) for Power9 VSX. Tune for Power9 pipeline, cache and memory hierarchy. Exploit new Power9 instructions.  Demonstrate equivalent performance to IBM ESSL.

quickwritereader:
Martin could you close the issue so that I could claim for bounty. Thanks

martin-frbg:
Certainly. (Methinks the whole process leaves something to be desired - it is not apparent here which tickets carry a bounty, and there does not appear to be any protocol on ticket closing w.r.t approval especially for relatively rare platforms where only those who pledged the bounty may be in a position to decide if the PR is correct and sufficient.)

edelsohn:
I'm not allowed to add labels to issues. All four of the Power9 optimization issues have bounties attached.

Acceptance is a combination of performance goals and OpenBLAS community acceptance of the patch (design, implementation, style, etc.)

PR #2070, opened 2019-03-29
power9 makefile. dgemm based on power8 kernel with following changes …

quickwritereader:
…: 32x unrolled 16x4 kernel and 8x4 kernel using (lxv stxv butterfly rank1 update). improvement from 17 to 22-23gflops. dtrmm cases were added into dgemm itself.
Though it is below desired 90% peak level and around 78-80% we decided to publish as it is still improved version above current kernel.


quickwritereader:
#2016

kernel/power/dgemm_kernel_power9.S
```diff
@@ -0,0 +1,249 @@
...
+    stxv    v29,  432(SP)
+    stxv    v30,  448(SP)
+    stxv    v31,  464(SP)
+
+
+	stfd	f1,  ALPHA_SP
+	stw	r0,  FZERO 
+
+	slwi	LDC, LDC, BASE_SHIFT
+
```
yuriy-yarosh, line 156: lwsync before a massive store might help with op ordering a bit.

kernel/power/dgemm_kernel_power9.S
```diff
@@ -0,0 +1,249 @@
...
+
+#ifndef NEEDPARAM
+
+	PROLOGUE
+	PROFCODE
+
+	addi	SP, SP, -STACKSIZE
+	li	r0, 0
+
```
yuriy-yarosh, line 93: It should be a good idea to add a cache hint with dcbtstds over here, with TH=0b01010 and the respective units count specified.

kernel/power/dgemm_kernel_power9.S
```diff
@@ -0,0 +1,249 @@
...
+	li	o48, 48
+
+
+	lxvdsx	alpha_r, 0, T1
+
+#include "dgemm_logic_power9.S"
+
+.L999:
+	addi	r3, 0, 0
+
```
yuriy-yarosh, line 188: Same goes for loads dcbtds, and TH=0b01010 would be great, but I think it worth also marking this load as Transient.

kernel/power/dgemm_logic_power9.S
```diff
@@ -0,0 +1,1981 @@
...
+	li		T1,	128
+	li		T2,	256
+ 
+	mr		AO,	A
+	mr		CO,	C
+	slwi		T3,	LDC	,	2
+	add		C,	C,	T3
+
+ 
+	dcbt		A,	T1
```
yuriy-yarosh, line 50: It worth using dcbtds with TH=0b01010 and Transient flag over here.
It's worth using extended mnemonics for dcbt - it makes code a bit more readable.

quickwritereader: @yuriy-yarosh Thanks for the valuable advice Frankly, While patching I went on ignoring and not adding prefetch instructions on both sgemm/dgemm to focus on correctness and initial architectural problems.  I will definitely consider adding them back in a  proper and readable way to improve performance further.

PR #2084, opened 2019-04-09
Add in runtime CPU detection for POWER.

RashmicaG:


brada4:
There is still ton of other CPUs in Power series starting around MAC G5...

martin-frbg:
@brada4 do you actually expect anybody to be using a 10+ years old PowerMac **and**  wanting DYNAMIC_ARCH to work on it ?

brada4:
Actuallu power6 is discontinued by manufacturer days ago....
Looking at debian just 8 and 9 is worth now....

martin-frbg:
Guess power6 still makes sense as a safe generlc fallback (more so than ppc970 or earlier) - at least I expect Rashmica will know better than eiher of us)

Issue #2066, opened 2019-03-21
Crash in ReLAPACK


Jellby:
I tried to use OpenBLAS with ReLAPACK in OpenMolcas (https://gitlab.com/Molcas/OpenMolcas) and got a crash, apparently in `RELAPACK_dgetrf_rec at dgetrf.c`.

I compiled OpenBLAS with:

```
make USE_OPENMP=1 INTERFACE64=1 NO_CBLAS=1 BUILD_RELAPACK=1 LIBPREFIX=libopenblas_i8
```

and the crash occurs, for example, when I run:

```
pymolcas verify 036
```

Without ReLAPACK, I don't see the problem.

What other information can I provide or how can I debug it further?

martin-frbg:
You could try building OpenBLAS with debug information (setting DEBUG=1, or adding `-g` to the compile flags) to get more detailed line information. (I assume your "crash" is a segmentation fault, or is it just that the calculation goes haywire ?) Which version of OpenBLAS did you try (not that much happened within ReLAPACK recently, but the problem could have started elsewhere in the code) ?

Jellby:
This is with `v0.3.5`. A crash is indeed a segmentation fault. With `DEBUG=1` I get this stack trace:

```
symbolized stack trace:
    #0 (?) /Programs/OpenBLAS/libopenblas_i8_nehalemp-r0.3.5.so 0x3ab4d9
    #1 (?) /Programs/OpenBLAS/libopenblas_i8_nehalemp-r0.3.5.so 0x8ca98
    #2 RELAPACK_dgetrf_rec at /Programs/OpenBLAS/relapack/src/dgetrf.c:67
    #3 RELAPACK_dgetrf_rec at /Programs/OpenBLAS/relapack/src/dgetrf.c:98 (discriminator 3)
    #4 RELAPACK_dgetrf_rec at /Programs/OpenBLAS/relapack/src/dgetrf.c:98 (discriminator 3)
    #5 RELAPACK_dgetrf_rec at /Programs/OpenBLAS/relapack/src/dgetrf.c:98 (discriminator 3)
    #6 RELAPACK_dgetrf_rec at /Programs/OpenBLAS/relapack/src/dgetrf.c:98 (discriminator 3)
    #7 RELAPACK_dgetrf at /Programs/OpenBLAS/relapack/src/dgetrf.c:35
    #8 dgetrf_ at /Programs/OpenBLAS/relapack/src/lapack_wrappers.c:373
[...]

PR #2093, opened 2019-04-23
Fix ReLAPACK compilation with INTERFACE64 

martin-frbg:
also correct cross-language calls to xerbla that would segfault due to missing length of name string.
For #2066


martin-frbg:
Probably better to split this into separate PRs, one for the basic OpenBLAS compatibility (lack of INTERFACE64 support and wrong interface to XERBLA) and the other(s) for probable ReLAPACK bugs encountered with the lapack testsuite...

PR #2098, opened 2019-04-28
Disable reallocation of work array in ReLAPACK xSYTRF

martin-frbg:
as it appears to cause memory management problems (seen in the LAPACK tests)


martin-frbg:
Well, that is a non-default configuration of OpenBLAS, and I'd still consider ReLAPACK a bit experimental. On the other hand the LAPACK testsuite may use small or otherwise unusual input sizes that are less likely to occur in actual production code so the xSYTRF problem certainly requires further attention. (Though my current understanding is that the behaviour of ReLAPACK is still valid with this silent internal resizing of the work array disabled)

PR #2105, opened 2019-05-01
Correct argument of CPU_ISSET for glibc <2.5

### LINKED ISSUE #2104: memory.c invalid argument type in v0.3.6
Bumping OpenBLAS builds from `v0.3.5` stable release to `v0.3.6` stable release in downstream scientific Python ecosystem (see PR [here](https://github.com/MacPython/openblas-libs/pull/2)) is causing a new error to crop up from the OpenBLAS source:

```
memory.c: In function ‘get_num_procs’:
memory.c:1775:10: error: invalid type argument of ‘->’ (have ‘cpu_set_t’)
      if (CPU_ISSET(i,cpuset)) n++;
          ^
```

Comparing the build log with the last successful build (`v0.3.5`), the first point of deviation appears to be this new warning, which may or may not be pertinent:
```
getarch.c:96:0: warning: "NO_AVX512" redefined [enabled by default]
 #define NO_AVX512
 ^
<command-line>:0:0: note: this is the location of the previous definition
```

Any suggestions for us?

martin-frbg:
fixes #2104

### DIFF
--- a/driver/others/memory.c
+++ b/driver/others/memory.c
@@ -229,7 +229,7 @@ int get_num_procs(void) {
   n=0;
   #if !__GLIBC_PREREQ(2, 6)
   for (i=0;i<nums;i++)
-     if (CPU_ISSET(i,cpuset)) n++;
+     if (CPU_ISSET(i,&cpuset)) n++;
   nums=n;
   #else
   nums = CPU_COUNT(sizeof(cpuset),&cpuset);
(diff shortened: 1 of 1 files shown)


tylerjereddy:
Yes, I was able to reproduce the originally reported OpenBLAS build failure and confirm a restored OpenBLAS build using the patch provided by this PR in a `manylinux1` container:

```
docker pull quay.io/pypa/manylinux1_x86_64
docker run -it quay.io/pypa/manylinux1_x86_64
git clone https://github.com/xianyi/OpenBLAS.git
make # with or without this patch
```

Issue #2015, opened 2019-02-14
Power9 VSX BLAS 3 single precision real


edelsohn:
Optimize BLAS 3 single precision real for Power9 VSX. Tune for Power9 pipeline, cache and memory hierarchy. Exploit new Power9 instructions. Demonstrate equivalent performance to IBM ESSL.

quickwritereader:
@martin-frbg After merge we can close this issue. As initial improvements are considered good for power9.

PR #2107, opened 2019-05-01
sgemm/strmm kernel for power9

quickwritereader:
sgemm/strmm kernel for power9. (note power8 kernels did not perform well on new architecture)


quickwritereader:
#2015 
@martin-frbg

PR #2110, opened 2019-05-03
Fix detection of Skylake processors when using GCC

michaellass:
21eda8b introduced a check in getarch.c to test if the compiler is capable of AVX512. This check is currently non-functional for GCC for two reasons:
* It checks for the `__AVX2__` macro instead of `__AVX512__`,
* these macros are only defined if getarch itself was compiled with AVX2/AVX512 support (e.g., with `-march=native`).

Instead, we can just test the GCC version as we do for clang. Support for `-march=skylake-avx512`, which is of interest here, was added in GCC 5.3: https://gcc.gnu.org/gcc-5/changes.html

While looking at the CPU detection, I noticed a small error in c_check, trying to unlink a file called `tmpf.o` instead of the actually created temporary object file. This is fixed as well.

### DIFF
--- a/Makefile.system
+++ b/Makefile.system
@@ -9,6 +9,11 @@ ifndef TOPDIR
 TOPDIR = .
 endif
 
+# If ARCH is not set, we use the host system's architecture.
+ifndef ARCH
+ARCH := $(shell uname -m)
+endif
+
 # Catch conflicting usage of ARCH in some BSD environments
 ifeq ($(ARCH), amd64)
 override ARCH=x86_64
@@ -137,6 +142,10 @@ endif
 endif
 
 
+# On x86_64 build getarch with march=native. This is required to detect AVX512 support in getarch.
+ifeq ($(ARCH), x86_64)
+GETARCH_FLAGS += -march=native
+endif
 
 
 ifdef INTERFACE64
--- a/c_check
+++ b/c_check
@@ -240,7 +240,7 @@ if (($architecture eq "x86") || ($architecture eq "x86_64")) {
 	} else {
 	    $no_avx512 = 0;
 	}
-	unlink("tmpf.o");
+	unlink("$tmpf.o");
     }
 }
 
--- a/cmake/system.cmake
+++ b/cmake/system.cmake
@@ -65,6 +65,11 @@ if (DEFINED TARGET)
   set(GETARCH_FLAGS "-DFORCE_${TARGET}")
 endif ()
 
+# On x86_64 build getarch with march=native. This is required to detect AVX512 support in getarch.
+if (X86_64)
+  set(GETARCH_FLAGS "${GETARCH_FLAGS} -march=native")
+endif ()
+
 if (INTERFACE64)
   message(STATUS "Using 64-bit integers.")
   set(GETARCH_FLAGS	"${GETARCH_FLAGS} -DUSE64BITINT")


michaellass:
@martin-frbg Could you have a look if this works for you?

martin-frbg:
Thanks. I had copied the version&macro test from fenrus75's SKX kernels where checking for (at least) AVX2 capability makes more sense (as `-march=` is expected to have been set up correctly by the time this is compiled). I am not convinced this makes the check entirely non-functional unless your default gcc specs does not even include avx2, but I agree that a plain version check would probably be sufficient.

michaellass:
For users that just download OpenBLAS / clone the git and enter make, most likely march is just x86-64 and getarch will see no AVX support. We actually ran into this issue on our systems where Skylake was mistaken for Haswell for this reason. If the CFLAGS are tweaked for the host or the GCC default specs were overridden then the current version should indeed work as expected.

Another question: Is the check for AVX512 support in c_check still required or could we just remove that entirely now that getarch takes care of it? It currently writes NO_AVX512=1 to Makefile.conf which is used at least in `driver/others/dynamic.c`. But a quick compilation test using an old GCC (4.8.5) showed no issues for me after removing it.

martin-frbg:
Good question, but probably yes. The code got so messy because I do not have any actual SkylakeX hardware to test it on (and the build system is a tangled mess anyway). We will certainly want to pick up all of - "SkylakeX with ancient compiler", "SkylakeX with a VM or OS that disabled the avx512 capability", "SkylakeX but the user expressly wants NO_AVX512" and add the required -march=skylakex-avx512 to CCOMMON_OPT when everything is perfect for AVX512 - with both CMAKE and plain make.
I'd rather have a little redundancy here than another round of compile failures.

martin-frbg (commented):
The version check here should be aligned with the #ifdef in the actual SkylakeX kernels (from which I borrowed the line in the first place). Not sure if fenrus75 actually wanted to exclude gcc 5 and 6, but his code imposes GCC > 6 - in #1721 he noted "I'm conservative in only picking recent compilers given the CI ICEs etc)

brada4:
Typical gcc5/6 packages do not include intrinsics headers for avx2
Version check is not indicative or sufficient

michaellass:
@brada4 Does this hold for GCC 5/6 only or in general, i.e., would a check for GCC >6 be sufficient? Alternatively, if we need to check for `__AVX2__` or `__AVX512__`, getarch needs to be compiled with `-march=native`, e.g., by adding it directly to https://github.com/xianyi/OpenBLAS/blob/3f427c0cf979dc3ba8bc0e48bda973dfdd6c35fb/Makefile.prebuild#L62.

brada4:
Like #1789 - gcc version not changed, but headers added to package.

martin-frbg:
I'd think a missing mmintrinsic.h would get caught by the compile test in c_check (as long as that stays), so keeping the original __GNUC__ >6 without my misguided __AVX2__ should be sufficient ?

michaellass:
I have no system with a recent enough compiler that does not support AVX512. But when I combine getarch.c from this pull request with a manipulated c_check that always puts `NO_AVX512=1` into Makefile.conf, Makefile.conf then still contains `SKYLAKEX` as `CORE` and the skylakex kernels are used (including immintrin.h).

Would compiling getarch with `march=native` be a viable option? Otherwise we would probably have to check for AVX512 support before compiling and running getarch.

martin-frbg:
I think the fundamental problem is that we do check for AVX512 before running getarch but there is still no (working) way to pass that information from c_check to getarch.

martin-frbg:
Probably easiest indeed to just add `-march=native` to GETARCH_FLAGS in Makefile.system - after all the getarch tool is intended to run only on the build host and explore its capabilities.

michaellass:
I agree. I'll update this PR accordingly.

martin-frbg:
Thanks. Somehow this topic keeps confusing me - I guess this still leaves the (contrieved) case where someone would want to build for SkylakeX on a system that has the required compiler but lacks AVX512 hardware, (I believe this would work in principle, just fail to run any of the tests - but a DYNAMIC_ARCH build would serve much the same purpose except for library size)

michaellass:
Yes, it's probably still not bullet proof. However, this should already help those on a Skylake system who just type in "make" and rely on the auto detection.

michaellass:
This seems to break clang, at least on arm64. So I guess we're still looking for a proper solution here that passes the result of c_check to getarch.

martin-frbg:
Probably safe to limit this to x86_64 architecture (though in the Makefile you would need to mess with uname directly, as ARCH is probably set by getarch - cmake at least  has CMAKE_SYSTEM_PROCESSOR set to one of x86_64, amd64 or AMD64.

michaellass:
Doesn't ARCH refer to the target architecture anyway? The decision has to be made based on the build host architecture, though. So, we'd need to have something like HOSTARCH, determined by `uname -m`. CMake's [CMAKE_HOST_SYSTEM_PROCESSOR](https://cmake.org/cmake/help/v3.15/variable/CMAKE_HOST_SYSTEM_PROCESSOR.html) unfortunately is based on `uname -p` which on macOS returns i386 even on 64 bit machines, so we'd have to use `uname -m` here as well.

Or am I confused and ARCH refers to the build host architecture?

martin-frbg:
Wouldn't the two ARCH types be equivalent for all practical purposes (unless you plan to cross-compile for SKX on your phone) ?

michaellass:
... or on a POWER or some upcoming ARM server. Still, that's an unlikely scenario.

Makefile.system
```diff
@@ -9,6 +9,11 @@ ifndef TOPDIR
 TOPDIR = .
 endif
 
+# If ARCH is not set, we use the host system's architecture.
+ifndef ARCH
+ARCH := $(shell uname -m)
```
michaellass, line 14: I hope this has no side effects when doing cross compilation. I have no setup to test this. If it has, of course we can use another variable here.

martin-frbg: My gut feeling is to use some local name that cannot ever conflict with the general ARCH setting here, but we will see how this goes.

PR #2112, opened 2019-05-05
Makefile.arm: remove -march flags

ffontaine:
The provided -march flags, especially for ARMv5 and ARMv6 may not
necessarily match the needed ones: for ARMv5, it might be armv5,
armv5te, armv5t, etc. If the wrong one is used, the incorrect toolchain
sysroot can be used in a multilib toolchain.

Therefore, let the user building OpenBLAS pass the appropriate -march
flag.

The other flags, such as -mfpu=vfp or -mfloat-abi=hard are kept, as they
are actually required for the build to proceed (OpenBLAS uses VFP
instructions, and assume an EABIhf ABI).

[Peter: update for v0.2.20]
Signed-off-by: Thomas Petazzoni <thomas.petazzoni@free-electrons.com>
Signed-off-by: Peter Korsgaard <peter@korsgaard.com>
[Retrieved from:
https://git.buildroot.net/buildroot/tree/package/openblas/0001-Makefile.arm-remove-march-flags.patch]
Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/Makefile.arm
+++ b/Makefile.arm
@@ -1,7 +1,7 @@
 ifeq ($(CORE), $(filter $(CORE),ARMV7 CORTEXA9 CORTEXA15))
 ifeq ($(OSNAME), Android)
-CCOMMON_OPT += -mfpu=neon  -march=armv7-a
-FCOMMON_OPT += -mfpu=neon  -march=armv7-a
+CCOMMON_OPT += -mfpu=neon
+FCOMMON_OPT += -mfpu=neon
 else
 CCOMMON_OPT += -mfpu=vfpv3 -march=armv7-a
 FCOMMON_OPT += -mfpu=vfpv3 -march=armv7-a
@@ -9,11 +9,6 @@ endif
 endif
 
 ifeq ($(CORE), ARMV6)
-CCOMMON_OPT += -mfpu=vfp -march=armv6
-FCOMMON_OPT += -mfpu=vfp -march=armv6
-endif
-
-ifeq ($(CORE), ARMV5)
-CCOMMON_OPT += -march=armv5
-FCOMMON_OPT += -march=armv5
+CCOMMON_OPT += -mfpu=vfp
+FCOMMON_OPT += -mfpu=vfp
 endif


brada4:
We should ask @rengolin who set flags for aarch64
-march= flag allows no-assembly code to use full ISA, for example in fortran as such

martin-frbg:
At least with V5 and V6 we are (probably) more likely to be talking about specialist equipment where the user will already know what flags to pass, compared to somebody just wishing to compile something for their phone. Still some architecture-specific guidelines would probably be a good addition to the faq/wiki.

martin-frbg:
Perhaps if any guidelines can be formulated, the appropriate flag selection can be derived from system or compiler characteristics in c_check and/or Makefile.system similar to the FP ABI setting in #1221 ? (What is currently in Makefile.arm looks like it came from the initial ARM support effort that may not have been so ambitious as to address all possible flavors of armv5 from the start, and even ashwinyes' later contributions seem to have been made without access to actual hardware)

rengolin:
Sorry for the delay, I didn't see the @ mention on my email. The change looks good. :)

martin-frbg:
Thanks for the confirmation. Actually I had almost forgotten about this PR until I tried to reproduce one of the open softfp problems in qemu (where I only found armv6 support) and promptly hit the inappropriate default -march flag. :)

martin-frbg:
As this is now causing problems for Android users, I wonder if the removal of the -march=armv7-a from the "OS_ANDROID" part of the conditional was really warranted ? If I interpret https://developer.android.com/ndk/guides/abis correctly, there should be no supported devices that are ARMV7 but not armv7-a ?

ffontaine:
Indeed, I sent back the original patch from buildroot that was applied since 2018: https://git.buildroot.net/buildroot/commit/package/openblas?id=dae0d7b3caae25246882455e6204fea9b367ab24. However, buildroot does not support Android so the OS_ANDROID part of the conditional could be reverted without impacting buildroot.

Issue #2113, opened 2019-05-06
getrf_parallel.c strange code


brada4:
Here the first "check" actually never checks if it is free
https://github.com/xianyi/OpenBLAS/blob/6a8b4269b522a8e1383e27e0ffef6111993a7264/lapack/getrf/getrf_parallel.c#L284
Id like to make a PR with either removing first no-check leaving just loop, or enclosing loop in if()?

martin-frbg:
Guess I had this as a while-do originally so needed to pull the initial value out of the loop, or I was just confused... Or maybe I was worried that optimization might change the order of execution ? Anyway there appear to be a few more in this style - _probably_ harmless but certainly pointless. In each case, the original (non-locking) code was just a one-liner while() with a dummy do() block

brada4:
I think not much space to make mistakes in PR.

PR #2120, opened 2019-05-07
Address redundant code concern #2113

brada4:
Remove extra unchecked checking step three times from getrf_parallel.c

### DIFF
--- a/lapack/getrf/getrf_parallel.c
+++ b/lapack/getrf/getrf_parallel.c
@@ -279,9 +279,6 @@ static int inner_advanced_thread(blas_arg_t *args, BLASLONG *range_m, BLASLONG *
     for (i = 0; i < args -> nthreads; i++)
 #if 1
     {
-	LOCK_COMMAND(&getrf_lock);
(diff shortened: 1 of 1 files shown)


brada4:
Trigger new CI

martin-frbg:
new CI had a dummy recipe only, abusing this PR to try an actual build now :slightly_smiling_face:

tylerjereddy:
I usually set up pipelines on my fork of a repo and debug there rather than pushing to develop, then open a PR when it is ready. I guess that has more upfront time cost though.

tylerjereddy:
I think you're making changes through the Azure web interface and then those are actually getting committed to develop--maybe you know that, but just checking. It can make the commit history a bit messy vs. adjusting the config yml in a PR.

martin-frbg:
> I think you're making changes through the Azure web interface and then those are actually getting committed to develop--maybe you know that, but just checking. It can make the commit history a bit messy vs. adjusting the config yml in a PR.

Err, actually I was doing direct commits to develop as I did not expect it to get so messy. Sorry.

PR #2122, opened 2019-05-08
Add ARMV6 build to azure CI setup

martin-frbg:
(and repair the mess I made there earlier today)


tylerjereddy:
Nice, I suppose once this is working, if it doesn't time out, you can remove the corresponding emulated entry in the Travis matrix, which consistently times out now (uses 50 full minutes), I think.

martin-frbg:
Took 53:33 after suppression of all fortran-based tests - the sole advantage of Azure here is that they (currently) allow up to 60 minutes, qemu is slow when it has to do cpu emulation. rengolin did some work towards setting up some sort of CI on actual arm hardware in #1867 but github integration was unsolved

tylerjereddy:
The free [Shippable service provides ARMv8 native CI](http://docs.shippable.com/platform/tutorial/workflow/run-ci-builds-on-arm/), so those two ARMv8 entries should be possible to migrate productively away from Travis. But I suspect @xianyi would have to set that up, and for NumPy I had to email them to get permission. But perhaps worthwhile eventually.

ARMv6 looks tricky indeed.

PR #2124, opened 2019-05-09
TST: Azure manylinux1 & clean-up

tylerjereddy:
* remove some of the steps & comments
from the original Azure yml template

* modify the trigger section to use
develop since OpenBLAS primarily uses
this branch; use the same batching
behavior as downstream projects NumPy/
SciPy

* remove Travis emulated ARMv6 gcc build
because this now happens in Azure

* use documented Ubuntu vmImage name for Azure
and add in a manylinux1 test run to the matrix
for older glibc and catching things
like #2104

[skip appveyor]

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -177,8 +177,8 @@ matrix:
       dist: trusty
       sudo: required
       services: docker
-      env: IMAGE_ARCH=arm32 TARGET_ARCH=ARMV6 COMPILER=gcc
-      name: "Emulated Build for ARMV6 with gcc"
+      env: IMAGE_ARCH=arm32 TARGET_ARCH=ARMV6 COMPILER=clang
+      name: "Emulated Build for ARMV6 with clang"
       before_install: sudo docker run --rm --privileged multiarch/qemu-user-static:register --reset
       script: |
         echo "FROM openblas/alpine:${IMAGE_ARCH}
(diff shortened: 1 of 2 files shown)


martin-frbg:
Thanks. (I had already wondered why the default choice of the dormant "master" branch as the trigger appeared to work)

martin-frbg:
Quick Q before I get tempted to mess with the azure file again - is it normal/intended that the Azure jobs show up as three individual lines in the github UI now (one for each individual job and one for the Azure task as such) ? I assume this is a byproduct of converting previous "script" steps to "jobs" ?

tylerjereddy:
> is it normal/intended that the Azure jobs show up as three individual lines in the github UI now (one for each individual job and one for the Azure task as such) ? I assume this is a byproduct of converting previous "script" steps to "jobs" ?

Yeah, we asked Microsoft about that downstream & they said the expanded list was working as intended. On balance, I think some maintainers didn't like the extra bloat in the immediate CI list, but it is 10 parallel jobs for open source so we live with it :)

And yes for adding jobs causing expansion of matrix  you see right in GitHub. You can also add a `matrix` directive within a job for sub-expansion if you want like here: https://github.com/numpy/numpy/blob/master/azure-pipelines.yml#L120

That will also have the same effect on display.

martin-frbg:
So this is just Microsoft trying to be different - it does look like unwanted extra bloat to me, but as long as the qemu jobs get their free 60 minutes of fame I can certainly live with it.

PR #2134, opened 2019-05-14
TST: add SkylakeX AVX512 CI test

tylerjereddy:
* adapt the C-level reproducer code for some
recent SkylakeX AVX512 kernel issues, provided
by Isuru Fernando and modified by Martin Kroeker,
for usage in the utest suite

* add an Intel SDE SkylakeX emulation utest run to
the Azure CI matrix; a custom Docker build was required
because Ubuntu image provided by Azure does not support
AVX512VL instructions

If you're like me and don't believe regression tests until you've seen the failure, [here's the same CI run](https://dev.azure.com/tylerjereddy/openblas-fork/_build/results?buildId=896) on a branch off OpenBLAS at 3f427c0c, the most problematic recent build used in NumPy ecosystem. After demonstrating that failure, I ported the relevant changes over to this PR branch.

The hope is that this can supersede the Intel SDE skx regression guards for:
- NumPy: https://github.com/numpy/numpy/pull/13466  
- SciPy: https://github.com/scipy/scipy/pull/10145
- etc.

and provide a faster feedback loop in SkylakeX AVX512 kernel issues directly / automatically at the source, rather than downstream.

This took rather longer to draft than I had hoped--one unfortunate issue is the way we `echo` to Dockerfiles rather than just putting well-formatted Dockerfiles into the repo proper, and the fact that `docker build` doesn't support `--privileged` as required for SDE process attachment.

Of course, there are still a few issues open about SkylakeX AVX512 matters, so this may simply be the start of related regression guards. And as Isuru previously noted, BLIS actually iterates over a few different emulation archs, so you / we might consider building up to that I suppose.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -25,3 +25,27 @@ jobs:
             make -C utest $COMMON_FLAGS $BTYPE" > Dockerfile
       docker build .
     displayName: Run manylinux1 docker build
+- job: Intel_SDE_skx
+  pool:
+    vmImage: 'ubuntu-16.04'
+  steps:
+  - script: |
+      # at the time of writing the available Azure Ubuntu vm image
+      # does not support AVX512VL, so use more recent LTS version
+      echo "FROM ubuntu:bionic
+      COPY . /tmp/openblas
+      RUN apt-get -y update && apt-get -y install \\
+          cmake \\
+          gfortran \\
+          make \\
+          wget
+      RUN mkdir /tmp/SDE && cd /tmp/SDE && \\
+          mkdir sde-external-8.35.0-2019-03-11-lin && \\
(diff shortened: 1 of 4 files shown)


martin-frbg:
>Of course, there are still a few issues open about SkylakeX AVX512 matters

Are there additional issues in numpy/scipy that are related to OpenBLAS SKX kernels apart from DGEMM ?

tylerjereddy:
> Are there additional issues in numpy/scipy that are related to OpenBLAS SKX kernels apart from DGEMM ?

As of 6a8b4269, both NumPy and SciPy full linear algebra test suites pass--I'm planning to use that commit to distribute with the final wheels for `SciPy 1.3.0`. So that's great!

I meant that there are more issues open to OpenBLAS itself about SKX and that there is at least a chance that some of them might be slightly different targets than the C reproducer used here? Maybe I'm wrong on that--the various discussions seem complicated by difficulties with reproduction.

As noted previously, when you deactivated the skx AVX512 kernels more selectively there was an uptick in NumPy skx test failures, and then the more comprehensive deactivation allowed the full linear algebra suite to pass.

tylerjereddy:
looks like an unrelated timeout on the mac job

Issue #1912, opened 2018-12-12
amax:samax utest failure


virtuald:
System configuration:

* Compile host is ODROID-C2, but I'm running the compile inside of a docker image that was created from a NI RoboRIO disk image (ARMv7, Cortex A9). I'm also using `setarch linux32`.
* Compilers are gcc 6.3.0, gfortran 6.3.0
* OpenBLAS 0.3.4 (but the error occurs on 0.2.20 also)

Compile was via:

```
make TARGET=CORTEXA9 PREFIX=/usr/local
```

Here's the output:

```
./openblas_utest 
TEST 1/23 amax:samax [FAIL]
  ERR: test_amax.c:44  expected 3.300e+00, got 4.204e-45 (diff 3.300e+00, tol 1.000e-04)
...
RESULTS: 23 tests (22 ok, 1 failed, 0 skipped) ran in 1860 ms
```

What steps should I take next to diagnose this issue? Thanks!

virtuald:
Looking more at the compile log, this pops out at me:

```
OPENBLAS_NUM_THREADS=1 OMP_NUM_THREADS=1 ./sblat3 < ./sblat3.dat
OPENBLAS_NUM_THREADS=1 OMP_NUM_THREADS=1 ./dblat3 < ./dblat3.dat
OPENBLAS_NUM_THREADS=1 OMP_NUM_THREADS=1 ./cblat2 < ./cblat2.dat
OPENBLAS_NUM_THREADS=1 OMP_NUM_THREADS=1 ./cblat3 < ./cblat3.dat
OPENBLAS_NUM_THREADS=1 OMP_NUM_THREADS=1 ./zblat2 < ./zblat2.dat
 TESTS OF THE COMPLEX          LEVEL 3 BLAS

 THE FOLLOWING PARAMETER VALUES WILL BE USED:
   FOR N                   0     1     2     3     7    31
   FOR ALPHA          ( 0.0, 0.0)  ( 1.0, 0.0)  ( 0.7,-0.9)  
   FOR BETA           ( 0.0, 0.0)  ( 1.0, 0.0)  ( 1.3,-1.1)  

 ERROR-EXITS WILL NOT BE TESTED

 ROUTINES PASS COMPUTATIONAL TESTS IF TEST RATIO IS LESS THAN   16.00

 RELATIVE MACHINE PRECISION IS TAKEN TO BE  1.2E-07

[...]

PR #2142, opened 2019-05-28
Add softfp support in min/max kernels

martin-frbg:
fix for #1912

### DIFF
--- a/kernel/arm/iamax_vfp.S
+++ b/kernel/arm/iamax_vfp.S
@@ -469,9 +469,11 @@ iamax_kernel_S10:
 
 
 iamax_kernel_L999:
-
+#if !defined(__ARM_PCS_VFP)
+	vmov    r0, s0
+#else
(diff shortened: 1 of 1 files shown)


ashwinyes:
@martin-frbg Could you please explain this change ?

iamax is supposed to return the index of the maximum value.

martin-frbg:
You are right that this was probably misguided, I did it in response to a SAMAX error on softfp by simple analogy to existing softfp changes in other kernels. (Either I miss something fundamental in the workings of this kernel, or perhaps arm32 should not use the same kernel for AMAX and IAMAX - probably the former.)

ashwinyes:
If we use this kernel for samax for hardfp case, it will work. Because the max value inadvertently ends up in s0. And index ends up in r0 which will work for iamax.

But for softfp case, iamax needs index to end up in r0, and amax also needs max value to end up in r0. So it cannot handle both.

Better to have a separate cleaner implementation for amax.

PR #2145, opened 2019-05-29
Separate implementations of AMAX and IAMAX on arm

martin-frbg:
As noted in #1912 and comment on #2142, the combined implementation happens to "do the right thing" on hardfp, but cannot return both value and index on softfp where they would have to share the return register

### DIFF
--- a/kernel/arm/KERNEL.ARMV6
+++ b/kernel/arm/KERNEL.ARMV6
@@ -1,20 +1,20 @@
 include $(KERNELDIR)/KERNEL.ARMV5
 
-SAMAXKERNEL  = iamax_vfp.S
-DAMAXKERNEL  = iamax_vfp.S
-CAMAXKERNEL  = iamax_vfp.S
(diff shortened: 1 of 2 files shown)


ashwinyes:
The implementation amax_vfp.S will not work for double precision or complex cases (both double and single).

martin-frbg:
Right, so far this is only an attempt to fix SAMAX for #1912 without breaking ISAMAX. So how would I handle double precision, would a single "vmov r0  r1 d0" be sufficient ? (I have even less of an idea how to handle the complex case, on my own I would probably drop them back to their generic C implementations)

ashwinyes:
For Double, see nrm2_vfp.S. "vmov r0, r1, d0" should be enough.

For Complex case, it would be better to write a file name izamax_vfp.S. For now move back to generic C implementation.

martin-frbg:
Thank you - that is actually where  I got the idea from. It will still be some time until I manage to write any meaningful assembly from scratch.

PR #2148, opened 2019-06-01
Thread safety tester using C++11 threading (cleaned history)

TiborGY:
As far as I could tell, there was no test that really tested the thread safety of the library, with many concurrent BLAS/LAPACK calls.
This small C++ program aims to remedy that. It uses C++11 threading to launch a configurably large number of simultaneous calculations on random matrices. All threads are given a copy of the same inputs, and after all threads are done, the results are compared and an error is shown if any of the threads return a different result. This process is repeated a configurably large number of times to try and catch quasi-random failures.

No LAPACK calls yet.
The test requires a considerable amount of RAM (~1300 MiB), so it is not enabled by default.
Using 512*512 matrices was not enough to reliably trigger thread safety problems.

Launching threads sequentially has a significant chance of giving a false negative. If an OMP parallel for construct is used to launch the concurrent calculations, then all threads start almost exactly at the same time, thus they have the highest chance of clashing.

A C++11 compiler and an OpenMP implementation needs to be present for this test to work, and CBLAS needs to be built.

closes #2025

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ endif
 
 LAPACK_NOOPT := $(filter-out -O0 -O1 -O2 -O3 -Ofast,$(LAPACK_FFLAGS))
 
-SUBDIRS_ALL = $(SUBDIRS) test ctest utest exports benchmark ../laswp ../bench
+SUBDIRS_ALL = $(SUBDIRS) test ctest utest exports benchmark ../laswp ../bench cpp_thread_test
 
 .PHONY : all libs netlib $(RELA) test ctest shared install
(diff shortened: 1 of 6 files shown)


brada4:
There are openmp implementations not based on pthreads, like iomp, or windows.

TiborGY:
> 
> 
> There are openmp implementations not based on pthreads, like iomp, or windows.

OK. Does that impact this PR?

brada4:
Yes, you are pulling pthread along omp, just decide which one is 'c++ threading' in your interpretation, because it is not anything of std::

TiborGY:
C++ threading in this case refers to std::async
I am not using pthread directly at all

PR #2162, opened 2019-06-13
Fixes for PGI compiler

martin-frbg:
fixes compile failure with pgi 18.10 as reported on OpenBLAS-users

### DIFF
--- a/Makefile.x86_64
+++ b/Makefile.x86_64
@@ -31,7 +31,7 @@ ifndef NO_AVX2
 ifeq ($(C_COMPILER), GCC)
 CCOMMON_OPT += -mavx2
 endif
-ifeq $(F_COMPILER), GFORTRAN)
+ifeq ($(F_COMPILER), GFORTRAN)
 FCOMMON_OPT += -mavx2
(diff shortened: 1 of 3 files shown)


Makefile.x86_64
```diff
@@ -28,11 +28,15 @@ endif
 ifeq ($(CORE), HASWELL)
 ifndef DYNAMIC_ARCH
 ifndef NO_AVX2
+ifeq ($(C_COMPILER), GCC)
 CCOMMON_OPT += -mavx2
+endif
+ifeq $(F_COMPILER), GFORTRAN)
```
tylerjereddy: I was a bit curious why CI was only failing for manylinux1, but it looks like it is just a missing opening bracket here maybe?

Perhaps only observed on manylinux1 because of the crude CI build code I wrote there, but maybe that's useful for this kind of detection anyway.

martin-frbg: Most likely it is just an older version of `make` with a different parser. But it is certainly good to have such a variety of systems available for CI.

PR #2170, opened 2019-06-28
Fix build on PPC970 for FreeBSD

pkubaj:
PPC970 on FreeBSD needs those macros too.

You could consider testing PPC970 on Linux to simplify this if:
`#if defined(POWER3) || defined(POWER6) || defined(PPCG4) || defined(CELL) || defined(POWER8) || defined(POWER9) || ( defined(PPC970) && ( defined(OS_DARWIN) || defined(OS_FREEBSD) ) )`


martin-frbg:
-Thanks.That ifdef does not look any worse than some of the x86_64 ones...

PR #2179, opened 2019-07-08
[WIP] Separate dgemv thread safety test

TiborGY:
The DGEMM thread safety test is a bit difficult to run due to its memory requirement.
Having a separate switch would allow running the much less demanding DGEMV test only.
This might allow thread safety testing to be enabled in one of the CI builds.


brada4:
Actually there was an idea of casting gemm special case to gemv, or you suspect some regression in this idea?

martin-frbg:
addressed in #2841

TiborGY:
Thanks Martin, I never got around to investing more work into OpenBLAS, but hopefully eventually I will.

Issue #2180, opened 2019-07-08
Test and tune for Zen 2


TiborGY:
Zen 2 is now released, bringing a number of improvements to the table.
Most notably, it now has 256 wide AVX units. This should in theory allow performance parity with Haswell-Coffee Lake CPUs, and initial results suggest this is true (at least for single thread).
https://i.imgur.com/sFhxPrW.png

The chips also have double the L3 cache, and a generally reworked cache hierarchy. One thing to note, is that these chips do not have enough TLB cache to cover all of L2 and L3, so hugepages might be a little more important.

I might be able to get my hands on a Zen 2 system in ~1-2 months.

brada4:
It is not L3 cache per core or NUMA domain, it is per socket, like 1-2MB per core, in place of haswell's 2.5MB
Smaller than zen1 L1d actually matches that of haswell....
Probably neither is considered even with zen1, just that some lengthy discussions how to work around BIOS with broken NUMA support.

TiborGY:
> 
> 
> It is not L3 cache per core or NUMA domain, it is per socket, like 1-2MB per core, in place of haswell's 2.5MB

I have no idea what you are talking about. The 3700X has 8 cores and a total of 32 MiB of L3. Internally each cluster of 4 cores share their L3, so its more like 2*16MiB of L3. That still works out to 4 MiB of L3 per core. No idea where you are getting the 1-2 MiB from.

L3 cache is not shared between the 4 core core complexes (CCXs), not even within the same die.

wjc404:
@TiborGY I also found tuning of kernel is required for zen2. I tested single-thread dgemm performance of OpenBLAS(target=Haswell) on a Ryzen 7 3700X at 3.0GHz fixed clock, got ~33GFLOPS, which was far behind the theoretical maximum (48GFLOPS at 3.0GHz). By the way I also tested [my dgemm subroutine](https://github.com/wjc404/DGEMM_AVX2) and got a speed of ~44GFLOPS.

wjc404:
![AIDA64_Cache_Mem_Test_R7-3700X](https://user-images.githubusercontent.com/52632443/61184159-01a19400-a67d-11e9-82ac-34f14c016816.png)
The speed of L3 in r7-3700X is fast, but the memory latency is still a problem.
I think the enhanced size of L3 allows larger blocks from matrix B to be packed, thus reducing the bandwidth requirement for accessing matrix A and C, eliminating the problem of slow memory access.

[...]

Issue #1459, opened 2018-02-13
make error


akshay-1993:
C:\Python27\OpenBLAS>make install
process_begin: CreateProcess(NULL, uname -s, ...) failed.
process_begin: CreateProcess(NULL, gcc -DGEMM_MULTITHREAD_THRESHOLD=4 -DNO_PARALLEL_MAKE=0 -o getarch getarch.c cpuid.S, ...) failed.
make (e=2): The system cannot find the file specified.
make: *** [getarch] Error 2
Makefile.system:171: Makefile.conf: No such file or directory
Makefile.:2: Makefile.system: Too many open files
Makefile.:90: warning: overriding commands for target `shared'
Makefile.:90: warning: ignoring old commands for target `shared'
Makefile.:116: warning: overriding commands for target `tests'
Makefile.:116: warning: ignoring old commands for target `tests'
Makefile.:138: warning: overriding commands for target `libs'
Makefile.:138: warning: ignoring old commands for target `libs'
Makefile.:169: warning: overriding commands for target `prof_blas'
Makefile.:169: warning: ignoring old commands for target `prof_blas'
Makefile.:180: warning: overriding commands for target `blas'
Makefile.:180: warning: ignoring old commands for target `blas'
Makefile.:188: warning: overriding commands for target `hpl'
Makefile.:188: warning: ignoring old commands for target `hpl'
Makefile.:202: warning: overriding commands for target `hpl_p'
Makefile.:202: warning: ignoring old commands for target `hpl_p'
Makefile.:215: warning: overriding commands for target `netlib'
Makefile.:215: warning: ignoring old commands for target `netlib'
Makefile.:228: warning: overriding commands for target `re_lapack'
Makefile.:228: warning: ignoring old commands for target `re_lapack'
Makefile.:232: warning: overriding commands for target `prof_lapack'
Makefile.:232: warning: ignoring old commands for target `prof_lapack'
Makefile.:236: warning: overriding commands for target `lapack_prebuild'
Makefile.:236: warning: ignoring old commands for target `lapack_prebuild'
Makefile.:279: warning: overriding commands for target `large.tgz'
Makefile.:279: warning: ignoring old commands for target `large.tgz'
Makefile.:286: warning: overriding commands for target `timing.tgz'
Makefile.:286: warning: ignoring old commands for target `timing.tgz'
Makefile.:293: warning: overriding commands for target `lapack-timing'
Makefile.:293: warning: ignoring old commands for target `lapack-timing'
[...]

Issue #1018, opened 2016-11-22
Can't compile openblas


WenqiangJiang:
getarch_2nd.c: In function ‘main’:
getarch_2nd.c:12:35: error: ‘SGEMM_DEFAULT_UNROLL_M’ undeclared (first use in this function)
     printf("SGEMM_UNROLL_M=%d\n", SGEMM_DEFAULT_UNROLL_M);
                                   ^
getarch_2nd.c:12:35: note: each undeclared identifier is reported only once for each function it appears in
getarch_2nd.c:13:35: error: ‘SGEMM_DEFAULT_UNROLL_N’ undeclared (first use in this function)
     printf("SGEMM_UNROLL_N=%d\n", SGEMM_DEFAULT_UNROLL_N);
                                   ^
getarch_2nd.c:14:35: error: ‘DGEMM_DEFAULT_UNROLL_M’ undeclared (first use in this function)
     printf("DGEMM_UNROLL_M=%d\n", DGEMM_DEFAULT_UNROLL_M);
                                   ^
getarch_2nd.c:15:35: error: ‘DGEMM_DEFAULT_UNROLL_N’ undeclared (first use in this function)
     printf("DGEMM_UNROLL_N=%d\n", DGEMM_DEFAULT_UNROLL_N);
                                   ^
getarch_2nd.c:19:35: error: ‘CGEMM_DEFAULT_UNROLL_M’ undeclared (first use in this function)
     printf("CGEMM_UNROLL_M=%d\n", CGEMM_DEFAULT_UNROLL_M);
                                   ^
getarch_2nd.c:20:35: error: ‘CGEMM_DEFAULT_UNROLL_N’ undeclared (first use in this function)
     printf("CGEMM_UNROLL_N=%d\n", CGEMM_DEFAULT_UNROLL_N);
                                   ^
getarch_2nd.c:21:35: error: ‘ZGEMM_DEFAULT_UNROLL_M’ undeclared (first use in this function)
     printf("ZGEMM_UNROLL_M=%d\n", ZGEMM_DEFAULT_UNROLL_M);
                                   ^
getarch_2nd.c:22:35: error: ‘ZGEMM_DEFAULT_UNROLL_N’ undeclared (first use in this function)
     printf("ZGEMM_UNROLL_N=%d\n", ZGEMM_DEFAULT_UNROLL_N);
                                   ^
[...]

PR #2186, opened 2019-07-16
Update "dgemm_kernel_4x8_haswell.S" for improving performance on zen2 chips

wjc404:
I tested dgemm performance of OpenBLAS previously on a ryzen 7 3700X cpu (see my comment on issue #2180) and got only 70% theoretical performance, in contrast to 90% on i9-9900K. 
Through controlled modification of kernel codes, I found the performance loss was mainly due to the high latency and low throughput of vpermpd instruction on zen2 (see the comments on issue #2180 for details). 
As a result I replaced most vpermpd instructions in the file "dgemm_kernel_4x8_haswell.S", with much cheaper instructions like vpermilpd and vperm2f128.
After the modification of "dgemm_kernel_4x8_haswell.S" and recompilation, the single-thread performance of dgemm on r7-3700x (turbo boost disabled, 3.6 GHz) improved substantially, from 40-41 GFLOPS to 51-52 GFLOPS (90% theoretical), without loss of the quality of results (test program and terminal outputs are in the attachment).
[test_results.zip](https://github.com/xianyi/OpenBLAS/files/3398291/test_results.zip)

### DIFF
--- a/kernel/x86_64/dgemm_kernel_4x8_haswell.S
+++ b/kernel/x86_64/dgemm_kernel_4x8_haswell.S
@@ -292,8 +292,8 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 	vblendpd $ 0x0a, %ymm7, %ymm6, %ymm2
 	vblendpd $ 0x05, %ymm7, %ymm6, %ymm3
 
-	vperm2f128 $ 0x01 , %ymm2, %ymm2
-	vperm2f128 $ 0x01 , %ymm3, %ymm3
+	vperm2f128 $ 0x01 , %ymm2, %ymm2 , %ymm2
+	vperm2f128 $ 0x01 , %ymm3, %ymm3 , %ymm3
 
 	vblendpd $ 0x03, %ymm0, %ymm2 , %ymm4
 	vblendpd $ 0x03, %ymm1, %ymm3 , %ymm5
@@ -330,8 +330,8 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 	vblendpd $ 0x0a, %ymm11, %ymm10, %ymm2
 	vblendpd $ 0x05, %ymm11, %ymm10, %ymm3
 
-	vperm2f128 $ 0x01 , %ymm2, %ymm2
-	vperm2f128 $ 0x01 , %ymm3, %ymm3
+	vperm2f128 $ 0x01 , %ymm2, %ymm2 , %ymm2
+	vperm2f128 $ 0x01 , %ymm3, %ymm3 , %ymm3
 
 	vblendpd $ 0x03, %ymm0, %ymm2 , %ymm4
 	vblendpd $ 0x03, %ymm1, %ymm3 , %ymm5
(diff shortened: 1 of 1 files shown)


wjc404:
The performance on i9-9900K (1 thread, 4.4 GHz) slightly improved after this update.
Screenshot from 2019-07-17 16-36-32

martin-frbg:
Thank you - unfortunately I did not get around to running any benchmarks on my hardware yet, but this looks great. (And if necessary at any time, it would be no problem to create Zen-specific kernels - the TARGET is already there with few differences in the parameters, it just happens to reuse Haswell kernels for now.)

wjc404:
In addition, I modified some prefetcht0 instructions in the dgemm kernel code to reduce LLC misses, observed an additional 0.4GFLOPS improvement on 9900K (1thread) after this change.

wjc404:
Adjusting the value of macro B_PRI can further reduce L1D cache misses (CYCLES_STALL_L1D_MISS decreased by 20-30Mcyc/s on 9900K as indicated by linux perf).

martin-frbg:
Pity there is no documentation why the original values were chosen (though it could be as simple as something copied from an earlier generation cpu kernel)

wjc404:
Some prefetcht0 instructions were added to the macro SAVE4x12 to prefetch the head part of b block matrix buffer as the access to the buffer will jump to its head (hardware prefetcher may not be cleaver enough to predict that) after SAVE4x12 completes. The single-thread speed improved by 0.2-0.3GFLOPS after the change.

wjc404:
The prefetch code dealing with elements of matrix C was further optimized, gained another 0.2GFLOPS improvement.

TiborGY:
> 
> 
> In addition, I modified some prefetcht0 instructions in the dgemm kernel code to reduce LLC misses, observed an additional 0.4GFLOPS improvement on 9900K (1thread) after this change.

Have you tested the impact of this change on Zen2?

wjc404:
After those optimizations of prefetch , the performance on zen2 increased by 0.5 GFLOPS (1 thread).

TiborGY:
Sounds great, thank you for your contribution!

stevenwong:
Sorry don't mean to hijack this thread. How did you get it to compile?

I'm following instructions in (https://github.com/xianyi/OpenBLAS/wiki/How-to-use-OpenBLAS-in-Microsoft-Visual-Studio), on Ryzen 3700x and I get the following error:

```
  C:\Users\Steven\Documents\OpenBLAS\getarch_2nd.c(12,35): error: use of
  undeclared identifier 'SGEMM_DEFAULT_UNROLL_M'

      printf("SGEMM_UNROLL_M=%d\n", SGEMM_DEFAULT_UNROLL_M);
                                    ^

  C:\Users\Steven\Documents\OpenBLAS\getarch_2nd.c(13,35): error: use of
  undeclared identifier 'SGEMM_DEFAULT_UNROLL_N'

      printf("SGEMM_UNROLL_N=%d\n", SGEMM_DEFAULT_UNROLL_N);
                                    ^
```
After a bit of googling, this error seems to be related to new CPUs #1018

[Edit] I'm building off a git checkout of v0.3.6

[Edit 2] Ok found `TargetList.txt`, seems like I need to set `TARGET=ZEN`

[Edit 3] Nevermind, looks like it's in the works  Commit 0ba29fd2625dfe405a08005a22d0fa21293cc16c

martin-frbg:
@stevenwong I think 0.3.6 should be able to autodetect your cpu (family 15, extended family 8, model 1 or 8 for any generation ZEN if https://en.wikichip.org/wiki/amd/cpuid is to be trusted). Possibly something went wrong with the cpuid call in getarch, which variant of the build methods described on the wiki page did you use ?

stevenwong:
@martin-frbg Method 1 Native (MSVC) ABI, WIndows 10 and VC++ 2015 build tools.

[Edit]
Ok, following the instructions in #1459 and removing config.h in the root directory fixed it.
[\Edit]

Rather than building it in miniconda/base, I've created an empty environment in my regular Anaconda install and done everything in there. I've added the environment to `LIB` and `CPATH`.

Actually, the output suggest there may have been an issue with the compiler. Output and error logs here:

https://gist.github.com/stevenwong/56fbd1fe33ebc6f58e301b4602c07bea
https://gist.github.com/stevenwong/4ba1f89c6d86b3bc9a03e5df44441c29

martin-frbg:
The logs look normal, just that cmake is trying different option dialects to identify the compiler and assembler before it even starts to process the OpenBLAS files. I wonder where the config.h in your root directory came from ?

PR #2191, opened 2019-07-23
MAINT: remove legacy CMake endif()

tylerjereddy:
* clean up a case where CMake endif()
contained the conditional used in the
if(), which is no longer needed /
discouraged since our minimum required
CMake version supports the modern syntax

Ok, so pretty small PR :)


martin-frbg:
Yeah, not quite the release blocker... Actually that old syntax would have been useful for keeping nested if() blocks readable, but this lone example does not quite look like it requires it. So let's remove it for consistency...

PR #2196, opened 2019-07-28
Add vbroadcastsd kernel to dgemm_kernel_4x8_haswell.S

wjc404:
By carefully examining the haswell dgemm kernel code, I found there's an equivalent micro kernel that load elements from matrix A through broadcast operations, to the current one which arranges the elements from A by permute instructions. 
I added the broadcast kernel to the current implementation, along with a "switch" (#define BROADCASTKERNEL) so developers can easily switch back to permute kernel (by disabling the definition) if it is more efficient on the target CPU.
When applying the broadcast kernel, I observed a performance gain of 0.4 GFLOPS (65.3 -> 65.7) on i9-9900K (1 thread, 4.4 GHz). On r7-3700x (1 thread, 3.6 GHz) the improvement was 0.1-0.4 GFLOPS.

### DIFF
--- a/kernel/x86_64/dgemm_kernel_4x8_haswell.S
+++ b/kernel/x86_64/dgemm_kernel_4x8_haswell.S
@@ -107,6 +107,7 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #define	A_PR1	512
 #define	B_PR1	160
+#define BROADCASTKERNEL
 
 /*******************************************************************************************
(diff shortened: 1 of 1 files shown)


wjc404:
Screenshot from 2019-07-28 16-52-32

wjc404:
Screenshot from 2019-07-28 16-55-19
99% MKL performance now

TiborGY:
> 
> 
> 99% MKL performance now

Nice!

martin-frbg:
Impressive. I wish I could do that...

Issue #2233, opened 2019-08-21
wrong answers with 0.3.7 on POWER8 and POWER9


cparrott73:
Not sure if you guys were already aware of this, but I wanted to pass this one along, just in case.

I was trying out the OpenBLAS 0.3.7 release on our POWER8 ("garrison") and POWER9 ("newell") OpenPOWER systems here this week.  I used GCC 9.2.0, although I have also observed similar behavior with GCC 8.3.0, which leads me to believe that this issue is not related to a particular release of GCC.

The library compiles fine, but several tests are failing on both platforms with accuracy errors:

```
 CGEMV  PASSED THE TESTS OF ERROR-EXITS

 ******* FATAL ERROR - COMPUTED RESULT IS LESS THAN HALF ACCURATE *******
                       EXPECTED RESULT                    COMPUTED RESULT
       1  (  -0.102385    ,   0.226428    )  (    0.00000    ,    0.00000    )
       2  (   0.108077    ,   0.805703E-01)  (  -0.102385    ,   0.226428    )
       3  (   0.992165E-02,  -0.840403E-01)  (   0.108077    ,   0.805703E-01)
       4  (   0.192725    ,   0.176668    )  (   0.992165E-02,  -0.840403E-01)
 ******* CGEMV  FAILED ON CALL NUMBER:
    512: CGEMV ('N',  4,  2,( 1.0, 0.0), A,  5, X,-1,( 0.0, 0.0), Y, 1)         .
```

I see similar issues with CGBMV, CHEMV, CHPMV, CTRMV, CTPMV, CTRSV, CTPSV, CGERC, CGERU, CHER, CHPR, CHER2, and CHPR2.

The cblas_cgemv test also fails:

```
 ******* FATAL ERROR - COMPUTED RESULT IS LESS THAN HALF ACCURATE *******
                       EXPECTED RESULT                    COMPUTED RESULT
       1  (  -0.342872    ,   0.714919E-02)  (  -0.342872    ,   0.714919E-02)
       2  (   0.233743    ,  -0.822020E-01)  (   0.233743    ,  -0.822020E-01)
       3  (   0.376182E-01,   0.161061    )  (   0.376182E-01,   0.161061    )
       4  (   0.144652    ,   0.137748    )  (    0.00000    ,    0.00000    )
 ******* cblas_cgemv  FAILED ON CALL NUMBER:
    440: cblas_cgemv (  CblasNoTrans,  4,  2,( 1.0, 0.0), A,
[...]

PR #2234, opened 2019-08-22
Try switching the ppc build to xenial

martin-frbg:
to get a newer gcc

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ matrix:
 
     - <<: *test-ubuntu
       os: linux-ppc64le
+      dist: xenial
       before_script:
         - COMMON_FLAGS="DYNAMIC_ARCH=1 TARGET=POWER8 NUM_THREADS=32"
       env:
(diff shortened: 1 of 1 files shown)


tylerjereddy:
Just a quick note that may be obvious--last time I checked the ppc64le Travis service was considered "beta" so I'm not sure there is much flexibility there--probably good to check the logs to make sure that anything actually changes the way you want it to.

I see a "bionic"  at the top, but then "xenial" actually being used below, maybe?

martin-frbg:
clearly this did not work - the first run was inconclusive as I read (later) that xenial comes with the same gcc5 as precise, but we should have seen gcc 7.4.0 used with bionic. It was just a quick test
anyway - I do not see myself rushing to buy a power8 workstation just to debug #2233.

tylerjereddy:
@martin-frbg Do you not have access to the gcc compile farm machines? You would certainly qualify for access as an open source maintainer for OpenBLAS. The [list of machines](https://cfarm.tetaneutral.net/machines/list/) includes POWER8 and some other useful archs. The [registration form](https://cfarm.tetaneutral.net/users/new/) is pretty simple too. I use those machines to build OpenBLAS for `ppc64le` and `ARMv8` binaries in the Python ecosystem.

PR #2252, opened 2019-09-08
Optimized ?trtrs

### LINKED ISSUE #2251: Slow ?trtrs implementation
openblas uses the reference implementation which is pretty slow since it uses ?trsm by default which is much slower than ?trsv in case the right hand side only has 1 column. This seems to come up regularly and surprise people (see [here](https://github.com/scipy/scipy/issues/10722) for a numpy discussion or [here](https://discourse.julialang.org/t/why-lapack-trtrs-not-blas-trsv/14239) for the same discovery in julia). I noticed ?getrs has an optimized implementation which uses this simple trick (?trsv in case of column vector b, ?trsm otherwise), so I propose to do the same for ?trtrs. Does it make sense?

thrasibule:
closes #2251 
I've written it in the same model as the optimized ?getrs implementation. The idea is to just use ?trsv instead of ?trsm when the right hand side only has 1 column. I've checked that this give the expected speedup.
The only sure I'm not quite sure about, and someone more familiar with the codebase could probably help me with are these lines: 
https://github.com/thrasibule/OpenBLAS/blob/trtrs/interface/lapack/trtrs.c#L142-L143.
I used the same parameters as in ?getrs
and here:
https://github.com/thrasibule/OpenBLAS/blob/trtrs/interface/lapack/trtrs.c#L165
Again I've used the same parameters as ?getrs. This looks like operation counts, so it probably should be half of that, but I'm not sure what these macros do.

This is a work in progress, I can reorganize the commits and maybe add some tests.


brada4:
quick exit on NOOP is missing from code - 
```
      IF( N.EQ.0 )
     $   RETURN
```

thrasibule:
@brada4 it is not. see https://github.com/thrasibule/OpenBLAS/blob/trtrs/interface/lapack/trtrs.c#L125

martin-frbg:
The appveyor failure is unrelated (and my fault), but I do not yet understand what happened to the travis OSX builds.

thrasibule:
This should fix it... I'm surprised it even got picked up, I thought extended precision was disabled.

brada4:
@thrasibule extended precision is alive on *BSD , just that OSX is only of a kind in CI
The other NOOP shortcut - ir could be few lines earlier - and -  does it qualify for upstream?

martin-frbg:
Unfortunately I am seeing a considerable number of test failures in the LAPACK EIG and LIN tests (available through `make lapack-test`) with this PR.

thrasibule:
I think I've fixed it all. I'm still seeing some error in lapack-test, but not more than before this PR.

martin-frbg:
Thank you. The remaining error in STFSM/CTFSM in indeed unrelated to the PR (probably caused by some inaccuracy in the corresponding trsm, and definitely not a recent regression).

PR #2253, opened 2019-09-11
fix error messages

thrasibule:
I noticed that the error messages in the reimplemented lapack funcions are null terminated. For instance, if you run make lapack-test, you'll see a lot of lines similar to this:
 *** XERBLA was called with SRNAME = SGETRF^@ instead      of    SGETRF ***
sizeof("SGETRF") is 1 + strlen("SGETRF"), so  the fix is to remove 1.

### DIFF
--- a/interface/lapack/gesv.c
+++ b/interface/lapack/gesv.c
@@ -44,19 +44,19 @@
 
 #ifndef COMPLEX
 #ifdef XDOUBLE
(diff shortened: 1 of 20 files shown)


brada4:
In original fortran source it is a 6 character string, fixed length, not terminated, padded with space char

PR #2256, opened 2019-09-12
Add vcpkg installation instructions

grdowns:
OpenBLAS is available as a port in vcpkg. Adding installation instructions will help users get started by providing a single command they can use to build OpenBLAS and include it into their CMake projects

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,11 @@ You can download them from [file hosting on sourceforge.net](https://sourceforge
 
 ## Installation from Source
 
+If you are using the [vcpkg](https://github.com/Microsoft/vcpkg/) dependency manager you can download and install OpenBLAS from source with CMake integration using a single command:
+```
+vcpkg install openblas
+```
+
 Download from project homepage, https://xianyi.github.com/OpenBLAS/, or check out the code
 using Git from https://github.com/xianyi/OpenBLAS.git.
 


README.md
```diff
@@ -24,6 +24,11 @@ You can download them from [file hosting on sourceforge.net](https://sourceforge
 
 ## Installation from Source
 
+If you are using the [vcpkg](https://github.com/Microsoft/vcpkg/) dependency manager you can download and install OpenBLAS from source with CMake integration using a single command:
+```
+vcpkg install openblas
```
isuruf, line 29: Isn't this an unoptimized build? I'm not sure an unoptimized build should be advertised without a warning.

martin-frbg: If this is a plain VC build (not using mingw or clang), it will indeed only contain generic C kernels as the Microsoft compiler does not handle the AT&T notation in which the assembly kernels are coded.

martin-frbg: Also not sure if such highly specific instructions for individual package managers should be in the README - we do not mention rpmbuild (etc.) there either ?

cbezault: vcpkg allows users to specify any compiler they want via toolchain files so MSVC can be replaced by clang-cl, though it isn't one-click. 

Even more problematic than that is we don't really have a good LAPACK solution in vcpkg.

I'm working towards allowing packages in vcpkg to be Fortran libraries (at that point we can use netlib's LAPACK) and allowing packages in vcpkg to specify that they REALLY need a specific compiler, so if a user is using MSVC by default if they try to install OpenBLAS we'll use clang-cl.

Until the above two issues are resolved, and vcpkg is a one-click OpenBLAS install/build solution, I don't think this PR should be merged.

grdowns: With @cbezault's input, I'm fine with closing this PR for now (or letting it sit) until we can get the one-click optimized OpenBLAS support within vcpkg. @isuruf and @martin-frbg, thoughts?

brada4:
You can add instructions in wiki, note that it is not extension to existing cmake support, it is just a windows-specific one of visual studio dependency managers , say previous one, nuget, seems quite abandoned.

README.md
martin-frbg: Can someone please update me on the current status of this ? If I read the vcpkg commit log correclty, it now supports at least mingw (and thus gfortran) - OTOH I am unsure if the suggested pointer to vcpkg should be in the OpenBLAS README at all, or belongs on the "Precompiled Installation Packages" page of the OpenBLAS wiki ?

JackBoosY:
I will take over this PR.

rgommers:
This content is already present in the docs. From http://www.openmathlib.org/OpenBLAS/docs/install/#windows:

image

I don't know at what point it got added to the wiki (which this docs content was derived from), but I think this PR can be closed now.

In case someone with `vcpkg` knowledge thinks that this content needs a tweak, I'm happy to take that along in a next PR if you suggest phrasing.

JackBoosY:
cc @Cheney-W to judge whether this PR is needed now.

Cheney-W:
I think this PR can be closed, and no changes are needed for the content on the wiki.

martin-frbg:
Thanks all for the feedback, closing accordingly

PR #2267, opened 2019-09-24
Use CPU count on ARMV8 to guide cache-dependent parameter choices

### LINKED ISSUE #2230: Unexpected slowdown on aarch64 using more than two cores/threads
Hello,

I was trying to compile hpl-2.3 on a Raspberry Pi4 with four core Cortex-A72 at 1.5 GHz using recent aarch64 Debian Buster. It doesn't matter if I use OpenMP or MPI, the results are the same. For more than two cores or threads (N) there is an unexpected slowdown in obtained GFlops:

```
N    MPI    OpenMP
1    4.8    4.8
2    8.7    8.5
3    6.2    6.4
4    4.2    4.7
```

This is for OpenBLAS-0.3.5 available through Debian and also for compiled versions 0.3.6 and 0.3.7.

While the same test using Blis-0.5.1 available through Debian gives expected increase using more cores/threads:

```
N    MPI    OpenMP
1    4.6    4.6
2    8.3    8.6
3    11.1   12.1
4    12.4   14.2
```

Moreover, I've run the test on the same machine on arm32 architecture using recent Raspbian. Here OpenBLAS-0.3.6 compiled from source behaves also as expected (the armv7 version in Raspbian is much slower, not using vfpv4 nor neon):

```
N    MPI    OpenMP
1    3.9    3.9
2    7.2    5.9
3    9.8    7.5
4    11.2   10.9
```

martin-frbg:
There is currently no simple way to query the cache sizes, so just assume that a target with more than eight cores has to be a server-class system with correspondingly big cache.
Fixes #2230 where performance on single-board computers was shown to suffer from an earlier decision to apply the GEMM_P and Q parameters for CortexA53 servers to the entire range of supported Cortex cpus. (This will err on the side of caution when server code is cross-compiled and the actual target cpu count is unavailable)


ashwinyes:
Sorry for replying late on this. 

When this change was made, I was trying on a 2 core system with a large shared L2 cache. Hence I could not check scaling beyond 2 cores for CortexA57.

Then my focus shifted to ThunderX2, so I never came back to this.

I think you can avoid the check for number of cores and reduce the blocking sizes for all cases. Reason being that in a server class machine where there are large number of cores, they could be either sharing a large L2 cache OR they would have smaller dedicated L2 cache and share a large L3 cache. So smaller blocking sizes are better for scaling in both the cases. And I think smaller blocking sizes will not significantly reduce the performance of 1-2 core cases.

PR #2273, opened 2019-09-30
Fix various source comment typos

luzpaz:
Found via `codespell -q 3 -L amin,als,ba,dum,mone,nd,nto,orign -S Changelog.txt,./lapack*`

### DIFF
--- a/driver/others/init.c
+++ b/driver/others/init.c
@@ -563,7 +563,7 @@ static void disable_affinity(void) {
 #endif
 
   /* if(common->final_num_procs > 64){ */
-  /*   fprintf(stderr, "\nOpenBLAS Warining : The number of CPU/Cores(%d) is beyond the limit(64). Terminated.\n", common->final_num_procs); */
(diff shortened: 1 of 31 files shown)


martin-frbg:
Thanks. I must admit I am of two minds when it comes to non-code fixes. On one hand, they tend to clutter up diffs and create a wrong impression that lots of things changed in the code. On the other, I once worked on code written by a brilliant engineer who was legasthenic so I understand that searching for a particular function based on likely code comments can be a huge pain.

brada4:
reference/ is remnant of never implemented test framework. Actually there was a issue reported that never materialized to get rid of it. It is a copy of upstream NetLib BLAS, thus immutable here, down stream.
For the rest - even imperfect, all the misteaks are pretty much understandable even for basic engilsh speaker.

luzpaz:
Close ticket then ?

martin-frbg:
My intention is/was to review/merge this shortly before the next release (provided it has no conflicts). But as stated above it seems pretty pointless to spell-check dead code (reference/) or commented-out messages (init.c) and the remaining transposition typos are not that hard to read.

PR #2274, opened 2019-09-30
Remove prototype of unused, unimplemented function

luzpaz:
Found via `codespell -q 3 -L amin,als,ba,dum,mone,nd,nto,orign -S Changelog.txt,./lapack*`

### DIFF
--- a/common_thread.h
+++ b/common_thread.h
@@ -194,10 +194,6 @@ int trsm_thread(int mode, BLASLONG m, BLASLONG n,
 
 int syrk_thread(int mode, blas_arg_t *, BLASLONG *, BLASLONG *, int (*function)(), void *, void *, BLASLONG);
 
-int beta_thread(int mode, BLASLONG m, BLASLONG n,
-		double alpha_r, double alpha_i,
-		void *c, BLASLONG ldc, int (*function)());
-
 int getrf_thread(int mode, BLASLONG m, BLASLONG n, BLASLONG k,
 		 void *offsetA, BLASLONG lda,
 		 void *offsetB, BLASLONG jb,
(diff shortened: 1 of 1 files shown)


luzpaz:
Wasn't sure about this one, so I split it off from #2273

martin-frbg:
Should be harmless (basically just a comment as well, the asterisk alone should be good enough for the compiler), but I agree it makes one wonder if the obvious typo has any ill effect.

brada4:
It is an unused prototype, you can delete it as well.

brada4:
Quite cute that you have to add paragraph of parameters to avoid calling library export 'main'

brada4 (changes requested):
Please change your patch into full removal of unused function prototype.

martin-frbg:
There is an entry for a (nonexisting) beta_thread in driver/level3/Makefile as well, maybe worth researching if an early GotoBLAS actually had the implementation (and if so, why it was dropped).
And we could just as well comment it as "unused" or "unimplemented" rather than delete a harmless bit of framework.

luzpaz:
@brada4 aka you want me to remove line 199 ? 
 
https://github.com/xianyi/OpenBLAS/pull/2274/files#diff-5bc29bcf15879fc42166282316b0602dL199

brada4:
Whole beta-thread function.

luzpaz:
Shall I squash it?

martin-frbg:
Thanks, no need to squash it as that can be done with the merge (if I remember to switch the combobox from "merge pull request" to "squash and merge" :slightly_smiling_face: )

brada4:
@luzpaz - good spotting, esp for an unorthodox tool.

Issue #2187, opened 2019-07-17
DGELSD gives incorrect results on v0.3.6 with SKYLAKEX


apataki:
It seems like the DGELSD least squares solver via SVD produces incorrect results when built with TARGET=SKYLAKEX or DYNAMIC_ARCH=1 and run on a SKYLAKEX node.  The processor we use is the Xeon 6148.

Here is a sample code trying to recover an exact result via fitting a small polynomial:

<pre>
#include &lt;stdio.h>

#define N 512
#define M 8

double X[N];
double Y[N];
double A[M][N];
double S[M];
double work[N*100];
int lwork = N*100;
int iwork[N*100];

void dgelsd_(int *MM, int *NN, int *NRHS, double *A, int *LDA, double *B, int *LDB,
             double *S, double *RCOND, int *RANK, double *WORK, int *LWORK, int *IWORK, int *INFO);

int main()
{
    int i, j;
    double x, dx, xn;

    dx = 1.0 / (N+1);
    for (i = 0; i < N; i++) {
        x = i*dx;
        Y[i] = 2*x + 3*x*x*x;

        // Generate the Vandermonde matrix
        xn = 1.0;
        for (j = 0; j < M; j++) {
            A[j][i] = xn;
            xn *= x;
        }
    }

    int NN = N;
    int MM = M;
    int NRHS = 1;
    int LDB = N;
    double RCOND = 1e-10;
    int r;
    int info;
    dgelsd_(&NN, &MM, &NRHS, A, &NN, Y, &LDB, S, &RCOND, &r, work, &lwork, iwork, &info);

    printf("info: %d\n", info);

    for (int i = 0; i < M; i++) {
        printf("    %9.6f\n", Y[i]);
    }

    return 0;
}
</pre>

The correct output (on Broadwell for example):

<pre>
info: 0
    -0.000000
     2.000000
    -0.000000
     3.000000
    -0.000000
     0.000000
    -0.000000
     0.000000
</pre>

On the 6148 with TARGET=SKYLAKEX I get:
<pre>
info: 0
    -1.842238
    -1.542870
     1.389274
    -1.695241
     0.884713
    -0.855344
    -0.331920
    -0.538012
</pre>

[...]

PR #2286, opened 2019-10-15
AVX512 DGEMM kernel

wjc404:
Originally designed for icopy_8 and ocopy_8, add transformation function to support icopy_4 (with ~5% performance loss), currently 80-85% MKL performance.


martin-frbg:
Great, thanks. I have been trying off and on to make fenrus75's dgemm code work but was not really getting anywhere with it.

martin-frbg:
Seems the OSX libc lacks aligned_malloc, and a gcc version later than 7 is required for the
"_mm256_cvtsd_f64" intrinsic.

martin-frbg:
And I am happy to confirm that this passes all tests in BLAS-Tester, the DGELSD test from #2187 and also isuruf's DTRMM reproducer from #1955 (showing that trmm was indeed "only" affected indirectly by the change of incopy, no other side effect). The lapack tests show two errors in DHS although the detailed testing_results.txt claims all tests passed.

isuruf:
I'm surprised that appveyor passes even though there's no `posix_memalign` on windows.

martin-frbg:
I suspect the one Appveyor job that uses DYNAMIC_ARCH=1 does not use an AVX512-aware clang.
(And I have restarted the xcode job that appeared to have timed out)

martin-frbg:
If https://stackoverflow.com/questions/196329/osx-lacks-memalign is to be trusted, posix_memalign is not available on older versions of OSX either (but aligned_malloc is available on Windows, so probably make that one #ifndef OS_DARWIN - not sure what to use on that platform though).
Update: OSX appears to have posix_memalign since 10.6 "Snow Leopard" from 2009 so there is 
probably no need for a workaround there, just use the aligned_alloc on OS_WINDOWS and posix_memalign elsewhere.

kernel/x86_64/dgemm_kernel_8x8_skylakex.c
```diff
@@ -429,7 +429,8 @@ static void KERNEL_MAIN(double *packed_a, double *packed_b, BLASLONG m, BLASLONG
     double *c_pointer = c;
     __mmask16 k01 = 0x00f0,k02 = 0x000f,k03 = 0x0033;
     BLASLONG ndiv8_count;
-    double *b_scratch = (double *)aligned_alloc(64,192*k);
+    double *b_scratch;
+    posix_memalign(&b_scratch,64,192*k);
```
martin-frbg: rather than replacing this, turning it into
```
#if defined(OS_WINDOWS)
 ... _aligned_malloc... (note the added undescore and "M") 
#else
... posix_memalign ...
#endif
```
would probably take care of platform issues ?

wjc404: Thank you very much for the suggestions. 
Later I found caching part of blocked B matrix into b_scratch actually makes the program slow down a bit... so no need for aligned allocation.

wjc404:
A 4x8 kernel is added for direct interface to icopy_4, which gives improved 1-thread performance (85 GFLOPS) on Intel Xeon Platinum 8269CY (theoretical 105 GFLOPS, MKL 95 GFLOPS).

martin-frbg:
Great, thanks. Just let me know when you think it is ready for merging. I started to write a simple 8x8 trmm kernel earlier but could not get that to work, not sure if I made a silly mistake or if the existing generic kernels I used as guidance are broken. (This project has accumulated a few kernels of questionable status that are not in use by any target.)

wjc404:
I've finished updating the 2 kernel files. They will soon be ready for merging after I finish reliability tests on them.

wjc404:
For the 2 kernels, 1-thread tests with all combinations of 1<=m<=1173 && 1<=n<=85 && 1<=k<=267 passed. OK to merge.
Screenshot from 2019-10-20 17-30-41
Screenshot from 2019-10-20 10-55-06
Testing program is attached here:
[dgemmtest.zip](https://github.com/xianyi/OpenBLAS/files/3747678/dgemmtest.zip)

PR #2288, opened 2019-10-22
NetBSD fix

gufe44:


martin-frbg:
Looks good to me, why did you close it ?

gufe44:
There are some issues when using a fortran compiler other than gfortran (broken for g95 due to name mangling), utest does not complete unless libgfortran/libg95 is manually added to LD_LIBRARY_PATH. This is on NetBSD version 9, x86_64.

martin-frbg:
Probably would need something like the 
https://github.com/xianyi/OpenBLAS/blob/eaa0be131365d60c5953fcb62d92a34d794b29f3/Makefile.system#L784

we have for gfortran (quite possible that building with g95 has not been tried recently)

gufe44:
How would that fix it? The problem is rather that utest misses an rpath or environment that allows finding the shared object at runtime. As for the other issue, the problem exists with some names that include an underscore. Those are exported with two trailing underscores.

``linktest.c:(.text.startup+0x3296): undefined reference to `slasyf_rk_'``

martin-frbg:
Looks like f_check made some wrong guesses then (FEXTRALIB setting written to Makefile.conf, NEED2UNDERSCORES written to config.h )

martin-frbg:
Seems the NetBSD `ld.so` has only /usr/lib hardcoded as its search patch, and adding entries to /etc/ld.so.conf appears to be discouraged, however the g95 package buries the library somewhere under /usr/pkg/lib/gcc-lib. Apparently the executable is expected to contain a suitable rpath, but that linker flag is either not  reported by g95 or not picked up by f_check.

martin-frbg:
And the underscores issue appears to be a fundamental problem within  LAPACKE - right now I do not see if/how the netlib reference implementation copes with a fortran compiler that adds a second trailing underscore only to function names that contain an embedded underscore. 
All the NEED2UNDERSCORES setting from config.h does is tell gensymbol to fix up such names in the linktest file that is created to check the completeness of the (shared) library. (Current develop branch does have a long-standing error there as the new functions from netlib 3.7.0 and 3.8.0 were all added to the general object list rather than the separate one for objects with an embedded underscore. )  But this does not address any lapack calls made from lapacke work functions, and the lapacke(mangling) headers seem to take an all-or-nothing approach when it comes to adding underscores in the LAPACK_GLOBAL macro.
Edit: g95 understands -fno-second-underscore to drop the odd behaviour, we could enforce this
in Makefile.system (at least when NO_LAPACKE is defined). Also f_check only needs to accept -R as an alternate spelling of -rpath to fix the dynamic linking problems.

martin-frbg:
Would you like to resurrect this so the commit gets credited to you ? I believe I have fixed all related issues with g95 and rpath now.

gufe44:
Great, thank you, the issue persists with gfortran however:

`./openblas_utest: Shared object "libgfortran.so.5" not found`

Also, I seem to have lost the fork and can't reopen, feel free to commit the changes yourself. I could also open a new PR if that's better.

Edit: In https://github.com/xianyi/OpenBLAS/issues/2287#issuecomment-542558094 you expressed the user or downstream is responsible for making sure the library is found. To clarify, why does this not apply here?

martin-frbg:
Works for me (with PR #2292 that was not yet merged yesterday). 
In #2287 I was assuming some sort of misconfiguration, the library not being installed where the compiler/linker assumed it would be. I was not aware that some implementations of ld specify the
rpath in a form that was not recognized by the f_check script (-R), until I installed NetBSD in a VM to debug "your" problem. Also most if not all Linux systems and at least some of the BSDs will use ldconfig and ld.so.conf to keep a list of library locations, something that appears to be frowned upon by the NetBSD crew.
 But Ubuntu will most likely be using both /etc/ld.so.conf and a GNU ld which uses -rpath, so this normally should have worked. The problem there is probably that the entire toolchain was located in the user's home directory rather than a system location (and tianlanli had mentioned that only later). The rpath reported by "gfortran -v ftest2.f"  was probably relative to the GNU default prefix of  /usr/local rather than /home/tianlanli/gcc.

gufe44:
Does this help?

```
Driving: gfortran -v ftest2.f -l gfortran -l m -shared-libgcc
Using built-in specs.
COLLECT_GCC=gfortran
COLLECT_LTO_WRAPPER=/usr/pkg/gcc8/libexec/gcc/x86_64--netbsd/8.3.0/lto-wrapper
Target: x86_64--netbsd
Configured with: ../gcc-8.3.0/configure --disable-libstdcxx-pch --with-system-zlib --enable-nls --with-libiconv-prefix=/usr --enable-__cxa_atexit --with-gxx-include-dir=/usr/pkg/gcc8/include/c++/ --disable-libssp --enable-languages='c obj-c++ objc fortran c++' --enable-shared --enable-long-long --with-local-prefix=/usr/pkg/gcc8 --enable-threads=posix --with-boot-ldflags='-static-libstdc++ -static-libgcc -Wl,-R/usr/pkg/lib ' --with-gnu-ld --with-ld=/usr/bin/ld --with-gnu-as --with-as=/usr/bin/as --with-arch=nocona --with-tune=nocona --with-fpmath=sse --prefix=/usr/pkg/gcc8 --build=x86_64--netbsd --host=x86_64--netbsd --infodir=/usr/pkg/gcc8/info --mandir=/usr/pkg/gcc8/man --enable-option-checking=yes
Thread model: posix
gcc version 8.3.0 (GCC)
COLLECT_GCC_OPTIONS='-v' '-shared-libgcc' '-mtune=nocona' '-march=nocona'
 /usr/pkg/gcc8/libexec/gcc/x86_64--netbsd/8.3.0/f951 ftest2.f -ffixed-form -quiet -dumpbase ftest2.f -mtune=nocona -march=nocona -auxbase ftest2 -version -fintrinsic-modules-path /usr/pkg/gcc8/lib/gcc/x86_64--netbsd/8.3.0/finclude -o /var/tmp//cctTcVm0.s
GNU Fortran (GCC) version 8.3.0 (x86_64--netbsd)
        compiled by GNU C version 8.3.0, GMP version 6.1.2, MPFR version 4.0.2, MPC version 1.1.0, isl version isl-0.16.1-GMP

GGC heuristics: --param ggc-min-expand=100 --param ggc-min-heapsize=131072
GNU Fortran2008 (GCC) version 8.3.0 (x86_64--netbsd)
        compiled by GNU C version 8.3.0, GMP version 6.1.2, MPFR version 4.0.2, MPC version 1.1.0, isl version isl-0.16.1-GMP

GGC heuristics: --param ggc-min-expand=100 --param ggc-min-heapsize=131072
COLLECT_GCC_OPTIONS='-v' '-shared-libgcc' '-mtune=nocona' '-march=nocona'
 /usr/bin/as -v -o /var/tmp//ccoFCUnu.o /var/tmp//cctTcVm0.s
GNU assembler version 2.31.1 (x86_64--netbsd) using BFD version (NetBSD Binutils nb1) 2.31.1
Reading specs from /usr/pkg/gcc8/lib/gcc/x86_64--netbsd/8.3.0/../../../libgfortran.spec
rename spec lib to liborig
COLLECT_GCC_OPTIONS='-v' '-shared-libgcc' '-mtune=nocona' '-march=nocona'
COMPILER_PATH=/usr/pkg/gcc8/libexec/gcc/x86_64--netbsd/8.3.0/:/usr/pkg/gcc8/libexec/gcc/x86_64--netbsd/8.3.0/:/usr/pkg/gcc8/libexec/gcc/x86_64--netbsd/:/usr/pkg/gcc8/lib/gcc/x86_64--netbsd/8.3.0/:/usr/pkg/gcc8/lib/gcc/x86_64--netbsd/
LIBRARY_PATH=/usr/pkg/gcc8/lib/gcc/x86_64--netbsd/8.3.0/:/usr/pkg/gcc8/lib/gcc/x86_64--netbsd/8.3.0/../../../:/lib/:/usr/lib/
COLLECT_GCC_OPTIONS='-v' '-shared-libgcc' '-mtune=nocona' '-march=nocona'
 /usr/pkg/gcc8/libexec/gcc/x86_64--netbsd/8.3.0/collect2 -plugin /usr/pkg/gcc8/libexec/gcc/x86_64--netbsd/8.3.0/liblto_plugin.so -plugin-opt=/usr/pkg/gcc8/libexec/gcc/x86_64--netbsd/8.3.0/lto-wrapper -plugin-opt=-fresolution=/var/tmp//ccwEYiI8.res -plugin-opt=-pass-through=-lgcc_s -plugin-opt=-pass-through=-lgcc -plugin-opt=-pass-through=-lquadmath -plugin-opt=-pass-through=-lm -plugin-opt=-pass-through=-lgcc_s -plugin-opt=-pass-through=-lgcc -plugin-opt=-pass-through=-lc -plugin-opt=-pass-through=-lgcc_s -plugin-opt=-pass-through=-lgcc --eh-frame-hdr -dc -dp -e _start -dynamic-linker /usr/libexec/ld.elf_so /usr/lib/crt0.o /usr/lib/crti.o /usr/pkg/gcc8/lib/gcc/x86_64--netbsd/8.3.0/crtbegin.o -L/usr/pkg/gcc8/lib/gcc/x86_64--netbsd/8.3.0 -L/usr/pkg/gcc8/lib/gcc/x86_64--netbsd/8.3.0/../../.. -R /usr/pkg/gcc8//lib/. -R /usr/pkg/gcc8/lib/. /var/tmp//ccoFCUnu.o -lgfortran -lm -lgcc_s -lgcc -lquadmath -lm -lgcc_s -lgcc -lc -lgcc_s -lgcc /usr/pkg/gcc8/lib/gcc/x86_64--netbsd/8.3.0/crtend.o /usr/lib/crtn.o
COLLECT_GCC_OPTIONS='-v' '-shared-libgcc' '-mtune=nocona' '-march=nocona'
```

```
F_COMPILER=GFORTRAN
FC=gfortran
BU=_
FEXTRALIB=-Wl,-rpath,/usr/pkg/lib -L/usr/pkg/gcc8/lib/gcc/x86_64--netbsd/8.3.0 -L/usr/pkg/gcc8/lib/gcc/x86_64--netbsd/8.3.0/../../.. -Wl,-rpath, -Wl,-rpath,  -lgfortran -lm -lquadmath -lm -lc
```

Edit: Is this worth considering?

> [Using -R has the advantage of working in older versions of NetBSD as well.](https://www.netbsd.org/docs/elf.html#elf-rpath)

martin-frbg:
Thanks. Apparently it is the space between the -R and the /usr/pkg/gcc8 rpath that makes it fail - I had tried with netbsd 8.1 and gfortran5 as I did not manage to install any packages with my initial
netbsd9.  (Lack of experience with netbsd, and I am not that great when it comes to regex either)

martin-frbg:
I fixed the regex in https://github.com/xianyi/OpenBLAS/commit/aeabe0a83fffce9ab43ab8d10795e1696574887c
so that it accepts -R with and without spaces before the path argument.
(simply change the silly    `$link =~ s/\-R+/\-rpath\@/g;` ("R+" would mean one or more "R" characters, which plainly makes no sense at all) in f_check to   `$link =~ s/\-R\s*/\-rpath\@/g;`)

PR #2296, opened 2019-10-28
Fixed a minor cmake problem, occuring when DYNAMIC_ARCH=ON and CMAKE_C_FLAGS was empty

kdunee:
Without the quotes, `${CMAKE_C_FLAGS}` could resolve to nothing, resulting in the following error:

```
CMake Error at cmake/arch.cmake:76 (string):
  string sub-command REGEX, mode REPLACE needs at least 6 arguments total to
  command.
```


martin-frbg:
Good catch, thanks.

PR #2313, opened 2019-11-20
Fix the integer overflow issue for large matrix size

ewanglong:
Signed-off-by: Wang, Long <long1.wang@intel.com>

We found and fixed this issue. Please help review it.

### DIFF
--- a/kernel/x86_64/sgemm_kernel_16x4_skylakex.c
+++ b/kernel/x86_64/sgemm_kernel_16x4_skylakex.c
@@ -1215,7 +1215,7 @@ CNAME(BLASLONG m, BLASLONG n, BLASLONG k, float alpha, float * __restrict A, flo
 
 int sgemm_kernel_direct_performant(BLASLONG M, BLASLONG N, BLASLONG K)
 {
-	int mnk = M * N * K;
+	unsigned long mnk = M * N * K;
 	/* large matrixes -> not performant */
 	if (mnk >= 28 * 512 * 512)
 		return 0;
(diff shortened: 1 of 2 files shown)


wjc404:
What about changing to "unsigned long long" which is at least 64-bit ?

ewanglong:
> What about changing to "unsigned long long" which is at least 64-bit ?

Thanks for review. I computed 64-bit data can handle matrix computing about 3 millions x 3 millions x 3 millions. It should be enough. Is there any real use case to process thus huge size matrix over it?

ewanglong:
> What about changing to "unsigned long long" which is at least 64-bit ?

Thanks for review. For x86 platform, unsigned long is 64 bit length. It's enough to handle large matrix computing reaching 3 millions x 3 millions x 3 millions.

ewanglong:
> > What about changing to "unsigned long long" which is at least 64-bit ?
> 
> Thanks for review. For x86 platform, unsigned long is 64 bit length. It's enough to handle large matrix computing reaching 3 millions x 3 millions x 3 millions.

Below is quick checking data length by attached code. 

Signed    : Size                    Min                    Max
--------------------------------------------------------------
char      :    1                   -128                    127
short     :    2                 -32768                  32767
int       :    4            -2147483648             2147483647
long      :    8   -9223372036854775808    9223372036854775807
long long :    8   -9223372036854775808    9223372036854775807

Unsigned  : Size                    Min                    Max
--------------------------------------------------------------
char      :    1                      0                    255
short     :    2                      0                  65535
int       :    4                      0             4294967295
long      :    8                      0   18446744073709551615
long long :    8                      0   18446744073709551615

Miscellaneous sizes:
--------------------------------------------------------------
Single precision float:                                      4
Double precision float:                                      8
size_t:                                                      8
ssize_t:                                                     8

wjc404:
(I don't know whether the answer below is correct)
Screenshot_20191120_181026_com huawei browser

martin-frbg:
32bit operating systems would currently see SkylakeX kernels replaced by Nehalem as fallback so it would not be an issue there. However Microsoft still appears to use just 4 bytes for "unsigned long" so "unsigned long long" (or uint64_t) would appear to be the safest solution.

ewanglong:
"unsigned long long" was committed.

martin-frbg:
Thanks for finding and fixing this

Issue #307, opened 2013-10-15
Openblas hangs in Windows


Maxxan:
I don't know if there is a way to free OpenBLAS if it's used in a DLL, but when I load a DLL that uses OpenBLAS , and then tries to free it, the FreeLibrary function hangs (deadlock?) and never returns.

I use pre-built package downloaded from http://sourceforge.net/projects/openblas/files/v0.2.8/
compiling the test program with Visual Studio 2010.

Here is code to reproduce the issue:
## Dummy DLL using OpenBLAS, does nothing:
# include "stdafx.h"
# include "cblas.h"

BOOL APIENTRY DllMain( HMODULE hModule, DWORD  ul_reason_for_call, LPVOID lpReserved)
{
    return TRUE;
}

void FunctionNeverCalled()
{
   float f = 1;
   cblas_sasum(1, &f, 1);
}

---
## Program calling the DLL:
# include "stdafx.h"
# include "Windows.h"

int _tmain(int argc, _TCHAR\* argv[])
{
   HINSTANCE hInstance = ::LoadLibraryExW(L"..\Debug\DllUsingOpenBlas.dll", NULL, LOAD_WITH_ALTERED_SEARCH_PATH);
   ::FreeLibrary(hInstance); // Hangs on FreeLibrary
    return 0;
}

PR #2314, opened 2019-11-20
Fix usage of TerminateThread() causing critical section corruption.

Jehan:
So here is the deal: we have crashes in GIMP on Windows where a specific security software is used. We got in touch with the publisher (through a user of both GIMP and this software) and they told us the problem does not come from GIMP, but from OpenBlas (which is an indirect dependency of GIMP through SuiteSparse) with the technical explanation which I gave in the commit message.

They also told us a similar problem was discussed here: https://stackoverflow.com/questions/39635817/windows-10-specific-crash-on-call-leavecriticalsection

So for some reason, they don't want to discuss this publicly or whatever, which is why I am doing the intermediary and why I don't even cite their name or the software's name, which is stupid — I know — as contributing to Free Software is to their credit. But whatever apparently they want to keep their contribution quiet.

------------------------------------------

This patch was submitted to the GIMP project by a publisher wishing to
keep confidentiality (hence anonymously). I just pass along the patch.
Here is the patch explanation which came with:

First they remind us what Microsoft documentation says about
TerminateThread:
> TerminateThread is a dangerous function that should only be used in
> the most extreme cases. You should call TerminateThread only if you
> know exactly what the target thread is doing, and you control all of
> the code that the target thread could possibly be running at the time
> of the termination.
(https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-terminatethread)

Then they say that 5 milliseconds time-out might not be long enough for
the thread to exit gracefully. They propose to set it to a much higher
value (for instance here 5 seconds).

And finally you should always check the return value of
WaitForSingleObject(). In particular you want to run TerminateThread()
only if WaitForSingleObject() failed, not on success case.

### DIFF
--- a/driver/others/blas_server_win32.c
+++ b/driver/others/blas_server_win32.c
@@ -462,11 +462,15 @@ int BLASFUNC(blas_thread_shutdown)(void){
 
     for(i = 0; i < blas_num_threads - 1; i++){
       // Could also just use WaitForMultipleObjects
-      WaitForSingleObject(blas_threads[i], 5);  //INFINITE);
+      DWORD wait_thread_value = WaitForSingleObject(blas_threads[i], 5000);
+
 #ifndef OS_WINDOWSSTORE
-// TerminateThread is only available with WINAPI_DESKTOP and WINAPI_SYSTEM not WINAPI_APP in UWP
-      TerminateThread(blas_threads[i],0);
+      // TerminateThread is only available with WINAPI_DESKTOP and WINAPI_SYSTEM not WINAPI_APP in UWP
+      if (WAIT_OBJECT_0 != wait_thread_value) {
+        TerminateThread(blas_threads[i],0);
+      }
 #endif
+
       CloseHandle(blas_threads[i]);
     }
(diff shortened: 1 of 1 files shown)


martin-frbg:
Interesting, this must have been fun to debug. The code in question seems to be mostly unchanged from GotoBLAS so may have appeared to do the right thing on XP or 7. Perhaps the earlier commit that changed the INFINITE wait to five milliseconds to solve a hang #307 was unknowingly papering over a side effect of this bug

Jehan:
> Perhaps the earlier commit that changed the INFINITE wait to five milliseconds to solve a hang #307

Yeah the developer who contributed this said ideally we should keep as infinite (as was in comment). But we all know that idealistic code doesn't exist, and infinite wait means potential blocking execution. So this person went for a longer time-out instead. Makes sense to me.

Jehan:
Anyway we hope this gets merged soon. This issue has quite a lot of people complaining about GIMP crashing. ;-(

PR #2315, opened 2019-11-21
revised fix windows compatible for #2313

ewanglong:
Signed-off-by: Wang, Long <long1.wang@intel.com>


martin-frbg:
Thanks. I should have looked closer at the previous diff as well

PR #2316, opened 2019-11-21
zarch: treat z15 as z14 instead of generic

sharkcz:


martin-frbg:
Thanks

martin-frbg:
BTW I added cpuid 8562 for the smaller model on top of your patch.

sharkcz:
> BTW I added cpuid 8562 for the smaller model on top of your patch.

sure, I missed that it has been announced too

martin-frbg:
I stumbled on an older QEMU patch thread where both numbers were revealed by someone with an ibm.com address. (Amusingly this triggered a discussion as the official marketing name had not been revealed at the time, or perhaps not even decided on)

PR #2317, opened 2019-11-21
Change bad usage of "asum" to "sum" in ZARCH versions of ?sum

aarnez:
The ZARCH implementations of ?sum contain a cut & paste-error: An inline
assembly argument is named "sum", but the assembly references "asum"
instead.  The mismatch causes a build error.  This is fixed.


martin-frbg:
Thanks 
Got another PR for this a moment later, did I miss a Z15 giveaway or something ? :)

aarnez:
> Got another PR for this a moment later, did I miss a Z15 giveaway or something ? :)
Yeah, I saw the other pull request with exactly the same patch as well.  Coincidence, it seems.
I'm not aware of a z15 giveaway, maybe I missed it ;-)

PR #2318, opened 2019-11-21
Correct Inline Assembly name mismatches (ZArch)

aubindetrez:
Solve compilation issue: "asum is not known" (Tested with GCC9.0 on s390x)


martin-frbg:
Thanks but it seems aarnez beat you by about an hour (#2317) :)

PR #2324, opened 2019-11-27
Fix SEGV in cdot_power9

antonblanchard:
We were corrupting r2 because the local entry wasn't being
setup correctly.

### DIFF
--- a/kernel/power/cdot_power9.S
+++ b/kernel/power/cdot_power9.S
@@ -13,10 +13,7 @@
 	
 cdot_k:
 .LCF0:
-0:	addis 2,12,.TOC.-.LCF0@ha
-	addi 2,2,.TOC.-.LCF0@l
-	.localentry	cdot_k,.-cdot_k
-	mr. 9,3
+0:	mr. 9,3
 	ble 0,.L10
 	cmpdi 7,5,1
 	beq 7,.L18


martin-frbg:
What I still do not quite understand here - when I run gcc -s on the cdot.c that would normally have been used, I get exactly the same instructions that your PR deletes. This is with gcc 9.2.1, binutils 2.32 on a Fedora Rawhide Power9 instance provided by OpenPower (unicamp). As you will certainly know your toys, am I doing something else wrong here or does this suggest a bug in etiher gcc or as ? (The release notes of binutis 2.33.1 do not mention any ppc changes, and I do not identify anything related among Alan Modra's recent commits in the git log on sourceware)

antonblanchard:
I think the issue is that you use the file multiple times without taking into consideration the ELF local entry point. An example, looking at readelf output:

This is correct, we have a local entry point 8 bytes (2 instructions) into the function:

```   111: 0000000000006580   572 NOTYPE  LOCAL  DEFAULT [<localentry>: 8]    12 cdot_k```

This is incorrect, there is no local entry point:

```   196: 0000000000006580     0 FUNC    GLOBAL DEFAULT   12 cdotc_k```

I presume you need to use the actual function name in the .localentry line, but I'd suggest just removing them completely.

antonblanchard:
Note that you wont see this SEGV if you are building with NOFORTRAN=1, because we get lucky and don't switch TOCs in the library.

martin-frbg:
Ah right, thanks, that would need to be `ifdef CONJ` for cdot/cdotc (and caxpy/caxpyc) as well then if we were to keep it. (There is another reference to cdot in a .size statement near L16)

PR #2338, opened 2019-12-09
Changes to build on AIX in POWER8 mode

kavanabhat:
This fix includes the following:

1) Macro changes so that it works with m4 macro preprocessor
2) Makefile changes to invoke m4 for macro processing
3) Use of macros for Power8 extended mnemonics so that it expands to basic mnemonics on AIX. Some of the Power8 Extended mnemonics are not supported by the AIX assembler.

### DIFF
--- a/common_power.h
+++ b/common_power.h
@@ -39,6 +39,35 @@
 #ifndef COMMON_POWER
 #define COMMON_POWER
 
+#define str(x)	#x
+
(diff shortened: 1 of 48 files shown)


martin-frbg:
Thank you very much - I believe this fixes #1997 ?

kavanabhat:
Yes, this fixes #1997.

PR #2340, opened 2019-12-16
[WIP] Use arm neon instructions to optimize gemm beta operation

Zeyiii:
This optimization can improve performance of  gemm_beta operation nearly twice. PTAL.

### DIFF
--- a/kernel/arm64/KERNEL
+++ b/kernel/arm64/KERNEL
@@ -34,7 +34,7 @@ ifndef SGEMM_BETA
 SGEMM_BETA = ../generic/gemm_beta.c
 endif
 ifndef DGEMM_BETA
-DGEMM_BETA = ../generic/gemm_beta.c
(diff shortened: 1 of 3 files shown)


brada4 (changes requested):
Guide says NEON is optional, so generic kernel should not assume it as default. It is present in all other real cores though

martin-frbg:
I believe the DGEMM_BETA could/should be declared in KERNEL.ARMV8 rather than the generic KERNEL. (IIRC neon extensions are already used in at least some of the optimized arm64 kernels)

Zeyiii:
> I believe the DGEMM_BETA could/should be declared in KERNEL.ARMV8 rather than the generic KERNEL. (IIRC neon extensions are already used in at least some of the optimized arm64 kernels)

You're right. I have corrected it.

martin-frbg:
Thanks

PR #2344, opened 2019-12-21
Optimize AVX2 ZGEMM

wjc404:
The current ZGEMM kernel for Haswell uses KERNEL_4x3 to carry most calculations, which reads packed matrix A (in L3 cache) 1/3 faster than packed matrix B (in L1 cache). One day I realized that the implementation of KERNEL_4x6 is possible with current i/o copy routines, which can slow down reading on packed A. A new ZGEMM kernel is written for the purpose. The 1-thread performance rose 7-10% after the change.

1-thread performance in GFLOPS (TARGET=HASWELL):
on r7-3700x at 3.6 GHz (MKL_DEBUG_CPU_TYPE=5): old kernel 12.2, new kernel 13.4, MKL2018 13.6, theoretical 14.4;
on Xeon Gold 6149 (MKL_CBWR="AVX2"): old kernel 10.5, new kernel 11.2, MKL2018 11.4, theoretical 12.4;
on i9-9900K (avx-offset=6): old kernel 15.0, new kernel 16.8, MKL2018 16.8, theoretical 17.6.


wjc404:
The new kernel passed reliability test.
Screenshot from 2019-12-25 18-49-30

PR #2346, opened 2019-12-25
LAPACK: avoid out-of-bound write in ?LANTR

vladimir-ch:
See https://github.com/Reference-LAPACK/lapack/pull/378 for background.

Note that there are other differences in LANTR routines that would go away after updating to 3.9.0 (or latest master) of netlib lapack.


martin-frbg:
Thanks, I already saw your PR and intend to import Reference-LAPACK master branch ASAP (next - i.e. this upcoming - weekend most likely).

vladimir-ch:
Sure, no problem, this isn't urgent and can wait.

Issue #1737, opened 2018-08-14
Dynamic runtime CPU detection for non-x86 architectures


susilehtola:
Hi,

what are your plans for dynamic CPU detection on non-x86 architectures? It would be great to have it, as it would enable efficient versions on linux distributions...

martin-frbg:
No concrete plans as far as I am aware, but patches welcome of course. Which architecture(s) do you have in mind ? A suggestion for implementing the runtime detection on ARM was posted quite non-recently in #989

susilehtola:
All of them? :laughing:

sharkcz:
/me agrees ;-) I'm looking into it, but no promises.

RashmicaG:
I was thinking of doing this for the POWER architecture.

Do you think its reasonable to assume that the glibc version on the system OpenBLAS is running on will be the same or higher than the system it was built on?

I was thinking of something like this:

```
int cpu_is_power8() 
{
#ifdef __BUILTIN_CPU_SUPPORTS__
        return __builtin_cpu_is("power8");
#elif ((__GLIBC__ > 2) || (__GLIBC__ == 2 && __GLIBC_MINOR__ >= 16))
        unsigned long v = getauxval( AT_PLATFORM );
        return !strcmp("power8", (char *)v);
#else
        return detect();
#endif
}

static FLOAT ddot_kernel_8(....)
{
        if (cpu_is_power9())
                //do power9 optimised code
        elif (cpu_is_power8())
                //do power8 optimised code
        else
                // default implementation
}

```
What are your thoughts @sharkcz? I'm not sure if there's a better way to determine the existence of getauxval()...

dnabanita7:
Hello! I am very new to this community. I want to learn and work on creating a dynamic cpu on s390x in OpenBLAS. Can I know where to get started? 
Thank you.

martin-frbg:
Hi, the corresponding work for PowerPC (#2084) can probably be used as a guideline. Basically, you would need to create driver/others/dynamic_zarch.c with some reliable runtime detection of the cpu (can be what is used in cpuid_zarch.c at build time, can be something else) and a little code to translate between the generic names of functions (sgemm) and their cpu-specific implementations (like sgemm_kernel_Z13). Do you have access to s390x hardware for testing ?

dnabanita7:
[...]

PR #2349, opened 2019-12-29
add in runtime cpu detection for zarch

nk521:
https://github.com/xianyi/OpenBLAS/issues/1737


nk521:
Thanks @martin-frbg for merging 😄

PR #2351, opened 2019-12-30
prefetching for dgemm_beta

Zeyiii:
PTAL.


Zeyiii:
@martin-frbg This patch can improve performance of dgemm_beta by about 20%.

wjc404:
Interesting observation.  Normally the performance of gemm_beta is limited by memory bandwidth where prefetch can have little effect.

martin-frbg:
Not sure I see any speedup on my (low-end) test hardware, but it does not appear to hurt either...

PR #2354, opened 2019-12-31
[WIP] Use arm neon instructions to optimize tcopy operation

ZuoQ3:
This optimization can improve performance of tcopy operation nearly twice. PTAL.


martin-frbg:
LGTM, thanks.

PR #2357, opened 2020-01-02
kernel/arm64/dgemm_beta.S: add beta == zero branch

chenxuqiang:
added beta == zero branch, and no need to load C matrix.

Signed by: Xuqiang Chen <chenxuqiang3@hisilicon.com>


martin-frbg:
Thanks

PR #2358, opened 2020-01-03
Test all 7 declared values of N in float and double cblas3 tests

shengyang-3390:
the number fit enumerated values of N


martin-frbg:
Thanks -  I had actually overlooked that you removed that unrelated part from #2356 by the time I merged it yesterday. (The higher N here is probably a leftover from private tests of one of the earlier developers - the corresponding tests in the "netlib" Reference-LAPACK project only have N=6)

PR #2367, opened 2020-01-13
Improve paralleled SGEMM performance on SKYLAKEX CPUs

wjc404:
Increase the parameter GEMM_Q from 192 to 320, to reduce the bandwidth required for reading and writing matrix C in parallel executions, thus alleviating pressure on main memory & LLC.
The kernel codes were also modified to improve performance (reduce branching, optimize loop unrolling and calculations with m<4, etc).
The performance with 8 threads on i7 9800x rose about 5% after this change (93% MKL2019 to 98% MKL2019).
In addition, the "direct sgemm" codes were extracted to a single file to reduce the total code size of sgemm kernel.


wjc404:
An 1-thread SGEMM test (with "direct sgemm" disabled) have passed with the new version of OpenBLAS.
Screenshot from 2020-01-14 13-40-06
(since I switch the CPU and mobo without reinstalling Ubuntu, the host name still contains "Z390 M GAMING")

PR #2372, opened 2020-01-19
Do not run any cleanup if the program is exiting anyway

martin-frbg:
From keno's PR #2350 - this avoids the potential hang in blas_thread_shutdown where we may wait for threads to exit while they are waiting on the loader lock from DllMain

### DIFF
--- a/exports/dllinit.c
+++ b/exports/dllinit.c
@@ -50,7 +50,10 @@ BOOL APIENTRY DllMain(HINSTANCE hInst, DWORD reason, LPVOID reserved) {
         gotoblas_init();
         break;
       case DLL_PROCESS_DETACH:
-        gotoblas_quit();
+        // If the process is about to exit, don't bother releasing any resources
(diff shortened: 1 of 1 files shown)


Keno:
We continued seeing hangs even after this patch and after some additional investigation, this patch turns out to be insufficient. We have another copy of DllMain here: https://github.com/xianyi/OpenBLAS/blob/develop/driver/others/memory.c#L1551 (as well as one more in that file), as well as a bunch of CRT tricks. In addition gotoblas_quit is declared as `DESTRUCTOR` when built with gcc on windows, so that'll run it yet again. Ideally all those instances in memory.c need to be removed. Also that file basically duplicates all functions in it. It would be nice to clean that up a bit.

martin-frbg:
Yes memory.c is messy, it is basically two versions of the same file rolled into one - owing to the bad experience with the TLS patches. And I suspect the DllMain there will only be invoked in static builds.

PR #2373, opened 2020-01-20
Optimize genenal Gemm Beta

Qiyu8:
make a small change, have many advantanges as below:
1. The code amount is reduced by one third.
2.The performance is improved by 5%, constant caculation moved out of the loop.
3.Tested in arm and x86 archtect.
4.East to maintain and understand.
5.provide a guide for assemble optimizition. I believe that there is room for further improment to dgemm_beta.S and sgemm_beta.S.


martin-frbg:
Thanks

PR #2375, opened 2020-01-22
fix a few performance drop in some matrix size per data type

ewanglong:
Signed-off-by: Wang,Long <long1.wang@intel.com>

Add this patch to fix a few performance drop in some matrix size. It was caused by the inappropriate GEMM_PREFERED_SIZE value(=32), which set the imbalance workload split per thread. 
For example 1280x1280 single precision matrix in 16 threads. It was split into 8 threads with 96 width single precision(32 bits)size and 8 threads with 64 width. But for avx512 alignment, the 16 threads with 80 width single precision(32 bits)size is also acceptable and more important in such case is that the workload is also balanced in each thread.

### DIFF
--- a/param.h
+++ b/param.h
@@ -1507,8 +1507,13 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #define SYMV_P  8
 
-#define SWITCH_RATIO	32
-#define GEMM_PREFERED_SIZE	16
+#if defined(XDOUBLE) || defined(DOUBLE)
+#define SWITCH_RATIO            4
+#define GEMM_PREFERED_SIZE      4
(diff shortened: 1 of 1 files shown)


TiborGY:
This is applied to all CPU types, right? You mentioned that this improves performance on AVX-512 CPUs, but what about AVX2 CPUs? (Haswell, Zen2, etc.)

martin-frbg:
@TiborGY just Haswell and SkylakeX (it was added by Arjan van de Ven in 5b708e5e ). I have not merged it yet as I did not have time to run any benchmarks but I assume someone with an Intel email will have thought of that ?

martin-frbg:
The gains seem small on my hardware (but at least I do not see a slowdown), would you happen to have a few numbers from your 16-thread case ?

PR #2376, opened 2020-01-22
Fix remaining bugs in parallel GEMM3M

wjc404:
BTW some thresholds were adjusted to improve performance.

### DIFF
--- a/driver/level3/level3_gemm3m_thread.c
+++ b/driver/level3/level3_gemm3m_thread.c
@@ -104,7 +104,7 @@ typedef struct {
 #define BETA_OPERATION(M_FROM, M_TO, N_FROM, N_TO, BETA, C, LDC) \
 	GEMM_BETA((M_TO) - (M_FROM), (N_TO - N_FROM), 0, \
 		  BETA[0], BETA[1], NULL, 0, NULL, 0, \
-		  (FLOAT *)(C) + (M_FROM) + (N_FROM) * (LDC) * COMPSIZE, LDC)
+		  (FLOAT *)(C) + ((M_FROM) + (N_FROM) * (LDC)) * COMPSIZE, LDC)
 #endif
 
 #ifndef ICOPYB_OPERATION
@@ -414,7 +414,7 @@ static int inner_thread(blas_arg_t *args, BLASLONG *range_m, BLASLONG *range_n, 
 
       for(jjs = xxx; jjs < MIN(n_to, xxx + div_n); jjs += min_jj){
 	min_jj = MIN(n_to, xxx + div_n) - jjs;
-	if (min_jj > GEMM3M_UNROLL_N) min_jj = GEMM3M_UNROLL_N;
+	if (min_jj > GEMM3M_UNROLL_N*3) min_jj = GEMM3M_UNROLL_N*3;
 
 	START_RPCC();
 
@@ -550,7 +550,7 @@ static int inner_thread(blas_arg_t *args, BLASLONG *range_m, BLASLONG *range_n, 
 
       for(jjs = xxx; jjs < MIN(n_to, xxx + div_n); jjs += min_jj){
 	min_jj = MIN(n_to, xxx + div_n) - jjs;
-	if (min_jj > GEMM3M_UNROLL_N) min_jj = GEMM3M_UNROLL_N;
+	if (min_jj > GEMM3M_UNROLL_N*3) min_jj = GEMM3M_UNROLL_N*3;
 
 	START_RPCC();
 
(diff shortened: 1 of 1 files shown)


martin-frbg:
Thanks. I think some files use both thresholds (if min > 3*X min=3*X else if min >X min=X), perhaps not worth doing that here ?

wjc404:
.. I'm sorry I didn't make myself clear for the change of thresholds (as well as that in PR #2348 , where I observed performance improvements after the change).  I need several paragraphs to do this... 
..The previous upper bound of min_jj (which is GEMM_UNROLL_N) may not be a problem with SSE or AVX builds, but can slow down AVX2-based kernels. 
..An efficient AVX2 gemm kernel requires usage of 10-13 ymm registers as accumulators (full usage of 2 fma units with latency ≥ 5cyc requires 10 registers, meanwhile at least 3 ymm registers are needed to hold the loaded elements from matrix A & B), to hold m x n elements of C being calculated(m and n should be at least 2, because the throughput of load instructions at run time can be lower than that of fma instructions on modern CPUs)(m should not be greater than n to avoid unbalanced reading rates on packed matrices).
..For single precision, possible combinations of m x n are 2x40, 2x48, 3x32, 4x24, 5x16, 6x16, 8x10 and 8x12. For double precision, possible combinations of m x n are 2x20, 2x24, 3x16, 4x12, 5x8 and 6x8. In all, m and n cannot become powers of 2 at the same time, which is required for traditional implementations where m=GEMM_UNROLL_M and n=GEMM_UNROLL_N. 
..There're 3 ways to deal with the inconsistency:
..(1). Use new value for GEMM_UNROLL_M or GEMM_UNROLL_N which is not a power of 2.  This requests the corresponding i/o copy routines for many level3 functions to be implemented(therefore a huge amount of debugging work is needed). 
..(2). Maintain old values and do icopy/ocopy transformation at run time. For example, the dgemm 4x8 kernel by Werner Saar does ocopy8 -> ocopy12 transformations to adapt ocopy_8 data to kernel_4x12. The transformation (by copying to a buffer) requires extra time. 
..(3). Read data from several neighboring tiles(3 or 5 tiles) at the same time. For example, when GEMM_UNROLL_M=GEMM_UNROLL_N=4, read 3 neighboring tiles (each has 4 columns) from packed B and 1 tile (4 rows) from packed A in the calculation of 4 x 12 elements of C is possible(i.e. fit to 4x12 kernel). Since maintaining sequential reading from packed matrix A (in lower level cache) is more important for performance than that from matrix B, concurrent reading of neighboring tiles of B is preferred, when n is 3 or 5 times of the width of a tile (GEMM_UNROLL_N). The current S/C/ZGEMM and C/ZGEMM3M AVX2 kernels adopt this approach. 
..The minimal dimension of N (which is min_jj in the driver code) for running the AVX2 GEMM kernel efficiently with the 3 approaches above are: 
(1) GEMM_UNROLL_N; 
(2) the least common multiple of GEMM_UNROLL_N and n, which is 3 times of GEMM_UNROLL_N in the current implementation of DGEMM; 
(3) n, which is 3 times of GEMM_UNROLL_N in current implementations of s/c/z GEMM and c/z GEMM3M.
..Perhaps in the future when a set of copy functions with unroll times like 6, 12(for avx2) and 24(for avx512) are available, the upper limit of min_jj in GEMM and GEMM3M driver codes will no longer need to be 3* GEMM_UNROLL_N.

wjc404:
The change of threshold doesn't have any effect on the performance of SSE-based OpenBLAS on i3-m380 with 2 threads.
(min_jj <= GEMM_UNROLL_N in OpenBLAS-xianyi; min_jj <= GEMM_UNROLL_N * 3 in OpenBLAS-wjc404)

cgemm3m (theoretically 13.3 equivalent cgemm GFLOPS):
Screenshot from 2020-01-23 10-47-33
Screenshot from 2020-01-23 10-47-39

zgemm3m (theoretically 6.7 equivalent cgemm GFLOPS):
Screenshot from 2020-01-23 11-12-47
Screenshot from 2020-01-23 11-12-36

However, raising the threshold do improve parallel GEMM3M performance of AVX2-based OpenBLAS, on Xeon Platinum 8163 (boost 2.7 GHz) with 4 threads (MKL_CBWR="AVX2", TARGET=HASWELL).

cgemm3m (theoretically 115.2 equivalent cgemm GFLOPS):
Screenshot from 2020-01-23 17-11-34
Screenshot from 2020-01-23 17-11-49

zgemm3m (theoretically 57.6 equivalent zgemm GFLOPS):
Screenshot from 2020-01-23 17-18-19
Screenshot from 2020-01-23 17-17-58

martin-frbg:
Thank you very much for the detailed explanation. (I feel this should really be part of the - currently almost non-existent - "developer manual" in the wiki)

PR #2389, opened 2020-02-05
Fix bugs in benchmark of gemv

Zeyiii:
These bugs cause access to uninitialized memory. Accessing uninitialized memory is not the same as accessing it with initialization, which can cause gemv performance differences.PTAL.


Zeyiii:
@martin-frbg  PTAL.

martin-frbg:
LGTM, thanks.

PR #2405, opened 2020-02-11
Fix typo in dynamic_zarch.c

susilehtola:


martin-frbg:
Err, right. :) Thanks

PR #2420, opened 2020-02-15
Correct generation of GETRF files by the CMAKE build

### LINKED ISSUE #2396: dgesv produces incorrect answer when compiling the library using CMake with USE_OPENMP=1

martin-frbg:
fixes #2396

### DIFF
--- a/lapack/CMakeLists.txt
+++ b/lapack/CMakeLists.txt
@@ -4,7 +4,6 @@ include_directories(${PROJECT_BINARY_DIR})
 
 
 set(LAPACK_SOURCES
-  getrf/getrf_single.c
   potrf/potrf_U_single.c
   potrf/potrf_L_single.c
   lauum/lauum_U_single.c
@@ -45,6 +44,10 @@ GenerateNamedObjects("${LAPACK_MANGLED_SOURCES}" "" "" false "" "" false 3)
 GenerateNamedObjects("laswp/generic/laswp_k_4.c" "" "laswp_plus" false "" ""  false 3)
 GenerateNamedObjects("laswp/generic/laswp_k_4.c" "MINUS" "laswp_minus" false "" ""  false 3)
 
+foreach (float_type ${FLOAT_TYPES})
+GenerateNamedObjects("getrf/getrf_single.c" "UNIT" "getrf_single" false "" "" false ${float_type})
+endforeach ()
+
 # dynamic_arch laswp needs arch specific code ?
 #foreach(TARGET_CORE ${DYNAMIC_CORE})
 #      set(TSUFFIX "_${TARGET_CORE}")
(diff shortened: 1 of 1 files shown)


marxin:
You found it, good job! It seems to me a more common problem and I have one suggestion. Instead of:
```
#define MAXPS	maxps
#define MAXSS	maxss
#ifdef USE_MIN
#define MAXPS	minps
#define MAXSS	minss
#endif
```

I would rather use:
```
#if defined(USE_MAX)
#define MAXPS	maxps
#define MAXSS	maxss
#elif defined(USE_MIN)
#define MAXPS	minps
#define MAXSS	minss
#else
#error Neither USE_MAX nor USE_MIN is defined
#endif
```

It will take some time to change it but it may pay off in the future?

martin-frbg:
Err, that comment belongs to the iamax_sse.S of #2414 or what ? I do not see a relevance to the -DUNIT problem here (where the #define plays a role in the included common.h) and neither will anybody who may try to understand code changes through reading the associated git history a few months from now.

marxin:
No, it's related to this issue as well. It's about a missing macro definition. In the case of `USE_MIN`, if you forget to define it somewhere you end up with a wrong code. With my suggestion, you will end up with a compilation error. That seems to me much nicer than a silent miscompilation.

PR #2425, opened 2020-02-17
Use -DUSE_MAX macro instead of -UUSE_MIN.

marxin:
As described here: https://github.com/xianyi/OpenBLAS/pull/2420#issuecomment-586644380
I'm suggesting to come up with more robust macro usage. So far, I've converted `USE_MIN` macro usage. The patch exposes one more issue in the current CMake files:

```
GenerateNamedObjects("${KERNELDIR}/${I${float_char}MINKERNEL}" "" "i*min_k" false "" "" false ${float_type})
```
there's missing `USE_MIN` macro.


marxin:
@martin-frbg What do you think about the suggested approach?

martin-frbg:
I'm not yet convinced that it is worth the trouble to change lots of existing files (with the risk of missing one), just to replace a "negative" define with the equivalent "positive" one. For new files, your suggestion to `#error` out if none of the possible options was seen is certainly superior but I feel that retroactively changing the logic will just make diffs harder to follow.

marxin:
Alright, I understand that the changes will be quite invasive.

PR #2427, opened 2020-02-17
Fix ISMIN and ISMAX kernel choices for POWER8

martin-frbg:
resolves ppc utest failures in the newly added ismin test. Note it is actually the defaults in Makefile.L1 that need to be corrected if this passes all tests.


marxin:
The tests look fine but you forgot to rebase the branch on top of the current master.

martin-frbg:
Huh ? I merged isuruf's PR in the meantime but that should not conflict. What am i missing ?

marxin:
> Huh ? I merged isuruf's PR in the meantime but that should not conflict. What am i missing ?

I see in this pull request linked 6 commits, but it's likely irrelevant.

martin-frbg:
Should be. Besides the actual rebases there was a mistaken commit of the KERNEL file to the develop branch instead of the newly created one, and its reversal. And I tried the imin_ppc440.S as well before setting on imin.S (Guess squash&merge might have looked cleaner, but as all changes affected the same file the result should be the same)

PR #2361, opened 2020-01-06
Optimize AVX2 SGEMM & STRMM

wjc404:
Replace KERNEL_16x6 with KERNEL_8x12 to slow down reading on packed matrix A(in L3 cache), as mentioned in issue #2210 .  The performance catchs up with MKL2019 after the change.

PR #2422, opened 2020-02-16
Adjust SkylakeX GEMM3M parameters, add an AVX512 STRMM kernel and fix performance bugs in AVX2 s/c/z GEMM

wjc404:
PS: I forgot to change SKX GEMM3M parameters when I added those kernels 2 month ago, sorry for that.  I fix it in this PR.

PR #2384, opened 2020-02-03
Optimize AVX512 DGEMM (& DTRMM)

wjc404:
Replace KERNEL_8x24 with KERNEL_16x12, which makes more room for raising GEMM_Q to slow down reading/writing on matrix C (in main memory), thus improves parallel performance.

PR #2430, opened 2020-02-18
Restore ZEN SGEMM speed after #2361.

marxin:
I partially reverted the changes in #2361 and I received the following
speed up on:
./xsl3blastst -R gemm -N 2048 2048 1 -a 5 1 1 1 1 1

AMD Ryzen 7 2700X (Zen+): 61400 to 63300 MFlops
AMD EPYC 7742 (Zen v2): 91400 to 94500 MFlops

These numbers are single-threaded performance.


wjc404:
@marxin Thanks for the information. Previously I tested the effect of #2361 on Ryzen 7 3700X (Ubuntu 19.04, 2ch DDR4-2133 (latency ~90ns, read ~30GB/s, write ~25GB/s), FCLK 1.6 GHz, turbo boost disabled (3.6 GHz base clock), mobo ab350m-gaming 3). With 1 thread the SGEMM performance (m=n=k=5999, transa=N and transb=N) rose from 103 GFLOPS to 108 GFLOPS after that PR.
  I'm curious about performance changes at dimensions greater than 5000.

TiborGY:
@marxin What is the memory clock in these two systems you have measured on?

marxin:
```
AMD Ryzen 7 2700X (Zen+):
8GiB DIMM DDR4 Synchronous Unbuffered (Unregistered) 1200 MHz (0.8 ns)

AMD EPYC 7742 (Zen v2):
16GiB DIMM DDR4 Synchronous Registered (Buffered) 2667 MHz (0.4 ns)
```

marxin:
And for `AMD Ryzen 7 2700X` I get speed improvement from 60500 to 63500.
Memory info:
```
16GiB DIMM DDR4 Synchronous Unbuffered (Unregistered) 2400 MHz (0.4 ns)
```

and I've also tried bigger matrices (4096) I get the same MFlops.

wjc404:
A SGEMM test on an 1c/1t KVM (tencent cloud SA2 instance), on an EPYC ROME processor:
[testlog_sgemm_epyc_rome.txt](https://github.com/xianyi/OpenBLAS/files/4220628/testlog_sgemm_epyc_rome.txt)
Screenshot_20200219_010903_com sonelli juicessh

wjc404:
A SGEMM test on an 1c/1t KVM (tencent cloud SA1 instance), on an EPYC NAPLES processor:
[testlog_sgemm_epyc_naples.txt](https://github.com/xianyi/OpenBLAS/files/4220916/testlog_sgemm_epyc_naples.txt)
Screenshot_20200219_021101_com sonelli juicessh

wjc404:
An another SGEMM test on the ROME processor I mentioned above (along with a typo, lol):
[testlog_sgemm_epyc_rome_2.txt](https://github.com/xianyi/OpenBLAS/files/4221339/testlog_sgemm_epyc_rome_2.txt)

Screenshot_20200219_033802_com sonelli juicessh

wjc404:
@marxin A kind reminder: if you're eager to turn back to the original 16x4 SGEMM kernel, please change SGEMM_DEFAULT_UNROLL_M (in param.h) to 16 and update SGEMM INCOPY & ITCOPY code paths in "KERNEL.ZEN", or the calculation will go off the rails.

wjc404:
@marxin I confirmed that in some cases with special dimensions (2048 & 4096) the performance drops. I'll try to figure out the cause and make a patch. Below is the terminal log of some SGEMM tests on an EPYC Rome processor.
[terminal_log_sgemm_tencent_sa2_nn.txt](https://github.com/xianyi/OpenBLAS/files/4222888/terminal_log_sgemm_tencent_sa2_nn.txt)

wjc404:
An update: the performance bug occurs when m is close to a multiple of 1024 (diff <3). It's an issue related to 32kB 8-way set associative L1D cache when >8 cache lines from a 8x12 block of matrix C move towards the same set (which is the case when m%1024==0) causing a conflict. I'll try to find a way to alleviate this problem. I have to admit that this bug is generic (it also occurred on a Cascade-Lake CPU taking away ~4% FLOPS, note that even MKL2019 with MKL_CBWR="AVX2" can neither get rid of that) and it also occurs in AVX512 SGEMM kernel (and in the AVX2 DGEMM kernel of OpenBLAS 0.3.7).  I'm sorry for that.
@marxin Thanks again for reporting the bug.

wjc404:
I just found that the AVX512 DGEMM kernel ( #2384 ) has the same problem, even the AVX512 DGEMM of MKL2019 has it too. 
[testlog_aliyun_avx512_serial_dgemm.txt](https://github.com/xianyi/OpenBLAS/files/4225829/testlog_aliyun_avx512_serial_dgemm.txt)

marxin:
> @marxin A kind reminder: if you're eager to turn back to the original 16x4 SGEMM kernel, please change SGEMM_DEFAULT_UNROLL_M (in param.h) to 16 and update SGEMM INCOPY & ITCOPY code paths in "KERNEL.ZEN", or the calculation will go off the rails.

Sure, the PR is just a kick off for discussion. I bet you'll come up with a better fix ;)

marxin:
> An update: the performance bug occurs when m is close to a multiple of 1024 (diff <3). It's an issue related to 32kB 8-way set associative L1D cache when >8 cache lines from a 8x12 block of 

Nice observation, can you see it with a L1D perf event? I can't for `L1-dcache-load-misses` perf event.

I can confirm your observation with the numbers close to a power of 1024:
```
   N: MFLOPS
2000: 65770.2
2048: 61781.2
2047: 62193.3
2046: 64012.4
2045: 65210.6
2044: 65828.2
```

> matrix C move towards the same set (which is the case when m%1024==0) causing a conflict. I'll try to find a way to alleviate this problem. I have to admit that this bug is generic and it also occurs in AVX512 SGEMM kernel (and in the AVX2 DGEMM kernel of OpenBLAS 0.3.7). I'm sorry for that.
> @marxin Thanks again for reporting the bug.

I'm happy we can both reproduce that.

wjc404:
I can't access my r7-3700x now because of 2019-nCoV.  Maybe Ryzen has a PMC counter of CPU cycles stall on L1D (I'm not sure, you can check it with libpfm4) that could indicate the bug. A PMC counter for store misses may also help.

marxin:
I should see it in `L1-dcache-load-misses` counter. You can try to find an event with `perf list` or do you see an event in AMD manual:
https://developer.amd.com/wp-content/resources/55803_0.54-PUB.pdf
?

wjc404:
I've worked out a new kernel that can reduce the influence of this bug by spreading the storage operations of each m8n12 block of C. 
[sgemm_kernel_8x4_haswell_2.c.txt](https://github.com/xianyi/OpenBLAS/files/4232231/sgemm_kernel_8x4_haswell_2.c.txt)
Currently only for SGEMM (no STRMM kernels), still have room for improvement. Any suggestions are welcome.

marxin:
Nice, Should I close this PR and will you create a separate one where you'll send the suggested change?

martin-frbg:
@marxin I believe that change is now part of #2422

PR #2436, opened 2020-02-19
Improve test coverage for utests.

marxin:


martin-frbg:
Not sure ismax and ismin really need to be separate files when they are primarily "variations on a theme", same for min and max ? And your Makefile change would drop the test_amax that you are adding the damax check to... also these would need to be added to utest_main2.c (which is the all-in-one version for platforms that do not support building the list of ctest function pointers on the fly)

marxin:
Ok, I've just updated the patch to reflect your comments.

martin-frbg:
Thank you. Happily, this time there were no errors exposed by the added tests.

PR #2443, opened 2020-02-24
[OpenBlas]:benchmark/copy.c has time，x,y data loop problems

aaawuanjun:
Problem 1、time accurate to Microseconds, when data amount is lower than 10000， the final result is not precise，and has large fluctuations in results.
so, change time accurate to Nanoseconds.

Problem 2、the x, y datas are changed when COPY every loops, this also results in large fluctuations.
So, change random get X, Y data out of "for (l=0; l<loops; l++)" loops.

### DIFF
--- a/benchmark/copy.c
+++ b/benchmark/copy.c
@@ -129,7 +129,10 @@ int main(int argc, char *argv[]){
   int step =   1;
 
   struct timeval start, stop;
-  double time1,timeg;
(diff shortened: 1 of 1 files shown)


martin-frbg:
Indeed, thanks. I suspect several of these old benchmarks that were created by wernsaar some five years ago will share the same problems.

PR #2451, opened 2020-02-27
Solve the problem of core dumped when using large-scale data in bench…

Darkness303:
…mark test

E.g ：  when running test calse such as below in benchmark:
./dgesv.goto 100000 100000 100000
From : 100000  To : 100000 Step = 100000 Trans = 'N' Inc_x = 1 Inc_y = 1 Loops = 1
   SIZE       Flops
 100000 : Segmentation fault (core dumped)
Because i+j*m has exceeded the maximum range of int

### DIFF
--- a/benchmark/gesv.c
+++ b/benchmark/gesv.c
@@ -177,20 +177,20 @@ int main(int argc, char *argv[]){
 
     for(j = 0; j < m; j++){
       for(i = 0; i < m * COMPSIZE; i++){
-	a[i + j * m * COMPSIZE] = ((FLOAT) rand() / (FLOAT) RAND_MAX) - 0.5;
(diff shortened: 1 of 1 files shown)


martin-frbg:
Right, these benchmarks were created several years ago and mostly in the context of desktop x86_64 hardware. As your changes are all basically about the same topic, perhaps it would make sense to bundle them instead of creating a separate PR for each single file you touch ?

martin-frbg:
BTW if you build OpenBLAS with -DINTERFACE64 , which you will eventually need to do when handling large arrays, the "blasint" in the benchmarks will automatically become BLASLONG (i.e. long) which should take care of this overflow.

xianyi:
@martin-frbg agree with you.

PR #2460, opened 2020-02-27
Solve the problem of core dumped when using large-scale data in bench…

Darkness303:
…mark test

E.g ：  when running test calse such as below in benchmark:
./ssyrk.goto 100000 100000 100000
From : 100000  To : 100000 Step = 100000 Uplo = U Trans = N
   SIZE       Flops
 100000 : Segmentation fault (core dumped)
Because i+j*m has exceeded the maximum range of int

image

### DIFF
--- a/benchmark/syrk.c
+++ b/benchmark/syrk.c
@@ -172,8 +172,8 @@ int main(int argc, char *argv[]){
 
     for(j = 0; j < m; j++){
       for(i = 0; i < m * COMPSIZE; i++){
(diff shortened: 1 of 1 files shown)


martin-frbg:
Did you see my comments on #2451 ? This is getting a bit ridiculous I think...

Darkness303:
sorry  i just see it

Darkness303:
sorry  i just see it

PR #2464, opened 2020-02-28
Add syr benchmark

Darkness303:
A syr benchmark was added because that is missing in benchmark, at the same time, the corresponding makefile file be modified .


martin-frbg:
Thanks

PR #2465, opened 2020-02-29
Add Neoverse-N1 core

AGSaidi:
The implementation is a hybird of the ARMV8 one with some of the
improved TX2 rountines along with specifying -march=v8.2-a


martin-frbg:
Thanks.

PR #2466, opened 2020-02-29
Switch blas_server to use acq/rel semantics

AGSaidi:
Heavy-weight locking isn't required to pass the work queue
pointer between threads and simple atomic acquire/release
semantics can be used instead. This is especially important as
pthread_mutex_lock() isn't fair.

We've observed substantial variation in runtime because of the
the unfairness of these locks which complety goes away with
this implementation.

The locks themselves are left to provide a portable way for
idling threads to sleep/wakeup after many unsuccessful iterations
waiting.

### DIFF
--- a/driver/others/blas_server.c
+++ b/driver/others/blas_server.c
@@ -572,7 +572,7 @@ int blas_thread_init(void){
       thread_status[i].status = THREAD_STATUS_WAKEUP;
 
       pthread_mutex_init(&thread_status[i].lock, NULL);
-      pthread_cond_init (&thread_status[i].wakeup, NULL)
+      pthread_cond_init (&thread_status[i].wakeup, NULL);
 
 #ifdef NEED_STACKATTR
       ret=pthread_create(&blas_threads[i], &attr,
--- a/driver/others/blas_server.c
+++ b/driver/others/blas_server.c
@@ -571,6 +571,9 @@ int blas_thread_init(void){
       atomic_store_queue(&thread_status[i].queue, (blas_queue_t *)0);
       thread_status[i].status = THREAD_STATUS_WAKEUP;
 
+      pthread_mutex_init(&thread_status[i].lock, NULL);
+      pthread_cond_init (&thread_status[i].wakeup, NULL)
+
 #ifdef NEED_STACKATTR
       ret=pthread_create(&blas_threads[i], &attr,
 		     &blas_thread_server, (void *)i);
--- a/driver/others/blas_server.c
+++ b/driver/others/blas_server.c
@@ -140,6 +140,16 @@ typedef struct {
 
 } thread_status_t;
 
+#if (__STDC_VERSION__ >= 201112L)
+#define	atomic_load_queue(p)		__atomic_load_n(p, __ATOMIC_RELAXED)
+#define	atomic_store_queue(p, v)	__atomic_store_n(p, v, __ATOMIC_RELAXED)
+#else
+#define	atomic_load_queue(p)		(blas_queue_t*)(*(volatile blas_queue_t**)(p))
+#define	atomic_store_queue(p, v)	(*(volatile blas_queue_t* volatile*)(p) = (v))
+#endif
+
+
+
 static thread_status_t thread_status[MAX_CPU_NUMBER] __attribute__((aligned(ATTRIBUTE_SIZE)));
 
 #ifndef THREAD_TIMEOUT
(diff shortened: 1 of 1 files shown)


martin-frbg:
Unfortunately this appears to increase the risk of a hang in the "fork safety" utest. (One process stuck in "wait" from test_fork.c:123, the other in the libpthread __lll_lock_wait due to getting signalled in blas_thread_shutdown by the fork handler. (It is entirely possible that your PR only unmasked a pre-existing bug by making things run faster)

AGSaidi:
What system/core count are you seeing this on? How many runs of utest are required to trigger it?

martin-frbg:
Ah, sorry. Haswell target with just 2 cores/4 threads (older Kaby Lake laptop), much to my surprise it appears to be reproducible on every run. (Compiling with gcc 9). Could possibly be a glibc bug, this is an older system with 2.22.

martin-frbg:
Seems the original process is still wait()ing for the forked child to exit, while in the child the originally active thread has exited and the three others appear to be stuck on the pthread_cond_signal(&thread_status[i].wakeup) in the blas_thread_shutdown procedure at line 949 of blas_server.c 
EDIT: sorry, one stuck on line 949 the other two at the pthread_cond_wait of line 332.

AGSaidi:
I reworked the code a little bit to test for C11. If it's present it uses __atomic... if not it uses a volatile pointer to try to convince the compiler to read it. 

I can't seem to replicate the hang. I've tried Ubuntu 19.04, 20.4, and Amazon Linux 2 and I've tried with 2c/4t and 18c/36t Skylake-based systems running openblas_utest 1000 times.

martin-frbg:
Hang seems to have caused the travis failures as well, probably need to make this depend on some yet to be determined value of __GLIBC_PREREQ as well.

AGSaidi:
@martin-frbg I did manage to get this to reproduce on CentOS6. I'll try to debug it.

AGSaidi:
What seems to be hanging is the nested fork in test_fork.c:104. Running with only two threads and `set follow-fork-mode child` in gdb:

```
(gdb) info thread
* 3 Thread 0x7ffff7ff2700 (LWP 27891)  0x00007ffff7949334 in __lll_lock_wait () from /lib64/libpthread.so.0
(gdb) bt
#0  0x00007ffff7949334 in __lll_lock_wait () from /lib64/libpthread.so.0
#1  0x00007ffff7946f81 in pthread_cond_signal@@GLIBC_2.3.2 ()
   from /lib64/libpthread.so.0
#2  0x0000000000459da5 in blas_thread_shutdown_ () at blas_server.c:963
#3  0x00007ffff7361d45 in fork () from /lib64/libc.so.6
#4  0x000000000040923d in __ctest_fork_safety_run () at test_fork.c:104
```

And the pthread_cond_t struct looks weird, particularly -1 for the total_seq. 
```
(gdb) p thread_status[i].wakeup
$2 = {
  __data = {
    __lock = 2,
    __futex = 0,
    __total_seq = 18446744073709551615,
    __wakeup_seq = 0,
    __woken_seq = 0,
    __mutex = 0x0,
    __nwaiters = 0,
    __broadcast_seq = 0
  },
  __size = "\002\000\000\000\000\000\000\000\377\377\377\377\377\377\377\377", '\000' <repeats 31 times>,
  __align = 2
}
```

martin-frbg:
Yes, this is what I am seeing as well, but I do not know what to make of it - I'd like to think of this as a long-fixed glibc bug but cannot find anything remotely matching in the glibc bugzilla. The bad part of course is that the original GotoBLAS has never been threadsafe (big surprise for me when I started to use OpenBLAS) and my sledgehammer approach to locking everything down was guided by partial knowledge more than anything else. Added to that the fork handler has always been a fragile kludge.

driver/others/blas_server.c
```diff
@@ -566,12 +568,9 @@ int blas_thread_init(void){
 
     for(i = 0; i < blas_num_threads - 1; i++){
 
-      thread_status[i].queue  = (blas_queue_t *)NULL;
+      atomic_store_queue(&thread_status[i].queue, (blas_queue_t *)0);
       thread_status[i].status = THREAD_STATUS_WAKEUP;
 
-      pthread_mutex_init(&thread_status[i].lock, NULL);
-      pthread_cond_init (&thread_status[i].wakeup, NULL);
```
geoffreyblake: @AGSaidi, `thread_status.lock` and `thread_status.wakeup` are now being used uninitialized in the code.  This may explain the odd garbage values seen and the cause of the hang.

martin-frbg: Good catch, don't know how often I have looked over that diff without spotting it. Sanity is indeed restored by restoring those two lines.

martin-frbg: Applied the changes to confirm this fixes the CI failures as well...

AGSaidi: Thanks @geoffreyblake. I too don't understand how I missed them

PR #2468, opened 2020-02-29
Use wait-for-event to not spin in the blas_lock

AGSaidi:


xianyi:
Hi, @AGSaidi @martin-frbg 

Do you test the performance of this lock on ARMv8 server?

Some users report the bad performance on ARMv8 SMP servers because of the lock overhead.

Xianyi

AGSaidi:
Do you have any more information you can share about the problem people are having with it? This is a pretty common way to do spin locks on arm systems. The lock isn’t fair and the load/store exclusives aren’t fair either. To make it fair we’d have to use a ticket lock or detect that the system supported LSE and use a swap instruction

martin-frbg:
Pretty sure I saw nothing suspicious before merging, but my only way to test big ARMv8 SMP is with the Drone.io CI which is probably far from ideal concerning overhead variability.

AGSaidi:
TravisCI now supports running on graviton2 hosts, so we should add a job to do that if that works for you @martin-frbg 

I tried the patch previously on a system and didn’t see an issue but I’m happy to try again.

martin-frbg:
@AGSaidi thanks for the hint, I have modified the travis setup in my fork, will copy the PR if and when I got the options right.

AGSaidi:
There are some caveats on which Ubuntu versions are supported [this blog](https://aws.amazon.com/blogs/opensource/getting-started-with-travis-ci-com-on-aws-graviton2/) describes some of those. If you want to use a VM,  Ubuntu focal is required.

Issue #2363, opened 2020-01-08
"FATAL ERROR" when testing "./xdcblat3 < din3"


njutcz:
"FATAL ERROR" when testing "./xdcblat3 < din3"
![error](https://user-images.githubusercontent.com/55725787/71945134-e4e7b280-3200-11ea-8932-c628a8cc2f3e.PNG)
Environment: ARMV8 CentOS 7.6, OpenBLAS-0.3.7
compile: make TARGET=ARMV8 -j

cd ctest
mycount=0; while (( $mycount < 100 )); do ./xdcblat3 < din3;sleep 1; ((mycount=$mycount+1)); done;

The error will occur about every 7 or 8 times! And when "export OMP_NUM_THREADS=1", the error doesn't occur! I'm not sure if this is a precision issue or a calculation error!

wjc404:
What about the test results of cblas_dgemm? It should have appeared above the results of cblas_dsymm.

njutcz:
Only cblas_dsymm has the error. Cblas_dgemm never occurs the same error!

martin-frbg:
What kind of hardware is this, how many cores and which compiler version ? I cannot reproduce this problem so far.

brada4:
There was something similar, solution that worked was to rpmbuild Fedora's SRPM/SPEC that likely has some compiler quirks included.
You can start here (or seek SRPM in rawhide repos)
https://src.fedoraproject.org/rpms/openblas/blob/HEAD/f/openblas.spec

njutcz:
> What kind of hardware is this, how many cores and which compiler version ? I cannot reproduce this problem so far.

kunpeng920 96cores gcc 4.8.5

njutcz:
> What kind of hardware is this, how many cores and which compiler version ? I cannot reproduce this problem so far.

When I build with “USE_SIMPLE_THREADED_LEVEL3=1”， the ERROR doesn't occur. But i don't know why.

martin-frbg:
Thanks for the additional information. Building with USE_SIMPLE_THREADED_LEVEL3=1 disables some additional splitting of the workload that could still contain race conditions - though I do not get any error messages from running valgrind/helgrind on the xdcblat3 test. 
[...]

PR #2469, opened 2020-02-29
Use acq/rel semantics to pass flags/pointers in getrf_parallel.

AGSaidi:
The current implementation has locks, but the locks each only
have a critical section of one variable so atomic reads/writes
with barriers can be used to achieve the same behavior.

Like the previous patch, pthread_mutex_lock isn't fair, so in a
tight loop the previous thread that has the lock can keep it
starving another thread, even if that thread is about to write
the data that will stop the current thread from spinning.

On a 64c Arm system this reduces variability by 20x on sgesv.goto.

### DIFF
--- a/lapack/getrf/getrf_parallel.c
+++ b/lapack/getrf/getrf_parallel.c
@@ -68,22 +68,13 @@ double sqrt(double);
 #define GETRF_FACTOR 1.00
 
 
-#if   defined(USE_PTHREAD_LOCK)
-static pthread_mutex_t    getrf_lock = PTHREAD_MUTEX_INITIALIZER;
-#elif defined(USE_PTHREAD_SPINLOCK)
-static pthread_spinlock_t getrf_lock = 0;
+#if (__STDC_VERSION__ >= 201112L)
+#define	atomic_load_long(p)		__atomic_load_n(p, __ATOMIC_RELAXED)
+#define	atomic_store_long(p, v)		__atomic_store_n(p, v, __ATOMIC_RELAXED)
 #else
-static BLASULONG  getrf_lock = 0UL;
+#define	atomic_load_long(p)		(BLASLONG)(*(volatile BLASLONG*)(p))
+#define	atomic_store_long(p, v)		(*(volatile BLASLONG *)(p)) = (v)
 #endif
-
-#if   defined(USE_PTHREAD_LOCK)
-static pthread_mutex_t    getrf_flag_lock = PTHREAD_MUTEX_INITIALIZER;
-#elif defined(USE_PTHREAD_SPINLOCK)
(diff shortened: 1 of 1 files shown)


martin-frbg:
I must admit I am a bit confused about the use of barriers vs. locks on ARMV8 - could you comment on #2363 where a memory barrier was seen to be insufficient due to the weak memory on that platform ? 
(Sorry, unintentionally closed your PR due to touchpad malfunction)

AGSaidi:
@martin-frbg Is there an official list of the compilers/versions that are supported? Given that manylinux1 is there, while I think it's clearer, this pre-dates support of the c11 atomics. While I think the acq/rel functions make it clearer, if that is requirement I'll probably re-work this to place the existing barrier types in the code in the correct places.

martin-frbg:
No concise list as such, just a few warnings in the readme for specific platforms. Unfortunately requiring specific minimum versions causes problems for non-priviledged users on sites that use (possibly older) "enterprise" distributions. So if it would be possible to enclose this in appropriate ifdefs (for __STDC_VERSION__ >= 201112L ) that would be preferable.

martin-frbg:
Looks like it's hitting a clang bug now. (I can make it compile by adding an explicit cast but need to convince myself that this is actually still the same thing)

AGSaidi:
I don't believe declaring these types as _Atomic is needed. _Atomic means that any operation on them is implicitly made atomic by the compiler (E.g. foo++ means  atomically read the current value of foo and increment it by one). If this works with just volatile in old compilers, we're not asking for those kinds of guarantees and the __atomic_* builtins don't require this either, "The ‘__atomic’ builtins can be used with any integral scalar or pointer type that is 1, 2, 4, or 8 bytes in length. " Removing the ifs declaring the types with _Atomic makes clang happy.

martin-frbg:
Alright thanks, so let's nuke the atomics then. BTW something in your force-push seems to have added a lot of unrelated changes that got committed to develop in the meantime. I assume git will manage to sort this out during a merge, but anybody trying to cherry-pick by downloading the current patch file would get a surprise. Have not seen this before, would you happen to know how to rectify it ?

AGSaidi:
Yea, this was my doing as I rebased on top of the develop branch. Let me see if I can undo it.

Seems like the #ifdef .. _Atomic should go universally?

AGSaidi:
Looks like the arm32 gcc build hung the utest fork test?

martin-frbg:
Looks like it's ready. Thanks a lot for your work on this.

AGSaidi:
Thanks @martin-frbg. Please feel free to ping me if you see other  similar issues, i'm happy to help.

PR #2478, opened 2020-03-02
Update benchmark statistical time function

MacChen02:
The function gettimeofday does not count the time，when testing the small data volume usecase of axpy.
old

Use the function clock_gettime to replace the gettimeofday function to count the time.

new


martin-frbg:
Thanks

PR #2479, opened 2020-03-02
Fix potential index overflows at large matrix sizes in the benchmark codes

Darkness303:
Hello!
First of all, i tried the way as you said before with DINTERFACE64=1 when i am building the Openblas,but it doesn't work in my Arm server (the core dumped error as the same before,it is still exits)，i have no ideal about that. secondly, it seems a bit inconvenient when users building the Openblas must be added the arg every time. so, all of the files have be modified and tested by ctest, the result show all passed. BTW, thanks for your suggestion bundle them as one PR. please review these changes , thanks.


martin-frbg:
Thanks.

PR #2484, opened 2020-03-03
Fix DYNAMIC_ARCH build for POWER9

RajalakshmiSR:
Setting DYNAMIC_ARCH=1 on POWER9 does not build POWER9 files due to some
compiler version checks.  This patch fixes some of the macros that are used
to check compiler version.  On fixing those checks, there are some new make
failures related to icamin, icamax, isamin, isamax and caxpy files on POWER9.
This patch fixes those failures as well.


martin-frbg:
Thanks (sorry, must have totally missed to push my assembler fixes for power9, guess technically these should also have _CALL_ELF==2 guards)

PR #2488, opened 2020-03-04
Modify the main Makefile in OpenBLAS 

liujingjue:
add c/fortran compiler version information in final note.
with these information , the user could know the exact verion for both c and fortran complier version which could avoid version conflict when there are more than two gcc tools in the complier environment.  especially the c complier using the command "cc" may lead to lots confusing actions while it links to the gcc the user don't intend to launch. the changed note the makefile show like this:

  OS                       ... Linux
  Architecture        ... arm64
  BINARY               ... 64bit
  C compiler          ... GCC  (cmd & version : cc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-39))
  Fortran compiler ... GFORTRAN  (cmd & version : GNU Fortran (GCC) 4.8.5 20150623 (Red Hat 4.8.5-39))
  Library Name      ... libopenblas_armv8p-r0.3.9.dev.a (Multi-threading; Max num-threads is 96)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,11 +56,21 @@ ifneq ($(INTERFACE64), 0)
 	@echo "  Use 64 bits int    (equivalent to \"-i8\" in Fortran)      "
 endif
 endif
-	@cverinfo=`$(CC) --version | sed -n '1p'`; \
-	echo "  C compiler       ... $(C_COMPILER)  (cmd & version : $${cverinfo})"
+	@$(CC) --version > /dev/null 2>&1;\
+	if [ $$? -eq 0 ]; then \
+	   cverinfo=`$(CC) --version | sed -n '1p'`; \
+	   echo "  C compiler       ... $(C_COMPILER)  (cmd & version : $${cverinfo})";\
+	else  \
+	   echo "  C compiler       ... $(C_COMPILER)  (command line : $(CC))";\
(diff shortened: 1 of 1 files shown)


martin-frbg:
Had been thinking about that as well, but can we be sure that `--version` returns something meaningful with all (non-gcc) compilers across the supported platforms ?

liujingjue:
once it is compiled with command line,  the compilers as far as I know all have the --version args and in fact almost every command (including "ls"/"sed"/"rm"/"touch")in the shell has the --version args. Or we can check the return value after --version executed , if the return code ==0 we can show the information otherwise we still show the origin one?

martin-frbg:
IBM xlc compiler for example seems to be using -qversion, with pgcc it is `-V` at least in older versions. So yes the result of this command would need to be checked, it would be a bit silly to crash out of the build process in the last second.

liujingjue:
add code to check the "--version" whether it is legal args for the compiler, it just show the version info once the --version args executed successfully. For the compiler which doesn't support the --version the users could add the branch for their platform specfic version args including the -qversion for IBM or -V for PGI you mentioned, at present the code just show the  command line like preview ones in this case

martin-frbg:
Thanks.

PR #2489, opened 2020-03-04
Remove redundant code

jijiwawa:
Remove redundant code in the benchmark directory

### DIFF
--- a/benchmark/hemm.c
+++ b/benchmark/hemm.c
@@ -178,8 +178,6 @@ int main(int argc, char *argv[]){
 
     time1 = (double)(stop.tv_sec - start.tv_sec) + (double)((stop.tv_usec - start.tv_usec)) * 1.e-6;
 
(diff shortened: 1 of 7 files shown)


martin-frbg:
That one supersedes #2487, right ? Pity I only realized this after deconflicting the older one. Need to check now that merging this would not throw out code from the others I guess...

martin-frbg:
It would probably be easier if you could just resubmit this based on the current develop and with only the redundant code removed.

jijiwawa:
> That one supersedes #2487, right ? 
yes

jijiwawa:
> It would probably be easier if you could just resubmit this based on the current develop and with only the redundant code removed.

Resubmit done

jijiwawa:
> That one supersedes #2487, right ? 

yes

martin-frbg:
Thank you.

PR #2498, opened 2020-03-09
Add benchmark for ?amax, ?max, ?amin, ?min, i?max, i?amin and i?min.

njutcz:
Add benchmark for ?amax, ?max, ?amin, ?min, i?max, i?amin， i?min and modify the Makefile.


njutcz:
When I call zamax, I can't get the right result. I found the MAXF was covered in every loop("fmaxp   MAXF, v1.2d"), so we can't get the right result. 
Solution: fmaxp   MAXF, v1.2d -> fmaxp   TMPF, v1.2d

xianyi:
@njutcz ，Thank you for the fix of zamax

PR #2504, opened 2020-03-11
Add benchmark test for SPMV and solve segmentation fault when data size >= 50000

xiaofengF:
Add benchmark test for SPMV and solve segmentation fault when data size >= 50000


martin-frbg:
Not sure what happened here, you may have uncovered an actual bug in the build process.

PR #2506, opened 2020-03-11
Add benchmark for SPMV and fix segmentation fault when data size >= 50000

xiaofengF:
Add benchmark for SPMV and fix segmentation fault when data size >= 50000

### DIFF
--- a/benchmark/Makefile
+++ b/benchmark/Makefile
@@ -117,7 +117,6 @@ acml :: slinpack.acml dlinpack.acml clinpack.acml zlinpack.acml \
        cherk.acml zherk.acml \
        cher2k.acml zher2k.acml \
        sgemv.acml dgemv.acml cgemv.acml zgemv.acml \
-	   sspmv.acml dspmv.acml cspmv.acml zspmv.acml \
        strmv.acml dtrmv.acml ctrmv.acml ztrmv.acml \
        strsv.acml dtrsv.acml ctrsv.acml ztrsv.acml \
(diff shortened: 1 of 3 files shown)


martin-frbg:
I'm beginning to suspect that it is actually a long-standing error in our version of the LAPACK Makefile. xSPMV is overloaded with an OpenBLAS implementation from driver/level2 so the original fortran version from the reference implementation needs to be excluded from the build.

xiaofengF:
> I'm beginning to suspect that it is actually a long-standing error in our version of the LAPACK Makefile. xSPMV is overloaded with an OpenBLAS implementation from driver/level2 so the original fortran version from the reference implementation needs to be excluded from the build.

But I can't understand why the fortran version doesn't affect others, but only the cspmv and zspmv.

xiaofengF:
> I'm beginning to suspect that it is actually a long-standing error in our version of the LAPACK Makefile. xSPMV is overloaded with an OpenBLAS implementation from driver/level2 so the original fortran version from the reference implementation needs to be excluded from the build.

 Maybe you can first pull the benchmark addition for sspmv and dspmv, because openblas only have s/d spmv in the cblas.h. c/z spmv is not that important.

martin-frbg:
Comparing with the reference implementation, it is indeed only SSPMV and DSPMV that are in BLAS (and consequently CBLAS), while CSPMV and ZSPMV are LAPACK auxiliaries that do not even have a LAPACKE interface for them. Probably some "historic" reason for this distinction...

PR #2508, opened 2020-03-13
[OpenBLAS]:fix the iamax benchmark error

liujingjue:
[Description]:the result for i?amax is not MFlops, it is MBytes

### DIFF
--- a/benchmark/iamax.c
+++ b/benchmark/iamax.c
@@ -181,7 +181,7 @@ int main(int argc, char *argv[]){
     timeg /= loops;
 
     fprintf(stderr,
-	    " %10.2f MFlops %10.6f sec\n",
+	    " %10.2f MBytes %10.6f sec\n",
 	    COMPSIZE * sizeof(FLOAT) * 1. * (double)m / timeg * 1.e-6, timeg);
 
   }


liujingjue:
the CI problem that miss the gfortran comand was fixed, and the i?amax code changed commit pass. 
iamax  benchmark test the show the capability of "MBytes" not "MFlops" which makes user confused because the number shown  usually passes  the peak of the calculation

martin-frbg:
Technically you are right, of course. But I wonder which of the two units is more confusing - probably the expectation is MFlops, and the number of bytes seems to be just an easy-to-obtain proxy value for the actual number of comparisons performed per unit time ?

liujingjue:
I think both "MFlops" and "Bytes"  are reasonable units to judge the performance once all your test cases just use the one of them. but there is a situation I encounter that this benchmark give me some confusion. since I know the peak MFlops value in my platform, this L1 benchmark just print the MFlops number greater than the theoritical limit(because of the extra sizeof(FLOAT) multiplication), which is impossible.  so I read the source code and found the print  is not rigorous to this case  and made this commit

martin-frbg:
Makes perfect sense, thanks

PR #2515, opened 2020-03-16
[OpenBLAS]: benchmark for her/her2 LEVEL2 functions

zelong-1024:
[description]: benchmark for her/her2
[solution]: added benchmark for her/her2, modified makefile in benchmark


martin-frbg:
thanks

PR #2520, opened 2020-03-20
Fix avx512 sgemm performance bug when ldc is a multiple of 1024

wjc404:
Previously the performance could drop 10% on a SKX CPU when ldc%1024 == 0 (causing 24 cache lines from matrix C moving towards the same set of L1D cache).  I rewrote the kernel to combine and split 2 consecutive m16n24 stores to 2 m32n12 ones. The performance drop at ldc%1024==0 reduced to 4% (similar to MKL2019) after this change. 

BTW some blocking parameters and micro kernels with m=8 or 4 were optimized.


wjc404:
The newly-compiled OpenBLAS library passed 1-thread SGEMM test.
[sgemm_testlog_20200321_aliyun.txt](https://github.com/xianyi/OpenBLAS/files/4362627/sgemm_testlog_20200321_aliyun.txt)
Screenshot from 2020-03-21 17-14-23

wjc404:
By the way I'm curious about the implementation of FUSED_GEMM_KERNEL functions. I saw the definitions of them in level3.c but I haven't found any kernel files related to them.

martin-frbg:
There are several such things in the code (QUAD_PRECISION comes to mind) that are just fragments of thoughts or plans Kazushige Goto had when he wrote the original GotoBLAS. Unfortunately there is no background information from that epoch beyond his publications and the scarce comments in the code - he does not appear to have used any kind of CVS, or if he did
its commit messages were never made public. All I can state with any certainty is that the same `ifdef FUSED_GEMM` without a matching implementation was already in GotoBLAS2 1.08 (the oldest GotoBLAS2 release I have kept, OpenBLAS started from the last available which was 1.13) but does not appear in the GotoBLAS (note no "2" here) that preceded it.

wjc404:
25-thread GEMM tests on a Cascade-lake XCC Die (on a KVM):
SGEMM(dim=36k) 2.88TFLOPS, 93% of MKL-2019;
DGEMM(dim=30k) 1.37TFLOPS, 92% of MKL-2019.

PR #2525, opened 2020-03-21
Fix ARCHCONFIG for Neoverse-N1

andreas-schwab:
../config_kernel.h:24:9: warning: missing whitespace after the macro name
   24 | #define ARMV8-march armv8.2-a
      |         ^~~~~


martin-frbg:
Ouch, thanks.

Issue #2524, opened 2020-03-21
reboots on haswell system


rgauges:
Hi,

for a few weeks I am experiencing funny behaviour which I think stems from openblas.
If I run machine learning workloads using numpy or tensorflow, my haswell system at home
reboots whenever I run certain tasks. This seems to depend on the environment I am running in.
When I set up a conda environment with numpy linked against mkl, the programs run just fine.
But when I run them in an environment where numpy is linked against openblas, my machine 
reboots.
As the machine immediately reboots, I do not see any error messages and I running it in a debugger
won't help either.
Running the code within a kvm virtual machine on the same host also reboots the host upon running the code.

All my other systems are not affected by this. So I first thought it was the CPU that is breaking down or overheating, but other CPU intensive workloads run just fine.

My system is:

Ubuntu 19.10 x64
Haswell i7-4790K
Different conda environments with numpy either from conda (mkl) or from pip (openblas). It's the pip installations that always make my machine reboot.

Any idea of how I could pinpoint this problem so I can send a useful bug report?

Thanks

Ralph

Diazonium:
This is extremely unlikely to be an OpenBLAS issue, you either have a hardware issue, bios setting issue or kernel issue.
Try running Prime 95 for a few hours, both with small and large FFTs, make sure you enable AVX.
You can also try running the Intel Linpack benckmark for a few hours, it uses MKL to perform an LU factorization and it is a good CPU+RAM stress test.

martin-frbg:
Anything in the system logs ? Normally a non-priviledged userspace program should never be able to bring down the kernel, though things might get a bit "interesting" if you run out of memory. Perhaps run memtest to make sure the hardware is healthy - though I'd expect you'd see similar failures with other software if memory or power supply had gone bad.

rgauges:
Thanks a lot for the fast response.
I also thought about thermal issues and memory issues first and I did run a memory check as well as several CPU stress tests. E.g. handbrake transcoding  or stress-ng. And even though CPU utilisation is at 100% on all cores for hours, the system runs stable.
Temperature goes up to 75°C after some time, but it takes a while and the system is not throttling or anything. So I guess, cooling is working.
I haven't run prime95 yet, so I will try that next, enabling AVX. But I guess Handbrake would have used AVX as well.

Yes, I also thought that user space programs shouldn't be able to crash the machine, but unfortunately it does so anyway.
[...]

PR #2527, opened 2020-03-22
Avoid calling DIRECT codepath in DYNAMIC_ARCH on non-SKX

martin-frbg:
#2526 (possibly #2524 as well)

### DIFF
--- a/interface/gemm.c
+++ b/interface/gemm.c
@@ -272,6 +272,9 @@ void CNAME(enum CBLAS_ORDER order, enum CBLAS_TRANSPOSE TransA, enum CBLAS_TRANS
   PRINT_DEBUG_CNAME;
 
 #if !defined(COMPLEX) && !defined(DOUBLE) && defined(USE_SGEMM_KERNEL_DIRECT)
+#ifdef DYNAMIC_ARCH
+ if (gotoblas == &gotoblas_SKYLAKEX)
+#endif  
   if (beta == 0 && alpha == 1.0 && order == CblasRowMajor && TransA == CblasNoTrans && TransB == CblasNoTrans && sgemm_kernel_direct_performant(m,n,k)) {
 	sgemm_kernel_direct(m, n, k, a, lda, b, ldb, c, ldc);
 	return;


mattip:
One way to test this would be to add a x64 run on azure pipelines, which I think has non-SKX hardware (at least that is my operative assumption from the bug report).

martin-frbg:
I have both kinds of hardware available locally, just need to get things organized as I need the SKX for other tasks as well.

mattip:
one macOS test did not start

martin-frbg:
Travis appears to have some ongoing problems with their MacOS hosts, I have been restarting CI jobs where necessary over the weekend. Still not sure if this PR is actually needed.

mattip:
This is failing to buld in some configurations of MacPython/openblas-libs#28:
```
make DYNAMIC_ARCH=1 USE_OPENMP=0 NUM_THREADS=64 BINARY=64

gemm.c: In function ‘cblas_sgemm’:

gemm.c:276:19: error: ‘gotoblas_SKYLAKEX’ undeclared (first use in this function); did you mean ‘gotoblas_t’?

  if (gotoblas == &gotoblas_SKYLAKEX)

                   ^~~~~~~~~~~~~~~~~

                   gotoblas_t
```

martin-frbg:
Thanks for the feedback. Could that be a configuration that does not have AVX512 ? Otherwise it is unclear to me why gotoblas_SKYLAKEX could be unknown.

mattip:
apparently :) I am not sure how travis CI chooses what hardware for each option in the matrix

martin-frbg:
Some pretty poor logic I put in there... this just won't be known at compile time unless the planets happen to be lined up correctly. I'll see if simply checking for core SKYLAKEX being defined does the trick.

mattip:
No pressure. Please xref #2526 from the new PR so I can try it out.

martin-frbg:
Probably makes more sense to expose the support_avx512 function that is already in dynamic.c and call that at runtime ... testing this now.

leezu:
@martin-frbg is there any plan to release an OpenBLAS version including this fix shortly? If not, how about backporting the fix to 0.3.9 and releasing 0.3.9b?

Thank you!

martin-frbg:
@leezu early June would certainly make sense (3 months since last release), not sure if I can make it happen next weekend. (Not in favor of doing "b" releases unless for a very bad bug that affects many/most platforms - otherwise there would have to be a microrelease every few days)

PR #2533, opened 2020-03-26
Use runtime check for AVX512 capability in DYNAMIC_ARCH builds made on SKX

martin-frbg:
replaces #2527 as the fix for #2526


martin-frbg:
@mattip could you check if this makes sense now ?

mattip:
In order to check this, it needs to be part of the repo. I first need to build the openblas-libs repo, upload the artifacts, then trigger a CI run in the NumPy repo using those artifacts.

mattip:
This seems to work so far. A CI run with the fix passes. xref numpy/numpy#15868

Issue #2545, opened 2020-04-08
Specifying thread affinity at runtime


sharvil:
It's important to specify the thread affinity at runtime for my application. I've locally modified OpenBLAS to expose a function to set the thread affinity (Linux-only) and I'd love to see it upstreamed. In `drivers/others/blas_server.c`:

```c
void openblas_setaffinity(const cpu_set_t* cpu_set) {
    for (int i = 0; i < blas_num_threads - 1; ++i) {
        pthread_setaffinity_np(blas_threads[i], sizeof(cpu_set_t), cpu_set);
    }
}
```

I'm happy to submit a PR. Is it sufficient to hide this function behind `#if __linux__`? Any other comments / suggestions about adding this function?

martin-frbg:
`#if defined(OS_LINUX) && !defined(NO_AFFINITY)` I would think... and expose it in cblas.h
ISTR there is a gotoblas_set_affinity already in the code but that is probably just an internal helper to apply the initial affinity mask to a thread.

sharvil:
Thanks for the quick response! Yes, there is a `gotoblas_set_affinity` and it is internal as you said.

I'm not sure gating on `!defined(NO_AFFINITY)` is right here. A caller explicitly specifying thread affinity seems independent of whether OpenBLAS sets thread affinities automatically or not (which, AFAICT, is the purpose of `NO_AFFINITY`).

brada4:
There are ways to prevent OS scheduler to migrate running processes off the caches. If something is badly sub-optimal there are ways with system tuning that benefits other software outside openblas alike.

martin-frbg:
Was not sure offhand if all the required code would be compiled with NO_AFFINITY set, but I guess if it already compiled and worked for you, you will be right. (And probably no reason not to provide an entirely optional interface in the library - not everybody will be able to make system-wide changes on their hardware, and not all OS may provide a working, scheduler-based solution.)

PR #2547, opened 2020-04-08
Add API to set thread affinity on Linux.

sharvil:
Issue: #2545

### DIFF
--- a/driver/others/blas_server.c
+++ b/driver/others/blas_server.c
@@ -72,6 +72,7 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "common.h"
 #if defined(OS_LINUX) || defined(OS_NETBSD) || defined(OS_DARWIN) || defined(OS_ANDROID) || defined(OS_SUNOS) || defined(OS_FREEBSD) || defined(OS_OPENBSD) || defined(OS_DRAGONFLY) || defined(OS_HAIKU)
 #include <dlfcn.h>
+#include <errno.h>
 #include <signal.h>
 #include <sys/resource.h>
 #include <sys/time.h>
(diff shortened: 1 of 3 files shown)


sharvil:
Just a quick note on this PR – I changed the API a little bit so it's more flexible and matches `sched_setaffinity`.

sharvil:
@martin-frbg, could you take a look at this PR? Thanks!

martin-frbg:
LGTM, I'm just stuck on some of the other issues at the moment - will merge this later today. pthread_setaffinity_np seems to have come with glibc 2.3.4 in 2005 so we probably do not need a glibc version check around this for compatibilty. (Would be interested in your use case for this iff you can disclose it)

sharvil:
Thanks a bunch, @martin-frbg!

In our case, we're building an ML inference library for a latency-sensitive model. That library can be linked in to various programs so using `taskset` on the entire process isn't great. We've also seen poor OS thread scheduling blowing away our latency budget. This change allows us to expose some tuning knobs that a) don't require OS-level changes and b) affect only our computations and not the rest of the process we're linked into.

driver/others/blas_server.c
```diff
@@ -279,6 +280,23 @@ int get_node(void);
 
 static int increased_threads = 0;
 
+#ifdef OS_LINUX
```
h-vetinari, line 283: Shouldn't this use the same variable as the other if-guard? I.e. `OPENBLAS_OS_LINUX` instead of `OS_LINUX`?

martin-frbg: The API header for OpenBLAS has OPENBLAS_ prepended to the configuration variables,
internally i.e. for building OpenBLAS itself it is just OS_LINUX etc.

mattip:
Could this get a mention in the wiki [extensions page](https://github.com/xianyi/OpenBLAS/wiki/OpenBLAS-Extensions)?

martin-frbg:
Certainly, except that it is not yet available in a formal release.

mattip:
Hmm. Is there a way to mark that as a post-release task?

driver/others/blas_server.c
h-vetinari: thanks for the response, learned something new again.

martin-frbg:
Hm, put a noose around my neck ? For the moment I hope to remember it when I create the changes document from the list of tagged tickets

driver/others/blas_server.c
martin-frbg: come to think of it, this could probably be added to the extremely sketchy "developer manual" in the wiki

mattip:
gentle ping about mentioning this in the wiki

martin-frbg:
Hrrggh ? :flushed:
Added now, thanks for the reminder

PR #2548, opened 2020-04-09
Add a GENERIC target for 64bit MIPS

XiWeiGu:
Using "make TARGET=GENERIC" on loongson platform will get the following
error messages:
"make[1]: *** No rule to make target 'sgemm_incopy.o', needed by 'libs'"
Add kernel/mips64/KERNEL.generic to slove the problem.


martin-frbg:
Thanks. I've changed the title as the problem is not specific to Loongson. The unrelated build failure is from issue 2444 - strange that it still happens here while I cannot reproduce it in my build. (Got it now I think - the test as committed requires more than 50 threads but drone.io builds can alternate between a 44-core and a 96-core host.)

XiWeiGu:
Thanks.

PR #2559, opened 2020-04-14
Add half precision gemm for bfloat16 in OpenBLAS

RajalakshmiSR:
This patch adds support for bfloat16 data type matrix multiplication kernel.
For architectures that don't support bfloat16, it is defined as unsigned short
(2 bytes).  Default unroll sizes can be changed as per architecture as done for
SGEMM and for now 8 and 4 are used for M and N.  Size of ncopy/tcopy can be
changed as per architecture requirement and for now, size 2 is used.

Added shgemm in kernel/power/KERNEL.POWER9 and tested in powerpc64le and
powerpc64.  For reference, added a small test compare_sgemm_shgemm.c to compare
sgemm and shgemm output.

This patch does not cover OpenBLAS test, benchmark and lapack tests for shgemm.
Complex type implementation can be discussed and added once this is approved.


martin-frbg:
This probably needs something like
```
ifndef SHGEMMKERNEL
SHGEMMKERNEL = ../generic/gemmkernel_2x2.c
SHGEMM_BETA = ../generic/gemm_beta.c
endif
```
in kernel/Makefile.L3 to make it compile on non-PPC

martin-frbg:
Ugh. DYNAMIC_ARCH is a mess - I think it is probably best to expand the table entries in common_param.h/setparam-ref.c to define a full set of SH... kernels even if there is "only" SHGEMM at the moment. And the bfloat16tof32 helper probably needs to be declared static to avoid multiple definition errors. Testing locally first...

martin-frbg:
What do you think of this ? 
[shgemm_fixes.tar.gz](https://github.com/xianyi/OpenBLAS/files/4480586/shgemm_fixes.tar.gz)

ceseo:
cc @ceseo

martin-frbg:
drone.io likes to blow up at the moment due to some unrelated problem with the Ubuntu repositories

martin-frbg:
drone.io should be fixed now, but I cannot get this PR to refresh just for the one CI job, so they are still showing as failed

And it looks like the typedefs for bfloat16 in common_level3.h, common_interface.h and common_param.h should be also `#ifndef COMMON_H` to make clang shut up - its redefinition warning is what is currently killing the two travis tasks by producing excessive output.

The failed appveyor is _probably_ missing the names of the new shgemm functions in exports/gensymbol (extra steps necessary for creating a DLL on Windows)

RajalakshmiSR:
Thanks @martin-frbg. Will update the headers.

martin-frbg:
Silly me, the dll export thing is actually a cmake problem, the CMakeLists.txt in kernel does not enumerate any SH names yet (and the GenerateNamedObjects function in utils.cmake obviously does know how to generate them either)

RajalakshmiSR:
@martin-frbg Do you know which make command will help to recreate that error locally? 
Do you think  if adding HALF in FLOAT_TYPES in CMakeLists.txt will help ?

martin-frbg:
That is certainly part of the solution, but also changes in cmake/utils.cmake and cmake/kernel.cmake. I _think_ I am almost there now but currently the compilation of shgemm fails with a strange complaint about a prototype mismatch.
If you really want to recreate the problem, just creating a build directory and running `cmake -DDYNAMIC_ARCH=ON ..` in it, followed by `make` will work even on Linux (and probably POWER as well). I hope to have something ready in an hour or so though

martin-frbg:
Unfortunately this is now leading me down a rabbit hole of missing declarations in common_macros.h, as fixing the cmake rules now makes it want to generate all the unimplemented SH  functions.

RajalakshmiSR:
Thanks @martin-frbg .. I am able to recreate this error.. 
[ 99%] Linking C executable openblas_utest... undefined reference to `shgemm_*'
 with cmake. Will check on this.

martin-frbg:
"in an hour or so" yeah right...
~~This at least makes everything compile - the compare job takes suspiciously long (in the DYNAMIC_ARCH case) so this may not be the end of the story yet~~ EDIT: actually all good when I make sure to link against the right build...
[cmakefiles.tar.gz](https://github.com/xianyi/OpenBLAS/files/4493511/cmakefiles.tar.gz)

RajalakshmiSR:
Thanks @martin-frbg . I have added this commit.

martin-frbg:
Hmm. Now it seems it does not pick up the right kernels from the sequence of applicable KERNEL files but uses what I put in cmake/kernel.cmake straight away. And that proves to be a poor choice for x86_64 where the base KERNEL does have axpy.S but any other target-specific KERNEL file overrides it with saxpy.c (and there is something in the windows clang-cl or its assembler that does not like the coding style in axpy.S - I wonder if it will complain about all the other .S files as well)

martin-frbg:
Close&reopen to rebase drone.io
(Meh. drone still trying to update ubuntu 19... will catch this when it gets merged then)

RajalakshmiSR:
Thanks @martin-frbg . Looks like appveyor jobs are queued. Can those be manually triggered ?

martin-frbg:
As I understand it, they are waiting for hosts to become available. As OpenBLAS is using their "free for open source projects" plan we cannot complain (and IDK if paying them money would allow the jobs to get queued faster)

martin-frbg:
Should probably try to move some of the Appveyor jobs to Azure if they keep getting queued for six hours or more now, but haven't found a good example for the required yml snippet yet.

PR #2563, opened 2020-04-16
[OpenBLAS]: benchmark error of potrf

zelong-1024:
[description]: when the matrix size goes higher than 5800 during the cpotrf test, error info, such as "Potrf info = 5679", will be returned on ARM64 and x86 machines. Uplo = L & F.
[solution]: changed the func for building the matrix so that the complex Hermitian matrix can stay positive definite during the computation.


zelong-1024:
Uplo = L or U.  Sorry for mistake.

PR #2565, opened 2020-04-18
Support MIPS32 24K family as P5600

### LINKED ISSUE #2564: MIPS32 24K / 24KE TARGET
Is there a target for MIPS32 24K / 24KE? 
http://en.techinfodepot.shoutwiki.com/wiki/MIPS32#MIPS32_24K_.2F_24KE

martin-frbg:
fixes #2564


bhack:
With this need we to specify `P5600` TARGET when we are cross-compiling?

martin-frbg:
With this it should be possible to specify either TARGET=1004K or TARGET=24K (which should both select the correct -mips32r2 compiler option), and it should not be necessary to specify a TARGET when compiling directly on the hardware. (Obviously you will always need to name a TARGET when cross-compiling on some completely different hardware)

bhack:
```
In file included from getarch_2nd.c:3:0:
config.h:8:9: error: macro names must be identifiers
 #define 24K
```

martin-frbg:
Damn. Identifiers are not allowed to start with a digit, I guess I have to label them MIPS1004K and MIPS24K  (or M1004K, M24K ?) throughout. Have not tried your docker image as I am actually working on something else at the moment, hence this is not even compile-tested, sorry.

bhack:
Yes they will not mach `march` names but if you need the macros you need to use one of that alternatives.

bhack:
Does `TargetList.txt` and Makefiles need to be updated?

bhack:
For the last RPCC commit why we have this ISSUE only with a multithread build?

martin-frbg:
Only the multithreaded code has a need to query the cpu time - mainly to decide when to timeout an idle thread. Elsewhere in the code this is/was only used for benchmarking and troubleshooting with a corresponding compile-time define (TIMING) set .

bhack:
So this native build was working cause  RPCC skip and was not related to CPU differences: https://github.com/xianyi/OpenBLAS/issues/2564#issuecomment-615961972

bhack:
I see @brada4 test on his CPU has `Library Name     ... libopenblas_p5600-r0.3.9.dev.a (Single-threading)`.
So I am not confident that Is a 24k related issue.

martin-frbg:
The -nan=2008 issue is _probably_ related to linking with musl-libc, the RPCC issue is from building multithreaded on an MIPS cpu where the rdhwr instruction is not supported for whatever reason - it is unclear at this time whether that is specific to the 24k cpu or a more general problem affecting P5600 etc. as well

bhack:
> it is unclear at this time whether that is specific to the 24k cpu or a more general problem affecting P5600 etc. as well

Yes this is my doubt. Cause the @brada4 native build test was single thread. Instead I am cross-compiling so multi-thread cannot be autodetected.

bhack:
With the last commit here I still get with `TARGET='MIPS24K'`:
```
getarch.c:1192:2: error: #error "The TARGET specified on the command line or in Makefile.rule is not supported. Please choose a target from TargetList.txt"
 #error "The TARGET specified on the command line or in Makefile.rule is not supported. Please choose a target from TargetList.txt"
```

bhack:
Still some macro problem `TARGET='MIPS24K'`
```In file included from getarch_2nd.c:3:0:
config.h:8:9: error: macro names must be identifiers
 #define 24K
         ^~~
```

brada4:
P5600 is a normal, non-embedded processor, something that drives thin clients usually. Not sure if it is correct to throw it together with less functiional series.

Compiler barfing out is not cpu-specific. I built for same 32v5 as you, not for native, and a single threaded one.

bhack:
Also with the last commit:
```
In file included from getarch_2nd.c:3:0:
config.h:8:9: error: macro names must be identifiers
 #define 24K
         ^~~
getarch_2nd.c: In 
```
Same
```
In file included from getarch_2nd.c:3:0:
config.h:8:9: error: macro names must be identifiers
 #define 1004K
```

bhack:
@brada4 Can you test your `32v5` multi-thread?

getarch.c
```diff
@@ -812,6 +812,34 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+#define CORENAME  "MIPS1004K"
+#else
+#endif
+
+#ifdef FORCE_MIPS24K
+#define FORCE
+#define ARCHITECTURE    "MIPS"
+#define SUBARCHITECTURE "MIPS24K"
+#define SUBDIRNAME      "mips"
+#define ARCHCONFIG   "-D24K " \
```
bhack: `-DMIPS24K`?

getarch.c
```diff
@@ -812,6 +812,34 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #else
 #endif
 
+#ifdef FORCE_MIPS1004K
+#define FORCE
+#define ARCHITECTURE    "MIPS"
+#define SUBARCHITECTURE "MIPS1004K"
+#define SUBDIRNAME      "mips"
+#define ARCHCONFIG   "-D1004K " \
```
bhack: MIPS1004K

martin-frbg:
`git blame` shows that the RPCC code containing the rdhwr instruction was added with/for P5600 support, so now the only question is how "my"1004K behaves in this regard. And I have now found me a MIPS32 programming document that states that unpriviledged access to individual hardware registers must be enabled by setting a bitmask in a special HWREna register (which is zeroed on boot and apparently expected to be managed by the operating system).

bhack:
Ok now macros are ok but i got:
```*** No rule to make target '../kernel/mips/amax.S', needed by 'samax_k.o'.  Stop.```

**Comment by martin-frbg** [comment]
That would suggest the build system is not reading kernel/mips/KERNEL.MIPS24K (which would be totally plausible as I seem to have done the rename from KERNEL.24K in my local tree only so far)

**Comment by bhack** [comment]
Ok with the last commit compile fine without fortran, `TARGET='MIPS24K'` and `USE_THREAD=0`.
I am trying setting `FC`.

File: common_mips.h [hunk 3]
```diff
@@ -43,6 +43,7 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #ifndef ASSEMBLER
 
+#if !defined(MIPS24K)
```
bhack, line 46: @brada4 Ok now it compiles fine also  without special flags. I suppose that this `if` removed assembly but I see:

```
static inline unsigned int rpcc(void){
  unsigned long ret;

  __asm__ __volatile__(".set   push    \n"
          "rdhwr %0, $30  \n"
          ".set pop" : "=r"(ret) : : "memory");

  return ret;
}
```
So seems is working on this PR. I don't know why it doesn't work on master.

bhack: Ok I suppose that the negation is wrong. But it is working :smile:

bhack:
@martin-frbg The rppc macro for 24k is not working here but on this PR it now compiles without special flags.

common_mips.h
martin-frbg: On master (or rather, develop branch) we have only the outer "ifndef ASSEMBLER" - meaning "when this is included from C code like blas_server.c" - the PR adds the "and we are not building for the 24K" condition so in your case this rpcc function is not included (and RPCC_DEFINED is not set, so the other code "knows" that it has to work without it)

bhack: ok

bhack: @martin-frbg So is it ok if I see this resolved as

```
cat ./build_dir/target-mipsel_24kc_musl/OpenBLAS-0.3.9/common_mips.h
#ifndef ASSEMBLER

static inline unsigned int rpcc(void){
  unsigned long ret;

  __asm__ __volatile__(".set   push    \n"
          "rdhwr %0, $30  \n"
          ".set pop" : "=r"(ret) : : "memory");

  return ret;
}
#define RPCC_DEFINED
```

martin-frbg: Sorry, I do not understand your question. What you have shown here looks like the "old" code from develop before this PR.

bhack: Yes sorry with all these building dirs on single commits I took an old one.

martin-frbg: Guess it is time to commit this to avoid further misunderstanding ?

bhack:
I think that we could merge probably.

Makefile.system
```diff
@@ -690,7 +690,12 @@ CCOMMON_OPT += -march=mips64
 FCOMMON_OPT += -march=mips64
 endif
 
-ifeq ($(CORE), 1004K)
+ifeq ($(CORE), MIPS24K)
+CCOMMON_OPT += -mips32r2 -mtune=24kc  $(MSA_FLAGS)
+FCOMMON_OPT += -mips32r2 -mtune=24kc  $(MSA_FLAGS)
```
brada4, line 695: It is mips32r5

cpuid_mips.c
```diff
@@ -146,7 +150,15 @@ void get_cpuconfig(void){
     printf("#define MIPS1004K\n");
     printf("#define L1_DATA_SIZE 32768\n");
     printf("#define L1_DATA_LINESIZE 32\n");
-    printf("#define L2_SIZE 26144\n");
+    printf("#define L2_SIZE 262144\n");
+    printf("#define DTB_DEFAULT_ENTRIES 8\n");
+    printf("#define DTB_SIZE 4096\n");
+    printf("#define L2_ASSOCIATIVE 4\n");
+  } else if (detect()==CPU_24K) {
```
brada4, line 157: 74k has same cache parameters, as much as OpenWRT is concerned it considers 74k to be another 24k

getarch.c
```diff
@@ -812,6 +812,34 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+#define CORENAME  "MIPS1004K"
+#else
+#endif
+
+#ifdef FORCE_MIPS24K
+#define FORCE
+#define ARCHITECTURE    "MIPS"
+#define SUBARCHITECTURE "MIPS24K"
+#define SUBDIRNAME      "mips"
+#define ARCHCONFIG   "-DMIPS24K " \
```
brada4, line 834: 74K would have 32 TLB entries and 4-way associativity (no L2 cache at all)
```
[    0.000000] Primary instruction cache 64kB, VIPT, 4-way, linesize 32 bytes.
[    0.000000] Primary data cache 32kB, 4-way, VIPT, cache aliases, linesize 32 bytes
```

Makefile.system
martin-frbg: The datasheet has "The 24Kc core implements the MIPS32 Release 2 Architecture" (and same for the instruction set later on).

cpuid_mips.c
martin-frbg: We can look into supporting 34K/74K as equivalent once this definitely works...

getarch.c
martin-frbg: IIRC most if not all of these values are never used in the code, my best guess is that K.Goto simply integrated some very detailed cpuid code from an earlier project.

brada4:
OpenWRT labels all non-kernel packages mips_24kc, it is OK as much as requested.
The change request is to not change rpcc assembler before OpenWRT packing system (or package) stops messing with CFLAGS and LDFLAGS

martin-frbg:
> The change request is to not change rpcc assembler before OpenWRT packing system (or package) stops messing with CFLAGS and LDFLAGS

So you think the rdhwr issue is just OpenWRT removing the `-mips32r2` from our CFLAGS ?

brada4:
> > The change request is to not change rpcc assembler before OpenWRT packing system (or package) stops messing with CFLAGS and LDFLAGS
> 
> So you think the rdhwr issue is just OpenWRT removing the `-mips32r2` from our CFLAGS ?

Yes, they export (makefile way) CFLAGS= LDFLAGS= , instead of unexport-ing them as they do with rest of typical linux system environment.

bhack:
Are you sue?
This is with master branch so (P5600) without unexport-ing:

```
mipsel-openwrt-linux-musl-gcc -Os -pipe -mno-branch-likely -mips32r2 -mtune=24kc -fno-caller-saves -fno-plt -fhonour-copts -Wno-error=unused-but-set-variable -Wno-error=unused-result -msoft-float -mips16 -minterlink-mips16 -iremap/home/build/openwrt/build_dir/target-mipsel_24kc_musl/OpenBLAS-0.3.9-8a6d26458b404dc0305e3b2c46cc37c665afe32c:OpenBLAS-0.3.9-8a6d26458b404dc0305e3b2c46cc37c665afe32c -Wformat -Werror=format-security -fstack-protector -D_FORTIFY_SOURCE=1 -Wl,-z,now -Wl,-z,relro  -I/home/build/openwrt/staging_dir/target-mipsel_24kc_musl/usr/include -I/home/build/openwrt/staging_dir/target-mipsel_24kc_musl/include -I/home/build/openwrt/staging_dir/toolchain-mipsel_24kc_gcc-7.5.0_musl/usr/include -I/home/build/openwrt/staging_dir/toolchain-mipsel_24kc_gcc-7.5.0_musl/include/fortify -I/home/build/openwrt/staging_dir/toolchain-mipsel_24kc_gcc-7.5.0_musl/include  -O2 -DMAX_STACK_ALLOC=2048 -Wall -mabi=32 -mips32r5 -mnan=2008 -mtune=p5600   -DF_INTERFACE_GFORT -fPIC -DNO_LAPACK -DNO_LAPACKE -DSMP_SERVER -DNO_WARMUP -DMAX_CPU_NUMBER=4 -DMAX_PARALLEL_NUMBER=1 -DVERSION=\"0.3.9.dev\" -DASMNAME= -DASMFNAME=_ -DNAME=_ -DCNAME= -DCHAR_NAME=\"_\" -DCHAR_CNAME=\"\" -DNO_AFFINITY -I. -O2 -DMAX_STACK_ALLOC=2048 -Wall -mabi=32 -mips32r5 -mnan=2008 -mtune=p5600   -DF_INTERFACE_GFORT -fPIC -DNO_LAPACK -DNO_LAPACKE -DSMP_SERVER -DNO_WARMUP -DMAX_CPU_NUMBER=4 -DMAX_PARALLEL_NUMBER=1 -DVERSION=\"0.3.9.dev\" -DASMNAME=parameter -DASMFNAME=parameter_ -DNAME=parameter_ -DCNAME=parameter -DCHAR_NAME=\"parameter_\" -DCHAR_CNAME=\"parameter\" -DNO_AFFINITY -I../..
```

bhack:
And so `unexport CFLAGS LDFLAGS` has the same:
```
{standard input}:209: Error: opcode not supported on this processor: mips32r5 (mips32r5) `rdhwr $2,$30'
{standard input}:233: Error: opcode not supported on this processor: mips32r5 (mips32r5) `rdhwr $2,$30'
{standard input}:319: Error: opcode not supported on this processor: mips32r5 (mips32r5) `rdhwr $2,$30'
```

bhack:
I am using a quite standard Makefile documented in openwrt cause my primary scope is to let Openblas be available for others users in openwrt more that just compile for 24k: 
https://openwrt.org/docs/guide-developer/packages

I think it is hard to be on the same page if we cannot be in that toolchain cause you are against Docker for reproducibility.

Plese `docker run --rm -it openwrtorg/sdk:ramips-mt76x8-19.07.2` if you can so we are on the same page.

martin-frbg:
Merging to make it easier to test - at least the rpcc topic should not affect anything except 24K

PR #2591, opened 2020-04-29
Add test for shgemm

RajalakshmiSR:
This patch has Makefile changes to add test for shgemm which
compares sgemm and shgemm result.


RajalakshmiSR:
Based on PR 2590, test is protected with BUILD_HALF check.

PR #2594, opened 2020-05-01
Delete a stray file and apply a few cosmetic fixes for build warnings

martin-frbg:


common_x86_64.h
```diff
@@ -1,4 +1,4 @@
-/*********************************************************************/
+ /*********************************************************************/
```
h-vetinari: This seems unintentional

martin-frbg: Yes, editing glitch. Guess I'll redo this PR without the latest set of changes, but I'd like to understand how changing blas_quickdivide to return an unsigned value (where it is only ever expected to be positive anyway) can blow up gemm on two of the travis VMs.

PR #2597, opened 2020-05-05
Use Clang 9.0.0 miscompilation fix for corresponding AppleClang version as well

martin-frbg:
As discussed in the original PR #2329, AppleClang 11.0.3 is derived from LLVM Clang 9.0.0 and displays the same (mis)optimization defect.


ax3l:
Awesome, thanks a lot!

PR #2613, opened 2020-05-12
Revise dynamic architecture support for IBM z (s390x)

ghost:
Extend, revise, and simplify the support for dynamic architecture on IBM z s390x.

Today, distributions build OpenBLAS on s390x with a fixed, conservative target, which provides compatibility for existing systems yet skips the potential for performance gains on the recent architectures. By switching to dynamic_arch support on s390x, we can address these downsides. Dynamically choosing the best-fitting kernels can tap into the SIMD features available on current generations while providing generic implementations as fallback helps maintain compatibility.

This pull request contains patches to:
- Add generic kernels the the choices available with dynamic_arch.
- Add checks for compiler versions, to avoid build failures with older toolchains and instead fallback gracefully to the feature set supported by a given compiler version.
- Add checks for the OS's advertisement of SIMD features before employing them. Note that we can only use the SIMD features on z13 or newer when the OS supports them (i.e., properly handles context-switching).
- Simplify the selection of kernels at run-time. Make a choice based on the available CPU features instead of model numbers (which amounts to the same result as the current code). As additional benefit, the code becomes more future-proof.

I am looking forward to your feedback and comments.
Marius Hillenbrand

EDIT: add more context.


ghost:
Closing this PR since all patches are also present in #2614

PR #2614, opened 2020-05-12
s390x: Improve performance of SGEMM and STRMM on z14 and newer

ghost:
Exploit the FP32 SIMD operations that were introduced with z14 for single-precision matrix-matrix multiplications by (1) adding a new GEMM kernel implementation that uses these instructions and (2) doubling the number of elements of the destination matrix that are held in registers. As a result, performance on z14 improves by around 2x and even more on z15 (in preliminary experiments).

Specifically,
- Add a new GEMM kernel implementation written in C with vector intrinsics, which is compiled into  FP32 SIMD operations.
- Apply the new kernel for SGEMM and STRMM on z14.
- Harvest the flexibility gained by the C code, change register blocking to 16x4, and improve performance even further.

Please note that the first four commit ins this PR are already covered by #2613 (I am not aware of a nicer way for "stacked" PRs).

Happy to address your questions, concerns, and feedback.
Thanks,
Marius


martin-frbg:
Thanks, I suspect the only people finding faults with this here would be your own coworkers, so committing to give it more exposure :-)
I take it that this supersedes your #2613 , so only this one should be merged ?

ghost:
Indeed, this PR contains all the patches from #2613. I filed two PRs to separate the two patch series, yet it just looks weird now.

PR #2615, opened 2020-05-14
s390x: improvise vector alignment hints for older compilers

ghost:
This PR includes a patch to "polish" the recent performance improvement for SGEMM on z14 for compilers before gcc-9.

Specifically, introduce inline assembly so that we can employ vector loads with
alignment hints on older compilers (pre gcc-9), since these are still
used in distributions such as RHEL 8 and Ubuntu 18.04 LTS.

Informing the hardware about alignment can speed up vector loads. For
that purpose, we can encode hints about 8-byte or 16-byte alignment of
the memory operand into the opcodes. gcc-9 and newer automatically emit
such hints, where applicable. Add a bit of inline assembly that achieves
the same for older compilers. Since an older binutils may not know about
the additional operand for the hints, we explicitly encode the opcode in
hex.

Signed-off-by: Marius Hillenbrand <mhillen@linux.ibm.com>

### DIFF
--- a/kernel/zarch/gemm_vec.c
+++ b/kernel/zarch/gemm_vec.c
@@ -159,6 +159,32 @@ static const bool backwards = false;
 typedef FLOAT vector_float __attribute__ ((vector_size (16)));
 
 /**
+ * Load a vector into register, and hint on 8-byte alignment to improve
+ * performance. gcc-9 and newer will create these hints by itself. For older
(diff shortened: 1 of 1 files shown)


ghost:
Minor fixup. Adding the alignment hint into the hex encoding (since it is constant anyway) to workaround a long-fixed bug in some ancient binutils.

PR #2617, opened 2020-05-17
Add workaround for unhandled  gmake jobserver flags in c_check/f_check

### LINKED ISSUE #2616: Don't parse `gcc -v` / `gfortran -v` blindly
`f_check` and `c_check` seems to parse the output of `gcc -v` / `gfortran -v` without any attempt to distinguish different part of the output. This seems to be extremely fragile and can break whenever new outputs were added that just happens to look like a linker flag even if it might be flags for something else or completely unrelated strings.

In particular, this cause a compile error under GCC 10 if running under a parallel make with job server and if the `-l` flag is in the `MAKEFLAGS`. Since `gcc` is started by `perl` and doesn't have the jobserver fd, GCC 10 correctly disabled job server support (after https://github.com/gcc-mirror/gcc/commit/e006ead52305) which causes the new `MAKEFLAGS` to be printed to the output when `-v` is on. The make flag is then being blindly interpreted as a linker flag and cause an error in the link test!

martin-frbg:
fixes #2616


yuyichao:
What's wrong with unsetting envs?

martin-frbg:
Going for minimal required change here - some system actually expecting to link lib1234.so seems much less likely than some combination of environment variables actually being vital for someone.

yuyichao:
But it seems more likely that some other make arguments will look like compiler arguments.

martin-frbg:
Did not come across any so far.

yuyichao:
What does "so far" mean? If it is about no one having hit this then that's not an argument since no one was printing it until GCC 10. If it's from reading of make flags then AFAICT `-L` is an example of make flag that can be incorrectly intnterpreted. `-R` seems like one too.

PR #2623, opened 2020-05-20
s390x: Use new sgemm kernel also for DGEMM and DTRMM on Z14 (+ small cleanup)

ghost:
Apply our new GEMM kernel implementation, written in C with vector intrinsics, also for DGEMM and DTRMM on Z14 and newer. As a result, we gain around 10% in performance on z15, in addition to improving maintainability. Also, improve that GEMM kernel by replacing the initialization-with-zero of registers with the peeled-off first loop iteration.


ghost:
@martin-frbg  thank you for merging!

PR #2626, opened 2020-05-25
make GCC version detection OS-independent

docularxu:
Previous design put GCC version detection inside of OSNAME 'WINNT'.
However, such detections are required for 'Linux' and possibly other
OS'es as well. For example, there is usage of the GCC versions
in Makefile.arm64. When compiling on Linux machine, in the previous
design, Markfile.arm64 will not know the correct GCC version.

The fix is to move GCC version detection into common part, not
wrapped by anything.

Signed-off-by: Guodong Xu <guodong.xu@linaro.com>

### DIFF
--- a/Makefile.system
+++ b/Makefile.system
@@ -277,6 +277,15 @@ NO_LAPACK = 1
 override FEXTRALIB = 
 endif
 
+ifeq ($(C_COMPILER), GCC)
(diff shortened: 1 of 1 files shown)


martin-frbg:
Indeed, thanks, this was quite an embarassing mistake in the makefile. Perhaps there should be comments on these kind of nested endifs to make it clearer where each conditional block ends.

PR #2590, opened 2020-04-29
Make building the bfloat16 functions conditional on option BUILD_HALF

martin-frbg:
(and off by default) 
for #2560 
Note this would need further changes in level2/Makefile etc. as more bfloat16 functions get added.

PR #2645, opened 2020-06-05
Miscellaneous fixes

martin-frbg:
Make blas_quickdivide return unsigned and ensure the result of the inline assembly is not lost
Make the half-precision GEMM test compile with older compilers
Remove a spurious copy of common_param.h (created by a botched upload for #2551), the intended changes (adding "UL" to constants to fix a compiler warning) were redone in #2590


martin-frbg:
Need to redo the shgemm patch as the branch it came from appears to be outdated

PR #2646, opened 2020-06-06
Optimize AVX512 parallel DGEMM performance

wjc404:
Reduce the amount of vector load instructions (from "12 vbroadcastsd + 2 vmovupd" to "6 vbroadcastf32x4 + 4 vmovddup" per iteration) to minimize power consumption and thermal throttling.

Performance of DGEMM with 26 threads on Intel Xeon Platinum 8269CY (transa = transb = N):

Dimension | GFLOPS-OpenBLAS-old | GFLOPS-OpenBLAS-new | GFLOPS-MKL2019
-- | -- | -- | --
10000 | 1348 | 1396 | 1527
20000 | 1401 | 1443 | 1551
30000 | 1400 | 1449 | 1555

### DIFF
--- a/kernel/x86_64/dgemm_kernel_16x2_skylakex.c
+++ b/kernel/x86_64/dgemm_kernel_16x2_skylakex.c
@@ -54,40 +54,40 @@
     #define kernel_kstart_n10(mdim,updk) ""
     #define kernel_kstart_n12(mdim,updk) ""
     #define kernel_kend_n4(mdim) "xorq %3,%3;"\
-      loada_kend_k1m##mdim acc_kend_nc2_k1m##mdim(0,8)\
-      loada_kend_k1m##mdim acc_kend_nc2_k1m##mdim(16,24)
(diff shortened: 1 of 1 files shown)


jeffhammond:
Have you measured real-time frequency data to prove that this change has any impact on power or thermals?

I don't think there is any basis to the suggest that.  Rather, you have optimized data movement and that improves performance.  This also has the effect of increasing the frequency of AVX512 FMA instructions, which - if anything - will increase power consumption, since FMAs are AVX512-heavy while load-store are AVX512-light.  However, given the overall intensity of FMAs in DGEMM, I would expect this change has no effect on frequency, because the timescale for frequency changes is much greater than that of load-store in the inner loop of DGEMM.

I work for Intel and consulted with our AVX512 DGEMM experts before posting this.  I would like this issue to be accurate because it has already convinced someone else that throttling, not reducing loads and stores, is the cause for the performance change (https://github.com/flame/blis/issues/441).

Issue #2650, opened 2020-06-08
libopenblas.so misses dependency declaration on libflang.so when compiling with flang


leezu:
Steps to reproduce

``` bash
cd /usr/local
wget https://github.com/flang-compiler/flang/releases/download/flang_20190329/flang-20190329-x86-70.tgz
tar xf flang-20190329-x86-70.tgz
make DYNAMIC_ARCH=1 USE_OPENMP=1 FC=flang
```

We can check the undefined symbols

```
nm -u libopenblas.so
                 U GOMP_parallel@@GOMP_4.0
                 w _ITM_deregisterTMCloneTable
                 w _ITM_registerTMCloneTable
                 U __assert_fail@@GLIBC_2.2.5
                 w __cxa_finalize@@GLIBC_2.2.5
                 U __errno_location@@GLIBC_2.2.5
                 U __fd_atan2_1
                 U __fd_cos_1
                 U __fd_cos_2
                 U __fd_exp_1
                 U __fd_log10_1
                 U __fd_log_1
                 U __fd_log_2
                 U __fd_pow_1
                 U __fd_powi_1
                 U __fd_sin_1
                 U __fd_sincos_1
                 U __fprintf_chk@@GLIBC_2.3.4
                 U __fs_atan2_1
                 U __fs_cos_1
                 U __fs_cos_4
                 U __fs_exp_1
                 U __fs_log10_1
                 U __fs_log_1
                 U __fs_log_4
                 U __fs_pow_1
                 U __fs_powi_1
                 U __fs_powi_4
                 U __fs_sin_1
                 U __fs_sin_4
                 U __fs_sincos_1
                 U __get_size_of
                 w __gmon_start__
                 U __kmpc_end_master
                 U __kmpc_fork_call
                 U __kmpc_global_thread_num
                 U __kmpc_master
                 U __kmpc_omp_task
                 U __kmpc_omp_task_alloc
                 U __memset_chk@@GLIBC_2.3.4
                 U __mth_i_cdabs
                 U __mth_i_cddiv
                 U __mth_i_cdiv
                 U __mth_i_cdsin
                 U __mth_i_cdsqrt
                 U __mth_i_csin
                 U __mth_i_csqrt
                 U __mth_i_idnint
                 U __mth_i_nint
                 U __printf_chk@@GLIBC_2.3.4
                 U __register_atfork@@GLIBC_2.3.2
                 U __sched_cpualloc@@GLIBC_2.7
[...]

PR #2665, opened 2020-06-13
Fix spelling of flang option -Mrecursive, add -Kieee and workaround for AOCC optimizer bug

martin-frbg:
fixes lapack-test errors seen at -O1 .
for #2650


martin-frbg:
updated to remove the restriction to -O1 again and add -fno-unroll-loops specifically for AMD AOCC flang to work around apparent over-optimizing bug of the latter seen in the lapack-test for complex and double complex.

PR #2669, opened 2020-06-16
Zarch fix gcc detection

ghost:
Fix the build failure with gcc-10 and DYNAMIC_ARCH=1 on s390x. Specifically, fix and tidy up the check for gcc version >= 5.2 by employing and extending the common version check variables, as introduced in #2626

### DIFF
--- a/Makefile.system
+++ b/Makefile.system
@@ -282,9 +282,11 @@ endif
 ifeq ($(C_COMPILER), GCC)
 GCCVERSIONGTEQ4 := $(shell expr `$(CC) -dumpversion | cut -f1 -d.` \>= 4)
 GCCVERSIONGT4 := $(shell expr `$(CC) -dumpversion | cut -f1 -d.` \> 4)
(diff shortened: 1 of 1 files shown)


Makefile.system
```diff
@@ -606,7 +606,6 @@ ifneq ($(C_COMPILER), GCC)
 DYNAMIC_CORE += POWER9
 endif
 ifeq ($(C_COMPILER), GCC)
-GCCVERSIONGT5 := $(shell expr `$(CC) -dumpversion | cut -f1 -d.` \> 5)
```
RajalakshmiSR, line 600: @mhillenibm This is needed for POWER as the variable was set under `ifeq ($(OSNAME), WINNT)` in the previous lines..

RajalakshmiSR: Ok.. they are not arch specific now.. This is fine then .

martin-frbg: it was actually rescued out of the WINNT-only section by #2626 so de-duplicating should be ok

ghost: note that the definition has moved out of the OS-specific section, in #2626

Issue #2668, opened 2020-06-15
openblas-0.3.10 build failure on dynamic zarch


susilehtola:
openblas-0.3.10 fails to build in Fedora s390x with `DYNAMIC_ARCH=1 DYNAMIC_OLDER=1`. I get a link error
```
/usr/bin/ld: ../libRblas-r0.3.10.a(dynamic_zarch.o): in function `gotoblas_corename':
/builddir/build/BUILD/openblas-0.3.10/Rblas/driver/others/dynamic_zarch.c:70: undefined reference to `gotoblas_Z13'
/usr/bin/ld: ../libRblas-r0.3.10.a(dynamic_zarch.o): in function `force_coretype':
/builddir/build/BUILD/openblas-0.3.10/Rblas/driver/others/dynamic_zarch.c:129: undefined reference to `gotoblas_Z13'
/usr/bin/ld: ../libRblas-r0.3.10.a(dynamic_zarch.o): in function `gotoblas_corename':
/builddir/build/BUILD/openblas-0.3.10/Rblas/driver/others/dynamic_zarch.c:70: undefined reference to `gotoblas_Z13'
/usr/bin/ld: ../libRblas-r0.3.10.a(dynamic_zarch.o): in function `force_coretype':
/builddir/build/BUILD/openblas-0.3.10/Rblas/driver/others/dynamic_zarch.c:129: undefined reference to `gotoblas_Z13'
collect2: error: ld returned 1 exit status
```

martin-frbg:
Have not checked why this happens, but DYNAMIC_OLDER does not currently make sense on zarch (or anything non-x86) anway. Does it still fail when you leave out that option ?

martin-frbg:
Does this fix it ? (dynamic_zarch.c from line 49)
```
 extern gotoblas_t gotoblas_ZARCH_GENERIC;
 #ifdef HAVE_Z13_SUPPORT
 extern gotoblas_t gotoblas_Z13;
+#else
+#define gotoblas_Z13 gotoblas_ZARCH_GENERIC
 #endif
 #ifdef HAVE_Z14_SUPPORT
 extern gotoblas_t gotoblas_Z14;
+#else
+#define gotoblas_Z14 gotoblas_ZARCH_GENERIC
 #endif
```

susilehtola:
@martin-frbg I think the main question is why `HAVE_Z13_SUPPORT` would not be true?

susilehtola:
[...]

PR #2670, opened 2020-06-16
RFC: Use -dumpfullversion to get minor version on gcc-7 and newer

ghost:
In gcc-7.1, the behavior of -dumpversion changed to be configured
at compile-time. On some distributions it only dumps the major version
(e.g., Ubuntu), so the current checks for the gcc minor version report
false negatives (though, none of them are used for gcc>= 7, as far as I am aware). As a replacement, gcc-7.1 introduced -dumpfullversion
which always prints the full version.
    
Update the gcc version detection in Makefile.system to employ
-dumpfullversion with gcc-7 and newer.
    
Posting this patch for discussion, since it emerged from discussions
around issue #2668 and PR #2669. It is not solving a problem right now,
but may be useful in the future.

(Note that the first two commits overlap with #2669, yet this PR is actually only about the last one)


RajalakshmiSR:
LGTM. Once this is merged I will reuse these new MINOR version macros to add new checks for POWER10.

PR #2677, opened 2020-06-24
address vs2019 C4293 warning

brada4:
Description claims shifting to set sign bit is undefined, though side effect is setting it.


brada4:
pipelines cannot pull intel_sde from docker, probably needs hash updated?

martin-frbg:
Bit of googling suggests the hash is that of a generic ubuntu 18 image, so a network hickup seems more likely. I have restarted the Azure job.

Issue #2671, opened 2020-06-16
Build of OpenBLAS 0.3.8 breaks on AIX


trex58:
Hi,
Between v0.3.7 and v0.3.8, a change was done in file kernel/Makefile.L3 , and thus building v0.3.8 breaks in my AIX 7.2 environment though building v0.3.7 builds/tests OK.

The following lines were added in line 529 of Makefile.L3 :
```
ifeq ($(OS), AIX)
	$(CC) $(CFLAGS) -UDOUBLE -UCOMPLEX -E $< -o cgemm_itcopy.s
	m4 cgemm_itcopy.s > cgemm_itcopy_nomacros.s
	$(CC) $(CFLAGS) -c -UDOUBLE -UCOMPLEX cgemm_itcopy_nomacros.s -o $@
	rm cgemm_itcopy.s cgemm_itcopy_nomacros.s
```
Which generates:
```
gcc -maix64 -O2 -DMAX_STACK_ALLOC=2048 -fopenmp -Wall -DF_INTERFACE_GFORT -fPIC -DSMP_SERVER -DUSE_OPENMP -DNO_WARMUP -DMAX_CPU_NUMBER=2 -DMAX_PARALLEL_NUMBER=1 -DVERSION=\"0.3.8\" -mpowerpc64 -maix64 -DASMNAME=cgemm_itcopy -DASMFNAME=cgemm_itcopy_ -DNAME=cgemm_itcopy_ -DCNAME=cgemm_itcopy -DCHAR_NAME=\"cgemm_itcopy_\" -DCHAR_CNAME=\"cgemm_itcopy\" -DNO_AFFINITY -I.. -UDOUBLE  -DCOMPLEX -UDOUBLE -UCOMPLEX -E ../kernel/power/../generic/zgemm_tcopy_2.c -o cgemm_itcopy.s

m4 cgemm_itcopy.s > cgemm_itcopy_nomacros.s

...
[...]

PR #2680, opened 2020-06-25
Fix for #2671

kavanabhat:


martin-frbg:
Not sure which version is "better" but I guess this is more robust ? AFAICT the only cases where the preprocessor would dump to stdout occured when it was erroneously invoked on the POWER6 files. I have put my current WIP in #2682 for reference, which builds with gcc/gfortran 7.2 (and the AIX `as`, as far as I can determine). xlc has an unrelated issue with vector statements

kavanabhat:
The fix here would work for AIX on all power platforms. It is generic enough to handle any macros introduced for any platform.

martin-frbg:
I guess what was confusing me is that the change is (currently) not needed for power8/gcc, but of course that fits in with your original commit back in december, and the problem only occurring when this code path is taken for power6.  I'll update my WIP accordingly (still need to figure out how to clean up the assembly for xlc builds following David Edelsohn's pointer)

Issue #2022, opened 2019-02-15
Makefile.rule: Typo error in the documentation


Naveenaidu:
Line- [183](https://github.com/xianyi/OpenBLAS/blob/develop/Makefile.rule#L183)
```
# If you need santy check by comparing reference BLAS. It'll be very
```

Santy needs to be changed to Sanity

martin-frbg:
Thanks. The entire SANITY_CHECK option could probably be removed, as it is an idea from the original GotoBLAS that was never implemented. (xianyi later created BLAS_Tester as a separate project for essentially this purpose)

Naveenaidu:
Do you want me to remove the option and send a PR?

brada4:
If you can...
```
./Makefile.rule:1
./CMakeLists.txt:1
./common.h:1
./Makefile.system:3
./Makefile:1
./cmake/os.cmake:3
```
Actually past file changes whole reference/ directory containing upstream BLAS becomes unused.
Good luck!

Naveenaidu:
I'll send in a PR soon ;)

martin-frbg:
Good point about the "reference" directory - on closer look the SANITY_CHECK option does appear to be at least partly implemented despite the comment in Makefile.rule claiming otherwise. (When it is defined, common.h includes the additional header file common_reference.h that adds a set of BLASFUNC_REF declarations for the reference functions - what appears to be missing is the code to actually switch between the two implementations and compare results.) 
Perhaps that option could still be (made) at least marginally useful for debugging cases where it is not trivial to extract the function and arguments that made a calculation go wrong ?

brada4:
Build tests take pre-computed results. Probably there is an elegant way to rename reference functions and base tests on comparison of two on the fly.

PR #2681, opened 2020-06-25
remove unused SANITY_CHECK option #2022

brada4:
Was never finalized to work.


martin-frbg:
so why remove it ? ISTR we disagreed about this back when #2022 was still active.

brada4:
OP sort of said will try but never made into a PR.
There is more recent version in netlib directories....

PR #2682, opened 2020-06-25
[WIP] fix compilation on AIX

martin-frbg:
Fixes several issues in conjunction with #2671 and #2694
- TARGET autodetection always returning POWER5
- Architecture detection failure setting the CROSS build flag
- utest getting built as an empty shell without actual tests
- lapack-test failures in COMPLEX with gfortran 7.3
- inadequate compiler options for xlc/xlf
- 32bit builds segfaulting in sgemm/dgemm/trsm (though this PR only drops 32bit-POWER8 to POWER6 for now)


trex58:
About c0afc11 (Fix POWERPC builds on AIX (gcc/gfortran 7)):
 - _I see MYOS and MYTARGET instead of OS and TARGET (OK ! Fixed in c854ef5 !!!_
 - Also, on AIX, we use the OpenSource GNU version of m4, not the AIX version. And, for the OpenSource GNU version (v1.4.18), the option -B does exist but is not documented (they say: " warning: `m4 -B' may be removed in a future release "). So, it should be better to use some variable which could be defined out of the Makefile.L3 file.

martin-frbg:
I guess I could ask the `m4` if it is GNU or AIX (probably not safe to assume everybody uses the GNU one). I have a few more compatibility fixes for assembly files that I need to append here, but I am still far from getting OpenBLAS to build with the AIX `as` and `xlc` (register naming, unimplemented constraints,...)

PR #2693, opened 2020-06-30
AIX build on POWER8 32bits

EGuesnet:
Hello,

OpenBLAS 0.3.10 does not compile in 32 bits on AIX 7.2, in Power 8 mode (works on Power 6), with GCC 8.4.

```
cgemm_kernel_r_nomacros.s: line 20833: Too few arguments
```

The `kernel/power/cgemm_kernel_8x4_power8.S` file that generates `cgemm_kernel_r_nomacros.s` has `addi    r11, 224`, whereas 3 argument are needed. The PR adds `SP` like in the 64 bits part.
A `^` character that causes a warning in GCC is also erased.

With this PR, OpenBLAS compiles in 32 bits in Power 8 mode, but some tests are failing for other reason.

### DIFF
--- a/kernel/power/cgemm_kernel_8x4_power8.S
+++ b/kernel/power/cgemm_kernel_8x4_power8.S
@@ -424,7 +424,7 @@ L999:
 	lwz	r16,  204(SP)
 	lwz	r15,  208(SP)
 	lwz	r14,  212(SP)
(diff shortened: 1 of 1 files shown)


EGuesnet:
@trex58

PR #2697, opened 2020-07-02
add benchmark for geqrf(Lapack)

jijiwawa:
add benchmark for geqrf(Lapack)
The number of floating-point operations comes from sopla.f (LAPACK3.0 TIMING catalog)


martin-frbg:
Wouldn't this be mainly a very indirect benchmark of gemm/trmm/gemv/ger, as all the lapack code involved is currently taken from the netlib reference implementation? Or are you suggesting/planning a reimplementation of the larf family of lapack functions in OpenBLAS as was done for e.g. potrf

jijiwawa:
Yes, we plan to reimplementation of the larf family of lapack functions in OpenBLAS, and this will be a difficult task. I will turn off this pr, and then resubmit it after reimplementation of the larf family of lapack functions.

PR #2699, opened 2020-07-02
Added benchmark for getrs

zelong-1024:
[OpenBLAS]: benchmark for getrs
[Description]: added benchmark for getrs


martin-frbg:
I believe this is already available as part of the linpack.c benchmark (from original GotoBLAS), though that is perhaps not so obvious ?

zelong-1024:
Oh okay, found it. Then it's not needed to do so. Thanks.

PR #2704, opened 2020-07-02
Delete test while make.

tanfenghua:
It takes a long time to test in many systems, such as ft2000 plus system.


martin-frbg:
Sorry I do not think it would be a good idea to leave out the BLAS tests by default. Can you tell more about the ft2000, what kind of hardware is this ? Normally I would not expect the tests to take appreciably longer than the compile itself. (lapack-test is a different matter, but that does not run automatically)

PR #2708, opened 2020-07-07
Changing mcpu option as power10

RajalakshmiSR:
As compiler enabled mcpu option as power10, changing it from future.


RajalakshmiSR:
GCC commit for reference : https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=9913aa6fbc42eab81fc9599a7d4636de2f92c00b

PR #2716, opened 2020-07-14
Add new linker option for POWER10

RajalakshmiSR:
While building with DYNAMIC_ARCH on POWER9 with POWER10
aware toolchain, new LDFLAG is needed to avoid POWER10
instructions on PLT calls .


RajalakshmiSR:
binutils commit for reference:  https://sourceware.org/git/?p=binutils-gdb.git;a=commit;h=d3b10ee787216d304a858246656ed2fdaecdfd93

PR #2734, opened 2020-07-25
Fix to store results in correct order for POWER10 GEMM kernels

RajalakshmiSR:
There is a recent compiler change in __builtin_mma_disassemble_acc() which
affects the order of storing result in POWER10. Also removing new LDFLAG
-mno-power10-stub as it is handled by linker automatically.


RajalakshmiSR:
Reference commit:  https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=5497677b497b95a261089d19f5295cc80f99a2b6

PR #2740, opened 2020-07-27
Fix compilation issues with clang on POWER

RajalakshmiSR:
clang does not support -malign-power and as gcc defaults to -malign-power, 
removing that option. Also adding -fno-integrated-as to use GNU assembler
for powerpc assembly optimization files. Fixed other compilation errors
reported in dgemv_t.c file.

### DIFF
--- a/Makefile.power
+++ b/Makefile.power
@@ -11,34 +11,34 @@ endif
 
 ifeq ($(CORE), POWER10)
 ifeq ($(USE_OPENMP), 1)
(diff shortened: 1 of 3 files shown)


martin-frbg:
Thanks, I did not yet get around to testing clang after adding PGI (which is clang-based I think). Makefile.power is getting uglier by the day, I will try later if I can regroup the compiler options to make it a bit more readable again...

PR #2743, opened 2020-07-28
Add AMD Matisse and Renoir ZEN2 cpu autodetection

### LINKED ISSUE #2738: Need support for Renoir CPUs
Nothing worth mentioning, it just does not recognize Renoir and uses Prescott binaries. I am able to get the performance acceptable by re-compiling with TARGET=ZEN yet it should be officially fixed since otherwise things like Julia will not work well out of box while the user might be suffering from the legacy binaries' slowness and still think them to be normal.

martin-frbg:
expected to fix #2738

### DIFF
--- a/cpuid_x86.c
+++ b/cpuid_x86.c
@@ -1456,7 +1456,7 @@ int get_cpuname(void){
       case  3:
       case  7:
       case 10:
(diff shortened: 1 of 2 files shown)


martin-frbg:
Nonsense, blame lack of sleep

PR #2758, opened 2020-08-02
Fix GCC ubsan warnings in x86_64 complex dot and gemv_t kernels

martin-frbg:
by replacing a left shift with the equivalent multiplication by 2
(to avoid invoking undefined behavior when the value happens to be less than zero, as seen in the BLAS tests).


brada4:
better += better;

Issue #2767, opened 2020-08-11
BFloat16 data type naming


Guobing-Chen:
In current code base, the BFloat16 data type is named as shxxxx (e.x.: shgemm), and related build flag as BUILD_HALF. Seems we simply take BF16 as half precision of float. This is not true as to the standard definition of IEEE. Half precsion should be FP16, which is different in format and content as compared with BFloat16. As for OpenBLAS, we may both support BFloat16 and and FP16 as valuable for different domains -- BFloat16 mostly valuable for Deep Learning and Machine Learning,  while FP16 more valuable for traditional scientific computation and telecom processing.

Suggest to change the naming of data type and build flag to be bxxxx (e.x.: bgemm) and BUILD_BF16. While we can leave shxxxx for the real half precision data type -- FP16, or even make it as hxxxx (e.x.: hgemm). Whatever, use shxxxx and BUILD_HALF is quite confusing to community, every other math libs like Eigen/oneDNN (previously mkldnn) use keywords like bf or bf16 or so.

I can submit PR to make this change if we are OK on this.

RajalakshmiSR:
The first 's' is for the result (single-precision). If b is for the input (bf16), bgemm would imply that both input and output are bfloat16, which is not the case. So 'sbgemm' would explain it better.

Guobing-Chen:
Good point! 'sb' prefix is good to me also as specific to bfloat16.

martin-frbg:
I remember suggesting to @edelsohn that function names be coordinated with other interested parties before the code lands,
but I have to admit that seeing the actual implementation might be a prerequisite. Also I have no clear idea what the appropriate panel for deciding naming and API for BLAS extensions would be these days ?

Guobing-Chen:
> I remember suggesting to @edelsohn that function names be coordinated with other interested parties before the code lands,
> but I have to admit that seeing the actual implementation might be a prerequisite. Also I have no clear idea what the appropriate panel for deciding naming and API for BLAS extensions would be these days ?

Do you mean naming thing and new APIs should be discussed in a specific panel? If Yes, can I attend? Besides this naming, I also have some new APIs (extensions) to suggest to add.

martin-frbg:
[...]

PR #2771, opened 2020-08-11
Change BFLOAT16 data type/API support naming and prefix

Guobing-Chen:
Existing BFLOAT16 support is named with HALF or sh, which is not good as half
precision usually points to FP16 according to IEEE definition. We'd
specify this as BFLOAT16 instead of HALF. According to the discussion in
issue #2767, update related naming and prefix as BFLOAT16 and sb.

Details as:
1. Change the generic control flag from BUILD_HALF to BUILD_BFLOAT16;
2. Change the prefix of BFLOAT16 API from shxxxx to sbxxxx;
3. Change related file names and other related with above prefix also;


martin-frbg:
Closing due to objection raised in #2767 (naming/API already agreed on with the netlib team)

martin-frbg:
Reopened after further discussions in #2767, though most files will have been altered in the meantime unfortunately. (Opening a new PR may be easier than updating this - please do whatever seems more practical)
(And perhaps BUILD_BFLOAT is clear enough without the 16 that already features in BUILD_COMPLEX16 ?)

martin-frbg:
superseding with #2886 due to the number of code changes since the original PR

PR #2775, opened 2020-08-12
Fix OMP num specify issue

Guobing-Chen:
In current code, no matter what number of threads specified, all
available CPU count is used when invoking OMP, which leads to very bad
performance if the workload is small while all available CPUs are big.
Lots of time are wasted on inter-thread sync. Fix this issue by really
using the number specified by the variable 'num' from calling API.

Signed-off-by: Chen, Guobing <guobing.chen@intel.com>

### DIFF
--- a/driver/others/blas_server_omp.c
+++ b/driver/others/blas_server_omp.c
@@ -335,7 +335,7 @@ int exec_blas(BLASLONG num, blas_queue_t *queue){
       break;
   }
 
-#pragma omp parallel for schedule(OMP_SCHED)
+#pragma omp parallel for num_threads(num) schedule(OMP_SCHED)
   for (i = 0; i < num; i ++) {
 
 #ifndef USE_SIMPLE_THREADED_LEVEL3


martin-frbg:
Did not notice this creating any overhead, as setting OMP_NUM_THREADS seemed to work as advertised ? But perhaps I am misreading your patch, or never tried on a really big big server

Guobing-Chen:
Yes, OMP_NUM_THREADS definitely can control this, but this is not always the case for end-usrs. Sometimes user just leave the default running which will invoke COUNT_OF_ALL_CPU threads, especially those not IT background users. And, as the function already accept 'num' as a parameter to specify the thread number, we'd better keep it work as expected, like what pthread mode running at driver/others/blas_server.c. 

As for the overhead, recent server platform like skylake is with 112 CPUs (HT on), and when the workload is not big, the cost will be very big which makes the multi-threading version running even slower than single-thread. A case in current code is the ddot, it starts to run multi-threading mode when input vector dim bigger than 10000. I've tested that the performance is very bad at 10000~100000 level, slower than single-thread version. Though to fix ddot issue, with this patch is not good enough, need to specify a tuned thread number per different input dim.

martin-frbg:
Ah, I see. Trying to guess intention from the code can be a bit tricky as some of it looks like young Kazushige Goto thinking about things that might be worthwile to implement "later", or just copypasting from a related function. The joys of legacy code without documentation... 
And "all the cores you can get" is the default for non-OMP operation unless OPENBLAS_NUM_THREADS is set - but I guess what your PR effectively does is make OPENBLAS_NUM_THREADS work for OMP builds as well, making a separate OMP_NUM_THREADS unnecessary ?
I suspect multiple switchover points governing thread number is what gives MKL an edge over OpenBLAS in at least some cases - for most functions this is handled in the interface code rather than the kernel file itself, and there are a few functions
there that still do not even have a trivial minimum size for going multi-threaded.

Guobing-Chen:
> Ah, I see. Trying to guess intention from the code can be a bit tricky as some of it looks like young Kazushige Goto thinking about things that might be worthwile to implement "later", or just copypasting from a related function. The joys of legacy code without documentation...
> And "all the cores you can get" is the default for non-OMP operation unless OPENBLAS_NUM_THREADS is set - but I guess what your PR effectively does is make OPENBLAS_NUM_THREADS work for OMP builds as well, making a separate OMP_NUM_THREADS unnecessary ?

No, I didn't change these Environment Variable behaviors. The only thing I changed is to make openmp lib really invokes "num" of threads to do the workloads instead of the_num_of_all_cpus threads. While here the 'num' is exactly the params passed to exec_blas function from high-level calling functions like ddot.c in kernerl/x86_64. I think the original purpose of this 'exec_blas' design is also that it should run 'num' of threads to do the workloads instead of invoking the_num_of_all_cpus threads. If we look into the exec_blas function in blas_server.c (the pthread mode version), it works as expected -- invoking 'num' of threads. 
The impact depends on the workload, if it is heavy enough, then generally no big difference before/after this PR. While when the workload is not that heavy, and multi-threading is triggered, then the performance drops dramatically, like the example I mentioned in previous comment about ddot. And also an additional scenario -- when there is some other non-openblas workloads running in parallel at the same host, using all CPU num of threads will definitely impact the non-openblas workloads badly, comparing with only use 'num' of threads -- assume openblas functions uses a reasonable 'num' of threads which I will talk more below.

> I suspect multiple switchover points governing thread number is what gives MKL an edge over OpenBLAS in at least some cases - for most functions this is handled in the interface code rather than the kernel file itself, and there are a few functions
> there that still do not even have a trivial minimum size for going multi-threaded.

Yes, that is an interesting thing I am tuning recently. E.x.: for dot, <10000 with single thread, [10000, 100000] with 2 threads, [100000, 10000000] 4 threads (Not real suggestion, just to tell the logic). One piece of code in my tuning of bf16 version dot is as below (n as input vector dim, nthreads as all cpu core numbers originally):

     int best_threads = (int) (n/(float)40960 + 0.5);

     if (best_threads < nthreads) {
         nthreads = best_threads;
     }
Strictly speaking, it should be per platform and per function, but more or less this kind of logic based on data from experiment . 

BTW, Yes, I found some multi-threading logic is in kernerl (usually level-1 APIs), and some in interface/driver (Level-2/Level-3). And some function like sdot even without multi-threading support, sound like openblas still has lots of potentials.

PR #2780, opened 2020-08-14
Enable COOPERLAKE build target

Guobing-Chen:
Enable new build target platform -- COOPERLAKE. This target platform supports all the SKYLAKEX supported ISAs + avx512bf16. So all the SKYLAKEX specific kernels/drivers and related code are now extended to be also active on COOPERLAKE. Besides, new BF16 related kernels are active under this target.


Guobing-Chen:
There are two build issues as by the CI checker:
1. Ubuntu platform build fail -- I checked the log, seems the ubuntu is 18, with an related old version GCC which does not support cooperlake platform. According to GCC, it starts to support cooperlake (-march=cooperlake) since 10.1 (For clang, since 9). So, could we upgrade our GCC in the CI/CD system?

2. MacOS build fail -- the error seems a bit strange to me (pasted the log below):
perl ./gensymbol osx x86_64 _ 1 0  1 1 0 0 "" "" 1 0 > osx.def
cc -O2 -DMAX_STACK_ALLOC=2048 -DEXPRECISION -Wall -m64 -DF_INTERFACE_GFORT -fPIC -DDYNAMIC_ARCH -DNO_LAPACK -DNO_LAPACKE -DSMP_SERVER -DNO_WARMUP -DMAX_CPU_NUMBER=4 -DMAX_PARALLEL_NUMBER=1 -DVERSION=\"0.3.10.dev\" -UASMNAME -UASMFNAME -UNAME -UCNAME -UCHAR_NAME -UCHAR_CNAME -DASMNAME=_ -DASMFNAME=__ -DNAME=_ -DCNAME= -DCHAR_NAME=\"_\" -DCHAR_CNAME=\"\" -DNO_AFFINITY -I..  -all_load -headerpad_max_install_names -install_name "/Users/runner/work/OpenBLAS/OpenBLAS/exports/../libopenblas.0.dylib" -dynamiclib -o ../libopenblasp-r0.3.10.dev.dylib ../libopenblasp-r0.3.10.dev.a -Wl,-exported_symbols_list,osx.def  
duplicate symbol '_sgemm_kernel_direct_performant' in:
    ../libopenblasp-r0.3.10.dev.a(sgemm_kernel_SKYLAKEX.o)
    ../libopenblasp-r0.3.10.dev.a(sgemm_kernel_COOPERLAKE.o)
duplicate symbol '_sgemm_kernel_direct' in:
    ../libopenblasp-r0.3.10.dev.a(sgemm_kernel_SKYLAKEX.o)
    ../libopenblasp-r0.3.10.dev.a(sgemm_kernel_COOPERLAKE.o)

Why both SKYLAKEX and COOPERLAKE built object are linked together? What is the build cmd executed here?

martin-frbg:
For (1) I can modify the configuration to add a new gcc version, but that error should still be handled e.g. by disabling COOPERLAKE or mapping it to SKYLAKEX when the compiler is too old.
My guess for (2) is that it might be a DYNAMIC_ARCH build - the job options should be visible at the start of the log, or in the relevant CI file itself. Which of the CI builds is it ?

Guobing-Chen:
> For (1) I can modify the configuration to add a new gcc version, but that error should still be handled e.g. by disabling COOPERLAKE or mapping it to SKYLAKEX when the compiler is too old.

Thanks for the help, so suppose I can just wait? LOL

> My guess for (2) is that it might be a DYNAMIC_ARCH build - the job options should be visible at the start of the log, or in the relevant CI file itself. Which of the CI builds is it ?

It is 'build (macos-latest, gfortran, make)'. Will try offline to see whether I can reproduce the same issue.

martin-frbg:
That's definitely DYNAMIC_ARCH=1, and the first time that this particular function gets called by more than one cpu target. Probably "only" an issue with a local forward declaration of that function in both source files, I'll look at that later.
And that Azure job is possibly the only CI configuration right now that has both DYNAMIC_ARCH=1 and a compiler new enough to support cooperlake targets at all.

Guobing-Chen:
> That's definitely DYNAMIC_ARCH=1, and the first time that this particular function gets called by more than one cpu target. Probably "only" an issue with a local forward declaration of that function in both source files, I'll look at that later.
> And that Azure job is possibly the only CI configuration right now that has both DYNAMIC_ARCH=1 and a compiler new enough to support cooperlake targets at all.

Quickly checked the code, sounds like below place is very suspected, which I highlighted as a temp patch below:

diff --git a/common_level3.h b/common_level3.h
index 4e44a5e7..0094a3f0 100644
--- a/common_level3.h
+++ b/common_level3.h
@@ -47,12 +47,12 @@ __global__ void cuda_dgemm_kernel(int, int, int, double *, double *, double *);
 extern "C" {
 #endif

-extern void sgemm_kernel_direct(BLASLONG M, BLASLONG N, BLASLONG K,
+void sgemm_kernel_direct(BLASLONG M, BLASLONG N, BLASLONG K,
        float * A, BLASLONG strideA,
        float * B, BLASLONG strideB,
        float * R, BLASLONG strideR);

-extern int sgemm_kernel_direct_performant(BLASLONG M, BLASLONG N, BLASLONG K);
+int sgemm_kernel_direct_performant(BLASLONG M, BLASLONG N, BLASLONG K);


 int shgemm_beta(BLASLONG, BLASLONG, BLASLONG, float,

These two functions are set as 'extern' which may conflict with two built objects.

martin-frbg:
More likely the kernel file has verbatim function names when it should be subject to CNAME substitution, at least the problem is vaguely familiar.
Nope, not just that - needs some refactoring before this kernel_direct stuff can coexist with clones of itself, probably needs to go through proper declarations in struct gotoblas of common_param.h etc.

brada4:
There are no revolutionary differences in Cooper Lake over Sky Lake II
There will be no use of BFloat16 any time soon as there is actually no agreed ABI on passing it to a library function (struct int16 aside)
Probably performance concerns could be levelled by simply making non-dynamic local build use march=native for outdated compilers to do their best on future architectures (think LTS/Enterprise distributions still at kernel 2.6 and gcc 4) , while still gaining in LAPACK part from all-latest setups.

Guobing-Chen:
> There are no revolutionary differences in Cooper Lake over Sky Lake II
> There will be no use of BFloat16 any time soon as there is actually no agreed ABI on passing it to a library function (struct int16 aside)

The major difference of CooperLake vs. SkylakeX is AVX512-BF16 support. And BF16 is now applied in Deep Learning domain already, while also good for Machine Learning. As for data type, Bfloat16 is typically represented as uint16_t at C level. Lots of libs support BF16 are using this, including Eigen/MKLDNN/Tensorflow ...

brada4:
The "supports" part does not extend to BLAS. Once infrastructure matures, like fortran compilers, to accomodate new type, I think no big problem having it. Maybe it is short-term stopgap feature, like float80, will see.

Guobing-Chen:
> The "supports" part does not extend to BLAS. Once infrastructure matures, like fortran compilers, to accomodate new type, I think no big problem having it. Maybe it is short-term stopgap feature, like float80, will see.

Yes, it is not by default currently. But the next generation BLAS API will consider that, @mgates3 from netlib team shared some new naming including bf16 of the BLAS G2 API at https://github.com/xianyi/OpenBLAS/issues/2767. 

And, as for now, we can take this BF16 data type related BLAS API as extension of standard BLAS API, which we already have (e.x.: shgemm) for openblas. And I'd contribute more as following.   

BTW, several other math libs are supporting BF16 BLAS level calculations. No reasons openBLAS not to.

brada4:
Can you:
- how it looks on other CPUs like ARMv5 and Pentium P5
- since there is no way at all to make fixed API because of lack of ABI - it is a call for certain future repair

Guobing-Chen:
Definitely there is new features coming at new platforms, that not supported by old platforms.

Anyway, I've described at above about the usage/current supporting in other math libs/the status of support in OpenBLAS now. As for platform, current code of BF16 related in OpenBLAS is for powerpc platform (Yes, there is generic version too), cooperlake is just one from X86_64 family, and ARM also have since ARMv8, which I believe will come in too in future, and Google's TPU and other XPU for deep learning family.

brada4:
But you mean general export for dynamic arch, at the moment nobody knows even if "standard" fallback is int16 or float32 or float64...
Once ABI part is clear - new functions using bad fallbacks (hope it is float32, not 64) and new architecture doing things optimal way?
In the meantime there is open market for cooper-lake-only extra library handling bfloat16 blas-like API.

martin-frbg:
brada4, while your input is appreciated just like anybody else's; I would appreciate if you would not create the impression that you are in any way governing the decision making in this project. Adding bfloat16 support is already a work in progress, and I 
do not see how it would necessarily affect legacy platforms as long as nobody tries by force to use it there. 

I see no harm in adding a new TARGET either, even when it adds "only" one potentially major new feature - by that same reasoning, you might as well have argued against SKYLAKEX (only AVX512) or even HASWELL (AVX2) back when those were new.
The only technical problem blocking the merge is the symbol name conflict caused by the kludgy proof-of-concept implementation of the direct SGEMM pathway for SKX, and that one needed to be fixed in any case, to allow applying it to other  platforms.

PR #2785, opened 2020-08-18
Do not require pkg-config to generate the *.pc file

albertziegenhagel:
Generating the pkg-config file does not actually depend on pkg-config being available, so we can always just generate it.

This has the benefit to make the output of the build process more consistent. The pkg-config files will always be generated, no matter whether I install pkg-config before building OpenBLAS or afterwards.

Additionally it is more consistent with the *.pc files that are generated by LAPACK if it is enabled: Those files are being generated unconditionally.


Neumann-A (approved):
I cannot remember if the prefix variable is defined in openblas.pc or not. If not it would be nice if you could add it to the upstream PR.
BTW: are you back due to lapack-reference being added to vcpkg? Haven't seen you around for a while. Do you have any knowledge of the progress of the flang driver on windows? I would like to test the lapack-reference with a fortran compiler outside of vcpkg.

Wrong github repo.

PR #2788, opened 2020-08-20
[Test,WIP] Add R benchmarks at higher core counts

martin-frbg:


martin-frbg:
pretty much useless as r-base pulls in a ton of dependencies, including a timezone packages that wants to be set up interactively

PR #2803, opened 2020-08-31
Implementaion of dasum, sasum with AVX2 & AVX512 intrinsic

xiegengxin:
Implement dasum, sasum with AVX2 & AVX512 intrinsic for x86_64 architecture to improve the performance.

### DIFF
--- a/kernel/x86_64/KERNEL.HASWELL
+++ b/kernel/x86_64/KERNEL.HASWELL
@@ -100,3 +100,5 @@ ZTRSMKERNEL_RT	= ../generic/trsm_kernel_RT.c
 CGEMM3MKERNEL    =  cgemm3m_kernel_8x4_haswell.c
 ZGEMM3MKERNEL    =  zgemm3m_kernel_4x4_haswell.c
 
+SASUMKERNEL = sasum.c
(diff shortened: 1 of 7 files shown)


brada4:
I think you need to force-align presumed jump targets to palign(5) or so , but measure.
What about AMD ZEN and Intel COOPERLAKE ?

brada4:
AVX2 and AVX512 needs some warmup, otherwise for small samples it is hundreds of cycles unproductive

xiegengxin:
1. Align to 64 byte or 32 byte.
2. When input array size is small (less than 256), using SSE intrinsic to instead of AVX512/AVX2. The performance won't worse too much.

| input size | Optimizaion vs Before |
|---|---|
|16|0.96x|
|32|1.03x|
|64|1.01x|
|128|0.99x|
|256|0.99x|
|512|0.95x|
|1024|1.16x|
|2048|1.18x|

brada4:
From excellent resource at agner.org
> The first YMM or ZMM instruction takes 150-250 clock cycles -probably to start a power-up process

How does it work with intentionally unaligned input, and not ending so round?

PR #2815, opened 2020-09-02
Fix build with clang on s390x

ghost:
Adapt s390-specific code for build with LLVM/clang. The changes in this PR fall into these categories:

- avoid C language extensions that are specific to gcc
- replace mnemonics where gas and LLVM's assembler differ
- fix compiler version checks to work with clang
- add -ffp-contract=fast optimization flag for clang to behave the same as gcc (which uses that setting by default)

Note that support for DYNAMIC_ARCH=1 for s390x is still dependent on gcc at this point.


ghost:
Thanks for merging!

martin-frbg:
Thanks for fixing :)

PR #2816, opened 2020-09-02
Add basic support for Apple Vortex (ARM64) cpu

martin-frbg:
cpu detection via sysctl, setting compiler option -march=armv8.3; initial support by ARMV8 kernels


cpuid_arm64.c
```diff
@@ -45,6 +50,8 @@
 #define CPU_TSV110        9
 // Ampere
 #define CPU_EMAG8180	 10
+// Apple
+#define CPU_SILICON      11
```
Keno: Apple Silicon will be the Marketing term for all of Apple's ARM64 chips. The particular chip in question is an A12Z, which has 4 Tempest microarchitecture and 4 Vortex microarchitecture cores. Is SILICON the best name choice here? All the other names generally refer to the microarchitecture. Maybe `APPLE_VORTEX`?

martin-frbg: Right - I was not sure about the name, though I saw CPU_TEMPEST_VORTEX somewhere in brew. VORTEX it is then.

PR #2829, opened 2020-09-08
Fix DYNAMIC_ARCH=1 with clang s390x

ghost:
Adapt the s390-specific part of the build system to support DYNAMIC_ARCH=1 on s390x. Specifically

- simplify the detection of the supported architecture generation in the compiler,
- make the detection robust enough to handle LLVM/clang 6.x, which supports -march=z13 yet cannot handle z13-specific inline assembly,
- simplify the code for dynamically detecting the architecture at run-time, and
- add a Travis build for clang on s390x.

WiP (I probably need to revise): can Travis provide an Ubuntu 20.04 (aka focal) with a more recent LLVM/clang version?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,18 @@ matrix:
         - BTYPE="BINARY=64 USE_OPENMP=1"
 
     - <<: *test-ubuntu
+      os: linux
+      dist: focal
+      arch: s390x
+      compiler: clang
+      before_script:
+        - COMMON_FLAGS="DYNAMIC_ARCH=1 TARGET=Z13 NUM_THREADS=32"
+      env:
+        # for matrix annotation only
+        - TARGET_BOX=IBMZ_LINUX
+        - BTYPE="BINARY=64 USE_OPENMP=0 CC=clang"
+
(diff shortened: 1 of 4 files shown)


ghost:
LLVM/clang currently does not support OpenMP on s390x, yet the first version enabled that in the Travis build. Disabled and updated the branch.

martin-frbg:
Thanks. Ubuntu "Focal" should be available in Travis according to https://docs.travis-ci.com/user/reference/linux

ghost:
So, disabling OpenMP makes the clang builds pass. Further, Travis-CI can provide Ubuntu 20.04 with a more recent LLVM/clang version. I propose that we use Ubuntu 20.04 for the build with clang on s390x and have updated my branch accordingly. As in the commit message:

    Explicitly request Ubuntu 20.04 (codename focal) to get a recent
    LLVM/clang version 10.x and thereby cover all s390x architecture
    generations supported in OpenBLAS. Ubuntu 18.10's LLVM/clang 6.x cannot
    build the inline assembly in some of the Z13 and Z14 kernels.

PR #2849, opened 2020-09-22
fix a bug of trmm

xiaojiayuan111:
[description]：s/d/c/ztrmm failed when S/D/C/Z_GEMM_DEFAULT_P is not the multiples of S/D/C/ZGEMM_DEFAULT_UNROLL_M.
[solution]:  modify trmm_L.c so that  copy is aligned by S/D/C/ZGEMM_DEFAULT_UNROLL_M and works properly.


martin-frbg (commented):
The deletion of the SKYLAKEX/COOPERLAKE-specific settings appears to be unrelated ?

PR #2850, opened 2020-09-22
fix a bug of trmm

xiaojiayuan111:
[description]: s/d/c/ztrmm failed when S/D/C/Z_GEMM_DEFAULT_P is not the multiples of S/D/C/ZGEMM_DEFAULT_UNROLL_M.

[solution]: modify trmm_L.c so that  copy is aligned by S/D/C/ZGEMM_DEFAULT_UNROLL_M and works properly.

### DIFF
--- a/driver/level3/trmm_L.c
+++ b/driver/level3/trmm_L.c
@@ -122,6 +122,9 @@ int CNAME(blas_arg_t *args, BLASLONG *range_m, BLASLONG *range_n, FLOAT *sa, FLO
     if (min_l > GEMM_Q) min_l = GEMM_Q;
     min_i = min_l;
     if (min_i > GEMM_P) min_i = GEMM_P;
+    if( min_i > GEMM_UNROLL_M){
+        min_i =  (min_i / GEMM_UNROLL_M) * GEMM_UNROLL_M;
+      }
 
     START_RPCC();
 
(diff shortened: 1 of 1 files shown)


driver/level3/trmm_L.c
```diff
@@ -397,9 +426,13 @@ int CNAME(blas_arg_t *args, BLASLONG *range_m, BLASLONG *range_n, FLOAT *sa, FLO
 	STOP_RPCC(trmmcost);
       }
 
-      for(is = ls - min_l + min_i; is < ls; is += GEMM_P){
+      for(is = ls - min_l + min_i; is < ls; is += min_i){
 	min_i = ls - is;
 	if (min_i > GEMM_P) min_i = GEMM_P;
+	if (min_i > GEMM_UNROLL_M){
+	    min_i = (min_i / GEMM_UNROLL_M) * GEMM_UNROLL_M;
```
brada4, line 433: Do you mean `min_i=min_i % GEMM_UNROLL_M` ?

xiaojiayuan111: The two are obviously not equivalent. min_i = (min_i / GEMM_UNROLL_M) * GEMM_UNROLL_M means rounding down to UNROLL_M.

xiaojiayuan111: you can test strmm_kernel_16x4.S to verify my merge.  Set SGEMM_DEFAULT_P  in param.h to 28 and set the scale of the test matrix to m = n = k = 44 in ctest/sin3 . Then, you'll find that ctest fails as follows:
imng
more details as follows:
https://groups.google.com/forum/#!topic/openblas-dev/O2NemfVkrI8

Issue #2588, opened 2020-04-28
way to compile only BLAS functions for a given precision ?


franko:
Normally when compiling OpenBLAS all the functions for single and double precision floating point are generated.

I would like to know if it is possible to compile only the functions for single or double precision. I didn't find any mention about this in the documentation but it seems that Netlib's LAPACK allow to choose the precision when building.

Thanks in advance for any help.

martin-frbg:
This is currently not supported in OpenBLAS - our cmake build files do copy the BUILD_SINGLE etc. from Netlib's LAPACK but they are only used in building LAPACK. The plain Makefiles do not support these options at all right now, although they mostly have the BLAS functions grouped by precision. #2560 is somewhat related

franko:
@martin-frbg Thank you for the message. So not supported and it could be implemented with the Makefiles but require some works.

Probably there are not a lot of requests in this sense.

martin-frbg:
I do not remember seeing such a request, but as I mentioned something similar just came up. And if  the reference implementation already supports it, perhaps it would make sense to add it here as well.

brada4:
Only place I know this is done is R where only double precision parts of reference BLAS are compiled from large concatenated blas.f and lapack.f files. Resulting library then can be safely replaced with any more complete BLAS (and LAPACK) implementation. In practice though, even some R CRAN modules expect libRblas library to export single-precision or even CBLAS symbols.

martin-frbg:
Picking them apart turns out to be a bit tricky, as some "real" functions like SSCAL are called from "complex" code.

martin-frbg:
Untwined the gmake build (except CBLAS/LAPACKE for now) so far but not cmake yet. (Conversely, the support for this in Reference-LAPACK appears to be cmake only)

martin-frbg:
Should work for CMAKE builds after #2833 (although with BUILD_COMPLEX one currently has to specify BUILD_SINGLE as well, and same for COMPLEX16/DOUBLE). Unfortunately need to redo the changes for gmake builds

franko:
@martin-frbg Thank you very much, that's great. Unfortunately I cannot test right now so feel free to make your own tests and close the issue if you think it is ok.

franko:
Hi @martin-frbg, just tested using the cmake command:

```sh
[...]

PR #2852, opened 2020-09-22
Support building only a subset of variable types

martin-frbg:
second set of patches for #2588 - improves on #2833 by adding support for DYNAMIC_ARCH builds - still cmake-only but does not affect gmake builds as long as they are done for all precision types.

There is still a bug here that makes lapack-test enter an infinite loop (probably in the non-complex GEMM) when BUILD_COMPLEX or BUILD_COMPLEX16 is used on its own, i.e. without the respective BUILD_SINGLE or BUILD_DOUBLE.


martin-frbg:
Added gmake support and fixed lots of loose ends.

Issue #2391, opened 2020-02-07
[WIP] Port OpenBLAS to CK860


xianyi:
Support C-SKY 860 architecture.

martin-frbg:
basic support for the CK860 through generic C kernels is now available through #4438

PR #2853, opened 2020-09-23
Optimize the performance of daxpy by using universal intrinsics

Qiyu8:
## BackGround

"Universal Intrinsics" is  a set of APIs instead of a real unified set of intrinsics. This set of APIs is in a higher level than latter. Because it is oriented to the specific scenario of numerical computation, it can be easily converged to fewer common functions. Simultaneously, there is hardly a problem of no equivalent implementations on different platforms.
For developers, this mechanism can reduce the effort of adaptation to different platforms and still produce good performance. According to some test cases in OpenCV/Eigen/Numpy, the performance of the optimized functions is almost as good as the performance of manually optimized ones.  Here we optimized the Level-one daxpy function to demonstrate the power of USIMD.
## Benchmark

  Each test has run for 10 loops and then take the average result, The radio is the MFlops division of Baseline and optimized branch(*The bigger* the radio is, the *better* performance is achieved).

- X86 BaseLine(auto vectorized)

 | data size(10^3) | develop(float)               |
 | --------------- | ---------------------------- |
 | 4000            | 900.20 MFlops   0.008887 sec |
 | 8000            | 969.41 MFlops   0.016505 sec |

- X86-SSE2 enabled

  | data size(10^3) | usimd-daxpy(float)              | ratio |
  | --------------- | ----------------------------- | ----- |
  | 4000            | 1153.80 MFlops   0.006934 sec | 1.28  |
  | 8000            | 1606.26 MFlops   0.009961 sec | 1.65  |
  
- X86-FMA3 enabled

  | data size(10^3) | usimd-daxpy(float)              | ratio |
  | --------------- | ----------------------------- | ----- |
  | 4000            | 1187.21 MFlops   0.006738 sec | 1.31  |
  | 8000            | 1231.84 MFlops   0.012989 sec | 1.27  |
 
  
- X86-AVX2 enabled

  | data size(10^3) | usimd-daxpy(float)              | ratio |
  | --------------- | ----------------------------- | ----- |
  | 4000            | 2559.92 MFlops   0.003125 sec | 2.84  |
  | 8000            | 2559.96 MFlops   0.006250 sec | 2.64  |

## System Info

|           | x86                                      |
| --------- | ---------------------------------------- |
| Processor | Intel(R) Xeon(R) Gold 6161 CPU @ 2.20GHz |
| OS        | Windows Server 2008 R2 Enterprise        |
| Compiler  | MSVC14.06                                |

## Conclusion

- Compare with develop branch, when operating float dot, the new implementation has about 3x faster in AVX2, about 24%~65% faster in SSE2/SSE3. 
- Other architect such as NEON/VSX/MIPS/RISC-V  can be easily extended, just need to add a corresponding head file in `simd` module 

## Co-Author

@seiko2plus 

## Reference

1. https://software.intel.com/sites/landingpage/IntrinsicsGuide/
2. https://www.thetopsites.net/article/58234904.shtml
3. https://github.com/numpy/numpy/tree/master/numpy/core/src/common/simd
4. https://github.com/opencv/opencv/tree/master/modules/core/include/opencv2/core/hal


martin-frbg:
Interesting concept. I wonder if this could be combined with the sse2neon.h from https://github.com/DLTcollab/sse2neon to port existing x86 optimizations to arm

Qiyu8:
@martin-frbg Sure, I'm working on another dot product function, which can easily extend to arm/power architect with the help of sse2neon, BTW, The opencv project is extending to [RISC-V](https://github.com/opencv/opencv/pull/18228) recently by using universal intrinsics.

martin-frbg:
That is good to know - xianyi is already working on optimizing OpenBLAS for RISC-V (C-Sky 860 cpu, #2391) now that it has vector extensions.

Diazonium:
I find it strange that you show results with FMA4 on an Intel system, no Intel CPU has ever supported FMA4.
Only the AMD Bulldozer series has FMA4, Zen1 has undocumented and possibly broken support, and Zen2 removed FMA4 completely.

martin-frbg:
@diazonium Probably misnomer (SSE3/SSE4 ?) - the PR itself parallels their (accepted) contributions to numpy

Qiyu8:
@Diazonium Thanks for pointing it out, I found that I actually running FMA3 benchmark two times,  as a response to your FMA4 question, please see [here](https://www.guru3d.com/news-story/fma4-instruction-set-hiddenbut-is-working-on-amd-zen-processors.html). Running an adapted script in the Zen processors that sends FMA4 instructions. The FMA4 task fired off at the processor surprisingly did not get refused and got executed successfully. And openblas has set the HAVE_FMA4 macro in Zen processors. so we will keep the HAVE_FMA4 instructions here.

xianyi:
@Qiyu8 I almost finished BLAS3 on RISC-V Vector 0.7.1 version

Diazonium:
> 
> 
> @Diazonium Thanks for pointing it out, I found that I actually running FMA3 benchmark two times, as a response to your FMA4 question, please see [here](https://www.guru3d.com/news-story/fma4-instruction-set-hiddenbut-is-working-on-amd-zen-processors.html). Running an adapted script in the Zen processors that sends FMA4 instructions. The FMA4 task fired off at the processor surprisingly did not get refused and got executed successfully. And openblas has set the HAVE_FMA4 macro in Zen processors. so we will keep the HAVE_FMA4 instructions here.

Yes Zen1 cores are able to execute FMA4 instructions, but we do not know why they are not advertised in the CPUID. It is possible that some of the instructions may trigger have rare corner cases where they are have bugs in Zen 1, which could manifest as precision issues or any other bugs.

I think it is a bad idea to enable any FMA4 codepath by default on Zen 1 cores, because we do not want to risk package maintainers shipping binaries that use undocumented instructions on some machines. Many people do not know or care about SIMD, because they are just using NumPy, SciPy, etc. that ends up using OpenBLAS, if FMA4 causes rare precision or stability issues it would be really confusing and annoying for them.

I do think having FMA4 is great, but enabling it on Zen 1 cores should be disabled by default in Makefile.rule. Users willing to compile their own build of OpenBLAS would be free to experiment with turning it on.

And FMA 4 is definitely removed in Zen 2 cores.

martin-frbg:
@Diazonium pretty sure HAVE_FMA4 is reported by getarch on Zen1 because the relevant CPUID bit is actually still set, but you are right in that e.g. Agner Fog considers it "not officially supported - do not use". Also pretty sure that none of the current ZEN-specific functions in OpenBLAS makes use of it.

Qiyu8:
@martin-frbg @Diazonium Good advice, I have removed the default support for FMA4 instruments. we should leave this switch to users who want to try it out. 
@xianyi Glad to see RISC-V Vector supported on BLAS3, we can use universal intrinsics framework to support RISC-V on BLAS1, what do you think?

PR #2873, opened 2020-09-30
Check for __linux rather than linux in cpuid code and benchmarks 

### LINKED ISSUE #2871: Wrong defines checked when compiling with -std=c11
There is some code which checks for C11 features, such as atomics. Usually C11 is enabled by passing `-std=c11`

However there are many checks like `#ifdef linux` which will then fail due to `linux` not being defined. In general (IIRC) the symbols from the compiler all are of the form `__foo__`, especially starting with a double underscore

I think it would be a good idea to check and replace such macros and add CI for using standard C11

martin-frbg:
std=c11 drops the non-underscored macros, potentially breaking build-time cpu identification (in the benchmarks,
only seeding of the random number generator would have been affected).
fixes #2871 as suggested by Flamefire there


Flamefire:
It seems that is not the full fix as the correct define is `__linux__` according to https://stackoverflow.com/questions/6649936/c-compiling-on-windows-and-linux-ifdef-switch#comment79735932_6649992 although using `__linux` seems to work better than just `linux`.
See also https://github.com/libunwind/libunwind/pull/201 stating the same as the SO comment

martin-frbg:
gcc, clang and nvcc (formerly PGI pgcc) definitely provide both

Flamefire:
Yes, I can confirm that too. But I don't have access to an  Ubuntu Trusty to verify what the others said (see links) so it might be possible, that on certain systems this could be an issue again. From the sources I could find the one with double underscore on both sides seems to be the safest bet.

PR #2874, opened 2020-10-01
Avoid out of bounds access on invalid memory free 

Flamefire:
I found an out of bounds access when `blas_memory_free` is called with an invalid free_area. In particular:

```
  while ((position < NUM_BUFFERS) && (memory[position].addr != free_area))
    position++;
```

After this loop `position` is either less than `NUM_BUFFERS` and pointing to a memory entry whose `addr` is equal to `free_area` or `position` is equal to `NUM_BUFFERS`. The following check `if (memory[position].addr != free_area) goto error;` checks the loop condition again in the "happy case" or will access memory outside of the `memory` array which has only `NUM_BUFFERS` entries. So the check should be `if (position >= NUM_BUFFERS) goto error;` instead, no need for checking `addr` again.

I did that change but moved the existing addr check inside the `ifdef DEBUG` block. I'd usually add an `assert(...)` but that doesn't seem to be used here which I find strange. IMO it's useful to check for preconditions that must be given so violations can be found when compiled without `-DNDEBUG`, but maybe OpenBLAS is usually compiled without that define so adding asserts without removing them in release/optimized builds could impact performance.

I also found that TABs were used at random locations making it hard to follow the code over multiple `#if` guarded blocks especially as the TAB width seemed to vary between 2 and 8. I did a search&replace from TABs to 8 spaces and fixed the differing tab sizes manually. I'd recommend disabling whitespace comparison to review this PR to verify that no other functional changes apply

### DIFF
--- a/driver/others/memory.c
+++ b/driver/others/memory.c
@@ -80,7 +80,7 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #undef COMPILE_TLS
 #endif
 
-#if defined(__GLIBC_PREREQ) 
+#if defined(__GLIBC_PREREQ)
 #if !__GLIBC_PREREQ(2,20)
 #undef COMPILE_TLS
 #endif
@@ -161,7 +161,7 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #if defined(OS_WINDOWS) && (defined(__MINGW32__) || defined(__MINGW64__))
 #include <conio.h>
 #undef  printf
-#define printf	_cprintf
+#define printf _cprintf
 #endif
 
 #ifdef OS_LINUX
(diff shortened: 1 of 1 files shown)


martin-frbg:
The general problem with assert in release builds of a library is that you never know what program you will be taking down with you - memory corruption is certainly not nice but it _could_ be happening in a context where it is harmless enough that it does not justify killing a long-running simulation. (Strange that this bug was still in the code after all the runs of valgrind and asan, maybe it was triggered by a recent change ?)

Flamefire:
> The general problem with assert in release builds of a library

That's why asserts are usually disabled in release builds by defining `-DNDEBUG`, see http://www.cplusplus.com/reference/cassert/assert/. Build systems such as CMake usually define this automatically for release builds and don't for debug builds. So I guess this would need to be added to the Makefiles here

> you never know what program you will be taking down with you - memory corruption is certainly not nice but it could be happening in a context where it is harmless enough that it does not justify killing a long-running simulation.

Well you can argue the same way the other way round: You can never know what trouble you can cause for a long-running simulation that after days of running spits out wrong results because some memory was overwritten without noticing or a should-be impossible race condition occurred.
The other benefit is a message to the programmer reading the code. Because the assert has the semantic "This cannot/must not happen" it allows for easier reasoning about the code following. Obviously the assert itself must be crafted so that it actually states a fact ensured by something else, like in this case.

> Strange that this bug was still in the code after all the runs of valgrind and asan, maybe it was triggered by a recent change ?

I think the reason is that this code never employs an OOB access unless something is seriously wrong, i.e. a non-allocated memory region is passed which I think is not tested. Maybe this should be tested? E.g. passing NULL or a bogus value? It doesn't seem to be a public API so maybe it is not required to be tested.

martin-frbg:
Looking at this again, I am fairly convinced that the "short-circuit" evaluation semantics of C should ensure that the right hand side of this conditional never gets executed when position is no longer <NUM_BUFFERS, so it looks more like an optimizer bug to me if the invaiid access actually occurs (or a false positive from the analyzer, if it only claims an invalid access here).

Flamefire:
No I'm talking about the if condition below the loop. Take a look at the diff without spaces to see what I changed. That should be fairly obvious I hope.
It's not from an analyzer, I discovered it manually. And as mentioned it is an almost impossible situation (eg double free)

martin-frbg:
I see now (after some sleep...), but I am not sure what the original intention was. This code does have a particularly colorful history (e.g. #1179), so has probably never been entirely correct (just that the failure became less likely with each iteration, but the code version before #1179 is already the original GotoBLAS2)

PR #2881, opened 2020-10-05
add fninit to reset fpu registers before assembler routines

### LINKED ISSUE #2709: Errors when using OpenBLAS through NumPy on Windows 10 2004
This issue only appears on Windows 10 2004 (19041). It does not appear on Windows 10 1909 (18363).

On a fresh install of NumPy from pip on a 2004 machine, e.g,

```
pip install numpy ipython
```

open `ipython` and enter

```
import numpy as np
a = np.arange(13 * 13, dtype=np.float64)
a.shape = (13, 13)
a = a % 17
va, ve = np.linalg.eig(a)
```

This produces an error:

```
 ** On entry to DGEHRD  parameter number  2 had an illegal value
---------------------------------------------------------------------------
LinAlgError                               Traceback (most recent call last)
<ipython-input-1-bad305f0dfc7> in <module>
      3 a.shape = (13, 13)
      4 a = a % 17
----> 5 va, ve = np.linalg.eig(a)

<__array_function__ internals> in eig(*args, **kwargs)

c:\anaconda\envs\py-pip\lib\site-packages\numpy\linalg\linalg.py in eig(a)
   1322         _raise_linalgerror_eigenvalues_nonconvergence)
   1323     signature = 'D->DD' if isComplexType(t) else 'd->DD'
-> 1324     w, vt = _umath_linalg.eig(a, signature=signature, extobj=extobj)
   1325
   1326     if not isComplexType(t) and all(w.imag == 0.0):

c:\anaconda\envs\py-pip\lib\site-packages\numpy\linalg\linalg.py in _raise_linalgerror_eigenvalues_nonconvergence(err, flag)
     92
     93 def _raise_linalgerror_eigenvalues_nonconvergence(err, flag):
---> 94     raise LinAlgError("Eigenvalues did not converge")
     95
     96 def _raise_linalgerror_svd_nonconvergence(err, flag):

LinAlgError: Eigenvalues did not converge
```

What makes me suspect that it is OpenBLAS related is:

1. It does not happen when using MKL.
2. It does not happen when using LAPACK-lite that ships with NumPy and is used as a fallback.

On the other hand, it may likely be a Windows bug since:

1. It does not occur before 2004.
2. It does not occur when using 32-bit Python on the same system.
3. It does not occur when running Pythin in WSL on the same machine whether using WSL1, which runs Linux binaries directly on Windows, or WSL2, which runs them in a hypervisor.

Any help is appreciated.

The related NumPy issue is numpy/numpy#16744.

mattip:
closes gh-2709 by adding a call to ~`fninit`~ `emms` to clear the FPU registers.

This seems to be the safe path, unless it can be proven that the kernels will only be called from code that clears the FPU registers and does not set them.

edit: wrong issue number
edit: the original PR had `fninit`, as per review it was changed to `emms`

### DIFF
--- a/kernel/x86_64/amax.S
+++ b/kernel/x86_64/amax.S
@@ -54,6 +54,8 @@
 
 	PROLOGUE
 	PROFCODE
+	
+	fninit
 
 	salq	$BASE_SHIFT, INCX
 
--- a/kernel/x86_64/amax.S
+++ b/kernel/x86_64/amax.S
@@ -55,7 +55,9 @@
 	PROLOGUE
 	PROFCODE
 	
-	fninit
+#ifdef WINDOWS_ABI
+	emms
+#endif
 
 	salq	$BASE_SHIFT, INCX
 
--- a/kernel/x86_64/asum.S
+++ b/kernel/x86_64/asum.S
@@ -49,7 +49,8 @@
 
 	PROLOGUE
 	PROFCODE
-
+	
+	fninit
 	fldz
 	testq	M, M
 	jle	.L999
--- a/kernel/x86_64/asum.S
+++ b/kernel/x86_64/asum.S
@@ -50,7 +50,10 @@
 	PROLOGUE
 	PROFCODE
 	
-	fninit
+#ifdef WINDOWS_ABI
+	emms
+#endif
+
 	fldz
 	testq	M, M
 	jle	.L999
--- a/kernel/x86_64/dot.S
+++ b/kernel/x86_64/dot.S
@@ -49,6 +49,7 @@
 
 	PROLOGUE
 	PROFCODE
+	fninit
 
 	salq	$BASE_SHIFT, INCX
 	salq	$BASE_SHIFT, INCY
--- a/kernel/x86_64/dot.S
+++ b/kernel/x86_64/dot.S
@@ -49,7 +49,10 @@
 
 	PROLOGUE
 	PROFCODE
-	fninit
+
+#ifdef WINDOWS_ABI
+	emms
+#endif
 
 	salq	$BASE_SHIFT, INCX
 	salq	$BASE_SHIFT, INCY
--- a/kernel/x86_64/iamax.S
+++ b/kernel/x86_64/iamax.S
@@ -59,6 +59,7 @@
 
 	PROLOGUE
 	PROFCODE
+	fninit
 
 	salq	$BASE_SHIFT, INCX
 
--- a/kernel/x86_64/iamax.S
+++ b/kernel/x86_64/iamax.S
@@ -59,7 +59,10 @@
 
 	PROLOGUE
 	PROFCODE
-	fninit
+
+#ifdef WINDOWS_ABI
+	emms
+#endif
 
 	salq	$BASE_SHIFT, INCX
 
--- a/kernel/x86_64/izamax.S
+++ b/kernel/x86_64/izamax.S
@@ -59,6 +59,7 @@
 
 	PROLOGUE
 	PROFCODE
+	fninit
 
 	salq	$ZBASE_SHIFT, INCX
 
--- a/kernel/x86_64/izamax.S
+++ b/kernel/x86_64/izamax.S
@@ -59,7 +59,10 @@
 
 	PROLOGUE
 	PROFCODE
-	fninit
+
+#ifdef WINDOWS_ABI
+	emms
+#endif
 
 	salq	$ZBASE_SHIFT, INCX
 
--- a/kernel/x86_64/nrm2.S
+++ b/kernel/x86_64/nrm2.S
@@ -50,6 +50,7 @@
 	PROLOGUE
 	PROFCODE
 
+	fninit
 	fldz
 	testq	M, M
 	jle	.L999
--- a/kernel/x86_64/nrm2.S
+++ b/kernel/x86_64/nrm2.S
@@ -50,7 +50,10 @@
 	PROLOGUE
 	PROFCODE
 
-	fninit
+#ifdef WINDOWS_ABI
+	emms
+#endif
+
 	fldz
 	testq	M, M
 	jle	.L999
--- a/kernel/x86_64/qconjg.S
+++ b/kernel/x86_64/qconjg.S
@@ -41,7 +41,10 @@
 
 	PROLOGUE
 	PROFCODE
-	fninit
+
+#ifdef WINDOWS_ABI
+	emms
+#endif
 
 	fldz
 	FLD	1 * SIZE(ARG1)
--- a/kernel/x86_64/qconjg.S
+++ b/kernel/x86_64/qconjg.S
@@ -41,6 +41,7 @@
 
 	PROLOGUE
 	PROFCODE
+	fninit
 
 	fldz
 	FLD	1 * SIZE(ARG1)
--- a/kernel/x86_64/qdot.S
+++ b/kernel/x86_64/qdot.S
@@ -58,7 +58,9 @@
 
 	PROLOGUE
 
-	fninit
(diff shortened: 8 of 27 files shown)


mattip:
My sed foo on windows was not strong enough to consistently format the code. Do you have a preference for blank lines before/after/none?

mattip:
xref numpy/numpy#16744

carlkl:
each fninit should follow something like that:

```C
#ifdef CONSISTENT_FPCSR
      __asm__ __volatile__ ("fnstcw %0"  : "=m" (queue -> x87_mode));
#endif
```

But I have no idea how to implement this directly in assembler code.

martin-frbg:
Note that this one-liner is only the instruction to grab the current setting and store it somewhere in memory (to then change individual bits and put it back via fnldcw). Anyway I am more in favor of using the generic C functions on Windows unless somebody demonstrates that a similar problem exists on other operating systems. I suspect only a handful of the 27 files is
actually used by any reasonably recent TARGET, about half of them appear to be placeholders for the unimplemented quadruple precision mode (nrm2.S being the big exception)

mattip:
Happy to close this in favor of an alternative. We may implement this (or a variant if numpy/numpy#16744 comes up with a NumPy-based solution) as a temporary patch for the upcoming NumPy release, pending a better fix in OpenBLAS or a OS fix from Microsoft. Pinging @charris for thoughts.

charris:
@mattip I don't have the background to judge the proposed fixes, but I am optimistic that things will get fixed at some point. Thanks for tracking down the problem, that was a nice bit of detective work. We can put a short test in `numpy/__init__.py` that detects the problem and recommends an upgrade if Microsoft provides a fix, or MKL if they don't. I won't complain if OpenBLAS fixes the problem, but working around compiler/library problems eventually clutters the code.  I'm pretty sure Microsoft will fix it at some point, it makes no sense to not fix a bug that appears to randomly manifest here or there, no one could trust their code after that.

martin-frbg:
Confirmed now that it is only (z)nrm2.S and (z)sum.S that are actually used by any x86_64 target. (Where the sum.S and zsum.S are trivial hacks of their respective asum files  that i made in #2072 to implement the non-absolute xSUM extensions - the active xASUM versions are SSE2 and presumably use some shift trickery to kill the signs)

martin-frbg:
Digging deeper, the CONSISTENT_FPCSR option does not need to manipulate individual bits in the fpu state, but what it does is
store the initial state seen when the thread server started, and push this to any newly created thread. Offhand I see no easy way to access what is `queue->x87_mode` in the context of blas_server_win32.c from where we are in nrm2.S. FNINIT would set the fpu mode to "round to nearest, all exceptions masked, 64bit precision" - which of these is incompatible with numpy expectations ?

mattip:
Truthfully I am not sure. I don't think NumPy consciously sets the FPU mode. There are probably users who set various compilation flags to specify floating point accuracy/speed, and this may mess with their expectations. Maybe @charris or @bashtage have more insight. 

The deeper we go into this rabbit hole, the more I think we should push Microsoft to fix the problem. Are the c-based loops that much less performant? Are there benchmarks that stress the [x]nrm2 loops, or could we write one?

charris:
NumPy uses round to even. The precision probably doesn't matter in practice, but the masking could be problematic, I don't know how it currently operates.

martin-frbg:
The C functions are probably not much worse than the assembly with an added FNINIT (hence my alternate PR) but carlkl pointed out that the MS-created problem had better be fixed "somewhere" rather than be left to blow up somebody else later. (OTOH i would really like to get the 0.3.11 release out, though it was not this issue in particular holding it up)

mattip:
Maybe that is an easier ask from Microsoft: could we get a small piece of code we could run after `fmod` that would clean up whatever mess they are leaving behind? We do this in NumPy to clear divide-by-zero and other errors. I think it is just the `ST[0]` register,  they would know better.

Edit: added this question to the microsoft developer [community issue](https://developercommunity.visualstudio.com/content/problem/1208774/fpu-exception-in-fmod0-x-in-windows-10-version-200.html)

mattip:
> hence my alternate PR

Which?

mattip:
Ahh, got it. gh-2882

martin-frbg:
I have merged "my" workaround for now but remain open to more efficient solutions.

carlkl:
The `finit` instruction at the beginning of each assembler file breaks OpenBLAS `CONSISTENT_FPCSR` and it does not close https://github.com/xianyi/OpenBLAS/issues/2709. The reason is that `finit` changes the fpu precision mode to extended mode and this would show up in the numpy tests. A similar problem is described here: https://github.com/numpy/numpy/pull/9580 _BUG: Add hypot and cabs functions to WIN32 blacklist._
A better solution is to add the `emms` instruction instead of `finit`.

bashtage:
Too bad `_mm_empty` was removed in VS 2015 which issues emms.

See https://github.com/avaneev/avir/issues/7#issuecomment-515835647 who were using it in a similar way it seems.

martin-frbg:
Are we sure now that fnclex+emms fixes it ?

mattip:
`fnclex` alone does not clear anything. `fnclex` + `emms` does clear the errors. Now trying just `emms` alone.

martin-frbg:
Ah, good. From what I read about `emms` I do not expect it to clear the exception flag on its own though. And maybe this should go somewhere in the initialization code of driver/others/blas_server_win32.c rather than the few assembly kernels that use the fpu themselves ?

mattip:
`emms` alone is enough. Should I update the PR to reflect that or close it?

mattip:
If I could, I would put a call to `emms` in NumPy just after the call to `fmod`. I wonder if I can compile with mingw just that routine and link it with MSVC

carlkl:
@mattip, thanks for testing. In the case  `emms` is sufficient means _only_ the stack was corrupted. 

'emms' can be compiled with MASM:

```asm


.code
_mm_empty PROC
     push rbp
     mov rbp , rsp 
     fnclex
     emms 
     nop
     pop rbp
     ret
_mm_empty endp
end
```
[_mm_empty_x64.zip](https://github.com/xianyi/OpenBLAS/files/5365927/_mm_empty_x64.zip)

I added the C sources for mingw-w64 and assembler code for MASM with the object-files in the attachment.

martin-frbg:
Update if you like, with an `ifdef OS_WINDOWS` around it (unless you can somehow put that `emms` into numpy, but OTOH some other half-MSVC-half-mingw code could create the same problem "in" OpenBLAS).

mattip:
@carlkl thanks, I will give it a shot over at NumPy. I see my previous mistake was I missed `ret`. 

I will update the PR.

carlkl:
@martin-frbg, all WIN64 programs using UCRT and OpenBLAS at the same time could fail the same way. So I think it's a good idea. to make OpenBLAS robst against such failures. This could be addressed in `blas_server_win32.c` as well IMHO.

Something like this:
```C
__asm__ __volatile__ ("emms");
```

martin-frbg:
@carlkl thanks, the "how" is clear to me, but I am unsure about the "where" - probably the gotoblas_init constructor in memory.c rather than the blas_server that should only get involved in the multithreaded case but I bet there are some MS subtleties w.r.t initializations of dlls vs. static code.

carlkl:
It should be enough to clean the stack once at the initialization of OpenBLAS. Maybe enclosed with something like that:

```bash
#if defined(OS_WINDOWS) && defined(__MINGW64__) && defined(__WIN64__)
```

martin-frbg:
#2889 has that "global" approach now (though untested)

Rolafawaz (approved):

PR #2883, opened 2020-10-10
Minor CMAKE fixes

### LINKED ISSUE #2872: Superflous statements in CMake code
I noticed that this line is not required: https://github.com/xianyi/OpenBLAS/blob/68b1713c300ac152d1efcb3c02f0c59fafcd39e1/cmake/system_check.cmake#L125 CMake already outputs:

> -- Looking for stdatomic.h
> -- Looking for stdatomic.h - found

And `if (HAVE_C11 EQUAL 1)` can be shortened to `if (HAVE_C11)`, similar for others

martin-frbg:
1. remove extra verbosity in check for stdatomic.h (fixes #2872)
2. Abort the configuration step for a DYNAMIC_ARCH build if the toplevel source directory already contains a file config_kernel.h
(from a preceding build attempt with gmake, and source of weird errors as some assembler files would be missing crucial defines if this stale file would get included instead of the dynamically generated one specific to each TARGET_CORE)

### DIFF
--- a/cmake/arch.cmake
+++ b/cmake/arch.cmake
@@ -79,7 +79,7 @@ if (DYNAMIC_ARCH)
       string(REGEX REPLACE "-march=native" "" CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
     endif ()
     if (DYNAMIC_LIST)
-	set(DYNAMIC_CORE ${DYNAMIC_LIST})
+      set(DYNAMIC_CORE PRESCOTT ${DYNAMIC_LIST})
     endif ()
   endif ()
 
--- a/cmake/arch.cmake
+++ b/cmake/arch.cmake
@@ -1,4 +1,3 @@
-##
 ## Author: Hank Anderson <hank@statease.com>
 ## Description: Ported from portion of OpenBLAS/Makefile.system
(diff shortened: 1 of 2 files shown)


cmake/arch.cmake
```diff
@@ -80,10 +79,15 @@ if (DYNAMIC_ARCH)
       string(REGEX REPLACE "-march=native" "" CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
     endif ()
     if (DYNAMIC_LIST)
-	set(DYNAMIC_CORE PRESCOTT ${DYNAMIC_LIST})
+      set(DYNAMIC_CORE PRESCOTT ${DYNAMIC_LIST})
     endif ()
   endif ()
 
+  CHECK_INCLUDE_FILE ("${PROJECT_SOURCE_DIR}/config_kernel.h" TRAP)
```
Flamefire, line 86: You probably wanted `if(EXISTS ${PROJECT_SOURCE_DIR}/config_kernel.h)`?

martin-frbg: CHECK_INCLUDE_FILE worked, but your suggestion certainly looks like the more elegant solution, thanks.

PR #2887, opened 2020-10-12
Use C kernels for x86_64 SUM on Windows

martin-frbg:
second part of workaround for #2709


mattip:
My local [build failed](https://paste.ubuntu.com/p/J6V9w9G7nJ/), any ideas ?

martin-frbg:
Looks like you are missing the
```
if (DEFINED HAVE_SSE3)
    set (KERNEL_DEFINITIONS "${KERNEL_DEFINITIONS} -msse3")
 endif()
```
addition to cmake/system.cmake (bb74dd2 - sorry for not going through the PR mechanism with that one but it was a late night fix for a merge conflict)

mattip:
No cookie. I still get the same errors as in #2882. Is there a more systematic way to do this?

martin-frbg:
hm strange. The other fpu-using .S files are all unused AFAICT, perhaps the uncleared exception is now triggered by something in gfortran's compilation of the lapack code.

martin-frbg:
btw this PR is likely to be obsolete as soon as #2888 gets merged. Now all we need is a decent reimplementation of nrm2 and OpenBLAS is going to be just another hapless victim of the Windows bug. :-/

PR #2891, opened 2020-10-13
Fix several bugs and omissions from the BFLOAT16 rename

martin-frbg:


RajalakshmiSR:
@martin-frbg Thanks for doing this. We  also have few old entries in interface/Makefile...*shstobf16 and *shdtobf16

RajalakshmiSR:
Looks like this is giving error in many files. I created a PR https://github.com/xianyi/OpenBLAS/pull/2892 for this. Tested in POWER9.

martin-frbg:
Yes, sorry for the somewhat chaotic partial fixing - other work kept intervening. Have a fix for test/Makefile that I will push after yours.

PR #2895, opened 2020-10-13
Fix remaining build errors related to bfloat16 and cmake

martin-frbg:


test/Makefile
```diff
@@ -214,9 +214,11 @@ endif
 
 
 
-ifeq ($(BUILD_BFLOAT16),1)
-level3 : test_sbgemm
-endif
+#ifeq ($(BUILD_BFLOAT16),1)
+#level3 : test_sbgemm sblat3 dblat3 cblat3 zblat3
```
RajalakshmiSR: @martin-frbg Can you help me understand why we are commenting this? I expect to run test_sbgemm when BUILD_BFLOAT16=1

martin-frbg: Something broken in github or my ways of interacting with it - this is absolutely not the version of test/Makefile that should be in this PR. Retrying..

martin-frbg: Seems to have worked now, hope it does not spontaneously revert again as I have done nothing differently now.

PR #2896, opened 2020-10-14
Add compiler flag for SSE4 where available

martin-frbg:


Qiyu8:
@martin-frbg Thank you for your attention, I will create a new PR adding double precision universal intrinsics for X86/ARM, The float64 implemention here is not optimized.

martin-frbg:
Sure - this was just a quick stopgap implementation to unblock 0.3.11 without having to revert your contributions. I have removed my hack here.

PR #2899, opened 2020-10-16
Add the support for RISC-V Vector.

damonyu1989:
Change-Id: Iae7800a32f5af3903c330882cdf6f292d885f266

Add the risc-v vector  support for openblas. And the vector spec we based is 0.7.1 .


xianyi:
Hi @damonyu1989 ,

I will merge this patch into risc-v branch at first. 

Xianyi

Issue #645, opened 2015-09-13
git describe shows the wrong tag


jeanfrancisroy:
Forgive me if the issue is only on my side, but it seems that there is a problem with the tags on this Git repository.

I am trying to create an git version of the ArchLinux OpenBLAS package, and the standard way of getting the version number is to execute the following :

$ git describe --tags

Currently, it gives the following :
v0.2.9.rc1-679-ga96a4cb

But it seems to me that the latest tag should be at least v0.2.14. Is something wrong with the way that tags are handled in the git repo?

Thanks!

martin-frbg:
This appears to be a peculiarity of the "develop" branch (which may rightly be git's default for "HEAD"), probably this was tagged only once very early in its lifetime. If I explicitly clone with "-b master" the "git describe" reports v0.2.14 as expected.

martin-frbg:
I remember the version number in the tag on the develop branch being "correct" for some while in the 0.2.16 timeframe, but a current git clone operation again gets an outdated version number in the tag - v0.2.17-268-g8f9975e (note the hash after the "g" correctly matches the current state)
I suspect this tagging business is a manual step that is all too easy to miss when doing releases, and version numbers probably make sense for static releases only, while the develop branch is best characterised by the hash only (though it may be nice to know what its last versioned ancestor was).
Thoughts ?

jeromerobert:
Yes you are right, this bug should be closed. In `v0.2.9.rc1-679-ga96a4cb` only `a96a4cb` is meaningful, that's how git and `git describes` works.

martin-frbg:
Thanks for your insight. Closing.

martin-frbg:
FWIW, in https://github.com/Reference-LAPACK/lapack/issues/123 this is now claimed to be the result of using "git tag" rather than "git tag -a"

zerothi:
@martin-frbg 

Well not exactly, `git describe --tags` takes _all_ tags into consideration. And the tags for `v0.2.19` in `master` is already annotated.  
However, the thing about the OpenBLAS development is that the `master` branch the release branch, whereas the `develop` branch is... well... the development branch. The tags are created on hash's in the `master` branch. Hence `git describe` in the `develop` branch cannot _reach_ the hash in the `master` branch.   
Do this and you will see:

    $> git clone https://github.com/xianyi/OpenBLAS.git
    $> git describe --tags # or git describe (`--tags` simply takes the most recent tag, lightweight or annotated)
[...]

PR #2903, opened 2020-10-16
Merge release

mattip:
This is needed to convince `git describe --tags --abbrev=8` to add the 0.3.10 tag. Before this PR HEAD on `develop` reported
`v0.3.9-849-gdf706670`, after this merge it reports `v0.3.10-493-g0eda7ac2`. 

Merging the release back into the develop branch would be helpful after each release for downstream packagers [like numpy](https://anaconda.org/multibuild-wheels-staging/openblas-libs/files): note the file name is `openblas-v0.3.9-452-g349b722d` even though it is built off a commit *after* the 0.3.10 release.


martin-frbg:
I am not entirely convinced this would work without messing up the develop branch shortly before 0.3.11 :(

mattip:
Sorry, feel free to close it then, or maybe leave it open as a reminder to merge the release branch back to develop shortly after the release?

martin-frbg:
Most likely that is just another git feature that I am not really familiar with... but I see we have been there before (#645) with some arguing that only the hash itself should be considered significant when building from `develop`

mattip:
In that issue there is this comment, which is an opinion I only partially agree with
>  Yes you are right, this bug should be closed. In v0.2.9.rc1-679-ga96a4cb only a96a4cb is meaningful, that's how git and git describes works.

That may be how `git describe` works, but it is not how your users expect `git describe --tags` to work. After a release, I would expect the active branch to continue after the tag. Is it is a big ask to merge the release branch back into develop for the convenience of downstream packagers after the release is tagged? This PR changes no actual files, it only changes the appearance of tags on branches.

martin-frbg:
I'm ok with merging back after the release, and I would merge your PR if I had a clear idea how the history of the `develop` branch would look after its application (i.e. what hash would get tagged as 0.3.10 retroactively, and if `git log` would still show the expected sequence of commits). With 0.3.9 I am pretty sure the mergeback of the tag must have happened as a side effect of a trivial merge conflict resolution - differing version strings in the (c)makefiles or something like that)

mattip:
cool thanks. I will close this and try to remember to make sure it is done after the upcoming release. I hope it goes smoothly.

Issue #2911, opened 2020-10-18
OpenBLAS-0.3.11 fails to compile on ppc64le


susilehtola:
Updating the Fedora package of OpenBLAS to version 0.3.11 fails, since the package does not build on ppc64le. I'm getting a bunch of errors like
```
../kernel/power/sgemm_kernel_power10.c:131:11: error: '__builtin_mma_xvf32gerpp' requires the '-mmma' option
  131 |           __builtin_mma_xvf32gerpp (&acc2, rowB[i], rowA[j+1]); \
      |           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
which means you should probably add this option to the makefiles.

Also, I noticed a bunch of macro redefinitions in the power8 kernels, e.g.
```
../kernel/power/dtrmm_kernel_16x4_power8.S:85: note: this is the location of the previous definition
   85 | #define STACKSIZE 320
      | 
../kernel/power/dtrsm_kernel_LT_16x4_power8.S:51: warning: "STACKSIZE" redefined
   51 | #define STACKSIZE 520
      | 
../kernel/power/dtrsm_kernel_LT_16x4_power8.S:50: note: this is the location of the previous definition
   50 | #define STACKSIZE 320
      | 
```
which you may want to address.

susilehtola:
Full build log
[build.log.gz](https://github.com/xianyi/OpenBLAS/files/5397873/build.log.gz)

martin-frbg:
Thanks, looks like 0.3.11 is going to set a new record for short-livedness. At least this mma error does not appear to be my fault (the redefinitions are, I'd left the old values in when I expanded the stack to save&restore additional registers). Good to know you have access to POWER10 though...

martin-frbg:
Ah, that is "only" a DYNAMIC_ARCH build on POWER8 ? Guess we may need to add some core-specific compiler options like it is already necessary for AVX512 on x86_64.

martin-frbg:
The POWER10 support is only included when GCC >10.2 , and unfortunately I cannot test my current PR on Fedora32 in the OpenPOWER cloud as even with gcc 10.2.1 and the latest binutils (2.34-5.fc32) I hit an assembler error `invalid switch -mpower10`
@RajalakshmiSR do we need to (test for) a specific `as` or `binutils` version for POWER10 as well ?

martin-frbg:
[...]

PR #2916, opened 2020-10-18
Clean up duplicate definitions in POWER8 kernels and fix power10 option passing

martin-frbg:
for #2911


Makefile.power
```diff
@@ -10,7 +10,7 @@ USE_OPENMP = 1
 endif
 
 ifeq ($(CORE), POWER10)
-COMMON_OPT += -Ofast -mcpu=power10 -mtune=power10 -mvsx  -fno-fast-math
+COMMON_OPT += -Ofast -mcpu=power10 -mtune=power10 -mvsx -mmma  -fno-fast-math
```
RajalakshmiSR: We dont need to add -mmma as -mcpu=power10 takes care of enabling mma options by default.
If we add the binutils check, we don't need this change here.

martin-frbg: Yes, this was only my initial reaction to the issue ticket - the actual problem must be that this line has no effect when COMMON_OPT is passed in via the `make` command line.

kernel/power/strmm_kernel_16x8_power8.S
```diff
@@ -12,7 +12,7 @@ the documentation and/or other materials provided with the
 distribution.
 3. Neither the name of the OpenBLAS project nor the names of
 its contributors may be used to endorse or promote products
-derived from this software without specific prior written permission.
+derived from this software without specific prior written permission.
```
h-vetinari: Spurious spaces here...

martin-frbg: Thanks, probably introduced them by using the github web editor. Wish we were back at that level of "problems" though.

PR #2925, opened 2020-10-19
Add binutils version check as prerequisite for POWER10 in DYNAMIC_ARCH build

### LINKED ISSUE #2911: OpenBLAS-0.3.11 fails to compile on ppc64le

martin-frbg:
fixes #2911


Makefile.system
```diff
@@ -641,18 +641,22 @@ DYNAMIC_CORE += POWER8
...
 endif
 ifeq ($(C_COMPILER), GCC)
 ifeq ($(GCCVERSIONGT5), 1)
 DYNAMIC_CORE += POWER9
 else
 $(info, OpenBLAS: Your gcc version is too old to build the POWER9 kernels.)
 endif
-ifeq ($(GCCVERSIONGTEQ11), 1)
+LDVERSIONGTEQ35 := $(shell expr ld --version | head -1 | cut -f2 -d "." | cut -f1 -d "-" >= 35)
+ifeq ($(GCCVERSIONGTEQ11)$(LDVERSIONGTEQ35), 11)
```
RajalakshmiSR, line 653: This if condition seems to not work even if both conditions are met.. checking with more options...

RajalakshmiSR: This is due to missing backquote. 
LDVERSIONGTEQ35 := $(shell expr \`ld --version | head -1 | cut -f2 -d "."  | cut -f1 -d "-"\` \>= 35)

martin-frbg: Indeed, thanks. Somehow they got lost in a copy+paste between two remote shells.

PR #2926, opened 2020-10-20
x86_64: clobber all xmm registers after vzeroupper

bartoldeman:
As observed using GCC 10 using -march=native -ftree-vectorize
on Knights Landing, it is now smart enough to find clobbers inside
non-inlined static functions.

In particular, sgemv counted on a kernel to preserve the whole
%ymm2 register (since it was not in the clobber list), but the top
part was destroyed by vzeroupper. This caused many tests to fail.

This patch makes sure all xmm (and ymm/zmm by extension) registers
are listed as clobbered to avoid this happening, as most kernels
already did correctly in fact.


martin-frbg:
Thank you. Hope this is the last in the "incomplete clobber lists" saga, but perhaps we should add a more aggressively optimized build to our CI to try and catch such things.

PR #2940, opened 2020-10-23
Refactor the performance measurement system

Qiyu8:
I'm trying to add some new tests into current benchmark system, found out that there were some places that can be improved.

- different time measure method in different files, some files uses `gettimeofday`, some files uses `clock_gettime`, other files uses both of them.

- common memory allocation module such as `huge_malloc` is defined in many places.

Here is the improvements.

1. Extract a single `bench.h` file that includes all of the common performance functions mentioned above, which can shield the differences of time measure method.

2. Add three common functions `begin()`, `end()`, `getsec()`, in order to simplify the test case invocation.

All of the bench cases were modified and tested in X86/ARM.

### DIFF
--- a/benchmark/amax.c
+++ b/benchmark/amax.c
@@ -25,125 +25,73 @@ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 *****************************************************************************/
 
-#include <stdio.h>
-#include <stdlib.h>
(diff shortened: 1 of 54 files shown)


brada4:
BG info: I am not the approver of any kind, just giving my humble opinion. I did some polishing of R scripts in subdirectory few times. 
Shortening code by using abstraction is very desirable because after:
* the benchmark code fits in few screens to spot bugs easier
* the abstraction does same
While you are at it if you have chance - try static checkers - cppcheck , clang-analyzer aka scan-build , gcc -std=c18 (or c11) , this code got less (to no) attention than main runtime in this regard, this is by no means showstopper, just do it if it is easy to do.

PR #2954, opened 2020-10-29
Implementation of BF16 based gemv

Guobing-Chen:
1. Add a new API -- sbgemv to support bfloat16 based gemv
2. Implement a generic kernel for sbgemv
3. Implement an avx512-bf16 based kernel for sbgemv


martin-frbg:
Sorry to bother you after so many years, but there seems to be something fundamentally wrong with the "N" part of this PR, producing wrong results for any combination of inputs in the "compare_sgemm_sbgemm" test case that has recently been amended to include sbgemv

PR #2955, opened 2020-10-29
Fix cooperlake compile issue

Guobing-Chen:
Add a missing macro which is required in Makefile.x86_64 due to recent clearnup, which causes cooperlake platform build failure.


martin-frbg:
Sorry for missing that

PR #2960, opened 2020-10-31
fix avx2 detection

### LINKED ISSUE #2957: Incorrect avx2 detection
This is a followup to issue #2933. I'm getting reports from users still having issues to compile after the fix that went in 3.12. Their cpu doesn't support avx2, yet HAVE_AVX2=1 is generated in Makefile.conf. I have the opposite situation: my cpu has the avx2 flag, yet `./get_arch 0` will set HAVE_AVX2 flag to 0. I'm starting to wonder if the `support_avx2` function defined [here](https://github.com/xianyi/OpenBLAS/blob/develop/cpuid_x86.c#L214) is correct. Based on intel docs https://software.intel.com/content/www/us/en/develop/articles/how-to-detect-new-instruction-support-in-the-4th-generation-intel-core-processor-family.html, I would expect to test something like ebx & (1 <<3 ) & (1 << 5) & (1 << 8) instead of ebx & (1 <<7). gcc also has some intrinsics for testing cpu features which might be easier to use.

thrasibule:
reword commits to make it clearer

closes #2957

### DIFF
--- a/cpuid_x86.c
+++ b/cpuid_x86.c
@@ -202,7 +202,7 @@ int support_avx(){
   if ((ecx & (1 << 28)) != 0 && (ecx & (1 << 27)) != 0 && (ecx & (1 << 26)) != 0){
     xgetbv(0, &eax, &edx);
     if((eax & 6) == 6){
-      ret=1;  //OS support AVX
+      ret=1;  //OS supports saving xmm and ymm registers (6 = (1<<1) | (1<<2))
     }
   }
   return ret;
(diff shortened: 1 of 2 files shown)


martin-frbg:
As noted in #2957 I am not yet convinced that checking only the cpu capability is sufficient (the original intention must have been to handle that via the cpu model, and only do an addtional check if the OS environment actually allows avx2 to be used)

martin-frbg:
IOW my spontaneous idea is to subsitute
```
-  if((ebx & (1<<7)) != 0)
+  if ( (ebx & (1<<5)) != 0 && (ebx & (1<<7)) != 0 )
```

thrasibule:
I don't see how checking bit 7 is related to avx2. According to https://en.wikipedia.org/wiki/CPUID#EAX=7,_ECX=0:_Extended_Features, it's the bit "Supervisor Mode Execution Prevention". I really think it was a typo originally, there is no reason that I can see to check that bit. This PR doesn't change at all checking for the xmm and ymm registers (that's the call to xgetbv), which I agree is required.

martin-frbg:
Right, sorry, must have been a typo indeed. Got confused over my initial confusion it seems :/

Issue #2958, opened 2020-10-31
AVX errors on compiling for OpenBSD


epsilon-0:
I'm trying to create a port for OpenBLAS on OpenBSD and as a starter I tried to do a simple compilation by hand
```
$ FC="egfortran" gmake 
```

And I've run into a weird problem of AVC inline errors

```
...
cc -O2 -DMAX_STACK_ALLOC=2048 -DEXPRECISION -Wall -m64 -DF_INTERFACE_GFORT -fPIC -DSMP_SERVER -DNO_WARMUP -DMAX_CPU_NUMBER=24 -DMAX_PARALLEL_NUMBER=1 -DBUILD_SINGLE=1 -DBUILD_DOUBLE=1 -DBUILD_COMPLEX=1 -DBUILD_COMPLEX16=1 -DVERSION=\"0.3.12.dev\" -msse3 -mssse3 -msse4.1 -UASMNAME -UASMFNAME -UNAME -UCNAME -UCHAR_NAME -UCHAR_CNAME -DASMNAME=sgemm_itcopy -DASMFNAME=sgemm_itcopy_ -DNAME=sgemm_itcopy_ -DCNAME=sgemm_itcopy -DCHAR_NAME=\"sgemm_itcopy_\" -DCHAR_CNAME=\"sgemm_itcopy\" -DNO_AFFINITY -I.. -UDOUBLE  -UCOMPLEX -c -UDOUBLE -UCOMPLEX ../kernel/x86_64/../generic/gemm_tcopy_16.c -o sgemm_itcopy.o
[...]

Issue #349, opened 2014-02-27
OS X Installation of OpenBLAS doesn't work


FortranRookie:
This is the response after I type 'make'

| => make
getarch.c:75:19: fatal error: stdio.h: No such file or directory
 #include <stdio.h>
                   ^
compilation terminated.
make: **\* [getarch] Error 1
Makefile.system:299: **\* OpenBLAS: Clang didn't support OpenMP yet..  Stop.

My Mac is a mid-2012 Macbook Pro
OS: Mavericks 10.9.2
Processor: Intel Core i7 (Dual Core)
8GB RAM

And I have the latest version of gfortran installed from http://gcc.gnu.org/wiki/GFortranBinaries#MacOS

What is to be done about this? (I know about the other thread on this, but none of the answers helped)

PR #2962, opened 2020-10-31
add openbsd 68+ gfortran name

brada4:
Avoid custom command in #2958 , in rest of places `/gfortran/` covers the detection correctly


martin-frbg:
Thx. OpenBSD still feels a bit... weird. System `as` appears to be ancient too, apparently need to redirect to clang as for anything AVX2 like xianyi noted ages ago for the then-current osx here https://github.com/xianyi/OpenBLAS/issues/349#issuecomment-37438914

brada4:
CC+gcc barfs against _TLS_ too in some places. Some scattered OMP headers around, though no compiler as it seems able to dig them.

PR #2972, opened 2020-11-05
Improve the performance of rot by using AVX512 and AVX2 intrinsic

xiegengxin:
Provides AVX2, AVX512 intrinsic code and SMP code for drot and srot.

**Performance** :
|size|AVX2, double| AVX2, float|AVX512, double|AVX512, float|
|----|---------------|-------------|-----------------|---------------|
|        8|         0.96x|           0.87x|                1.33x|              0.84x|
|      16|         1.05x|           0.93x|                1.51x|              1.30x|
|      32|         1.30x|           0.73x|                1.97x|              0.75x|
|      64|         1.41x|           1.22x|                2.56x|              2.45x|
|    128|         1.31x|           1.39x|                3.08x|              2.79x|
|    256|         1.48x|           1.28x|                2.93x|              3.08x|
|    512|         1.35x|           1.46x|                3.09x|              2.83x|
|  1024|         1.40x|           1.34x|                3.19x|              3.12x|
|  2048|         1.39x|           1.40x|                3.20x|              3.20x|
|  4096|         1.26x|           1.36x|                1.83x|              3.10x|
|  8192|         1.27x|           1.25x|                1.83x|              1.78x|
|16384|         1.18x|           1.20x|                1.66x|              1.77x|
|            7|     0.66x|           0.72x|                0.90x|              1.03x|
|          10|     0.76x|           0.62x|                0.62x|              0.95x|
|        100|     1.22x|           1.13x|                2.15x|              1.48x|
|      1000|     1.43x|           1.35x|                2.95x|              2.09x|
|    10000|     1.37x|           1.33x|                1.84x|              1.78x|
|  100000|     1.04x|           1.24x|                1.14x|              1.66x|
|1000000(SMP)|   11.20x|           5.60x|              10.75x|              5.36x|

**Test on:**
CPU: Intel(R) Xeon(R) Platinum 8280L, 112 cores
Memory: 12 * 32 GB @2933 MT/s DDR4 
OS: CentOS 8 Stream
Kernel: 4.18.0-193

### DIFF
--- a/kernel/x86_64/srot.c
+++ b/kernel/x86_64/srot.c
@@ -1,0 +1,139 @@
+#include "common.h"
+
+#if defined(SKYLAKEX)
+#include "srot_microk_skylakex-2.c"
+#elif defined(HASWELL)
+#include "srot_microk_haswell-2.c"
+#endif
+
+#ifndef HAVE_SROT_KERNEL
+
+static void srot_kernel(BLASLONG n, FLOAT *x, FLOAT *y, FLOAT c, FLOAT s)
+{
+    BLASLONG i = 0;
+    FLOAT f0, f1, f2, f3;
+    FLOAT x0, x1, x2, x3;
+    FLOAT g0, g1, g2, g3;
+    FLOAT y0, y1, y2, y3;
+
+    FLOAT* xp = x;
+    FLOAT* yp = y;
+
+    BLASLONG n1 = n & (~7);
+
+    while (i < n1) {
+        x0 = xp[0];
+        y0 = yp[0];
+        x1 = xp[1];
+        y1 = yp[1];
+        x2 = xp[2];
+        y2 = yp[2];
+        x3 = xp[3];
+        y3 = yp[3];
+
+        f0 = c*x0 + s*y0;
+        g0 = c*y0 - s*x0;
+        f1 = c*x1 + s*y1;
+        g1 = c*y1 - s*x1;
+        f2 = c*x2 + s*y2;
+        g2 = c*y2 - s*x2;
+        f3 = c*x3 + s*y3;
+        g3 = c*y3 - s*x3;
+
+        xp[0] = f0;
+        yp[0] = g0;
+        xp[1] = f1;
+        yp[1] = g1;
+        xp[2] = f2;
+        yp[2] = g2;
+        xp[3] = f3;
+        yp[3] = g3;
+
+        xp += 4;
+        yp += 4;
+        i += 4;
+    }
+
+    while (i < n) {
+        FLOAT temp = c*x[i] + s*y[i];
+        y[i] = c*y[i] - s*x[i];
+        x[i] = temp;
+
+        i++;
+    }
+}
+
+#endif
+static void rot_compute(BLASLONG n, FLOAT *x, BLASLONG inc_x, FLOAT *y, BLASLONG inc_y, FLOAT c, FLOAT s)
+{
+    BLASLONG i = 0;
+    BLASLONG ix = 0, iy = 0;
+
+    FLOAT temp;
+    
+    if (n <= 0)
+        return;
+    if ((inc_x == 1) && (inc_y == 1)) {
+            srot_kernel(n, x, y, c, s);
+    }
+    else {
+        while (i < n) {
+            temp = c * x[ix] + s * y[iy];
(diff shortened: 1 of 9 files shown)


martin-frbg:
seems we now need to add `-mfma` wherever `-mavx2` gets added already

Qiyu8:
FMA instruction is an extended set of AVX, so if avx2 is detected, fma should enabled as well, the compiler is not smart enough to handle this situation.

xiegengxin:
Should I remove the FMA intrinsic code?

martin-frbg:
Let me try to fix the cpuflags code - there is something broken in the setup for DYNAMIC_ARCH, probably has been broken for a very long time already but only uncovered by the recent use of intrinsics. (With direct assembly coding it "almost" never mattered if the compiler had been told that fma or even just sse was available. There are a few `ifdef HAVE_SSE3` etc in the .S files that are read from the respective config_kernel.h correctly, but the way the Makefiles handle the same information must be broken. Nobody cared as along as that was not used to drive the compiler)

carlkl:
The best strategy for cpuflags is to use `-march` instead of explicitely set the instructions individually.

`gcc -march=haswell` i.e. sets avx2 as well as fma instructions.

For gcc on x86/x64 this is: (clang has -march as well)

**-march=pentium3**
    Intel Pentium 3 CPU  _since Katmai_
    MMX, SSE

**-march=pentium4**
    Intel Pentium 4 CPU
    MMX, SSE, SSE2

**-march=prescott’**
    Intel Pentium 4 CPU
    MMX, SSE, SSE2, SSE3

**-march=nocona**
    Intel Pentium 4 CPU with 64-bit extensions
    MMX, SSE, SSE2, SSE3
    _similar to: -march=barcelona’_

**-march=core2**
    Intel Core 2 CPU with 64-bit extensions
    MMX, SSE, SSE2, SSE3, SSSE3
    _SSE4.1 is not supported by all Core2 CPUs_

**-march=nehalem**
    Intel Nehalem CPU with 64-bit extensions
    MMX, SSE, SSE2, SSE3, SSSE3, 
    SSE4.1, SSE4.2, POPCNT

**-march=sandybridge**
    Intel Sandy Bridge CPU with 64-bit extensions
    MMX, SSE, SSE2, SSE3, SSSE3, 
    SSE4.1, SSE4.2, POPCNT, 
    AVX, AES, PCLMUL

**-march=haswell**
    Intel Haswell CPU with 64-bit extensions
    MOVBE, MMX, SSE, SSE2, SSE3, SSSE3, 
    SSE4.1, SSE4.2, POPCNT, 
    AVX, AES, PCLMUL, 
    AVX2, FSGSBASE, RDRND, FMA, BMI, BMI2, F16C

martin-frbg:
Thx, had thought about that already. Historically we have the various HAVE__feature entries created by getarch, which _should_ work in principle but it looks as if they need to be `undef`d between cycles of the kernel Makefile but are not, so it ends up adding features that may not actually be present. In 0.3.11/12 I had worked around this at the cost of silly duplications and long target lists. Will do a PR fixing the current feature-based code once i've checked if CMAKE gets it right.

kernel/x86_64/srot.c
```diff
@@ -0,0 +1,139 @@
...
+    BLASLONG i = 0;
+    FLOAT f0, f1, f2, f3;
+    FLOAT x0, x1, x2, x3;
+    FLOAT g0, g1, g2, g3;
+    FLOAT y0, y1, y2, y3;
+
+    FLOAT* xp = x;
+    FLOAT* yp = y;
+
+    BLASLONG n1 = n & (~7);
```
jinboson, line 22: n1 = n & (~3) ?

xiegengxin: This won't cause error result. May be not good enough for performance.

jinboson: Thx, I was curious about that your step was 4 in the loop but ~7 was taken.

xiegengxin: I reference it from https://github.com/xianyi/OpenBLAS/blob/7f11e33e8d3c979183a4785c9d4519555ccb1bba/kernel/power/drot.c#L105

I'm not sure which one is good. If n1 = n & (~3) is better, I'll fix this code.

jinboson: Thanks for setting me straight. No need to fix.

PR #2978, opened 2020-11-07
Fix handling of cpu capability flags in DYNAMIC_ARCH builds

martin-frbg:
- In both make and cmake builds, the `HAVE_feature` entries kept accumulating from one cpu build cycle to the next
- In cmake builds, the property flags were evaluated and transformed into compiler flags even before `getarch` was invoked, thereby misapplying the properties of the preceding cpu target (the common type in the case of the first DYNAMIC_ARCH target)
- The HAVE_FMA3 flag was not converted into -mfma


cmake/system.cmake
```diff
@@ -211,6 +146,63 @@ else()
...
+#    if (${CMAKE_C_COMPILER_ID} STREQUAL "GNU")
+      execute_process(COMMAND ${CMAKE_C_COMPILER} -dumpversion OUTPUT_VARIABLE GCC_VERSION)
+        if (${GCC_VERSION} VERSION_GREATER 10.1 OR ${GCC_VERSION} VERSION_EQUAL 10.1)
+          set (KERNEL_DEFINITIONS "${KERNEL_DEFINITIONS} -march=cooperlake")
+        else()
+          set (KERNEL_DEFINITIONS "${KERNEL_DEFINITIONS} -march=skylake-avx512")
+        endif()
+#    elseif (${CMAKE_C_COMPILER_ID} STREQUAL "CLANG")
+#      set (KERNEL_DEFINITIONS "${KERNEL_DEFINITIONS} -mavx2")
+#    endif()
```
xoviat, line 160: @martin-frbg What if the compiler is not GCC? Such as in this case:

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=271807&view=logs&j=8f2007e9-8cd3-534e-f684-684150acfa25&t=0b7eae37-0441-5a37-ab2a-71a810374446&l=936

martin-frbg: Right I forgot the `if (${CMAKE_C_COMPILER_ID} STREQUAL "GNU")` in cc.cmake (unlike os.cmake)

PR #2983, opened 2020-11-11
Optimize the performance of rot by using universal intrinsics

Qiyu8:
## Introduction

Inspired by #2972 , we can extend the advantage of intrinsics code to other x86 architectures by using usimd. Compared with develop branch,  the new implementation has about **_1.6x_** faster in `AVX2`, about **_1.26x_** faster in `FMA3`, Other intrinsics set such as `SSE2/NEON` has a negative impact on performance. it is worth mentioning that `AVX2 `has negative impact while handling double data types(I guess the reason would be too many time-consuming memory operations). @xiegengxin what do you think?

## Benchmark

  Each test has run for 10 loops and then take the average result, The radio is the MFlops division of Baseline and optimized branch(*The bigger* the radio is, the *better* performance is achieved).

- X86 BaseLine(loop unrolled)

| data size(10^3) | develop(float)               |
| --------------- | ---------------------------- |
| 4000            | 384.59 MFlops   0.020802 sec |
| 8000            | 389.15 MFlops   0.041115 sec |

- X86-FMA3 enabled

  | data size(10^3) | usimd-sum(float)             | ratio |
  | --------------- | ---------------------------- | ----- |
  | 4000            | 484.72 MFlops   0.016504 sec | 1.26  |
  | 8000            | 483.28 MFlops   0.033107 sec | 1.24  |
  
- X86-AVX2 enabled

  | data size(10^3) | usimd-sum(float)             | ratio |
  | --------------- | ---------------------------- | ----- |
  | 4000            | 597.93 MFlops   0.013380 sec | 1.55 |
  | 8000            | 635.01 MFlops   0.025196 sec | 1.63  |
  

## System Info

|           | Arm                                                          | x86                                      |
| --------- | ------------------------------------------------------------ | ---------------------------------------- |
| Hardware  | KunPeng                                                      |                                          |
| Processor | ARMv8 2.6GMHZ 8 processors                                   | Intel(R) Xeon(R) Gold 6161 CPU @ 2.20GHz |
| OS        | Linux ecs-9d50 4.19.36-vhulk1905.1.0.h276.eulerosv2r8.aarch64 | Windows Server 2008 R2 Enterprise        |
| Compiler  | gcc (GCC) 7.3.0                                              | MSVC14.06                                |


xiegengxin:
Sounds great. AVX512 has an intrinsic, `_mm512_maskz_loadu_pd`, which can improve the performance when data size is small (<8 for double, < 16 for float). Is it possible to implement such intrinsic by using usimd?

Qiyu8:
@xiegengxin Thanks for your quick reply, I think we can simulate `_mm512_maskz_loadu_pd `by using [partial load and store universal intrinsics](https://github.com/numpy/numpy/pull/17340) in SSE2/AVX2/NEON, but Predictably, The performance improvements can be almost ignored because the simulation and relatively small pack size(SSE2/Neon only handles 2 double at one time, AVX2 handles 4 double at one time)

martin-frbg:
Strange that it should need the `-mfma` in kernel/Makefile (again) when that is supposed to be handled in Makefile.x86_64 now.
(Actually the failing builds appear to be CMAKE, so cmake/system.cmake should be adding the `-mfma' unless NO_AVX2 was set)

Qiyu8:
Althouth Makefile.x86_64 and system.cmake added `fma` compile tag, both `clang` and `gcc ` can't recognize the `fma `intrinsic because of different reason, clang warned of missing `<immintrin.h>` header ,  while gcc reported missing compiler tag `-mfma`.

martin-frbg:
So that particular clang version on Appveyor is probably too old ? And I still wonder why gcc did not see `-mfma` when it should have been added based on the HAVE_FMA3 property. I see your latest revision now adds it when the host has AVX, which does not look correct to me.

martin-frbg:
Looks more like a missing codepath in the cmake files - HAVE_FMA3 somehow not used when it is not a DYNAMIC_ARCH build.
Seen it now - missing in cc.cmake, will do a PR in a minute. When I have more time for this again I need to check if the duplication in cc.cmake and system.cmake for dynamic and non-dynamic is still necessary.

Qiyu8:
OK, since the CI failure has resolved by #2989 , I will remove the `-mfma` flag in when the host has AVX.

PR #2994, opened 2020-11-19
Power10 fixes

antonblanchard:
Here are a couple of POWER10 fixes that improve the dynamic detection code.

### DIFF
--- a/driver/others/dynamic_power.c
+++ b/driver/others/dynamic_power.c
@@ -52,6 +52,9 @@ static gotoblas_t *get_coretype(void) {
 	if (__builtin_cpu_supports ("arch_3_1") && __builtin_cpu_supports ("mma"))
 		return &gotoblas_POWER10;
 #endif
+	/* Fall back to the POWER9 implementation if the toolchain is too old or the MMA feature is not set */
+	if (__builtin_cpu_is("power10"))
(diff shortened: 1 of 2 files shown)


driver/others/dynamic_power.c
```diff
@@ -52,6 +52,9 @@ static gotoblas_t *get_coretype(void) {
 	if (__builtin_cpu_supports ("arch_3_1") && __builtin_cpu_supports ("mma"))
 		return &gotoblas_POWER10;
 #endif
+	/* Fall back to the POWER9 implementation if the toolchain is too old or the MMA feature is not set */
```
RajalakshmiSR, line 55: Thanks  @antonblanchard for correcting the check for ld version.

If the toolchain is old, I think we might get error like this. 
       error: cpu ‘power10’ is an invalid argument to builtin ‘__builtin_cpu_is’
In which distro are you seeing this?  I remember we added this P10 check specifically to work for fedora 32, which has gcc10.2 and ld 2.34.

martin-frbg:
@RajalakshmiSR very sorry for merging this without heeding your objection - obviously you are right, the builtin_cpu_is(power10) will fail with gcc < 11 as the corresponding gcc patch was merged only in june 2020. Bumping the version number in the conditional should fix this I guess...

Issue #2970, opened 2020-11-02
maybe a problem with 0.3.12 and NumPy?


mattip:
Just a heads up, that NumPy is seeing some problems with OpenBLAS 0.3.12.

We released NumPy 1.19.3 with OpenBLAS 0.3.12 in order to fix the windows `fmod` problems. We had to back it out and are releasing a 0.19.4 with the previous OpenBLAS and code to error out if the `fmod` bug is detected instead. We got reports that 1.19.3 crashes, xref numpy/numpy#17674 and numpy/numpy#17684. In the first issue, the reporter provided ~a docker image and code to repoduce~ output of some system diagonstics. The second issue is a lack of memory in `hstack`, which is difficult to attribute to OpenBLAS, so there may be something else going on.
Both issues were "fixed" (or covered up) by the 1.19.4 release candidate with an older OpenBLAS.
Does any of this make sense to you?

Edit: no docker image, just diagnostics

martin-frbg:
No idea. If anything, thread memory requirements of 0.3.12 should be less than before, and I thought NumPy was satisfied with the state of OpenBLAS after application of the fmod workarounds. Are you bundling the actual 0.3.12 release or some snapshot from around that time ? And the older OpenBLAS you went back to is what - 0.3.10 or even older ?

charris:
@martin-frbg Our CI testing showed no problems and the wheels builds succeeded, so 0.3.12 looked good prior to release.

EDIT: I also expect a fair number of people tried 1.19.3 because of the Python 3.9 support. If the problem was universal there would probably have been more bug reports.

martin-frbg:
Well the only change in 0.3.11/12 that I would expect to have any effect at library/thread initialization time is the reduction of the BLAS3_MEM_ALLOC_THRESHOLD variable, and its previous default is stil available in Makefile.rule (or as a parameter to `make`. Reducing this actually made crashes in other people's code go away however. If your "working" OpenBLAS is older than 0.3.10 then we are looking at a scary number of changes, and probably need a `git bisect` to get anywhere.

h-vetinari:
Unfortunately, it's [0.3.9](https://github.com/numpy/numpy/pull/17680/files#diff-2032d49a4193331916561c2246dd208ad8b27aea8cc47c3ef87bc9f0f3fe297bR14)

brada4:
Original issues got closed without code change
https://github.com/numpy/numpy/issues/17674#issuecomment-720637865
[...]

PR #2995, opened 2020-11-19
Don't overwrite blas_thread_buffer if already set

### LINKED ISSUE #2993: Segfault when NUM_THREADS is smaller than host vCPUs
Most context can be found here: https://github.com/xianyi/OpenBLAS/pull/2982

But in summary:

A segfault will occur if the NUM_THREADS used to build openblas is lower than the host's vCPU count when using OMP.

```
[Switching to Thread 0x7ffebba6e640 (LWP 104473)]
0x00007fffe95cbe70 in dgemm_itcopy_ZEN () from /nix/store/3v2i5ga27y9al3z0d8ccwaa389qz3ma6-lapack-3/lib/liblapack.so.3
(gdb) backtrace
#0  0x00007fffe95cbe70 in dgemm_itcopy_ZEN () from /nix/store/3v2i5ga27y9al3z0d8ccwaa389qz3ma6-lapack-3/lib/liblapack.so.3
#1  0x00007fffe882416a in inner_thread () from /nix/store/3v2i5ga27y9al3z0d8ccwaa389qz3ma6-lapack-3/lib/liblapack.so.3
#2  0x00007fffe8956bf6 in exec_blas._omp_fn () from /nix/store/3v2i5ga27y9al3z0d8ccwaa389qz3ma6-lapack-3/lib/liblapack.so.3
#3  0x00007fffe347f916 in ?? () from /nix/store/f23sq7lk6xfrvz467ffkpzackyk5q8dm-gfortran-9.3.0-lib/lib/libgomp.so.1
#4  0x00007ffff7c20e9e in start_thread () from /nix/store/5didcr1sjp2rlx8abbzv92rgahsarqd9-glibc-2.32/lib/libpthread.so.0
#5  0x00007ffff793866f in clone () from /nix/store/5didcr1sjp2rlx8abbzv92rgahsarqd9-glibc-2.32/lib/libc.so.6
```

This currently has a "workaround" by setting the environment variable `OMP_NUM_THREADS` to a value equal or lower to the `NUM_THREADS` used during build. However, desired behavior is that it doesn't crash.

EDIT:
Also of note, this fails when running the numpy test suite. Openblas is able to run it's test suite without crashing.
blas, openblas, and lapack are all aliased to openblas
```
$ cat numpy/site.cfg
[blas]
include_dirs=/nix/store/bm0mjpbnaxixma2wvmj99zq2kfq1017h-blas-3-dev/include
library_dirs=/nix/store/35lvfykv13mnva6sipnrqks05mvrldi0-blas-3/lib
runtime_library_dirs=/nix/store/35lvfykv13mnva6sipnrqks05mvrldi0-blas-3/lib

[lapack]
include_dirs=/nix/store/s9rb0jhplgsfcnay1mp7kn1zhxyppq0n-lapack-3-dev/include
library_dirs=/nix/store/w6xjrzfnxsi7jwrlvfy10qcljm6gyhss-lapack-3/lib
runtime_library_dirs=/nix/store/w6xjrzfnxsi7jwrlvfy10qcljm6gyhss-lapack-3/lib

[openblas]
include_dirs=/nix/store/bm0mjpbnaxixma2wvmj99zq2kfq1017h-blas-3-dev/include:/nix/store/s9rb0jhplgsfcnay1mp7kn1zhxyppq0n-lapack-3-dev/include
libraries=lapack,lapacke,blas,cblas
library_dirs=/nix/store/35lvfykv13mnva6sipnrqks05mvrldi0-blas-3/lib:/nix/store/w6xjrzfnxsi7jwrlvfy10qcljm6gyhss-lapack-3/lib
runtime_library_dirs=/nix/store/35lvfykv13mnva6sipnrqks05mvrldi0-blas-3/lib:/nix/store/w6xjrzfnxsi7jwrlvfy10qcljm6gyhss-lapack-3/lib
```

## System / Build Info
```
$ nix eval -f default.nix openblas.makeFlags
[ "BINARY=64" "CC=cc" "CROSS=0" "DYNAMIC_ARCH=1" "FC=gfortran" "HOSTCC=cc" "INTERFACE64=1" "NO_AVX512=1" "NO_BINARY_MODE=" "NO_SHARED=0" "NO_STATIC=1" "NUM_THREADS=64" "TARGET=ATHLON" "USE_OPENMP=1" ]
```
CC = gcc-9.3.0
arch = x86_64
platform = linux
CPU = 3990X ( 64 cores / 128 thread)

Flamefire:
After a fork it is possible that blas_thread_buffer has already allocated memory buffers: goto_set_num_threads does allocate those
already and it may be called by num_cpu_avail in case the OpenBLAS NUM_THREADS differ from the OMP num threads.
This leads to a memory leak which can cause subsequent execution of BLAS kernels to fail.

Fixes #2993

I basically factored out the code from `goto_set_num_threads`. It might be marginally slower than the code from `blas_thread_init` but is correct in all cases and simpler to maintain (having it in 1 place)

BTW: Looking at the code, wouldn't it need a `WMB` or so after the array is filled to make sure all threads see the new buffers? Hasn't been a problem in practice though.

### DIFF
--- a/driver/others/blas_server_omp.c
+++ b/driver/others/blas_server_omp.c
@@ -76,9 +76,27 @@ static atomic_bool blas_buffer_inuse[MAX_PARALLEL_NUMBER];
 static _Bool blas_buffer_inuse[MAX_PARALLEL_NUMBER];
 #endif
 
-void goto_set_num_threads(int num_threads) {
+static void adjust_thread_buffers() {
 
   int i=0, j=0;
+
+  //adjust buffer for each thread
+  for(i=0; i < MAX_PARALLEL_NUMBER; i++) {
+    for(j=0; j < blas_cpu_number; j++){
+      if(blas_thread_buffer[i][j] == NULL){
+        blas_thread_buffer[i][j] = blas_memory_alloc(2);
+      }
+    }
+    for(; j < MAX_CPU_NUMBER; j++){
+      if(blas_thread_buffer[i][j] != NULL){
+        blas_memory_free(blas_thread_buffer[i][j]);
+        blas_thread_buffer[i][j] = NULL;
+      }
+    }
+  }
+}
+
+void goto_set_num_threads(int num_threads) {
 
   if (num_threads < 1) num_threads = blas_num_threads;
 
@@ -92,20 +110,7 @@ void goto_set_num_threads(int num_threads) {
 
   omp_set_num_threads(blas_cpu_number);
 
-  //adjust buffer for each thread
-  for(i=0; i<MAX_PARALLEL_NUMBER; i++) {
-    for(j=0; j<blas_cpu_number; j++){
-      if(blas_thread_buffer[i][j]==NULL){
-        blas_thread_buffer[i][j]=blas_memory_alloc(2);
-      }
-    }
-    for(; j<MAX_CPU_NUMBER; j++){
-      if(blas_thread_buffer[i][j]!=NULL){
-        blas_memory_free(blas_thread_buffer[i][j]);
-        blas_thread_buffer[i][j]=NULL;
-      }
-    }
-  }
+  adjust_thread_buffers();
 #if defined(ARCH_MIPS64)
   //set parameters for different number of threads.
   blas_set_parameter();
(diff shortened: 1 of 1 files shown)


Flamefire:
Tested with 64 threads OpenBLAS on a 256 threads system and no crash any longer

jonringer:
@martin-frbg I know that you have a few more items in your milestone https://github.com/xianyi/OpenBLAS/milestone/32 , but it would be nice to have this fix in a patch release

martin-frbg:
The fix will be part of the planned release. I understand that there is always somebody waiting for one particular bugfix or feature but IMHO it makes no sense to do rush releases every few days (and I cannot work on OpenBLAS 24/7 anyway - the milestone date was set to somewhat coincide with having time available for a halfway orderly release). Most importantly it appears to be still unclear if the "other" numpy problems in #2970 are actually fixed through this and/or the reduction of the default BUFFERSIZE.

jonringer:
> The fix will be part of the planned release. I understand that there is always somebody waiting for one particular bugfix or feature but IMHO it makes no sense to do rush releases every few days

I completely understand. Thanks for responding :)

PR #2997, opened 2020-11-19
Add reproducer test for crash after fork

Flamefire:
See #2993 for an analysis

This reproduces the issue in #2993 without resorting to numpy. I confirmed that the test fails prior to applying #2995 and succeeds afterwards independent of the number of threads that OpenBLAS was build with


Flamefire:
The Drone CI needs to be restarted manually

PR #3013, opened 2020-11-29
Fix 32bit x86 builds and add workaround for x86_64 miscompilations by gcc 4.6 (including our Travis setup)

martin-frbg:
gcc 4.6.3 as used in Ubuntu Precise miscompiles e.g. the Sandybridge kernels with the -mavx option, leading to SIGILL in the zgemm tests (probably running out of registers)


conradsnicta:
@martin-frbg This may add unnecessary [technical debt](https://en.wikipedia.org/wiki/Technical_debt) and increase the risk of (present and future) configuration bugs.  
Is there any substantial practical reason to support such an ancient release of gcc?

Ubuntu Precise (12.04) is no longer supported upstream.  It's EOL as of April 2017 (see https://wiki.ubuntu.com/Releases).  Paid "extended security maintenance" finished on April 2019.

[gcc 4.6.0](https://gcc.gnu.org/news.html) was released on 2011-03-25 (roughly 10 years ago), with the last update (v. 4.6.4) released on 2013-04-12 (roughly 8 years ago).  

It might be better to simply place a cut-off on the lowest version of gcc that's still in wide use.  In this case it would be gcc 4.8, which is part of [RHEL 7](https://en.wikipedia.org/wiki/Red_Hat_Enterprise_Linux) / CentOS 7.  This OS is supported until 30 June 2024.

martin-frbg:
There are a few tests for gcc <= 4.7 in the build system already, so the increase in technical debt should be small (and I have gotten used to seeing ancient systems still running - though obviously shielded from the internet - in industrial settings). 
Of course I might as well just update the Travis configuration once I am certain that this is the reason behind the unexpected build failures. (Travis has dropped from the PR checks overview for some reason, but it is still alive behind the badge in the README)

conradsnicta:
Is anybody likely to install new versions of OpenBLAS on these old systems?  I'd envisage that once something is "stable" on such old systems, stuff wouldn't get upgraded -- otherwise there is a risk of breaking a production system.

It might be worth cleaning out stuff related to gcc <= 4.7, which would decrease the maintenance burden (and hence free up time for other things).

PR #3025, opened 2020-12-07
MIPS: Fix two bugs

TiredNotTear:
Commit ID 47b639c:
The swap test case will call sswap_msa.c and dswap_msa.c files in MIPS environmnet.
When inc_x or inc_y is equal to zero, the calculation result of the two functions will be wrong.
This patch adds the processing of inc_x or inc_y equal to zero, and the swap test case has passed.
Commit ID ad38bd0:
The cgemv and zgemv test case will call cgemv_n/t_msa.c zgemv_n/t_msa.c files in MIPS environment.
When the macro CONJ is defined, the calculation result will be wrong due to the wrong definition of OP2.
This patch updates the value of OP2 and passes the corresponding test.


TiredNotTear:
@yinshiyou @gxw-loongson @jinboson, I'd like you to do a code review.

yinshiyou:
LGTM.

PR #3027, opened 2020-12-09
Add msa support for loongson

XiWeiGu:
1. Using core loongson3r3 and loongson3r4 for loongson
2. Add DYNAMIC_ARCH for loongson


martin-frbg:
Thank you - I wonder if we could/should keep the old 3A, 3B names around for backwards compatibility ? (Could treat them as aliases in getarch.c, so that `make TARGET=LOONGSON3A` gets mapped to LOONGSON3R3)

XiWeiGu:
Thank you for your advice，I will keep the old 3A and 3B.

PR #3035, opened 2020-12-13
Define BLAS acronym in README

misyltoad:
BLAS is a fairly common acronym in other parts of computing (ie. raytracing for bottom level acceleration structure.)

I saw an article about the project on Phoronix and has no idea as to what this project was about as it expected me to know what BLAS meant without ever defining it.

This PR makes it more explicit in the README about what the library does.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Drone CI: [![Build Status](https://cloud.drone.io/api/badges/xianyi/OpenBLAS/sta
 
 ## Introduction
 
-OpenBLAS is an optimized BLAS library based on GotoBLAS2 1.13 BSD version.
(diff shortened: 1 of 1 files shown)


martin-frbg:
Hm, we have this information in the FAQ, but I was not aware that the acronym has been appropriated for other fields lately.
Generally the fundamental documentation is a bit scarce here as OpenBLAS sees itself as the continuation of GotoBLAS which was a novel concept to speed up BLAS, all under the assumption that "you already know what BLAS is" and are familiar with THE reference implementation at netlib.org and its accompanying documentation :smiley:

brada4:
It is initialism for particular API, not referring to any Basic set of linear subroutines, like sandard C math libraries that are able to calculate simple linear projection.

PR #3050, opened 2020-12-29
getarch.c: define OPENBLAS_SUPPORTED for riscv64

aurel32:
For some reason, `getarch.c` doesn't define `OPENBLAS_SUPPORTED` for the riscv64 CPU, despite this CPU being supported. It looks like it has just been forgotten, so just add it.


martin-frbg:
Indeed this looks like an oversight, thank you.

xianyi:
haha, I forgot to edit `getarch.c`.

Issue #3020, opened 2020-12-03
Patches for OpenBLAS 0.3.12 to support NVIDIA HPC SDK Compilers release 20.11 and later


cparrott73:
Good news!  I have been working with our developers to clean up a number of the residual issues involved with compiling OpenBLAS with the NVIDIA HPC SDK Compilers, and as of the 20.11 release (due out shortly) - there are only a small number of changes needed against the OpenBLAS 0.3.12 source tree to allow it to compile it out of the box with our compilers.

Summary of these changes:

1. `-Mnollvm` is no longer needed to compile OpenBLAS with our compilers.  In fact, the old NoLLVM compilers do not even exist anymore.  This flag no longer has any effect and should be removed from the OpenBLAS build with modern releases of our compilers.

2. The `-D__MMX__` flag used to work around a previous issue is also no longer needed.  That issue has been fixed.

3. `f_check` needs a small tweak to correctly detect `nvfortran` as a PGI variant, and thus switch in the appropriate calling conventions for complex numbers.

4. In order to compile kernels that use `_mm256` and `_mm512` intrinsics with our compilers, you have to pass the `-mavx512f` flag.  Otherwise, these intrinsics are not enabled in our compilers by default.  I have modified the SKYLAKEX and COOPERLAKE build rules in `kernel/Makefile` to pass this flag.  I also tested this flag with GCC 10.2.0, and confirmed that it does not adversely affect compiling OpenBLAS with GCC.

Note I built OpenBLAS 0.3.12 with these patches with our 20.11 compilers as follows:

`make CC=nvc FC=nvfortran HOSTCC=gcc DYNAMIC_ARCH=1 USE_OPENMP=1 NUM_THREADS=512 all`

Please give these patches a test with our 20.11 compilers once they are released, and if you agree they work well for you, consider integrating them into the OpenBLAS source tree ahead of the next OpenBLAS release.  I think they will greatly help our customers who also use OpenBLAS.  Cheers!

[openblas-nvhpc.patch.txt](https://github.com/xianyi/OpenBLAS/files/5639534/openblas-nvhpc.patch.txt)

cparrott73:
Update - there is some internal discussion that `-march=skylake-avx512` should imply `-mavx512f` automatically, as it apparently does in GCC.  We may try to fix this ahead of the next release.  In that case, the patches that add this flag will no longer be necessary.

brada4:
Version detection would be handy, mm256 iz like AVX2 thing?

[...]

PR #3061, opened 2021-01-12
Support NVIDIA HPC SDK on ARM64

martin-frbg:
based on patch provided by cparrott73 in #3020


martin-frbg:
Current build failures are caused by cmake using a CMAKE-prefixed variable for C_COMPILER, thus failing to interpret the new conditional in the KERNEL files...

Issue #2257, opened 2019-09-13
TARGET=GENERIC produces errors with TRMM()


BarrySmith:
Building with 

`make TARGET=GENERIC MAKE_NB_JOBS=13  USE_OPENMP=0  USE_THREAD=0  NO_EXPRECISION=1 `

produces errors with TRMM(). This happens for both Gnu and Apple CLANG compilers on both Linux and Apple. Both v0.3.6 and v0.3.7. Same error also with USE_THREAD=1 (haven't tested USE_OPENMP=1). 

`OPENBLAS_NUM_THREADS=1 OMP_NUM_THREADS=1 ./sblat3 < ./sblat3.dat

OPENBLAS_NUM_THREADS=1 OMP_NUM_THREADS=1 ./dblat2 < ./dblat2.dat

 TESTS OF THE REAL             LEVEL 3 BLAS

 THE FOLLOWING PARAMETER VALUES WILL BE USED:

   FOR N                   0     1     2     3     7    31

   FOR ALPHA             0.0   1.0   0.7

   FOR BETA              0.0   1.0   1.3

 ROUTINES PASS COMPUTATIONAL TESTS IF TEST RATIO IS LESS THAN   16.00

 RELATIVE MACHINE PRECISION IS TAKEN TO BE  1.2E-07

 SGEMM  PASSED THE TESTS OF ERROR-EXITS

 SGEMM  PASSED THE COMPUTATIONAL TESTS ( 17496 CALLS)

 SSYMM  PASSED THE TESTS OF ERROR-EXITS

 SSYMM  PASSED THE COMPUTATIONAL TESTS (  1296 CALLS)

 STRMM  PASSED THE TESTS OF ERROR-EXITS

 ******* FATAL ERROR - COMPUTED RESULT IS LESS THAN HALF ACCURATE *******

           EXPECTED RESULT   COMPUTED RESULT

       1      0.186813          0.373626 
   
 ******* STRMM  FAILED ON CALL NUMBER:

    506: STRMM ('L','U','N','U',  1,  1, 1.0, A,  2, B,  2)        `

Similar errors for all the variants of Xtrmm()

Error does not occur if TARGET=GENERIC is not used on the same systems.

martin-frbg:
This was apparently caused by me with PR #2051

PR #3067, opened 2021-01-14
Fix building "generic" TRMM kernel with CMake

albertziegenhagel:
The CMake "TARGET_CORE" variables stores the "generic" target name in all lowercase letters, but gets compared to an all uppercase string, which results in the wrong TRMM kernel being selected.

This MR converts the TARGET_CORE to all uppercase before comparing its value to make sure case mismatches are not an issue in the future anymore.

The selection of the wrong TRMM kernel can be seen when compiling OpenBLAS with CMake and the option `-DTARGET=GENERIC`.

In this case the tests `sblat3`, `dblat3`, `cblat3` and `zblat3` all fail with an error similar to the following:

```
       1      0.186813          0.373626
 ******* STRMM  FAILED ON CALL NUMBER:
    506: STRMM ('L','U','N','U',  1,  1, 1.0, A,  2, B,  2)        .
```

which is basically the same what could be observed in #2257.


martin-frbg:
Oops - thanks.

PR #3073, opened 2021-01-18
add embedded option

xoviat:
the main thing this does is removes the mmap memory allocator so that the standard "malloc" is used, and removes the "gettimeofday" which is not available on embedded. The CMakeLists.txt modification is wrong, but I would like some advice about what that should look like.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,3 +91,4 @@ benchmark/*.goto
 benchmark/smallscaling
 CMakeCache.txt
 CMakeFiles/*
+.vscode
--- a/README.md
+++ b/README.md
@@ -212,7 +212,8 @@ Please note that it is not possible to combine support for different architectur
 - **Android**: Supported by the community. Please read <https://github.com/xianyi/OpenBLAS/wiki/How-to-build-OpenBLAS-for-Android>.
 - **AIX**: Supported on PPC up to POWER8
 - **Haiku**: Supported by the community. We don't actively test the library on this OS.
-- **SunOS**: Supported by the community. We don't actively test the library on this OS:
+- **SunOS**: Supported by the community. We don't actively test the library on this OS.
+- **Cortex-M**: Supported by the community. Please read <https://github.com/xianyi/OpenBLAS/wiki/How-to-use-OpenBLAS-on-Cortex-M>.
 
 ## Usage
 
--- a/cmake/system.cmake
+++ b/cmake/system.cmake
@@ -233,6 +233,11 @@ if (BINARY64)
   endif ()
 endif ()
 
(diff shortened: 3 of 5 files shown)


martin-frbg:
The way this is currently done is probably less than elegant, but I would think adding your target options to the  CCOMMON_OPT variable in cmake/system.cmake like it is done for e.g. NEED_PIC should work, so
```
if (OS_EMBEDDED)
set (CCOMMON_OPT "${CCOMMON_OPT} -mthumb -mcpu=cortex-m4 -mfloat-abi=hard -mfpu=fpv4-sp-d16")
endif ()
```
(though this may need coupling with `if ($TARGET STREQUAL ARMV6)` or similar if we ever get any other embedded targets)
And I am actually unsure if anybody tried to build for one of the newer ARMV8 targets with CMAKE - I notice their `-march` and `-mtune` are supplied via Makefile.arm64 but there are no corresponding provisions in the cmake build files. This is probably much less of an issue for ARMV6/ARMV7.

xoviat:
Note: ARMV6 does not work on cortex-M, even though the README claims that it does. Cortex-M requires "thumb" instructions, which are not present in the assembly files.

martin-frbg:
Oh, that's interesting - but I suspect wernsaar only ever tried with the Raspberry Pi1 and assumed the Cortex M products to be similar enough. So you are using TARGET=ARMV5 (which has only generic C kernels) for your build ?

xoviat:
Yes, I am. Supposedly there is some optimization possible (https://talmetz.tech/libtela/), but the first step is to get it working. I also considered other options, but having everyone focus on a single numerical library (OpenBLAS) seems to be the best approach.

brada4:
Very good wiki page for others to re-use.

There is a long  road ahead to make most essential loops auto-vectorizable by compiler like here on rarity FPU ISAs.

Probably double precision is not worth the flash space used, FPU does not help it, and integer emulation takes unpredictable time, like 10-20 insns per add  rendering it useless performance or realtime wise.

xoviat:
@brada4 IIUC, ARMv7 (both A and M) supports thumb instructions, and there is no downside to using them. Therefore, it would be better to switch all ARMv7 kernels to thumb instructions. Of course that is a big undertaking, but I'm just describing the ideal scenario. I'm not too familiar with assembly, but maybe you have some tips on how to do this.

brada4:
Many things will need to be nop-padded to maintain existing natural alignment for the normal consumer CPUs (like jmp targets in loops)
For getting formally good numeric result, it is almost like sed s/QXX/T2QXX/g

PR #3080, opened 2021-01-22
remove extra copies of omatcopy kernels

brada4:
* generic originals stay in arm & generic the way Makefile.L3 understands it
* "no change" testing wise.


brada4:
will do more profound approach.

PR #3091, opened 2021-01-29
Fix calculation of the non-exceptional shift values in LAPACK complex QZ

martin-frbg:
from the updated version of Reference-LAPACK PR 477, solving occasional (rare?) convergence failures in complex GGEV/GGES
Original issue is https://github.com/Reference-LAPACK/lapack/issues/475


MigMuc:
There is a spelling mistake LSAME instead of LLSAME.

pablosanjose:
> There is a spelling mistake LSAME instead of LLSAME.

Could somebody please confirm or refute this? I'm proposing this as a patch downstream (for Julia)

martin-frbg:
Oh, right you are - seems gfortran is not visibly annoyed as it "only" affects the declaration as EXTERNAL.

pablosanjose:
Thanks for the confirmation

Issue #3100, opened 2021-02-09
Use of configure_file to change file in include_directory (leading to issues building in Bazel)


jake-arkinstall:
I hope this isn't a big ask.

Some build systems place the source of projects in a read-only environment to make certain guarantees about the build process for caching purposes. Bazel is one example of this - it allows building CMake projects (and [the rules_foreign_cc project that enables this](https://github.com/bazelbuild/rules_foreign_cc) even uses this OpenBLAS repository as an example).

This kind of approach _usually_ works, because it's rather rare for a CMake project to make in-source writes. Most instead produce their intermediate steps in the build directory.

You may notice the above project passes `BUILD_WITHOUT_LAPACK` - possibly because LAPACK isn't needed for Eigen, but mainly because trying to build LAPACK under bazel fails due to a use of `configure_file` [to modify a file in the include directory](https://github.com/xianyi/OpenBLAS/blob/f6e4cf2f9dac6f90b36f49659d883767d63791f5/cmake/lapacke.cmake#L2502). This doesn't appear to be necessary, and other uses of `configure_file` in this project that I can see seem to place the changes in the output directory and later use them from there.

It isn't this project's responsibility to support Bazel (though it'd be awesome if it did), and I don't know enough about the OpenBLAS build to know how difficult it would be to adjust this step to load the target file from a space in the build directory instead, but I'm just throwing it out there to see if it's something that might be possible.

martin-frbg:
Thanks for pointing this out - this instance of configure_file appears to be redundant (with https://github.com/xianyi/OpenBLAS/blob/3e1e74fca64767833252aa594f2e673fb7d41fb4/CMakeLists.txt#L392 ) in any case - it may have been intended as a slightly more elegant replacement(?).

PR #3101, opened 2021-02-10
Addressed issue #3100 - removing an unnecessary write to the include directory

jake-arkinstall:
Removing an unnecessary write to the include directory, which @martin-frbg pointed out to appear redundant, and which prevents building when the source is read-only.

After removing the configure_file from line 2502 in `cmake/lapacke.cmake`, the build output is identical under the default configuration. When using CMake from the project root, the version in the build directory is used anyway - but this change should be tested on other configurations to ensure that this doesn't have unintended consequences. In particular, it could impact anyone using `cmake/lapacke.cmake` directly.


martin-frbg:
Thank you. I do not think anybody should be expecting to call lapacke.cmake directly here

Issue #2270, opened 2019-09-27
Deadlock in blas_thread_shutdown_() triggered by parallel calls to fork()


madMAx43v3r:
```
#0  __lll_lock_wait () at ../sysdeps/unix/sysv/linux/x86_64/lowlevellock.S:135
#1  0x00007fb456ee3a7d in pthread_cond_signal@@GLIBC_2.3.2 () at ../sysdeps/unix/sysv/linux/x86_64/pthread_cond_signal.S:140
#2  0x00007fb428bdb68e in blas_thread_shutdown_ () from /usr/lib/libopenblas.so.0
#3  0x00007fb4559563a5 in __libc_fork () at ../sysdeps/nptl/fork.c:96
```
I am getting this deadlock when spawning two threads which each call fork() at (presumably) almost the same time.
If I insert a 100 ms sleep between spawning the threads I do not get the deadlock.

https://linux.die.net/man/3/pthread_cond_signal

>It is not safe to use the pthread_cond_signal() function in a signal handler that is invoked asynchronously. Even if it were safe, there would still be a race between the test of the Boolean pthread_cond_wait() that could not be efficiently eliminated.

>Mutexes and condition variables are thus not suitable for releasing a waiting thread by signaling from code running in a signal handler.

As you can see in that link calling pthread_cond_signal() in a signal handler is not safe. I assume the fork callback is similar to a signal handler.

`/usr/lib/libopenblas.so.0 -> libopenblasp-r0.2.18.so`

martin-frbg:
Could you please check if the problem still occurs with the current r0.3.7  ?

madMAx43v3r:
Not easily, openblas is liked by cuBLAS which is already the newest version for Ubuntu 16.04 ...
I would have to recompile cuBLAS with a newer version, which probably wont work because of incompatibilities...

Ah, even worse:
>CUBLAS is not an open-source library, and the source code for it is not published by NVIDIA.

brada4:
You can symlink that libopenblas.so.0 to latest openblas library which will be installed by `make ; sudo make install` in /opt/openblas directory

PR #3111, opened 2021-02-18
Fix race in blas_thread_shutdown.

hawkinsp:
`blas_server_avail` was read without holding `server_lock`. If multiple threads call `blas_thread_shutdown` simultaneously, for example, by calling `fork()`, then they can attempt to shut down multiple times. This can lead to a segmentation fault.

This may be the same issue as #2270 although in my case it exhibited as a segmentation fault, not a deadlock.

### DIFF
--- a/driver/others/blas_server.c
+++ b/driver/others/blas_server.c
@@ -1024,38 +1024,39 @@ int BLASFUNC(blas_thread_shutdown)(void){
 
   int i;
 
-  if (!blas_server_avail) return 0;
-
   LOCK_COMMAND(&server_lock);
 
-  for (i = 0; i < blas_num_threads - 1; i++) {
+  if (blas_server_avail) {
+
+    for (i = 0; i < blas_num_threads - 1; i++) {
 
 
-    pthread_mutex_lock (&thread_status[i].lock);
+      pthread_mutex_lock (&thread_status[i].lock);
 
-    atomic_store_queue(&thread_status[i].queue, (blas_queue_t *)-1);
-    thread_status[i].status = THREAD_STATUS_WAKEUP;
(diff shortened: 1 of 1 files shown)


martin-frbg:
Ouch, thanks. Don't know how often I (and certainly many others) have looked at that part of the code without realizing.

hawkinsp:
Thanks!

I should add: I'm pretty sure some of the other code patterns in that file related to `blas_server_avail` and `server_lock` are also racy.

For example, suppose another thread calls `fork()` (and hence `blas_thread_shutdown`) immediately after the call to `blas_thread_init` in this line: https://github.com/xianyi/OpenBLAS/blob/1caa44bea999e196918afc095d26ad418db4ba19/driver/others/blas_server.c#L698
If I'm reading the code right, we might easily deadlock because the thread pool will be shutdown right before we enqueue work to it. That said, I don't know this code that well and it wasn't the bug I was hitting, so I didn't try for a wider fix.

In general it looks fairly suspicious to me to read `blas_server_avail` non-atomically without holding a lock (which happens in several functions) and without some sort of memory fence. This is the classic "double-checked locking" pattern and it is unsafe for the usual reasons. It might well work on x86 in practice with the right C compiler, but it is risky.

More broadly, I think it might be a mistake to shut down the parent's thread pool on fork; you most likely only really want to ensure that the child starts without a threadpool configured, and I think that could be done with a slightly more sophisticated `pthread_atfork()` handler. That said, calling `fork()` in a multithreaded program without following it with `exec()` is a fundamentally unsafe thing to do, so these kind of things are band-aid fixes at best. However, I don't know the history of this code and perhaps it is something users are relying on.

PR #3117, opened 2021-02-24
use /usr/bin/env perl

haampie:
The user might not have perl installed in /usr/bin/perl


martin-frbg:
thank you (weird, I thought we already did this ?)

PR #3134, opened 2021-03-10
update the amin function for riscv platform

wzgpeter:
improve the amin function with less vector instruction.
only verify it on QEMU.
build with C910 toolchain.
please help do more verify it.


martin-frbg:
Sorry, I do not have C910 hardware so "looks good in qemu" is only sufficient for bug fixes unless somebody endorses it.

PR #3143, opened 2021-03-14
Resolve circular dependency between common.h and param.h

martin-frbg:
caused by #3088


xoviat:
note: common.h was included make available BLASLONG

martin-frbg:
Yes, but common.h itself tries to include param.h  again. When getarch_2nd.c includes common.h instead of param.h, BLASLONG is available by the time param.h is seen. 
Unfortunately there is now also some other problem that caused the build to be broken after merging your patch, probably some conflict with one of the other recent pull requests that breaks one of the appveyor CI jobs

xoviat:
@martin-frbg I will take care of the other problem today. Give me ~30 minutes.

martin-frbg:
My fault for relying on git that did not flag a conflict. (I had run a local build before merging, but did not remember to test with OpenMP as well). Not quite sure I understand why iparam2stage does not get found in that one CI job

xoviat:
the cause of the problem is that lapack was updated after I submitted the pull request. flang cannot currently build one of the tests in the new lapack for some reason. my plan is to disable it for flang. I need to build locally, so I can avoid hogging the CI though.

martin-frbg:
Hm, I still do not get it - the only thing your PR changed was the "helper_prefix" stuff for the powershell script to run the xlintstc test, and the only related thing that changed in LAPACK at the same time was the source code for xlintstc (and xlintstz) - but no difference in the commands to compile and link that test.

xoviat:
see #3144

martin-frbg:
Ok, thanks but this still does not explain the "why"... to me at least ?

xoviat:
after #3088, more tests are enabled. see 

https://ci.appveyor.com/project/xianyi/openblas/builds/38214243/job/n9boy3y67yabhy28#L115746

vs 

https://ci.appveyor.com/project/xianyi/openblas/builds/38217979/job/x3xcdrpn8fygwui1#L107472

the cause of the failure is that #3122 modified the lapack files in such a way that flang could not compile them with openmp enabled. before #3088, testing could not detect this.

martin-frbg:
Agreed, my point is just that I do not see which change from #3122 could have caused flang (on Windows) to fail ? In particular with regard to iparam2stage.F that was not changed at all (and was compiled without generating any warnings).

PR #3144, opened 2021-03-14
disable openmp

xoviat:


xoviat:
cc @martin-frbg 

This should pass.

Issue #3160, opened 2021-03-21
int32 multiplication overflow at a fat side of 32bit input


brada4:
I think both inputs should be cast to native pointer prior to multiplication (they are checked unsigned already at that red point)
https://lgtm.com/projects/g/xianyi/OpenBLAS/latest/files/interface/imatcopy.c?sort=name&dir=ASC&mode=heatmap
https://lgtm.com/projects/g/xianyi/OpenBLAS/latest/files/interface/zimatcopy.c?sort=name&dir=ASC&mode=heatmap

martin-frbg:
Would you mind proposing a patch ? (Though I suspect any lda or ldb large enough to cause an overflow here would fail elsewhere a bit later when not building with INTERFACE64=1)

brada4:
This would get one bit more before failing, will do tomorrow, like in 10h

PR #3161, opened 2021-03-22
Rectify _?matcopy size calculation

brada4:
Shall address #3160 , avoiding oldest bit mis-handled as the sign.

### DIFF
--- a/interface/imatcopy.c
+++ b/interface/imatcopy.c
@@ -149,10 +149,7 @@ void CNAME( enum CBLAS_ORDER CORDER, enum CBLAS_TRANSPOSE CTRANS, blasint crows,
 
 #endif
 
-	if ( *lda >  *ldb )
-		msize = (*lda) * (*ldb)  * sizeof(FLOAT);
-	else
-		msize = (*ldb) * (*ldb)  * sizeof(FLOAT);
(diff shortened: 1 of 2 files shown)


martin-frbg (changes requested):
Hmm.. I do not like the unrelated&unannounced change to the allocation here - have to admit wernsaar's code looks a bit unexpected but I guess technically the output matrix is allowed to be bigger than the input (with some elements undefined)

brada4:
codeql demo says multiplication result is signed then assigned to unsigned eventually losing sign bit via overflow that might have been there. It is just arithmetic problem, sad it happens in the face of malloc.
this is just to clear the warning, quite obvious some sanity checking would be handy, but then we have no way to give controlled feedback

martin-frbg:
> this is just to clear the warning,

my point is that you removed the `if *lda > *ldb` logic for the calculation of msize. I am still not entirely sure if it is really needed, but it looks too elaborate to be a random mistake - and it has nothing to do with the intermediate overflow that you set out to fix. 
(probably it would even be enough to put parentheses around the `ldb * sizeof` part so that everything gets cast to size_t automatically)

brada4:
OK, can do double way if that is f concern.

brada4:
Thanks, fixed with other PR

PR #3166, opened 2021-03-24
Add OSX/Clang/OpenMP build to Azure

martin-frbg:


h-vetinari:
I saw you did a huge amount of debugging here - did you manage to find out something beyond what you commented on the feedstock issue?

martin-frbg:
#3184 has the actual find and fix - this problem must have hidden in plain sight since the days of GotoBLAS, seems yours was the first case where the memory overwrite actually hit something important

h-vetinari:
Sorry, hadn't immediately seen this because it had been merged already 😅 
Thanks so much for cracking this very tricky problem 🥳

PR #3205, opened 2021-04-30
optimize on sgemv_n for small n

intelmy:
This patch includes:
1.    Comparing with original sgemv_n implementation, change to not to divide on M and N for small N (N<=48), directly use small n kernel to calculate column by column, and use AVX512 registers to process maximum 128 floats at one time. The improvement can reach to 2.97x based on the validation on CascadeLake 8280L @2.7G
2.    Replace the original 4x2 and 4x1 kernel used in sgemv_n with our newly implemented kernel, the improvement when N > 48 is up to 10% on CascadeLake 8280L @2.7G

### DIFF
--- a/kernel/x86_64/sgemv_n_4.c
+++ b/kernel/x86_64/sgemv_n_4.c
@@ -417,7 +417,11 @@ int CNAME(BLASLONG m, BLASLONG n, BLASLONG dummy1, FLOAT alpha, FLOAT *a, BLASLO
 
 			if ( n2 & 2 )
 			{
+#ifdef HAVE_SGEMV_N_SKYLAKE_KERNEL				
 				sgemv_kernel_n_64(NB, 2, alpha, a_ptr, lda, x_ptr, ybuffer);
+#else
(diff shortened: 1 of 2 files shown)


kernel/x86_64/sgemv_n_4.c
```diff
@@ -305,9 +343,6 @@ int CNAME(BLASLONG m, BLASLONG n, BLASLONG dummy1, FLOAT alpha, FLOAT *a, BLASLO
 	BLASLONG lda8 =  lda << 3;
 	FLOAT xbuffer[8],*ybuffer;
 
-        if ( m < 1 ) return(0);
-        if ( n < 1 ) return(0);
```
martin-frbg, line 309: Why this deletion - it looks like a safe optimization to me to do a quick return here if there is no work to be done ?

intelmy: Thanks for review, Martin. Yes, they are useful, not deleted, but revised to be one line at line 297 as below:
  if ( m < 1 || n < 1) return(0);

martin-frbg: Ah, silly me - sorry.

PR #3206, opened 2021-04-30
Import packing improvements to LAPACK xLAQR from Reference-LAPACK (PR 480+535)

martin-frbg:


h-vetinari:
Just a quick question: instead of backporting individual PRs from lapack, would it not perhaps make more sense to bump the vendored lapack to 3.9.1?

martin-frbg:
I'm cherrypicking from 3.9.1 _plus_ subsequent fixes right now, plan to do the full "bump" at 3.10.0 (There are a few differences between OpenBLAS and Reference particularly but not exclusively in the build Makefiles, and also the final diff will be huge as Reference-LAPACK has decided to remove version numbers from all source files. I do not have the time and energy to tackle that (and the almost inevitable fallout from such a big merge) right now but still want to provide the improved functionality)

PR #3207, opened 2021-05-01
x86: Enable Intel CET

hjl-tools:
When Intel CET is enabled, we need to include <cet.h> in assembly codes
to mark Intel CET support and place _CET_ENDBR at the function entry.


martin-frbg:
Thank you very much

PR #3220, opened 2021-05-04
Delete lapack_wrappers.c.orig

drhpc:
This looks like a leftover from patching and confuses further patching;-)


martin-frbg:
Right, thanks.

PR #3260, opened 2021-06-08
Optimized sgemv_t for small N based on AVX512

intelmy:
This patch is to optimized sgemv_t for N=1-8, as original algorithm would divied M and N no matter how big M and N is. This patch is to customize kernel for individual N size to get  much better performance with vectorization . 
The performance improvement is up to 8.16x per the verification on Cascade server


martin-frbg:
Unfortunately there appears to be a problem with the sgemv_t microkernel for SkylakeX that you introduced with this PR, causing about 1060 failures in the LAPACK testsuite. Apologies for not catching this earlier, and I wonder if you could recheck your contribution ? Running `make lapack-test` after a successful build should be sufficient to demonstrate the issue.

martin-frbg:
Now fixed by guowangy's #3348

PR #3265, opened 2021-06-10
Removed use of non portable '-p' arg to install

TAAPArthur:
Not all versions of install support '-p' flag and it isn't worth failing
the build in the installed files' timestamps get updated.

### DIFF
--- a/Makefile.install
+++ b/Makefile.install
@@ -74,17 +74,17 @@ endif
 ifneq ($(OSNAME), AIX)
 ifndef NO_LAPACKE
 	@echo Copying LAPACKE header files to $(DESTDIR)$(OPENBLAS_INCLUDE_DIR)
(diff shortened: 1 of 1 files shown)


martin-frbg:
Curious to know whose `install` does not support the `-p` option (besides AIX, which is already handled in Makefile.install) ?

TAAPArthur:
@martin-frbg [suckless's sbase](https://core.suckless.org/sbase/). 
```
$ install --help
usage: install [-g group] [-o owner] [-m mode] (-d dir ... | [-D] (-t dest source ... | source ... dest))
```

martin-frbg:
Oops, a self-inflicted bavarian tool to turn any system into an early HPUX ? (only kidding)

brada4:
Would be nice if them to discard `-c` and `-p` options to mimic solaris version of `install`

PR #3266, opened 2021-06-10
Remove spurious casts from PPC parameters and fix compilation for older targets

martin-frbg:
addresses recent comments on #3148


cptnapalm:
Could you do this for SPARC V9.  Was trying to compile on my UltraSparc laptop and got the same errors as PPC64 big endian.  I don't have a V7, so I can't check if that needs it too.

PR #3278, opened 2021-06-19
Add CORTEXA55 cpuid 0xd05 support

brada4:
- based on A53
- cache sizes snooped off wikipedia
- compiler parameters reduced for older versions spotting neoverse
btw aarch64 enhanced ISAs do not apply with CLANG


martin-frbg:
could you add the changes for dynamic_arm64.c please ?

Djip007:
ODROID C4 build: 
```

  OS               ... Linux             
  Architecture     ... arm64               
  BINARY           ... 64bit                 
  C compiler       ... GCC  (cmd & version : cc (Ubuntu 10.2.0-13ubuntu1) 10.2.0)
  Fortran compiler ... GFORTRAN  (cmd & version : GNU Fortran (Ubuntu 10.2.0-13ubuntu1) 10.2.0)
  Library Name     ... libopenblasp-r0.3.15.dev.a (Multi-threading; Max num-threads is 4)
  Supporting multiple arm64 cpu models with minimum requirement for the common code being CORTEXA55
```

brada4:
Oops, now corrected @martin-frbg

brada4:
:crossed_fingers:

driver/others/dynamic_arm64.c
```diff
@@ -99,6 +99,11 @@ extern gotoblas_t  gotoblas_NEOVERSEN1;
 #else
 #define gotoblas_NEOVERSEN1 gotoblas_ARMV8
 #endif
+#ifdef DYN_CORTEX_A55
+extern gotoblas_t  gotoblas_CORTEXA55;
+#else
+#define gotoblas_NEOVERSEN1 gotoblas_ARMV8
```
martin-frbg: gotoblas_CORTEXA55 not neoverse

Djip007:
https://github.com/xianyi/OpenBLAS/issues/3279#issuecomment-864596533
look good for me... dynamic_arm64 look good now

If you want I made some more/special test ... ask!

brada4:
Thank you for taking time to test. Intent of this is to make sure OpenBLAS builds & works on casual hardware, so that with next cheaper release of same SoC bugtracker does not get flooded with identical requests.

brada4:
Let it rest a bit to free drone slot

cpuid_arm64.c
```diff
@@ -284,6 +289,8 @@ void get_cpuconfig(void)
       printf("#define %s\n", cpuname[d]);
       // Fall-through
 		case CPU_ARMV8:
+		// case CPU_CORTEXA53;
+		// case CPU_CORTEXA55;
```
martin-frbg: this needs to be an actual "case CPU_CORTEXA55:" (and ending with a colon) else it will not set DTB_DEFAULT_ENTRIES for the A55 (and I think the commented-out case line for A53 should not be added, we are falling through from CORTEXA53 just a couple of lines above).
Or perhaps the "case CPU_CORTEXA55:" could be added to the A57/A72/A73 block of cases below, unless we are
certain that a "typical" A55 chipset will be as cache-limited as the older A53 designs ?

martin-frbg: and actually that "case CPU_CORTEXA55:" needs to come directly after the "case CPU_CORTEXA53:" line (so that in both cases the "print #define cpuname" is executed - else you end up with wrong UNROLL parameters later)

brada4: I tried to mark the sport that both A53 and A55 fall through, probably needs own section.

martin-frbg: the point is that the sequence needs to be
```
case CPU_CORTEXA53:
case CPU_CORTEXA55:
 (write the cpuname define to the config file)
case ARMV8:
(the fall-through, write what is needed by all three)
```
so no real need to mark anything (except note the fact that it is intentional fall-through and not a missing break there)

brada4: OK, got it.

brada4:
Now got your idea, sorry for being slow, and thank you for patience.

PR #3280, opened 2021-06-21
Support CORTEXA55 0xd05 and reduce drone resource usage

brada4:
And let clang++ compile if needed.


brada4:
Ill make separate if this gets through, at least default was to much for 128GB swap on raspberry

martin-frbg:
Pretty sure there is something else going on at drone.io - they are using serious hardware and there is no way addition of just one more DYNAMIC_ARCH target could have driven up build times like that. Perhaps it is just rate limiting kicking in due to the number of rebuilds in fairly quick succession.

brada4:
Ill wait some before re-opening A55

martin-frbg:
Looks like closing the PR does not (always) cancel the associated CI on drone (which I am pretty sure it did in the past).

brada4:
Does not let me operate cancel button either ;-)

martin-frbg:
Interesting. I managed to cancel your pending re-run of this PR, and its original invocation finished successfully a minute ago. I think there are no pending jobs now.

brada4:
I will push both, if it fails I will follow the slow way one a day.

brada4:
Looks like cached result is stuck.

brada4:
Rationale of having it in the low-end pool:
- mass market pieces will be always cache-challenged
- it is power-efficient piece, unlikely to scale into upper segments.
I suppose it needs a rewrite around cpuid (and dynamic thereof) to be in best shape for mass market SoC products and enormous network switches alike.

PR #3281, opened 2021-06-22
reduce drone ci cmake processes to number of CPUs mimicking normal make

brada4:
Moderate drone's number of cpus instead of priming up all at once.


brada4:
96 CPUs, useless to limit.

martin-frbg:
If I recall correctly, drone.io alternates between 96-core ThunderX and 48-core Falkor for ARMV8, and provides 1st generation AMD Threadripper for x86_64.

brada4:
It is a no concern by large, but could save raspberries of copypasters like me ;-)

PR #3297, opened 2021-07-10
fix compilation with musl libc

rdck:
Programs including the installed `<openblas_config.h>` fail to compile with musl libc:
```
test_sched_include.c: In function 'main':
test_sched_include.c:6:2: error: unknown type name 'cpu_set_t'
    6 |  cpu_set_t* cpu_set = NULL;
      |  ^~~~~~~~~
```
This is because `<sched.h>` only defines `cpu_set_t` when the feature test macro `_GNU_SOURCE` is set.

I added a test for this in https://github.com/xianyi/OpenBLAS/commit/447084b95a0da34f852086c35effe56f20653614. Testing requires performing an installation, so I used a separate Make target from the existing tests, and added a new directory `test_install` for tests specific to installation. Installation is done to `install/`. I don't really think testing for this is necessary, though. Feel free to remove it.

Additionally, musl doesn't provide `<linux/unistd.h>`. As far as I can tell, it's not actually needed for anything as long as `<sys/syscall.h>` is included, but it would be good to get confirmation of this.

I tested this on Alpine:
```
OS               ... Linux
Architecture     ... x86_64
BINARY           ... 64bit
C compiler       ... GCC  (cmd & version : cc (Alpine 10.2.1_pre1) 10.2.1 20201203)
Fortran compiler ... GFORTRAN  (cmd & version : GNU Fortran (Alpine 10.2.1_pre1) 10.2.1 20201203)
Library Name     ... libopenblas_haswellp-r0.3.15.dev.a (Multi-threading; Max num-threads is 8)
```


martin-frbg:
From looking at similar musl issues in other projects you appear to be right about unistd.h vs. syscall.h
I do think the install test should be done in the appropriate CI job rather than added to the build system though.

martin-frbg:
install test moved to Azure CI in #3300

rdck:
> I do think the install test should be done in the appropriate CI job rather than added to the build system though.

That makes more sense, thanks!

PR #3306, opened 2021-07-12
Add cpu detection support for Zhaoxin processors

### LINKED ISSUE #3295: Add cpu detection support for Zhaoxin processors
The current openblas version cannot recognize the Zhaoxin CPU. Now add cpu detection support for Zhaoxin processors. Set the default TARGET to NEHALEM.

jonaszhou1:
Fix #3295

Signed-off-by: JonasZhou <JonasZhou@zhaoxin.com>

### DIFF
--- a/cpuid.h
+++ b/cpuid.h
@@ -54,6 +54,7 @@
 #define VENDOR_TRANSMETA  9
 #define VENDOR_NSC	 10
 #define VENDOR_HYGON	 11
+#define VENDOR_ZHAOXIN   12
 #define VENDOR_UNKNOWN   99
 
 #define BITMASK(a, b, c) ((((a) >> (b)) & (c)))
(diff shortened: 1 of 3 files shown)


martin-frbg:
Thank you, that is certainly a bit cleaner than my guesses. I wonder if there is AVX/AVX2 support in newer models (which would allow identifying them as Sandybridge/Haswell for somewhat faster code) ?

jonaszhou1:
> 
> 
> Thank you, that is certainly a bit cleaner than my guesses. I wonder if there is AVX/AVX2 support in newer models (which would allow identifying them as Sandybridge/Haswell for somewhat faster code) ?

Family 7 processors support AVX, but it performance is weak.

jonaszhou1:
> 
> 
> Thank you, that is certainly a bit cleaner than my guesses. I wonder if there is AVX/AVX2 support in newer models (which would allow identifying them as Sandybridge/Haswell for somewhat faster code) ?

Support AVX, and AVX2 is partially supported, but SSE instructions perform better on some platforms.

martin-frbg:
I see, thanks.

PR #3308, opened 2021-07-12
[WIP,Testing] run lapack testsuite on OSX

martin-frbg:


h-vetinari:
You might be interested in a recent [bug](https://github.com/flame/blis/issues/514#issuecomment-876582767) in blis that only appeared on AVX512, and was also changed by switching the calling convention to HASWELL.

martin-frbg:
Thanks for the pointer, though I guess that case is a bit different, the AVX512 code here is using Intel intrinsics rather than handcrafted assembler. So either I have inadvertently pulled the plug on code that only happened to work due to an unrelated allocation of stack space to scribble on, or I am misunderstanding who is responsible for allocatng stack space for AVX512 register saves when using intrinsics in general.
.

h-vetinari:
> [...] or I am misunderstanding who is responsible for allocating stack space for AVX512 register saves when using intrinsics in general.

It seems you're not alone on that, at least, another very similar discussion is [happening](https://github.com/conda-forge/numpy-feedstock/pull/238#issuecomment-879840507) regarding pypy (but on windows).

PR #3320, opened 2021-07-18
Empirical workaround for numpy SVD NaN problem from issue 3318

### LINKED ISSUE #3318: nan in svd N for Haswell core
Please forgive me for this bug-report - coming from Numpy - but I think we've hit a bug in the singular value calculation for the Haswell core type - see https://github.com/numpy/numpy/issues/19473

Specifically, only on the Haswell core, we are getting unexpected `nan` values in the case where we compute the singular values without the UV matrices.   The files to reproduce in Numpy are here: 
[svd_check.zip](https://github.com/xianyi/OpenBLAS/files/6830660/svd_check.zip)

The check file  is `svd_check.py` in the archive - and contains:

```python
import numpy as np
from numpy.linalg._umath_linalg import svd_n

mat = np.loadtxt('1.csv', delimiter=',')
s = svd_n(mat, signature='d->d')
print(np.count_nonzero(np.isnan(s)))
```

We run with many cores using the `test_svds.sh` script:

```bash
#!/bin/bash
export OPENBLAS_VERBOSE=2
for core in prescott dunnington penryn core2 nehalem sandybridge haswell; do
    export OPENBLAS_CORETYPE=$core
    python svd_check.py
done
```

This gives 5 `nan` values for the Haswell core, and 0 otherwise.   If we compute the UV matrices at the same time, we never get `nan` in the SVs, regardless of core.

Could this be a bug in the SV-only computation in OpenBLAS?

martin-frbg:
more Voodoo than fix, actual problem probably just papered over by the -mfma option
fixes #3318 (for now)


matthew-brett:
Please forgive my ignorance about the testing framework, but it is practical add a test to exercise this, perhaps with the original array from #3318 ?

martin-frbg:
Only if I manage to convert it to C - right now I am still wondering what this is trying to tell me (or if is just a persistent gcc bug)

carlkl:
@martin-frbg, using GCC one would very much prefer to use `-march=` options instead of `-msse2`, `-mavx` and so forth. `-mavx2` for example does not switch on `-mfma`, you have to use `-march=haswell` (Corrected).  The `-march=` flags controls much more instructions.

This behaviour is different to clang and MSVC. On the other hand with GCC  you have fine control over the instructions set.

I'm able to prepare a PR for that if desired.

martin-frbg:
I am not convinced as the original issue appeared to have been caused by applying `-mfma` to "too many" files so I do not yet think an even less fine-grained application of options would help.

mattip:
Stupid question that probably is already present in testing: is there a way to check that the xmm/ymm registers are being properly saved/restored and wrap the kernel calls with it for testing?

martin-frbg:
Unfortunately not, so far it has always been retroactive fixing as things blew up with the next smarter release of gcc and/or somebody shouting "oi, you need to save that". (I do believe most such bugs have been fixed in my time here, but I fear I still know just enough assembly to be dangerous. And actually the last change to the Haswell DGEMV microkernel has been to proactively save all xmm/ymm registers instead of just those directly touched by the code - but reverting that had no bearing on the issue)

mattip:
A wild theory (I have lots of these): maybe it is outside the kernel, and somewhere else someone is not restoring a (different?) register. By adding the flag the registers are used differently and so the problem does not appear. Not sure what would be the easiest way to verify this and how much effort it would be.

PR #3326, opened 2021-07-27
Include Haiku in processor count checks

wattoc:
Haiku should be performing processor count checks and using them to limit thread creation as the other OS's do.

This is currently being patched in on HaikuPorts against 0.3.17: https://github.com/haikuports/haikuports/pull/6094

### DIFF
--- a/driver/others/memory.c
+++ b/driver/others/memory.c
@@ -428,7 +428,7 @@ extern int openblas_goto_num_threads_env();
 extern int openblas_omp_num_threads_env();
 
 int blas_get_cpu_number(void){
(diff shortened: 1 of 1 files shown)


brada4:
Is this correct CPU number detection?
```
#ifdef OS_HAIKU
int get_num_procs(void) {
  static int nums = 0;
  if (!nums) nums = sysconf(_SC_NPROCESSORS_CONF);
  return nums;
}
#endif
```

wattoc:
Yes, Haiku supports this call and it is used by other apps to count CPUs:

https://github.com/haiku/haiku/search?q=_SC_NPROCESSORS_CONF

Issue #2587, opened 2020-04-28
[WIP]Optimize gemm for small matrix


xianyi:
- [x] Add basic implementation ( please check aae6af94bbe4f7ad97c417e40fe6a7d4a2798b79 )

- [ ] Merge sgemm_kernel_direct implementation

- [ ] Work for DYNAMIC_ARCH

- [ ] Tune the input matrix size.

- [ ] Add optimized kernel for architecture.

antoscha:
Probably, you're aware of this project, but I'm leaving it here just for reference:
https://github.com/hfp/libxsmm

ValZapod:
See https://github.com/xianyi/OpenBLAS/issues/3783

PR #3335, opened 2021-08-05
Add GEMM optimization for small matrix and single/double kernel for skylakex

guowangy:
GEMM optimization for small matrix without COPY (#2587).

1. cherry-pick and rebase reference design from `small_matrices` branch 
2. add S/DGEMM small matrix kernel (nn, nt, tn, tt) for skylakex
3. add `GEMM_SMALL_MATRIX_PERMIT` interface to tune small matrices
4. support DYNAMIC_ARCH and cmake build
5. enable SMALL_MATRIX_OPT by default for x86_64 arch (if not customized kernel defined, it will just by-pass to normal path)


brada4:
Do you have any benchmark or indicative matrix sizes where simple unrolled loop beats optimized assembly?

guowangy:
For skylakex platform, we define the matrix size (`M * N * K <= 10^6`, except TN kernel: `M * N <= 1200 && K >= 32`) will be applied to small matrix optimization. The size can be tuned in files:
* `sgemm_small_kernel_permit_skylakex.c`
* `dgemm_small_kernel_permit_skylakex.c`

For benchmark, each kernel I test S/DGEMM with different M, N, K combination: 
* `M=x, N=100, K=100`
* `M=100, N=x, K=100`
* `M=100, N=100, K=x`
 where x = 1, 2, 4, 6, 8, 12, 16, 32, 48, 64, 100

From my testing, it shows that small matrix optimization can reach 1.1x ~ 3.x in GFlops compared to original design in single thread mode.

brada4:
What I am saying that skylakex small kernels are heavily optimized, while for all others they are on par with very generic implementation. Thus the question id if on e.g. HASWELL and ZEN things are not made slower. Just small test with TARGET=HASWELL should tell on samke skylakex system.

guowangy:
Whether to use small matrix kernel is determined by `GEMM_SMALL_MATRIX_PERMIT` function.
The generic kernel is very inefficient, so I disabled them in 6b58bca by default. 

Unless customized  `GEMM_SMALL_MATRIX_PERMIT` is defined (like we did in  `s/dgemm_small_kernel_permit_skylakex.c`), it will just by-pass to normal kernel.

I also verify in my machine with `TARGET=HASWELL`, it works as expected: the generic small matrix kernel is not used and performance is not made slower.

brada4:
Those unoptimized kernels actually resemble generic/* , I dont think those shall be repeated. I dont know.

guowangy:
The small matrix framework and generic kernel come from @xianyi.
By design, it needs various kernel although they looks quite similar. 

If the number of source files is really a problem, my suggestion is to make *b0* kernel to share the source file with base kernel by define an extra marco like below:
```cpp
#ifdef B0  // B0 routine
int CNAME(BLASLONG M, BLASLONG N, BLASLONG K, FLOAT * A, BLASLONG lda, FLOAT alpha, FLOAT * B, BLASLONG ldb, FLOAT * C, BLASLONG ldc)
#else
int CNAME(BLASLONG M, BLASLONG N, BLASLONG K, FLOAT * A, BLASLONG lda, FLOAT alpha, FLOAT * B, BLASLONG ldb, FLOAT beta, FLOAT * C, BLASLONG ldc)
#endif
```
and add `-DB0` flags in *Makefile* when building *b0* kernel.

This can reduce the source file number by half.

guowangy:
@martin-frbg 
Do you have any suggestion or comment for this PR?
Is it ready for merge?

martin-frbg:
have not had time to review this fairly big change (also would like to see xianyi's take on this, as it started from his now somewhat stale branch)

martin-frbg:
Looks good to me, thank you very much.

PR #3350, opened 2021-08-26
[WIP] malloc a new control structure when we run out of thread space

martin-frbg:
for #3321


martin-frbg:
bah, broken upload again

PR #3354, opened 2021-08-31
[win/arm64]: Explicit casting for GEMM_DEFAULT_ALIGN to create 64-bit value

niyas-sait:
Win64 uses the LLP64 data model and unsigned long is only 32-bit. For 64-bit architecture, we need a 64-bit mask to correctly generate the address.


martin-frbg:
Probably safer to make this casting `ifdef __WIN64` or something ? I had to throw out a lot of the earlier, generous sprinkling of casts (#3088) in the meantime as some compilers/preprocessors choked on it. (And I suspect now that all the (BLASLONG) added for x86 back then should have been BLASULONG as well ?)

niyas-sait:
> Probably safer to make this casting `ifdef __WIN64` or something ? I had to throw out a lot of the earlier, generous sprinkling of casts (#3088) in the meantime as some compilers/preprocessors choked on it. (And I suspect now that all the (BLASLONG) added for x86 back then should have been BLASULONG as well ?)

Thanks, @martin-frbg. I have added a new commit to make it conditional on `_WIN64 `flag. 

Yes using BLASULONG for x86 make sense to me. Do you want me to update it in this patch or shall i leave it for another PR ?

martin-frbg:
Thank you. 
As you prefer - I could also do this later (and remove any remaining casts from non-Win64 architectures)

Issue #2949, opened 2020-10-26
OpenBLAS 0.3.12 skylakex build fail


minchaoyue:
In file included from ../kernel/x86_64/dasum.c:8:0:
../kernel/x86_64/dasum_microk_skylakex-2.c: In function ‘dasum_kernel’:
../kernel/x86_64/dasum_microk_skylakex-2.c:42:38: error: incompatible type for argument 1 of ‘_mm512_abs_p’
             accum_0 += _mm512_abs_pd(_mm512_load_pd(&x1[i + 0]));
                                      ^~~~~~~~~~~~~~
cc -c -O2 -DMAX_STACK_ALLOC=2048 -Wall -m64 -DF_INTERFACE_GFORT -fPIC -DSMP_SERVER -DNO_WARMUP -DMAX_CPU_NUMBER=64 -DMAX_PARALLEL_NUMBER=1 -DBUILD_SINGLE=1 -DBUILD_DOUBLE=1 -DBUILD_COMPLEX=1 -DBUILD_COMPLEX16=1 -DVERSION=\"0.3.12\" -msse3 -mssse3 -msse4.1 -march=skylake-avx512 -mavx2 -UASMNAME -UASMFNAME -UNAME -UCNAME -UCHAR_NAME -UCHAR_CNAME -DASMNAME=camax_k -DASMFNAME=camax_k_ -DNAME=camax_k_ -DCNAME=camax_k -DCHAR_NAME=\"camax_k_\" -DCHAR_CNAME=\"camax_k\" -DNO_AFFINITY -I.. -UDOUBLE  -DCOMPLEX -DCOMPLEX -UDOUBLE  -DUSE_ABS -UUSE_MIN ../kernel/x86_64/zamax_sse.S -o camax_k.o
In file included from /opt/rh/devtoolset-7/root/usr/lib/gcc/x86_64-redhat-linux/7/include/immintrin.h:45:0,
                 from ../kernel/x86_64/dasum_microk_skylakex-2.c:6,
                 from ../kernel/x86_64/dasum.c:8:
/opt/rh/devtoolset-7/root/usr/lib/gcc/x86_64-redhat-linux/7/include/avx512fintrin.h:7301:1: note: expected ‘__m512 {aka __vector(16) float}’ but argument is of type ‘__m512d {aka __vector(8) double}’
 _mm512_abs_pd (__m512 __A)
 ^~~~~~~~~~~~~
[...]

PR #3357, opened 2021-09-02
Add a Windows build with mingw gcc and make to AzureCI

martin-frbg:


martin-frbg:
Hitting gcc bug from #2949 on SKX now due to the preinstalled mingw 8.1.0...

Issue #2754, opened 2020-07-31
Adding POWER9 to travis CI


RajalakshmiSR:
It would be nice to add a P9 system in our travis CI.
There are POWER9 systems at OSU. We can  request for VM here:
https://osuosl.org/services/powerdev/

RajalakshmiSR:
cc @ceseo

martin-frbg:
I must admit it is unclear to me how this meshes with travis CI - as far as I know their P8 offering (still a semi-official beta ?) is using osuosl but no idea if they would be easily convinced to expand it to P9. Or are you suggesting setting up a generic CI of our own using a github workflow or whatever?

ceseo:
@martin-frbg I think that, as a community, submitting a request for adding P9 to Travis CI wouldn't hurt. I understand that may take a while to happen.

Meanwhile, adding a way to test on P9 in the workflow sounds like a very good idea. Is it something that you would consider?

martin-frbg:
@ceseo certainly - I "only" need to read up on how to do it with an external "self-hosted" site.

ceseo:
@martin-frbg cool! If you need help setting up a P9 at OSU, please let me know.

martin-frbg:
Access request submitted, still a bit sketchy about setup but found their powerci wiki documentation already

martin-frbg:
Request got approved and both xianyi's and my github username added to powerci. No idea yet how to match the "freestyle setup" described in the wiki to what is actually available on the Jenkins dashboard.

martin-frbg:
@ceseo if you could point me to some documentation for setting up jobs with the osuosl Jenkins specifically that would be great - right now I do not even see anything in their interface that would allow me to do anything except add a bio to my username.
BTW I have also asked about P9 support in the travis community forum but judging from their blog they seem to be focused on expanding their aarch64 portfolio.

martin-frbg:
Just received a reply from travis - with their recent move to the POWERVS cloud all their POWER  backend is actually P9 now apparently.

ceseo:
@martin-frbg which request form did you submit? The POWER CI form, or the OpenStack one?

Regarding Travis, it seems the problem is that the build script is passing `-mcpu=power8`, as you can see [here](https://travis-ci.org/github/xianyi/OpenBLAS/jobs/716881083)

martin-frbg:
[...]

PR #3367, opened 2021-09-08
POWER: Fixing syntax error in makefile

RajalakshmiSR:
Fixing syntax issue in Makefile.power added by recent commit
af19cda65aef4d033ae33213013c88b0a99f9da2

### DIFF
--- a/Makefile.power
+++ b/Makefile.power
@@ -12,7 +12,7 @@ endif
 ifeq ($(CORE), POWER10)
 ifneq ($(C_COMPILER), PGI)
 CCOMMON_OPT += -Ofast -mcpu=power10 -mtune=power10 -mvsx -fno-fast-math
-ifeq ($(F_COMPILER, IBM)
+ifeq ($(F_COMPILER), IBM)
 FCOMMON_OPT += -O2 -qrecur -qnosave
 else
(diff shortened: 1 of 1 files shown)


martin-frbg:
Aww, sorry for that.

RajalakshmiSR:
Do you know how we can avoid these issues with our Travis CI  as we have POWER9 added in Travis. Ref : https://github.com/xianyi/OpenBLAS/issues/2754

martin-frbg:
I do not think Travis (or actually OSUOSL providing the POWER9 hardware to them) is offering POWER10 yet. Perhaps I can set up qemu (POWER10 support appears to be very limited but may be enough to get a build started), but primarily I need to be more
careful with edits (or stop using github's simple web editor for good)

RajalakshmiSR:
Yes we don’t have POWER10 yet in OSUOSL, but the question is about general POWER testing and not related to this issue/PR.
Do we have an active POWER build job?

martin-frbg:
The one in Travis is active again since a few days (needed xianyi to initiate the migration to travis.com and apply for free credits). But it cannot catch bugs that I introduce in P10-specific sections of the Makefile.

RajalakshmiSR:
Thanks @martin-frbg

PR #3387, opened 2021-09-26
Makefile.system: adjust mipsel/mips64el ARCH variables

commodo:
When building for MIPS{64} little-endian variants, the included makefiles
should be the same as for the big-endian.

There are already some adjustments being done for some ARCH names.
This change adds the ones for the `mipsel` and `mips64el` names, so that
the Makefile.mips{64} files get included.

This comes as a result of: https://github.com/openwrt/packages/issues/16649

Signed-off-by: Alexandru Ardelean <ardeleanalex@gmail.com>

### DIFF
--- a/Makefile.system
+++ b/Makefile.system
@@ -33,6 +33,10 @@ else ifeq ($(ARCH), armv7)
 override ARCH=arm
 else ifeq ($(ARCH), aarch64)
 override ARCH=arm64
+else ifeq ($(ARCH), mipsel)
+override ARCH=mips
+else ifeq ($(ARCH), mips64el)
+override ARCH=mips64
 else ifeq ($(ARCH), zarch)
 override ARCH=zarch
 endif


commodo:
thank you :)

martin-frbg:
well, thank _you_
(btw I have a patchset lying around for basic support of Octeon3, but I guess it will not add anything worthwile for openwrt?)

commodo:
> well, thank _you_
> (btw I have a patchset lying around for basic support of Octeon3, but I guess it will not add anything worthwile for openwrt?)

i'm not fully aware of the differences between Octeon versions;
I don't use MIPS{64} unfortunately;
so, if you add it, it won't be used (just yet), but maybe someone can benefit from it at some point;

i try to maintain OpenBLAS on OpenWrt, because it was requested, and i (also) maintain numpy on OpenWrt;
I try to keep OpenBLAS at a good-enough level that it can be used on OpenWrt;
and users can tweak the build target if they want;
and if it's present, people can request/submit improvements for low-level optimizations later;

PR #3390, opened 2021-09-30
Make sure that Netlib LAPACK respects FFLAGS

Keno:
OpenBLAS allows users to specify `FFLAGS` and then uses `override` to append additional
options. However, without such an override in lapack's make.inc, lapack would use the external
FFLAGS, rather than the ones being computed by OpenBLAS. For example the `DEBUG=1` flag
would not apply to LAPACK code. This is all a bit messy but forced by the integration with netlib
lapack. Note that `CFLAGS` already has this override for the same reason. It is possible that
other variables here should have a similar override, but I think for most of the other ones, OpenBLAS's
build system does not append to the flags passed in by the user.

cc @staticfloat @ianatol

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -269,7 +269,7 @@ prof_lapack : lapack_prebuild
 lapack_prebuild :
 ifeq ($(NOFORTRAN), $(filter 0,$(NOFORTRAN)))
 	-@echo "FC          = $(FC)" > $(NETLIB_LAPACK_DIR)/make.inc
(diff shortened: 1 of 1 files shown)


staticfloat (approved):

martin-frbg:
Right, thanks. The overrides have been added on a case-by-case basis so far, and I do not think any of the remaining variables need them. 
(DEBUG=1 may be a corner case, as in if you set your own FFLAGS you might as well put the `-g` there if you know you need it, but things could get interesting if the predefined FFLAGS dropped the `recursive` that a number of functions in Reference-LAPACK need)

PR #3410, opened 2021-10-16
Fix MinGW/Clang 64 bits detection.

MehdiChinoune:
CMAKE_COMPILER_IS_GNUCC is only valid for GCC.


MehdiChinoune:
Travis-CI failue on PPC64LE is not related to my changes

martin-frbg:
Thanks - the mingw/clang combination has not come up before. (I wonder if your change would exclude non-mingw builds of gcc - assuming they exist?)

martin-frbg:
> Travis-CI has PPC64LE is not related to my changes

Right - looks like a broken network link. And now Travis tells me we have run out of credits again - despite using only third-party hardware for which they claim not to charge anything at all.

MehdiChinoune:
> Thanks - the mingw/clang combination has not come up before. (I wonder if your change would exclude non-mingw builds of gcc - assuming they exist?)

The conditions/tests were made to set MINGW64.

martin-frbg:
I was just wondering if the original code would happen to support more than just mingw, but I guess any others would probably be taken care of by the CMAKE_CL_64 later on. Thanks again for the patch.

PR #3412, opened 2021-10-16
Silent warning from cmake about version<2.8.12

MehdiChinoune:


martin-frbg:
I know cmake 2.8.5 is about ten years old, but if this is just to silence a harmless warning I do not think it is worth it to potentially cut off anybody who cannot update their old system for some reason. From https://discourse.cmake.org/t/how-to-fix-cmake-minimum-required-deprecation-warning/2487 the alternative would appear to be to use something like `2.8...3.13` instead to indicate compatibility with both ancient and recent versions (?)

MehdiChinoune:
> I know cmake 2.8.5 is about ten years old, but if this is just to silence a harmless warning I do not think it is worth it to potentially cut off anybody who cannot update their old system for some reason. From https://discourse.cmake.org/t/how-to-fix-cmake-minimum-required-deprecation-warning/2487 the alternative would appear to be to use something like `2.8...3.13` instead to indicate compatibility with both ancient and recent versions (?)

The oldest supported OS CentOS/REHL 7 has cmake 2.8.12
Do you know any (still supported) OS that has cmake<2.8.12?

martin-frbg:
I'm thinking more about unsupported OS versions (which should be more or less fine if sufficiently isolated from the internet). As long as we do not need cmake functionality not provided by 2.8.5, I do not see the need to bump the minimum version requirement

PR #3417, opened 2021-10-18
[cmake] Build only for targets newer than TARGET

MehdiChinoune:
Build only for targets newer than TARGET when DYNAMIC_ARCH is enabled and a TARGET is selected.
I don't know how to implement this for Makefiles.

This will also reduce buildtime.


martin-frbg:
In Makefile syntax one could probably achieve this with the "word" and "wordlist" functions but I expect it would be a bit messy.
Not sure if this is really useful - in particular, it is not always the case that a cpu later in the list is actually newer (though for x86_64 this happens to be mostly true at the moment).

MehdiChinoune:
It needs more tinkering.
Close it for now

PR #3423, opened 2021-10-26
s390x: use DYNAMIC_ARCH's cpu detection for compile-time choice

mhillenbrand:
On s390x, the run-time detection for DYNAMIC_ARCH and the compile-time choice in cpuid_zarch use different methods for identifying the supported CPU features. 

To make cpuid_zarch.c future-proof and both cpuid_zarch.c and dynamic_zarch.c easier to maintain, switch cpuid_zarch to the same mechanism as DYNAMIC_ZARCH (i.e., derive the supported CPU features from hwcap flags) and share code between both (in a new header cpuid_zarch.h).

Signed-off-by: Marius Hillenbrand <mhillen@linux.ibm.com>

### DIFF
--- a/cpuid_zarch.c
+++ b/cpuid_zarch.c
@@ -27,57 +27,11 @@
 
 #include <string.h>
 
-#define CPU_GENERIC     0
-#define CPU_Z13         1
-#define CPU_Z14         2
-#define CPU_Z15         3
+#include "cpuid_zarch.h"
 
-static char *cpuname[] = {
-  "ZARCH_GENERIC",
-  "Z13",
-  "Z14",
-  "Z15"
-};
-
-static char *cpuname_lower[] = {
-  "zarch_generic",
-  "z13",
-  "z14",
-  "z15"
-};
-
-int detect(void)
-{
-  FILE *infile;
-  char buffer[512], *p;
-
-  p = (char *)NULL;
-  infile = fopen("/proc/sysinfo", "r");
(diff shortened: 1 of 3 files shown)


mhillenbrand:
Hello,
As context, I should have included this patch in my PR #2613 for dynamic arch support on IBM Z last year (commits were merged via #2614) as I see no reason why the detection mechanisms should differ between run-time detection for dynamic arch support and compile-time detection when building for a single target architecture.

In addition to reducing maintenance effort, there is another benefit to checking feature flags in hwcaps over deducing supported features from the model number: The vector extensions require both hardware and OS support -- you may encounter a version of the Linux kernel that does not yet support or that disabled support for the vector extension. The model number can only indicate hardware support whereas hwcaps also tells you that the OS is ready and has enabled the vector extensions.

This patch looks more intrusive than it is: most of the visible code change moves common definitions and code into `cpuid_zarch.h`.

martin-frbg:
No objections from me - I do not have zarch access except through the IBM-sponsored CI services anyway, detection by cpuid was probably done for "historic" reasons alone (and cpuid numbers for new models are somewhat easier to google than hwcaps).

mhillenbrand:
Thanks for merging. You have a point that model numbers are somewhat more accessible than the hwcaps. I have added notes on "where they come from" and code for dumping them on OPENBLAS_VERBOSE in #3427. Please let me know whether that helps.

PR #3425, opened 2021-10-27
Add dgemm kernel for arm64 SVE

binebrank:
This PR is to add a dgemm kernel for arm64 SVE. I do not aim to fully optimize it for any real hardware, it is more of a generic SVE implementation that will serve as a starting point for optimization for future SVE processors. 

This is still a work-in-progress, but I wanted to give an idea and ask you for some feedback. Since SVE is vector-length-agnostic architecture, the kernel will not have a fixed size. One dimension is equal to the size of the SVE register, and the other dimension is kept fixed. (similar approach was used in BLIS library) 

The kernel is actually simplified a lot. For example in the loop over M, the last M % sve_size iterations can be done with a single sweep using appropriate predicate register created with _whilelt_ instructions.

In order for this to work, I also have to write SVE dgemm_ncopy and dgemm_tcopy to scale with SVE size. Do you foresee any other problems that might occur? Also, is this an acceptable approach?

TODO:

- [x]  Write dgemm_ncopy_sve and dgemm_tcopy_sve
- [x]  Write dtrmm_kernel_sve
- [x]  Write dtrmm  and symm SVE copy routines
- [x]  Add arm64_sve target (calls dgemm v1x8)
- [x]  Add size of v2x8
- [x]  Commentary & cleanup
- [x]  CMake


martin-frbg:
Looks good to me, thank you very much. I cannot think of any problems with this approach right now, but I am certainly much less familiar with SVE coding than you. (It occurs to me that we might want to add a CI job running the Arm Instruction Emulator eventually)

martin-frbg:
regarding arm64_sve target,I guess we could use the A64FX for now (which is probably the only living example at this time) ?

binebrank:
A CI job running ArmIE would be great.
Regarding arm64_sve, my idea was that that would be a generic SVE target, since these are not yet optimised for A64FX. But I can also keep only A64FX for now. (Yes, it is the only SVE target currently.)

martin-frbg:
No problem with adding another target for generic ARMV8SVE, but the current A64FX is nothing but generic ARMV8 plus fancy cpuid recognition and announcing SVE compatibility to the (gcc) compiler anyway. Any code that actually _is/uses_ SVE should be a major improvement...

martin-frbg:
Prepared an ArmIE job for Travis but tests that run flawlessly outside the emulator (so gcc did not even add SVE instructions) currently segfault with it

binebrank:
> Prepared an ArmIE job for Travis but tests that run flawlessly outside the emulator (so gcc did not even add SVE instructions) currently segfault with it

This is odd. I'll try to look into that.

martin-frbg:
> > Prepared an ArmIE job for Travis but tests that run flawlessly outside the emulator (so gcc did not even add SVE instructions) currently segfault with it
> 
> This is odd. I'll try to look into that.

Probably Travis-specific - same sequence of commands runs fine locally.

binebrank:
Yes, these commands run for me also. 

Here just some updates regarding this PR:
The dgemm kernel runs fine and all dgemm tests pass, but I am now realising the effect of making the m_unroll_factor vector-length-agnostic. 
It means that I also have to write SVE specific functions for dtrmm_{ut|un|lt|ln}copy (and similarly trsm_copy and symm_copy) which is a lot more work than first anticipated. This will take some time.

martin-frbg:
This is unfortunate, I did not realize this consequence earlier as it is so different from all other architectures that "just work" with the "generic" C implementations of these copy functions. (But there is probably no merit in trying to hack the required functionality into those).

binebrank:
I finished writing the dtrmm and dsymm sve copy routines. Now all tests pass for 256 and 512-bit SVE. In the end, I had to modify kernel/Makefile.L3 to separate SVE and generic copy routines (hope this is not too ugly).

Also, I realised that this implementation is not yet completely vector-length-agnostic. For that, one also has to write SVE dtrsm_copy and change some functions to completely remove DGEMM_UNROLL_M. But this is work for another PR. I'll also write single-precision SVE kernels (it's not much work to transform dgemm kernels to sgemm).

martin-frbg:
Thank you very much for all the work you have put into this. Unfortunately I had lost network connection to my office with the ARM64 system where I had setup ARMIE already, and the Raspberry Pi I have dug up here at home is still busy compiling gcc10.
I was about to upload my suggestion for Makefile.L3 (which I think is a bit more general as it checks for DTRMMLNCOPY etc being defined in the `KERNEL.<cpu>` file, but maybe i overengineered it... also this is not compile-tested yet. I plan to tackle the CMakeLists.txt for CMAKE later this evening.
[Makefile.L3.txt](https://github.com/xianyi/OpenBLAS/files/7576866/Makefile.L3.txt)

binebrank:
> Thank you very much for all the work you have put into this. Unfortunately I had lost network connection to my office with the ARM64 system where I had setup ARMIE already, and the Raspberry Pi I have dug up here at home is still busy compiling gcc10. I was about to upload my suggestion for Makefile.L3 (which I think is a bit more general as it checks for DTRMMLNCOPY etc being defined in the `KERNEL.<cpu>` file, but maybe i overengineered it... also this is not compile-tested yet. I plan to tackle the CMakeLists.txt for CMAKE later this evening. [Makefile.L3.txt](https://github.com/xianyi/OpenBLAS/files/7576866/Makefile.L3.txt)

Thanks, this looks much nicer. I actually had to further split DTRMMLNCOPY and others into DTRMMLNCOPY_M (and DTRMMLNCOPY_N). I only use these special copy in "M direction" which is SVE specific, whereas the direction in N still uses generic ones.
And yea, I totally forgot about the CMAKE part...

martin-frbg:
Unfortunately I am getting a TRSM failure in test/DBLAT3.SUMM

martin-frbg:
kernel/CMakeLists.txt (already anticipating TRSMxxCOPY_M functions but I have not quite figured out SVE intrinsics, so failed at deriving them from your TRMM ones)
[CMakeLists.txt](https://github.com/xianyi/OpenBLAS/files/7590277/CMakeLists.txt)

binebrank:
> Unfortunately I am getting a TRSM failure in test/DBLAT3.SUMM

I assume you are running tests with armie 128-bit SVE? I see all tests passing for 256 and 512bit. For 128-bit to work I had to reduce DGEMM_DEFAULT_UNROLL_M to 2. This is what I meant with this implementation not yet being completely vector length agnostic. This macro is still used in generic kernels and I suspect there is some mixing of generic and sve copy routines somewhere, but can't figure it out exactly where. As said, I will also write these dtrsm_copy in sve. Now with this macro set to 2. all tests pass for 128, 256, 512, but please double-check this.

binebrank:
> kernel/CMakeLists.txt (already anticipating TRSMxxCOPY_M functions but I have not quite figured out SVE intrinsics, so failed at deriving them from your TRMM ones) [CMakeLists.txt](https://github.com/xianyi/OpenBLAS/files/7590277/CMakeLists.txt)

Thanks, I'll make this work.

martin-frbg:
Thank you. (I may have gotten a trsm_uncopy_sve working right now, but I bet it will be much less elegant and efficient than what you come up with). The generic trsm kernels call DGEMM internally, perhaps this is where a mixing of copy routines can occur (I did indeed try with 128 and even wondered why I did not see the failure previously when I probably used 256). 
cmake build still needs a bit of work as it is not adding the `-march=armv8.2-a+sve` automatically yet.

binebrank:
I have left the trsm_copy part unchanged as it is not necessary yet. I put the SVE flags `-march=armv8.2-a+sve` into cmake/cc.cmake (as the TODO comment mentions, one should probably transform all Makefile.${ARCH} into cmake/${ARCH}.cmake and add flags there at one point).

martin-frbg:
Great, thank you. I was just about to upload my version of cc.cmake, but it is almost identical (mine matched on HAVE_SVE and used the CORE name only for adding the `-mtune`). I think this is now ready for merging. Now if you would like to add yourself to the CONTRIBUTORS.md file, as this is a fairly significant improvement for OpenBLAS ?

binebrank:
Thanks, I'll try to do the SGEMM and dtrsm also once I have a bit more time.

wxmwy:
> > Unfortunately I am getting a TRSM failure in test/DBLAT3.SUMM
> 
> I assume you are running tests with armie 128-bit SVE? I see all tests passing for 256 and 512bit. For 128-bit to work I had to reduce DGEMM_DEFAULT_UNROLL_M to 2. This is what I meant with this implementation not yet being completely vector length agnostic. This macro is still used in generic kernels and I suspect there is some mixing of generic and sve copy routines somewhere, but can't figure it out exactly where. As said, I will also write these dtrsm_copy in sve. Now with this macro set to 2. all tests pass for 128, 256, 512, but please double-check this.

I used this code to run DGEMM tests on an ARMV8 CPU, and SVE got about a 20% to 30% performance improvement. But I still have a question, if I want to test the performance of fixed SVE register length, such as '-msve-vector-bits=256', will kernel still be valid? Thanks.

binebrank:
> > > Unfortunately I am getting a TRSM failure in test/DBLAT3.SUMM
> > 
> > 
> > I assume you are running tests with armie 128-bit SVE? I see all tests passing for 256 and 512bit. For 128-bit to work I had to reduce DGEMM_DEFAULT_UNROLL_M to 2. This is what I meant with this implementation not yet being completely vector length agnostic. This macro is still used in generic kernels and I suspect there is some mixing of generic and sve copy routines somewhere, but can't figure it out exactly where. As said, I will also write these dtrsm_copy in sve. Now with this macro set to 2. all tests pass for 128, 256, 512, but please double-check this.
> 
> I used this code to run DGEMM tests on an ARMV8 CPU, and SVE got about a 20% to 30% performance improvement. But I still have a question, if I want to test the performance of fixed SVE register length, such as '-msve-vector-bits=256', will kernel still be valid? Thanks.

The code is vector-length-agnostic and should run for all sizes. I tested it with Armie for 128, 256 and 512bit SVE and all was fine. May I ask on which cpu you are running this? I assume Graviton 3, if the vector length is 256?

wxmwy:
Raspberry Pi 4B (Cortex A-72). But core dump happened when I compiled directly on the Raspberry PI. The version I'm using is cross-compiled. Perhaps some variables are not well set, so the performance improvement is just a reference.

wxmwy:
> Raspberry Pi 4B (Cortex A-72). But core dump happened when I compiled directly on the Raspberry PI. The version I'm using is cross-compiled. Perhaps some variables are not well set, so the performance improvement is just a reference.



> > > > Unfortunately I am getting a TRSM failure in test/DBLAT3.SUMM
> > > 
> > > 
> > > I assume you are running tests with armie 128-bit SVE? I see all tests passing for 256 and 512bit. For 128-bit to work I had to reduce DGEMM_DEFAULT_UNROLL_M to 2. This is what I meant with this implementation not yet being completely vector length agnostic. This macro is still used in generic kernels and I suspect there is some mixing of generic and sve copy routines somewhere, but can't figure it out exactly where. As said, I will also write these dtrsm_copy in sve. Now with this macro set to 2. all tests pass for 128, 256, 512, but please double-check this.
> > 
> > 
> > I used this code to run DGEMM tests on an ARMV8 CPU, and SVE got about a 20% to 30% performance improvement. But I still have a question, if I want to test the performance of fixed SVE register length, such as '-msve-vector-bits=256', will kernel still be valid? Thanks.
> 
> The code is vector-length-agnostic and should run for all sizes. I tested it with Armie for 128, 256 and 512bit SVE and all was fine. May I ask on which cpu you are running this? I assume Graviton 3, if the vector length is 256?

Raspberry Pi 4B (Cortex A-72). But core dump happened when I compiled directly on the Raspberry PI. The version I'm using is cross-compiled. Perhaps some variables are not well set, so the performance improvement is just a reference.

binebrank:
Hey, I think you mixed a few things. Cortex A-72 does not support SVE instructions so the code in this PR is not the right place. You should compile OpenBLAS for A72 which will only include NEON instructions.

wxmwy:
> Hey, I think you mixed a few things. Cortex A-72 does not support SVE instructions so the code in this PR is not the right place. You should compile OpenBLAS for A72 which will only include NEON instructions.

Thanks. It was a terrible mistake on my part.

PR #3064, opened 2021-01-12
Add cblas_crotg, cblas_zrotg, cblas_csrot and cblas_zdrot

martin-frbg:
the `cmake` build already generated the interface functions, but neither the `gmake` build nor the cblas.h had them. Fixes #3063

PR #3434, opened 2021-11-01
Add cblas_{c/z}srot cblas_{c/z}rotg support

XiWeiGu:
Fixed the following issue：
```
/usr/bin/ld: /tmp/ccOl9R8p.o: in function `.LM6':
test.c:(.text+0x90): undefined reference to `cblas_crotg'
```

### DIFF
--- a/interface/zrot.c
+++ b/interface/zrot.c
@@ -42,13 +42,19 @@
 #include "functable.h"
 #endif
 
+#ifndef CBLAS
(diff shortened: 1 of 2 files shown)


martin-frbg:
Ouch, looks like my earlier #3064 lacked the updated files for some reason. Thanks for fixing this. (As these cblas functions are not in Reference-LAPACK (and nobody there responded to my query if they should be), few will have missed them I guess)

PR #3450, opened 2021-11-15
cmake: Set SUFFIX64 also for NOFORTRAN

mmuetzel:
When building OpenBLAS with `cmake -DINTERFACE64=1` using `gcc`, a suffix is added to the library and other files. That is a good thing imho.
However when doing the same using `clang`, that suffix isn't used.

Afaict, that is because `NOFORTRAN` is set in the latter case and setting these suffices is done depending on `NOFORTRAN` not being set.

IIUC, it would be ok to set these suffices independent of `NOFORTRAN`.

### DIFF
--- a/cmake/fc.cmake
+++ b/cmake/fc.cmake
@@ -3,11 +3,6 @@
 ## Description: Ported from portion of OpenBLAS/Makefile.system
 ##              Sets Fortran related variables.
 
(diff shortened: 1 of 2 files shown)


martin-frbg:
Right, thanks. This must have been an oversight when the automatic suffixing was added three years ago.

PR #3451, opened 2021-11-17
MOD: optimize DGEMM of large matrices on cortex A53 & A55

wjc404:


wjc404:
image
Test programs: 
[OpenBLAS-test.tar.gz](https://github.com/xianyi/OpenBLAS/files/7557166/OpenBLAS-test.tar.gz)
(src/test_kernel.cpp for the correctness of kernel function, src/test_dgemm.cpp for the result and performance of DGEMM)

wjc404:
I need some help for figuring out the syntax error in inline asm:) The asm block can be complied with NDK-r23b and gcc-7, but fails in the CI job.

martin-frbg:
Sorry I do not see why it does not like the local labels here. As I read the error messages, it seems to think that jumping forward to 3: in the `"blt 3f; beq 2f; .align 4; 1:\n\t"` of line 84 would cross into a different loop section, and anything behind the "blt3f" on that line seems to get ignored so label 1: does not get defined at all(?) The only thing that stands out to me is the ".align 4" (and I remember earlier Mac assemblers being peculiar in their handlng of .align, making it necessary to use .p2align instead).

martin-frbg:
Looks like the Apple Clang assembler simply wants to see the labels on a separate line (cf my test in #3452) - go figure...

wjc404:
Thank you very much @martin-frbg :)

PR #3455, opened 2021-11-19
Fix unsafe read during final iteration of zsymv_L_sse2.S

cenewcombe:
The unsafe read and fix are demonstrated with test_zhemv.c

### DIFF
--- a/kernel/x86_64/zsymv_L_sse2.S
+++ b/kernel/x86_64/zsymv_L_sse2.S
@@ -451,7 +451,6 @@
 #endif
 
 	MOVDDUP(4 * SIZE, A1, a1)
-	MOVDDUP(6 * SIZE, A2, a2)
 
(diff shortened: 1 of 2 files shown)


martin-frbg:
Thank you. This bug is probably at least 15 years old (original GotoBLAS), I wonder how you came across it - valgrind or an actual program where the excessive read did cause problems ? 
Not sure the testcase needs to be in the codebase though...

martin-frbg:
extracting the test case to here so that it does not get lost
[test_zhemv.c.txt](https://github.com/xianyi/OpenBLAS/files/7609300/test_zhemv.c.txt)

cenewcombe:
Thanks! This bug causes one of the test programs in Magma (http://icl.utk.edu/magma/) to fail. I only recently became aware of it; however, this may have been an issue with this particular test case for some time.

Issue #3458, opened 2021-11-24
Issues when building OpenBLAS with cmake


rafaelcfsousa:
Hi, I tried to build OpenBLAS using the **CMakefile** and I realized that the build does not work as expected for most of the architectures (e.g., ARM64, RISCV, POWER, MIPS, etc). In addition, for some cases the build crashes, e.g., -DTARGET=POWER8.

After a careful analysis, I figured out that the issue described above comes from the fact that the parser used by **cmake** to read the **kernel/arch/KERNEL\*** files is not working properly.

Below I list some of the issues that may be found:
1) The parser does not treat statements such as ifdef and ifndef.
2) The code below crashes if CMAKE_MATCH_1 is not defined.
```
 if (DEFINED ${${CMAKE_MATCH_1}} AND ${${CMAKE_MATCH_1}} STREQUAL ${CMAKE_MATCH_2})
```
3) The way that the code above checks if the variable is defined should be:
```
if (DEFINED ${CMAKE_MATCH_1} ...)
```
4) The code below does not check if CMAKE_MATCH_1 is defined so that, if it is not defined, the build crashes.
```
if (NOT ( ${${CMAKE_MATCH_1}} STREQUAL ${CMAKE_MATCH_2}))
```

Besides that, there are some other missing features, as shown below:

1. Comments are not treated
2.  Fails to parse more complex structures, like the following:
```
ifeq(...)
    ifneq(...)
        SGEMM_KERNEL=sgemm_kernel_option1.c
        DGEMM_KERNEL=dgemm_kernel_option1.c
    else ()
        SGEMM_KERNEL=sgemm_kernel_option2.c
        DGEMM_KERNEL=dgemm_kernel_option2.c
    endif()
else ()
    SGEMM_KERNEL=sgemm_kernel_option3.c
    DGEMM_KERNEL=dgemm_kernel_option3.c
endif()
```
3. Even though the statement ifdef is implemented, some patterns are not considered, like the one below:
```
ifeq ($(SGEMM_UNROLL_M)x$(SGEMM_UNROLL_N), 8x8)
```

I have already worked on a solution to solve all points mentioned above (#https://github.com/rafaelcfsousa/OpenBLAS/commit/fb891f33da3428602d765905b1395a9a7e798e4a). I would like to know if it is ok to PR it?

**Side-effect:**
[...]

PR #3459, opened 2021-11-25
Fix issues when building OpenBLAS with cmake

rafaelcfsousa:
This patch solves the issues described here: #3458


martin-frbg:
Thank you very much, looks quite a bit cleaner now :slightly_smiling_face:

PR #3465, opened 2021-12-01
Fix truncated assembler checks used to build Power10 Kernels

kavanabhat:


Makefile.system
```diff
@@ -15,7 +15,7 @@ ifeq ($(HOSTARCH), amd64)
 HOSTARCH=x86_64
 endif
 
-HAVE_GAS := $(shell as -v < /dev/null 2>&1 | grep GNU 2>&1 >/dev/null)
+HAVE_GAS := $(shell as -v < /dev/null 2>&1 | grep GNU 2>&1 >/dev/null ; echo $$?)
```
RajalakshmiSR: Can we move this after line 347 and use it as $(AS) instead of 'as' ?

PR #3466, opened 2021-12-01
[POWER] Add small matrix for sgemm/dgemm on Power10

rafaelcfsousa:
This patch adds support for small matrix kernels on Power10, as follow:
- Add sgemm small matrix kernels (NN, NT, TN and TT)
- Add dgemm small matrix kernels (NN, NT, TN and TT)

See below some benchmarking details:

**1) SGEMM:** 

Combining the dimensions below:
- 0 < M <= 100;
- 0 < N <= 100;
- K ∈ {1, 2, 8, 16, 32, 48, 64, 96, 100}

where M\*N\*K <= 100\*100\*100, we achieve an average speedup of up to 2.09x

**2) DGEMM:**

Combining the dimensions below:
- 0 < M <= 100;
- 0 < N <= 100;
- K ∈ {1, 2, 8, 16, 32, 48, 64, 96, 100}

where M\*N\*K <= 54\*54\*54, we achieve an average speedup of up to 1.65x


martin-frbg:
Thank you very much

PR #3471, opened 2021-12-07
[Testing] add thread safety test to Travis

martin-frbg:


martin-frbg:
turns out to be pointless as the neoverse lease is limited to 4 cpus

PR #3472, opened 2021-12-08
Fallback for Power kernels

kavanabhat:
1) Changes to fallback to P6 mode when building 32-bit binary on P9/P10 platform
2) Changes to use P8 kernels when GNU assembler is not available on P10.


kavanabhat:
@martin-frbg , two of the checks have failed. I am still looking into the failures. Please delay the merge till these are resolved.

martin-frbg:
? The failures I saw were clearly unrelated (OSX homebrew apparently updated their build of the Android NDK, breaking the OSX ARMV7 crossbuild)

kavanabhat:
I wasn't aware that these were existing failures. Then, the fix looks fine and can be merged. Thanks

PR #3478, opened 2021-12-14
Makefile: also consider -O, -Og and -Os when stripping flags

ffontaine:
gcc also supports `-O`, `-Og` and `-Os` as optimization flags.
They may be given on the make command-line by users.

For the calculation of `LAPACK_NOOPT`, all such flags should be considered.

Signed-off-by: Thomas De Schampheleire <thomas.de_schampheleire@nokia.com>
[Retrieved from: https://git.buildroot.net/buildroot/tree/package/openblas/0003-Makefile-also-consider-Os-when-determining-LAPACK_NO.patch]
Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ export NOFORTRAN
 export NO_LAPACK
 endif
 
-LAPACK_NOOPT := $(filter-out -O0 -O1 -O2 -O3 -Ofast,$(LAPACK_FFLAGS))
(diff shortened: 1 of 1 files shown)


martin-frbg:
Thanks. Looking into the significance of this variable for Reference-LAPACK, it turns out it is used for only a handful of sources that were probably miscompiled at some time in the past - but only in the (older) gmake Makefiles, not in the CMAKE environment that seems to be their preferred build method nowadays. Probably worth discussing there...

PR #3480, opened 2021-12-15
fix bug in zscal function

wzgpeter:
memset can not be used in zscal because of
the stride parameters.

Signed-off-by: Wu Zhigang <zhigang.wu@starfivetech.com>


martin-frbg:
Oh, right. Thank you very much.

PR #3485, opened 2021-12-16
Add feature-based fallback for unknown x86_64 cpus

### LINKED ISSUE #3453: ddot_k_PRESCOTT segfaults when called on non-memory aligned data
As discussed in https://github.com/scipy/scipy/issues/14886 and https://github.com/scikit-learn/scikit-learn/issues/21361.

Given that prescott is an old architecture, I don't think it impacts many users. We only discovered it because older version of OpenBLAS would wrongly detect SkylakeX CPU as Prescott, but this is no longer the case in 0.3.17 and 0.3.18 (and maybe 0.3.16 as well IIRC).

I am not sure if OpenBLAS officially supports being called with non-memory aligned data. This fact that we face this problem in the scikit-learn test suite is a consequence of a low level issue in joblib https://github.com/joblib/joblib/issues/563 that we intend to fix anyways (but I don't know yet how...).

martin-frbg:
should offer better performance than dropping down to Prescott, and also avoid potential issues with old Goto kernels that do not support unaligned data. Fixes #3453 as suggested by ogrisel there

### DIFF
--- a/driver/others/dynamic.c
+++ b/driver/others/dynamic.c
@@ -1061,7 +1061,13 @@ void gotoblas_dynamic_init(void) {
 #ifdef ARCH_X86
   if (gotoblas == NULL) gotoblas = &gotoblas_KATMAI;
 #else
-  if (gotoblas == NULL) gotoblas = &gotoblas_PRESCOTT;
+  if (gotoblas == NULL) {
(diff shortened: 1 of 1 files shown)


ogrisel:
Thank you very much!

Issue #3337, opened 2021-08-06
test failed for ctbsv on blas tester


qqqil:
bin/xcl2blastst -R tbsv -U 1 U -A 1 N -D 2 U N -n 4 -X 2 2 1  -q 3

bin/xcl2blastst -R tbsv -U 1 U -A 1 N -D 2 U N -n 4 -X 2 2 1 -q 3

----------------------------- TBSV ------------------------------
TST# UPLO TRAN DIAG    N    K  LDA INCX   TIME  MFLOP  SpUp  TEST
==== ==== ==== ==== ==== ==== ==== ==== ====== ====== ===== =====
ERROR:   resid=98473.945312, normD=0.469560, normA=0.393167, normX=0.539283, eps=1.192093e-06
   resid=98473.945312
   0    U    N    U    4    3    4    2   0.00    5.6  1.00 -----
   0    U    N    U    4    3    4    2   0.00    3.1  0.56 FAIL
   1    U    N    U    4    3    4    1   0.00   75.5  1.00 -----
   1    U    N    U    4    3    4    1   0.00   75.5  1.00 PASS
ERROR:   resid=28358.013672, normD=0.188386, normA=1.393167, normX=0.539283, eps=1.192093e-06
   resid=28358.013672
   2    U    N    N    4    3    4    2   0.00    0.0  1.00 -----
   2    U    N    N    4    3    4    2   0.00   75.5  0.00 FAIL
   3    U    N    N    4    3    4    1   0.00    0.0  1.00 -----
   3    U    N    N    4    3    4    1   0.00   75.5  0.00 PASS

4 tests run, 2 passed

brada4:
Hi,
What kind of CPU you are using (/proc/cpuinfo)?
Which version of OpenBLAS? (like last block from compilation output)
Parameters given to CMAKE or MAKE?
Compilers? (if from system - just roughly version of that system)

Please copy over extra information you gave in nearly identical report at https://github.com/xianyi/BLAS-Tester/issues/5

brada4:
Sandybridge
```
bin/xcl2blastst -R tbsv -U 1 U -A 1 N -D 2 U N -n 4 -X 2 2 1 -q 3


bin/xcl2blastst -R tbsv -U 1 U -A 1 N -D 2 U N -n 4 -X 2 2 1 -q 3 


----------------------------- TBSV ------------------------------
[...]

PR #3492, opened 2021-12-21
SVE zgemm&cgemm (and other BLAS 3 complex)

binebrank:
Create SVE kernels for zgemm&ztrmm,  cgemm&ctrmm.

- [x]  Zgemm kernel
- [x]  Ztrmm kernel
- [x]  Zgemm copy
- [x]  Ztrmm copy
- [x]  Cgemm kernel
- [x]  Ctrmm kernel
- [x]  Cgemm copy
- [x]  Ctrmm copy
- [x]  C|Zsymm copy
- [x]  C|Zhemm copy
- [x]  Configure Kernel files
- [x]  Adapt Makefile
- [x]  Adapt CMakeLists
- [x]  Add SVE trsm kernels


binebrank:
I believe this is ready for merge :)

kernel/Makefile.L3
```diff
@@ -1691,29 +1691,61 @@ $(KDIR)qtrmm_oltucopy$(TSUFFIX).$(SUFFIX) : generic/trmm_ltcopy_$(QGEMM_UNROLL_N
...
-$(KDIR)ctrmm_ilnucopy$(TSUFFIX).$(SUFFIX) : generic/ztrmm_lncopy_$(CGEMM_UNROLL_M).c
+ifdef CTRMMLNCOPY_M
+$(KDIR)ctrmm_ilnucopy$(TSUFFIX).$(SUFFIX) : $(KERNELDIR)/$(CTRMMLNCOPY_M)
 	$(CC) -c $(CFLAGS) $(NO_UNINITIALIZED_WARN) -UDOUBLE -DCOMPLEX -UOUTER -DLOWER -DUNIT $< -o $@
 
-$(KDIR)ctrmm_ilnncopy$(TSUFFIX).$(SUFFIX) : generic/ztrmm_lncopy_$(CGEMM_UNROLL_M).c
+$(KDIR)ctrmm_ilnncopy$(TSUFFIX).$(SUFFIX) : $(KERNELDIR)/$(CTRMMLNCOPY_M)
 	$(CC) -c $(CFLAGS) $(NO_UNINITIALIZED_WARN) -UDOUBLE -DCOMPLEX -UOUTER -DLOWER -UUNIT $< -o $@
+else
+$(KDIR)ctrmm_iunucopy$(TSUFFIX).$(SUFFIX) : generic/ztrmm_uncopy_$(CGEMM_UNROLL_M).c
```
martin-frbg: copypasta needs to read "ilnucopy" here (and "ilnncopy","iutucopy" "iutncopy" etc. on subsequent lines)

binebrank: ah of course don't know how this slipped through

martin-frbg:
Unfortunately I am getting a bunch of errors for COMPLEX in lapack-test, apparently related to CPOTRF failing with the new kernels.

binebrank:
I think this is because of the mixing of generic TRSM and sve DGEMM kernels. (both are used simultaneously in potrf and copy functions are different for the two).  I'll write trsm kernels for SVE. Hopefully, this will solve the problem.

martin-frbg:
Still not sure what is causing it - notably ZPOTRF appears to be fine, it is only the single-precision complex lapack-test that shows failures. (Sorry for not making that clearer, also my tests yesterday were run with a vector length argument of 128 only). BLAS-Tester i.e. the ATLAS testsuite did not show any errors for the combinations of options I tried so far.

binebrank:
For 256 and 512bit, ZPOTRF are also failing in lapack-test. Another thing I have to look into is the macro GEMM_UNROLL_MN which is also used in potrf. This macro is set as MAX(GEMM_UNROLL_M, GEMM_UNROLL_N), but this doesn't work for the SVE approach because GEMM_UNROLL_M depends on SVE size. For now, I set this explicitly to some higher value with CGEMM_DEFAULT_UNROLL_MN. But again, I think having SVE TRSM will fix all this.

binebrank:
I now added SVE for TRSM and this solves the problem with lapack.
There are still 2 tests failing, namely:
Testing REAL              RFP-linear-equation-routines-LIN/xlintstrfs < stest_rfp.in > stest_rfp.out   STFSM auxiliary routine:     1 out of  7776 tests failed to pass the threshold
Testing COMPLEX           RFP-linear-equation-routines-LIN/xlintstrfc < ctest_rfp.in > ctest_rfp.out   CTFSM auxiliary routine:     1 out of  7776 tests failed to pass the threshold

But I also obtain these errors when targeting TSV110, so they are not SVE related. You maybe know anything about these?

martin-frbg:
Thank you very much. The two lapack-test failures in STFSM/CTFSM are present on most architectures, so far I have preferred to explain them away as a numerical instability in the test code itself. (As a case in point, my local testing of this PR with ARMIE on an Odroid board shows only the STFSM failure)
BTW is there a particular reason why you tried with TSV110 target, i.e. do you have access to that hardware ? (Its kernels are generic enough to allow testing on other ARMV8, but we have an open issue ticket claiming failures on an actual TSV110)

martin-frbg:
For STFSM, the "failed" result is 30.604 on the Odroid with ARMIE, where stest_rfp.in defines the threshold as 30.0

binebrank:
> Thank you very much. The two lapack-test failures in STFSM/CTFSM are present on most architectures, so far I have preferred to explain them away as a numerical instability in the test code itself. (As a case in point, my local testing of this PR with ARMIE on an Odroid board shows only the STFSM failure) BTW is there a particular reason why you tried with TSV110 target, i.e. do you have access to that hardware ? (Its kernels are generic enough to allow testing on other ARMV8, but we have an open issue ticket claiming failures on an actual TSV110)

Ok good to know. Regarding TSV110 there is no particular reason. I just have access to it and usually do my Arm stuff on that machine.

martin-frbg:
>  Regarding TSV110 there is no particular reason. I just have access to it and usually do my Arm stuff on that machine.

ah good. the reason I'm asking is #3337 - I could not reproduce that problem in tsv110 builds on other arm server hardware (#3340)

wxmwy:
For my graduation project, I am studying SVE optimization of DGEMM. Based on the current code, I think the future optimization directions are as follows: 

1. Use assembly to rewrite sve_ncopy and sve_tcopy 
2. For assembly code like KERNELv2x8_I, we can use LD1RQD instead of LD1RD, so the load of pB can be reduced by half. 

change:


    ld1rd  z8.d, p0/z,  [pB]
    ld1rd  z9.d, p0/z,  [pB, 8]
    fmla z16.d, p0/m, z0.d, z8.d
    fmla z17.d, p0/m, z1.d, z8.d
    fmla z18.d, p0/m, z0.d, z9.d
    fmla z19.d, p0/m, z1.d, z9.d

to:


    ld1rqd	z8.d, p0/z, [pB]
    fmla z16.d, z0.d, z8.d[0]
    fmla z17.d, z1.d, z8.d[0]
    fmla z18.d, z0.d, z8.d[1]
    fmla z19.d, z1.d, z8.d[1]

Any suggestions? Thank you. （From a student who is having trouble graduating）

binebrank:
Hi there, nice to see someone working on this :). Below are some of my comments:

1. ncopy and tcopy. This is a relatively simple one. Rewrite those with assembly and be sure to use proper unrolling (I would suggest 2- or 4-way).
2. LD1RQD. I was thinking about this when implementing the kernels, but in the end, I decided on the simpler approach. Feel free to test this to see what kind of effect it has.
3. Another possible point of optimization is prefetching. Admittedly, I was a bit sloppy with prefetch instructions when writing kernels. (I knew what I was doing but there are probably some mistakes in implementation) Feel free to experiment with different prefetch distances and placements of these operations. This could potentially have a big effect on performance, probably more than 1. and 2.
4. Optimize the kernels for A64FX. This is a big one but requires a lot of work. The main problem is that the current kernel (2vx8) is simply too small. A64FX has big latency for fmla (9 cycles but increases to 10 if both SVE pipelines are fully occupied) and the minimum kernel that would not cause any stalls in SVE pipelines would have to be at least 2vx10. This would then also require special routines for copying data and everything else that depends on it (there are no generic copy functions for width 10). Additionally getting top performance on A64FX also requires using sector cache and some other advanced things...

I will probably get back to some of these once the Graviton3 processor is available.

wxmwy:
> Hi there, nice to see someone working on this :). Below are some of my comments:
> 
> 1. ncopy and tcopy. This is a relatively simple one. Rewrite those with assembly and be sure to use proper unrolling (I would suggest 2- or 4-way).
> 2. LD1RQD. I was thinking about this when implementing the kernels, but in the end, I decided on the simpler approach. Feel free to test this to see what kind of effect it has.
> 3. Another possible point of optimization is prefetching. Admittedly, I was a bit sloppy with prefetch instructions when writing kernels. (I knew what I was doing but there are probably some mistakes in implementation) Feel free to experiment with different prefetch distances and placements of these operations. This could potentially have a big effect on performance, probably more than 1. and 2.
> 4. Optimize the kernels for A64FX. This is a big one but requires a lot of work. The main problem is that the current kernel (2vx8) is simply too small. A64FX has big latency for fmla (9 cycles but increases to 10 if both SVE pipelines are fully occupied) and the minimum kernel that would not cause any stalls in SVE pipelines would have to be at least 2vx10. This would then also require special routines for copying data and everything else that depends on it (there are no generic copy functions for width 10). Additionally getting top performance on A64FX also requires using sector cache and some other advanced things...
> 
> I will probably get back to some of these once the Graviton3 processor is available.

Thank you for reply.
I did a simple test that I hope will help you if you come back.
https://github.com/xianyi/OpenBLAS/issues/3581

PR #3502, opened 2021-12-29
Fix cmake crosscompilation for core2 target

jgillis:
Missing HAVE_SSE* cmake variables cause cc.cmake to forget about `-msse*` flags


martin-frbg:
Thank you.

PR #3508, opened 2022-01-07
OpenBLAS: aarch64: Add neoverse-v1/n2 architecture specifics

snadampal:


martin-frbg:
Wouldn't it make more sense to use the new ARMV8SVE target for the V1 ? Also rather than copying an entire KERNEL file, a single-line "#include KERNEL.whatever" would probably make it easier to track cpu-specific changes.

PR #3516, opened 2022-01-21
cmake: Check if Fortran compiler is usable before enabling it.

mmuetzel:
When configuring with `cmake -DBUILD_WITHOUT_LAPACK=ON`, the Fortran compiler is forced to be `gfortran`. That causes the configuration (for the tests) to fail if no Fortran compiler is installed.
I don't understand why the compiler is forced to `gfortran` in that case. Afaict, that shouldn't be necessary.

The proposed change lets the build system check if a Fortran compiler is available before using it.

### DIFF
--- a/cmake/f_check.cmake
+++ b/cmake/f_check.cmake
@@ -20,19 +20,16 @@
 # NEEDBUNDERSCORE
 # NEED2UNDERSCORES
 
-if (NOT NO_LAPACK)
-  include(CheckLanguage)
-  check_language(Fortran)
-  if(CMAKE_Fortran_COMPILER)
-    enable_language(Fortran)
-  else()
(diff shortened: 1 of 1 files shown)


martin-frbg:
Yeah, this is some oddity left over from GotoBLAS (which then got faithfully translated to cmake syntax - only I suspect in the gmake case setting NOFORTRAN=1 is a stronger deterrent for the tests). Probably some higher-order cleanup should be done (could be the dummy compiler entry is just created to avoid syntax errors from half-empty statements in the original f_check script).

mmuetzel:
Just to clarify: Is the higher-order cleanup you are hinting at a prerequisite for this PR? The proposed change works for me. But I might be missing something.

martin-frbg:
No, just a note to myself...

PR #3527, opened 2022-02-06
Treat AVX512-enabled Alder Lake like Cooper Lake/Sapphire Rapids

### LINKED ISSUE #3490: Propose optional CPUARCH target AlderLakeAVX512, as aliase of SapphireRapids
## Context 

It's been reasonably well documented that on alder lake, if, and only if, the GraceMount Ecore's are disabled, it becomes possible to enable all of the available AVX512 instruction available on the Golden Cove Pcore's the same core's used in SapphireRapids. 

For relevant workloads, many of which OpenBLAS has AVX512 accelerated code paths, this can lead to a significant performance uplift. 

As of release 0.3.19, even if ecores are disabled and AVX512 is available, the build system will not make use of them automatically. 
 
### Current workaround

 The user has the option of either passing: 
 
```
CFLAGS='-O3 -march=sapphirerapids -mno-amx-tile -mno-amx-int8 -mno-amx-bf16'
CXXFLAGS='-O3 -march=sapphirerapids -mno-amx-tile -mno-amx-int8 -mno-amx-bf16'
FFLAGS='-O3 -march=sapphirerapids -mno-amx-tile -mno-amx-int8 -mno-amx-bf16'
```
or alternatively passing
```
CFLAGS='-O3 -march=native -mavx512f -mavx512dq -mavx512ifma -mavx512cd -mavx512bw -mavx512vl -mavx512bf16 -mavx512vbmi -mavx512vbmi2 -mavx512vnni -mavx512bitalg -mavx512vpopcntdq -mavx512vp2intersect'
CXXFLAGS='-O3 -march=native -mavx512f -mavx512dq -mavx512ifma -mavx512cd -mavx512bw -mavx512vl -mavx512bf16 -mavx512vbmi -mavx512vbmi2 -mavx512vnni -mavx512bitalg -mavx512vpopcntdq -mavx512vp2intersect'
FFLAGS='-O3 -march=native -mavx512f -mavx512dq -mavx512ifma -mavx512cd -mavx512bw -mavx512vl -mavx512bf16 -mavx512vbmi -mavx512vbmi2 -mavx512vnni -mavx512bitalg -mavx512vpopcntdq -mavx512vp2intersect'
```

**NOTE:** It is preferred to use the -march=sapphirerapids option as GCC, Clang, LLVM based ICX and ICC will then preserve the relevant AVX512 cost functions for auto-vectorization vs other instruction pathways, instead of generic costs meant as a catch all across all supported architectures. 
**NOTE:** We must disable the AMX tile and AMX instructions as the additional hardware was not built into Alder Lake. 

## Proposed solution/Request

I'd like to propose that, if the user supplies an architecture flag AlderLakeAVX512, it be aliased to be enable all features of Sapphire Rapids not explicitly requiring the AMX Tile. 

### Sources on AVX512 support on alder lake: 

[Anandtech article](https://www.anandtech.com/show/17047/the-intel-12th-gen-core-i912900k-review-hybrid-performance-brings-hybrid-complexity/2)

[Phoronix Article](https://www.phoronix.com/scan.php?page=article&item=alder-lake-avx512&num=1)
[
OpenBenchmarking Results of disabling 8 ecores to enable AVX512 on 8 pcores](https://openbenchmarking.org/result/2111077-TJ-ALDERAVX567)

[Testing of AVX512 per instruction cost and pipeline on Alderlake ](https://twitter.com/InstLatX64/status/1456365451513040896?s=20)

[discussion of AVX512 performance support and performance](https://twitter.com/FelixCLeClair/status/1467276502156910600?s=20)

martin-frbg:
fixes #3490 for Alder Lake systems with E cores disabled and AVX512 enabled in (suitable older) BIOS

### DIFF
--- a/cpuid_x86.c
+++ b/cpuid_x86.c
@@ -1495,6 +1495,10 @@ int get_cpuname(void){
         switch (model) {
         case 7: // Alder Lake desktop
         case 10: // Alder Lake mobile
+	  if(support_avx512_bf16())
+            return CPUTYPE_COOPERLAKE;	
+          if(support_avx512())
+            return CPUTYPE_SKYLAKEX;
           if(support_avx2())
             return CPUTYPE_HASWELL;
           if(support_avx())
--- a/driver/others/dynamic.c
+++ b/driver/others/dynamic.c
@@ -708,8 +708,11 @@ static gotoblas_t *get_coretype(void){
 	
       case 9:
         if (model == 7 || model == 10) { // Alder Lake
+	   if(support_avx512_bf16())
+             return &gotoblas_COOPERLAKE;
+          if (support_avx512()) 
+	    return &gotoblas_SKYLAKEX;
           if(support_avx2()){
-            openblas_warning(FALLBACK_VERBOSE, HASWELL_FALLBACK);
             return &gotoblas_HASWELL;
           }
           if(support_avx()) {


FCLC:
Thanks @martin-frbg 

If I understand this merge correctly, for now SR has yet to be broken out into a separate unit from Cooper Lake. 

As of now AL is being treated like the CL spec based on having AVX512-BF16 if available, otherwise it falls back on Haswell AVX-2

Rocket Lake, Ice Lake and Tiger Lake are all treated as Skylake X due to the lack of AVX-512BF16

With AL, RL, IL, TL and when released, SR  having additional support for ```avx512ifma avx512vbmi avx512_vbmi2   avx512_bitalg avx512_vpopcntdq  avx512_vp2intersect avx512_fp16``` amongst others over Cooper Lake, with many of those being shared with each other, will it make sense to break those out? since they all support a common set of AVX512 with the exception of bf16, where as Cooper lake is the inverse?

Of specific interest for BLAS would be the avx-512 FMA extensions available on Cannon Lake, Ice/ lake/Rocket lake, tiger lake and Alder lake/SR.

Also of interest are AVX512-FP16 extensions available on AL/SR could be of interest as an addition to augment any subroutines currently utilizing AVX512-bf16 intrinsic's in need of more significands instead of exponent bits.  (Also includes FMA specific for these sort of operations)

P.S.

I'm keeping an eye on the secondary die being released for Alder Lake that is only made of Pcores that will make up the lower portion of Intels stack. As of now no word on AVX-512 support, but We will have to wait and see.

martin-frbg:
if it has avx512-bf16 it is seen as CL, if "only" avx512 it is SKX, otherwise it falls back to Haswell as before. Handling the various other Lakes differently would make sense if there were any BLAS kernels that make use of their added instruction sets, but this is currently not the case. Similarly there is no support for fp16, nor any plans to add it. There is no big organisation or research group behind OpenBLAS, just the spare time of a very small number of individuals plus code contributions from interested vendors.

FCLC:
> just the spare time of a very small number of individuals plus code contributions from interested vendors.

Sounds good. 

Any advice on where to get started for submitting contributions? 

I've read through the wiki entry and pulled the git. 

It seems to me like the headers in /kernel/simd/intrin_[foo].h define the intrinsics to be v_[instruction_datatype]. 

From there at build time based on the selection script one of these headers is set to active and then used as the defining intrinsics header.

As a base POC, am I right in thinking that I could take OpenBLAS/kernel/x86_64/sbgemv_n_microk_cooperlake_ which is designed for bfloat16 avx512 micro kernels, convert them to their FP16 mem, fmad load str etc. equivalents to create a baseline AVX512-FP16 kernel within the OpenBLAS stack?

As far as I can tell for that cooper lake avx512 specific bfloat16 kernel, OpenBLAS currently uses 4 files: 

One header for macros which mainly serves to define matrix layouts. 

One templating C file

One file for bf16 type conversion functions

Finally the actual bfloat16 mikro kernels themselves.

Is that the standard approach for adding new intrinsics/kernels for a given data type? 

I also stumbled upon #2767 attempting to define the standard for bfloat naming and fp16 naming.

A search of the git returns no existing fp16 kernels for any architecture for any of the naming schemes proposed, let alone anything on x86_64.

Should I assume that a name of shgemv_n_microk_sapphirerapids.c is acceptable?

SH being Single Half as mentioned in the above issue? 

Should that be acceptable and the above noted assumptions hold true, using the table mentioned in https://github.com/xianyi/OpenBLAS/issues/2767#issuecomment-692442506

Where in:
hgemm: fp16 input, fp16 output
shgemm: fp16 input, fp32 output
sgemm: fp32 input, fp32 output
dgemm: fp64 input, fp64 output

Would serve as the formatting for the IEEE half precision FP GEMM kernels. 

Does that seem right?

martin-frbg:
> It seems to me like the headers in /kernel/simd/intrin_[foo].h define the intrinsics to be v_[instruction_datatype].

Yes. "Universal intrinsics" as originally codified by Intel.

> As a base POC, am I right in thinking that I could take OpenBLAS/kernel/x86_64/sbgemv_n_microk_cooperlake_

Yes, that would probably work (you do not have to follow that exact four-file layout, I guess that is more a question of personal coding style and practicality)

> A search of the git returns no existing fp16 kernels for any architecture
Correct, this has not yet been implemented (and at the time of the discussion in #2767 seemed a bit hypothetical). I believe using H and SH as prefixes was what was agreed upon. (I would hate having to do another mass rename if a new controversy
broke out, but it would certainly be doable again if necessary).

Besides the actual kernels themselves, there would need to be a bunch of interface code, additions to various headers etc. especially for DYNAMIC_ARCH builds, but I can take care of that

PR #3539, opened 2022-02-23
Use f2c translations of LAPACK when no Fortran compiler is available

### LINKED ISSUE #1284: When I compile OpenBLAS with NOFORTRAN=1 some LAPACK (e.g. dpotrf) routines are not available
Hi,
I'd like to build complete version of openBlas with maximum support of lapack/blas routines without using gfortran in runtime.

When I'm building openBlas from: 
develop branch from commit 50715e8945d24d648c12b5f79bfcdbc570052f3a 
with specifying NOFORTRAN=1 in make command

In this circumstances I don't have several routines in the final static library. They are:

1. dpotrf
2. dpotrs
3. dgels
4. dpotrf
5. zpotrf
6. dlarfg
7. dlarf
8. zlarfg
9. zlarf
10. dlarft
11. zlarft
12. zlarfb
13. dlarfb
14. dlarft

What is interesting this all symbols are defined if build withtout `NOFORTRAN=1`.
Unfortunately I want to have OpenBlas library without dependencies in gfortran

### LINKED ISSUE #3509: clang64 flang compile failure
it works fine when I use mingw64 to compile openblas on windows, but it compiles failure when I use clang64 flang with Ninja generator .

I try to fix the issue, and add the FC environment variable, it can build successfully, but it seems the lapack isn't compiled. the prebuild log also shows the info :
-- Looking for a Fortran compiler - NOTFOUND
-- No Fortran compiler found, can build only BLAS but not LAPACK

Is threre a solution that can build LAPACK successfully on windows with clang64 ? thanks in advance!

martin-frbg:
translations done with a modified version of the f2clapack script from PRIMME (which in turn based on the one from f2clapack)
and additional hand-editing to work around use of F90 idiom in current LAPACK.
- the current translation drops OpenMP parallelism from the few .F files that used it (f2c limitation, could be put back by hand)
- currently omits cgbrfsx,cgerfsx,cherfsx,cporfsx,csyrfsx, zgbrfsx,zgerfsx,zherfsx,zporfsx,zsyrfsx due to TRANSFER intrinsic
~~LAPACK-TEST currently shows failures in ?gelsd~~ only observed with gcc7.5.0
~~need to update most files to support INTERFACE64 - right now only the absolute minimum needed to run DGESVD (as used to prepare the DGEMM "kernel_regress" test in openblas_utest) is updated as a proof of concept~~
- currently does not work with MSVC (which uses structs to represent complex numbers) 
fixes  #3509 and fixes #1284


emmenlau:
This looks awesome! Will there be support to build this with cmake too?

martin-frbg:
The last bunch of additions contained the cmake-related changes. However  I've neglected to support the INTERFACE64 option so far(which is probably the reason for the failed checks on Mac).

emmenlau:
> The last bunch of additions contained the cmake-related changes. 

Ah, I overlooked the changes in `cmake/lapack.cmake`. Thanks! A slightly related note, currently when setting `ONLY_CBLAS=ON` in cmake, then `system.cmake` will set `NO_LAPACK` to `1`, which disables `include("${PROJECT_SOURCE_DIR}/cmake/lapack.cmake")` in `CMakeLists.txt`.

So maybe a change is required around https://github.com/xianyi/OpenBLAS/blob/1ef97c470ccf23d06d3bb0597c52df4a2e9093b9/cmake/system.cmake#L122?
 I'm not 100% certain I understand the implications of `BUILD_WITHOUT_LAPACK`, `BUILD_WITHOUT_CBLAS`, `NOFORTRAN`, `ONLY_CBLAS` and `C_LAPACK` well enough to propose something, but there seems to be some overlap between `NOFORTRAN`, `ONLY_CBLAS` and `C_LAPACK`?

emmenlau:
In any case, I can confirm that this PR works like magic to enable use of lots of BLAS functionality on Android (cross-compiled with NDK r23b using a pure cmake build). Awesome work!

djacobs010:
@emmenlau can you explain the process you used to xc and use on Android?

I tried to xc via:
```sh
export NDK=/path/to/NDK/r23b
make HOSTCC=gcc CC=$NDK/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android28-clang NO_FORTRAN=1 TARGET=ARMV8
make PREFIX=$NDK/toolchains/llvm/prebuilt/linux-x86_64/sysroot/openblas_f2c_lapack install
```
But I think I'm building with the wrong command (and likely also installing to a weird location), as the following command prints nothing, when I would expect the `dgesv` symbols to be in the built library:
```
readelf -s $NDK/toolchains/llvm/prebuilt/linux-x86_64/sysroot/openblas_f2c_lapack/lib/libopenblas.so | grep dgesv
```

emmenlau:
@danielcjacobs , I build using cmake. My cmake build instructions are a bit generalized because they come from a continuous integration system. But I think they should work for anyone, if you adjust the paths accordingly. My Android SDK is in `/opt/android-sdk`, and I have NDK `23.1.7779620` installed. The build is for `arm64-v8a` and platform `android-31`. The generator is `Ninja` but you can pick Makefiles or something else too.

```{bash}
cmake \
    /home/user/src/OpenBLAS \
    -GNinja \
    -DCMAKE_INSTALL_PREFIX=/data/user/Debug \
    -DCMAKE_BUILD_TYPE=Debug \
    -DCMAKE_CXX_STANDARD=17 \
    -DCMAKE_CXX_STANDARD_REQUIRED=ON \
    -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
    -DANDROID_SDK_ROOT="/opt/android-sdk" \
    -DCMAKE_TOOLCHAIN_FILE="/opt/android-sdk/ndk/23.1.7779620/build/cmake/android.toolchain.cmake" \
    -DANDROID_PLATFORM=android-31 \
    -DANDROID_ABI=arm64-v8a \
    -DBUILD_SHARED_LIBS=OFF \
    -DBUILD_TESTING=ON \
    -DUSE_OPTIMIZED_BLAS=OFF \
    -DBUILD_WITHOUT_LAPACK=OFF \
    -DBUILD_WITHOUT_CBLAS=OFF \
    -DNOFORTRAN=ON \
    -DONLY_CBLAS=OFF \
    -DC_LAPACK=ON
```
Note: I've just updated this comment after first copy-pasting a few wrong cmake options (different project).

martin-frbg:
@danielcjacobs perhaps try adding `C_LAPACK=1` to your build options (I have not tried cross-compiling, or indeed compiling on Android with these changes yet, could be there is still a spot where NO_FORTRAN categorically means NO_LAPACK).
Note this is still very much a work in progress - on x86_64 it passes all tests in the LAPACK testsuite except for the GELSD functions, but I have not checked other platforms.

djacobs010:
Ah, both `NO_FORTRAN=1` and `C_LAPACK=1` "work", I was just on the wrong branch lol.
Except it doesn't actually succeed, as I get the following error when running the `make` command I posted above:
```
/home/djacobs/Android/Sdk/ndk/23.1.7779620/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android28-ranlib: No such file or directory
```
There is a file called `llvm-ranlib`, and older NDK versions have `aarch64-linux-android-ranlib`, but `aarch64-linux-android28-ranlib` doesn't exist. Is there a way to specify the `RANLIB` path like I do with the `CC` argument?

djacobs010:
> @danielcjacobs , I build using cmake. My cmake build instructions are a bit generalized because they come from a continuous integration system. But I think they should work for anyone, if you adjust the paths accordingly. My Android SDK is in `/opt/android-sdk`, and I have NDK `23.1.7779620` installed. The build is for `arm64-v8a` and platform `android-31`. The generator is `Ninja` but you can pick Makefiles or something else too.
> 
> ```
> cmake \
>     /home/user/src/OpenBLAS \
>     -GNinja \
>     -DCMAKE_INSTALL_PREFIX=/data/user/Debug \
>     -DCMAKE_BUILD_TYPE=Debug \
>     -DCMAKE_CXX_STANDARD=17 \
>     -DCMAKE_CXX_STANDARD_REQUIRED=ON \
>     -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
>     -DANDROID_SDK_ROOT="/opt/android-sdk" \
>     -DCMAKE_TOOLCHAIN_FILE="/opt/android-sdk/ndk/23.1.7779620/build/cmake/android.toolchain.cmake" \
>     -DANDROID_PLATFORM=android-31 \
>     -DANDROID_ABI=arm64-v8a \
>     -DBUILD_SHARED_LIBS=OFF \
>     -DBUILD_TESTING=ON \
>     -DUSE_OPTIMIZED_BLAS=OFF \
>     -DBUILD_WITHOUT_LAPACK=OFF \
>     -DBUILD_WITHOUT_CBLAS=OFF \
>     -DNOFORTRAN=ON \
>     -DONLY_CBLAS=OFF \
>     -DC_LAPACK=ON
> ```
> 
> Note: I've just updated this comment after first copy-pasting a few wrong cmake options (different project).

@emmenlau have you tried installing yet? I tried `cmake --install ..`, but I get `CMake Error: Error processing file: /home/user/src/OpenBLAS/cmake_install.cmake`. Also, I need the shared lib, but removing `-DBUILD_SHARED_LIBS=OFF` still doesn't build the shared lib.

emmenlau:
Everything works for me smoothly. I use ninja as I was saying, so after cmake I run:
```
ninja -j8
ninja install
```
and thats it. Can you share your actual commands the exact errors you're getting?

djacobs010:
ah `ninja install` worked, so I'll just go with that!

Just to summarize, here's what I did:
```
cmake \
    /home/user/src/OpenBLAS \
    -GNinja \
    -DCMAKE_INSTALL_PREFIX=/home/user/installed_openblas \
    -DCMAKE_CXX_STANDARD=17 \
    -DCMAKE_CXX_STANDARD_REQUIRED=ON \
    -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
    -DANDROID_SDK_ROOT="/path/to/Android/Sdk" \
    -DCMAKE_TOOLCHAIN_FILE="/path/to/ndk/23.1.7779620/build/cmake/android.toolchain.cmake" \
    -DANDROID_PLATFORM=android-28 \
    -DANDROID_ABI=arm64-v8a \
    -DBUILD_SHARED_LIBS=ON \
    -DBUILD_TESTING=ON \
    -DUSE_OPTIMIZED_BLAS=OFF \
    -DBUILD_WITHOUT_LAPACK=OFF \
    -DBUILD_WITHOUT_CBLAS=OFF \
    -DNOFORTRAN=ON \
    -DONLY_CBLAS=OFF \
    -DC_LAPACK=ON
ninja -j8
ninja install
```
(before trying `ninja`, I tried `cmake ..` followed by `cmake --install ..` to no avail, but since it works now I don't really care lol)

djacobs010:
@martin-frbg when you get this fully working, would you be willing to add a note to the “compile for Android” docs on using clapack?

mmuetzel:
Is this change also necessary?
``` diff
--- OpenBLAS-0.3.20/cmake/lapack.cmake.orig	2022-03-20 12:57:35.728321300 +0100
+++ OpenBLAS-0.3.20/cmake/lapack.cmake	2022-03-20 13:08:48.512121500 +0100
@@ -990,4 +990,8 @@
 foreach (LA_FILE ${LA_GEN_SRC})
   list(APPEND LA_SOURCES "${NETLIB_LAPACK_DIR}/TESTING/MATGEN/${LA_FILE}")
 endforeach ()
-set_source_files_properties(${LA_SOURCES} PROPERTIES COMPILE_FLAGS "${LAPACK_FFLAGS}")
+if (NOT C_LAPACK)
+  set_source_files_properties(${LA_SOURCES} PROPERTIES COMPILE_FLAGS "${LAPACK_FFLAGS}")
+else ()
+  set_source_files_properties(${LA_SOURCES} PROPERTIES COMPILE_FLAGS "${LAPACK_CFLAGS}")
+endif ()
```

mmuetzel:
Is this change also necessary?
``` diff
--- OpenBLAS-0.3.20/cmake/lapack.cmake.orig	2022-03-20 12:57:35.728321300 +0100
+++ OpenBLAS-0.3.20/cmake/lapack.cmake	2022-03-20 13:08:48.512121500 +0100
@@ -990,4 +990,8 @@
 foreach (LA_FILE ${LA_GEN_SRC})
   list(APPEND LA_SOURCES "${NETLIB_LAPACK_DIR}/TESTING/MATGEN/${LA_FILE}")
 endforeach ()
-set_source_files_properties(${LA_SOURCES} PROPERTIES COMPILE_FLAGS "${LAPACK_FFLAGS}")
+if (NOT C_LAPACK)
+  set_source_files_properties(${LA_SOURCES} PROPERTIES COMPILE_FLAGS "${LAPACK_FFLAGS}")
+else ()
+  set_source_files_properties(${LA_SOURCES} PROPERTIES COMPILE_FLAGS "${LAPACK_CFLAGS}")
+endif ()
```

martin-frbg:
I think so (FFLAGS not identical to CFLAGS)

mmuetzel:
> I think so (FFLAGS not identical to CFLAGS)

One additional data point: Compilation with clang for Windows 32-bit i686 fails without that change. But it succeeds with it.
The error without that change is:
```
clang: error: unknown argument: '-mincoming-stack-boundary=2'
```

mmuetzel:
Re-reading the last few comments, I realize this was kind of ambiguous. Trying again:
Would you like to amend your PR with this change?
``` diff
--- OpenBLAS-0.3.20/cmake/lapack.cmake.orig	2022-03-20 12:57:35.728321300 +0100
+++ OpenBLAS-0.3.20/cmake/lapack.cmake	2022-03-20 13:08:48.512121500 +0100
@@ -990,4 +990,8 @@
 foreach (LA_FILE ${LA_GEN_SRC})
   list(APPEND LA_SOURCES "${NETLIB_LAPACK_DIR}/TESTING/MATGEN/${LA_FILE}")
 endforeach ()
-set_source_files_properties(${LA_SOURCES} PROPERTIES COMPILE_FLAGS "${LAPACK_FFLAGS}")
+if (NOT C_LAPACK)
+  set_source_files_properties(${LA_SOURCES} PROPERTIES COMPILE_FLAGS "${LAPACK_FFLAGS}")
+else ()
+  set_source_files_properties(${LA_SOURCES} PROPERTIES COMPILE_FLAGS "${LAPACK_CFLAGS}")
+endif ()
```

mmuetzel:
Is this PR in a state where it would be helpful to give feedback on test results?
If it is not ready yet, I'll wait a little longer.

martin-frbg:
My primary focus is on getting the xGELSD functions to pass their tests (on x86_64, but I expect other platforms to be mostly the same unless their lapack-test was already broken before)

mmuetzel:
I tested with msys2's CLANG64 environment. That is a build environment that uses a LLVM toolchain. Currently, it doesn't have a Fortran compiler that actually produces binaries (only `flang-new`). So, these changes would be very interesting imho.

`ctest` segfaulted for a build with 64-bit Fortran integers in that environment. But that also happened with the Fortran LAPACK bundled with OpenBLAS in a MINGW64 environment using a GCC toolchain with `gfortran` (see #3575).
I haven't tested again on top of the changes of that PR. Please let me know if you think that would be helpful.

If you are interested, I could try to add an action for GitHub runners that builds OpenBLAS in different configurations in some of the build environments provided by MSYS2:
https://www.msys2.org/docs/environments/
I thought of maybe MINGW64 with 32-bit and with 64-bit Fortran integers, CLANG64 with 32-bit and with 64-bit Fortran integers, and MINGW32 with 32-bit integers.
Would that be helpful for test coverage in your opinion?

martin-frbg:
Well, I guess both the original Fortran and the converted version would be affected by any mis-definition of the 64bit-integer type, so if you have the time and resources it would be useful to confirm these problems are gone with your PR applied now.
More tests would be a plus - we do have some Windows configurations on Azure, but I believe from your list only MINGW32 with 32bit integers (and conda-provided clang/flang) is currently covered.

mmuetzel:
I applied the changes from this PR on top of ad4598143bfbf77acd16f4ae884462edc69b496e locally with the additional change from https://github.com/xianyi/OpenBLAS/pull/3539#issuecomment-1073238003 (otherwise it wouldn't compile).
In msys2's CLANG64 build environment, configured with (similar to the configuration used for the OpenBLAS shipped by msys2):
```
cmake -DBUILD_SHARED_LIBS=ON -DBUILD_STATIC_LIBS=ON -DDYNAMIC_ARCH=ON -DUSE_THREAD=ON -DNUM_THREADS=64 -DTARGET=CORE2 -DBINARY=64 -DINTERFACE64=1 -DC_LAPACK=ON ..
```
Unfortunately, `ctest` still segfaults. Test log: 
[LastTest.log](https://github.com/xianyi/OpenBLAS/files/8381321/LastTest.log)

`gdb` backtrace from the crash:
```
Thread 1 received signal SIGSEGV, Segmentation fault.
0x00007ff6899ac43e in dnrm2_k_ZEN ()
(gdb) bt
#0  0x00007ff6899ac43e in dnrm2_k_ZEN ()
#1  0x00007ff689b38c89 in dnrm2_ ()
#2  0x00007ff689b322e8 in dlarfg_ ()
#3  0x00007ff689b27dd6 in dgebd2_ ()
#4  0x00007ff688249e99 in dgebrd_ ()
#5  0x00007ff68822a6b0 in dgesvd_ ()
#6  0x00007ff6881e6d69 in LAPACKE_dgesvd_work ()
#7  0x00007ff68816d5dc in LAPACKE_dgesvd ()
#8  0x00007ff6881663ba in __ctest_kernel_regress_skx_avx_run ()
#9  0x00007ff688161f4f in ctest_main ()
#10 0x00007ff6881621fe in main ()
```

This was on:
```
PS> Get-CIMInstance -Class Win32_Processor | Select-Object -Property Name

Name
----
AMD Ryzen 7 5800X 8-Core Processor
```

Did I do something wrong?

Edit: Oops. Realized now that this part of the patch didn't apply:
```
--- CMakeLists.txt
+++ CMakeLists.txt
@@ -19,6 +19,7 @@ include(CMakePackageConfigHelpers)
 
 if(MSVC AND NOT DEFINED NOFORTRAN)
   set(NOFORTRAN ON)
+  set(TIMER "NONE")
 endif()
 
 #######
```
Could that be the reason? (Guess not since I don't use MSVC anyway...)

martin-frbg:
No idea, may even be unrelated to the LAPACK integer size (that test uses CBLAS_DNRM2 to check the output of a round-trip calculation involving LAPACKE_DGESVD and CBLAS_DGEMM). Does it also fail with the Fortran LAPACK ?
At least it is not related to the TIMER setting - this is just a compile-time choice of a source file in LAPACK (to select among various fortran dialects for reading cpu time in benchmarks)

CMakeLists.txt
```diff
@@ -17,11 +17,18 @@ include(GNUInstallDirs)
 
 include(CMakePackageConfigHelpers)
 
+if(MSVC AND NOT DEFINED NOFORTRAN)
+  set(NOFORTRAN ON)
+  set(TIMER "NONE")
+endif()
+
```
mmuetzel: IIUC, this effectively reverts ed8c028f7fc30cc2327410b8051c99f8eb1f58cb.

Should this line be removed from the patch?
``` diff
+  set(NOFORTRAN ON)
```

mmuetzel:
> Does it also fail with the Fortran LAPACK ?

The CLANG64 build environment doesn't have a Fortran compiler. So, I rebuilt (without `-DC_LAPACK=ON`) in a new build tree with the MINGW64 build environment. With that configuration `ctest` didn't segfault, and all tests did pass.
To rule out that the segfault depends on the build tools (LLVM vs. GCC toolchain), I rebuilt in a new build tree with the MINGW64 build environment but this time *with* `-DC_LAPACK=ON`. With that configuration, the first test resulted in a segfault again. And it looks like it is now stuck indefinitely at `Start  39: SSVD:_Testing_Singular_Value_Decomposition_routines`. The other tests up to that point (excluding the first one that resulted in a segfault) did pass.

So, there might be an issue with the 64-bit indexing version of C_LAPACK (at least on Windows).

Edit: The gdb backtrace (with MINGW64) looks very similar to the one with CLANG64:
```
TEST 33/34 potrf:smoketest_trivial [OK]
TEST 34/34 kernel_regress:skx_avx
Thread 1 received signal SIGSEGV, Segmentation fault.
0x00007ff65bfe7cc0 in dnrm2_k_ZEN ()
(gdb) bt
#0  0x00007ff65bfe7cc0 in dnrm2_k_ZEN ()
#1  0x00007ff65c7537bd in dnrm2_ ()
#2  0x00007ff65c74e950 in dlarfg_ ()
#3  0x00007ff65c767c76 in dgebd2_ ()
#4  0x00007ff65c7475e8 in dgebrd_ ()
#5  0x00007ff65c711e00 in dgesvd_ ()
#6  0x00007ff65b16d56a in LAPACKE_dgesvd_work ()
#7  0x00007ff65b16cf96 in LAPACKE_dgesvd ()
#8  0x00007ff65b086b69 in __ctest_kernel_regress_skx_avx_run ()
#9  0x00007ff65b0822b0 in ctest_main ()
#10 0x00007ff65b08242e in main ()
```

martin-frbg:
Fixed a few files I had missed in the big edit for INTERFACE64, but probably not the cause of the utest failure.

mmuetzel:
Built again with `-DCMAKE_BUILD_TYPE=Debug` after updating to your latest changes:
```
TEST 34/34 kernel_regress:skx_avx
Thread 6 received signal SIGSEGV, Segmentation fault.
[Switching to Thread 12812.0x3380]
0x00007ff6a72a8fb0 in dscal_kernel_8_zero (n=27111981056, alpha=0xd4de1ffa98, x=0x2b585b24020) at D:/OpenBLAS/kernel/x86_64/dscal_microk_haswell-2.c:150
150             __asm__  __volatile__
(gdb) bt
#0  0x00007ff6a72a8fb0 in dscal_kernel_8_zero (n=27111981056, alpha=0xd4de1ffa98, x=0x2b585b24020)
    at D:/OpenBLAS/kernel/x86_64/dscal_microk_haswell-2.c:150
#1  0x00007ff6a72a93b0 in dscal_k_ZEN (n=27111981056, dummy0=0, dummy1=0, da=-0, x=0x2b585b24020, inc_x=1, y=0x0, inc_y=0, dummy=0x0, dummy2=0)
    at D:/OpenBLAS/kernel/x86_64/dscal.c:217
#2  0x00007ff6a63aaa2d in legacy_exec (func=0x7ff6a72a91da <dscal_k_ZEN>, mode=32771, args=0xd4dd7fb6c8, sb=0x282bdb70000)
    at D:/OpenBLAS/driver/others/blas_server_win32.c:99
#3  0x00007ff6a63ab024 in blas_thread_server (arg=0x1) at D:/OpenBLAS/driver/others/blas_server_win32.c:318
#4  0x00007fffbb7c54e0 in KERNEL32!BaseThreadInitThunk () from C:\Windows\System32\kernel32.dll
#5  0x00007fffbc9c485b in ntdll!RtlUserThreadStart () from C:\Windows\SYSTEM32\ntdll.dll
#6  0x0000000000000000 in ?? ()
Backtrace stopped: previous frame inner to this frame (corrupt stack?)
(gdb)
```

It crashes in a different function now.
That value of `n` looks pretty high. Is it supposed to be this large?

martin-frbg:
Stack overwritten somewhere, so unfortunately that does not tell much. (You'd get the same signature from running multithreaded GEMM with a huge matrix and a small BUFFERSIZE).
(That "somewhere" is most likely the f2c-converted DGESVD, but I do not notice anything obviously wrong there. The problem does not occur on Linux, so valgrind is not an option)

mmuetzel:
This took me a while. But I think I finally got it.
`OS_WINDOWS` and `__64BIT__` are never defined in the C_LAPACK sources. I find them defined in `config.h` and `config_kernel.h`. But those aren't included in these "upstream" files.
That means that `integer` in those files is `long` (but it should be `long long` on Windows 64-bit).

Should the following be used instead in all files in the conditions that select how to typedef `BLASLONG` and `BLASULONG` and also for `blasabs(x)`?
``` diff
-#if defined(OS_WINDOWS) && defined(__64BIT__)
+#if defined(_WIN64)
```

Edit:
I made that change locally and compiled with the tools/compilers in the CLANG64 and MINGW64 environments. Both times, with 64-bit indexing and with C_LAPACK=On. `openblas_utest` no longer segfaults. 🎉
But still, it looks like the tests are stuck indefinitely at `Start  39: SSVD:_Testing_Singular_Value_Decomposition_routines` for the MINGW64 environment (100% load of one core).

I don't know how to propose changes to an open PR correctly. I'm uploading a patch with those changes. I hope that is ok.
~~[0001-C_LAPACK-Check-for-_WIN64-in-preprocessor-conditions.zip](https://github.com/xianyi/OpenBLAS/files/8399616/0001-C_LAPACK-Check-for-_WIN64-in-preprocessor-conditions.zip)~~

Edit 2:
Oops. Forgot to adapt those conditions in the files in the TESTING folder. Updated patch:
[0001-C_LAPACK-Check-for-_WIN64-in-preprocessor-conditions.zip](https://github.com/xianyi/OpenBLAS/files/8399758/0001-C_LAPACK-Check-for-_WIN64-in-preprocessor-conditions.zip)

With it, `ctest` finishes without any failed tests. Hurray! 🎉

martin-frbg:
Ah, sorry, I copied that block from common.h but it would need config.h to see OS_WINDOWS... no wonder I could not reproduce this on Linux. (I'm still trying to track down the xGELSD bug)

mmuetzel:
I don't see the failure in `xGELSD`. Is that part of the tests run by `ctest`?

martin-frbg:
Yes, the LAPACK testsuite - but only available if you have a Fortran compiler installed as the testsuite is not translated. I'm currently working with a gmake build, but I guess with ctest you would see e.g. one test in the 40s or 50s range failing (SLS)

martin-frbg:
Oh sh... it is a gcc problem. Present with 7.5 (suse 15.x default) and absent from at least 10.2 onwards. I really need to take some time off as soon as possible.

martin-frbg:
Now I wonder why MSVC's inability to handle _Complex has not come up earlier...

mmuetzel:
This would already be very useful for the CLANG* build environments in MSYS2 in its current state.
Do you think it could make sense to disable C_LAPACK for MSVC for now (and maybe fix that in a follow up)?

CMakeLists.txt
```diff
@@ -175,8 +177,8 @@ endforeach ()
 
 # Can't just use lapack-netlib's CMake files, since they are set up to search for BLAS, build and install a binary. We just want to build a couple of lib files out of lapack and lapacke.
 # Not using add_subdirectory here because lapack-netlib already has its own CMakeLists.txt. Instead include a cmake script with the sources we want.
-if (NOT NOFORTRAN AND NOT NO_LAPACK)
-  include("${PROJECT_SOURCE_DIR}/cmake/lapack.cmake")
+if (NOT NO_LAPACK)
+    include("${PROJECT_SOURCE_DIR}/cmake/lapack.cmake")
```
mmuetzel: Nit: Should the whitespace change in line 181 be reverted?

Makefile
```diff
@@ -246,10 +249,10 @@ netlib :
 
 else
 netlib : lapack_prebuild
-ifeq ($(NOFORTRAN), $(filter 0,$(NOFORTRAN)))
+#ifeq ($(NOFORTRAN), $(filter 0,$(NOFORTRAN)))
```
mmuetzel: Would this condition be better here?
``` diff
-ifeq ($(NOFORTRAN), $(filter 0,$(NOFORTRAN)))
+ifneq ($(NO_LAPACK), 1)
```

cmake/f_check.cmake
```diff
@@ -26,10 +26,15 @@ if(CMAKE_Fortran_COMPILER)
   enable_language(Fortran)
 else()
   if (NOT NO_LAPACK)
-    message(STATUS "No Fortran compiler found, can build only BLAS but not LAPACK")
+	  message(STATUS "No Fortran compiler found, can build only BLAS and f2c-converted LAPACK")
```
mmuetzel: Nit: Whitespace?

cmake/system.cmake
```diff
@@ -284,8 +284,12 @@ if (NOT NOFORTRAN)
   # Fortran Compiler dependent settings
   include("${PROJECT_SOURCE_DIR}/cmake/fc.cmake")
 else ()
-set(NO_LAPACK 1)
-set(NO_LAPACKE 1)
+set(C_LAPACK 1)
+if (INTERFACE64)
+  set (CCOMMON_OPT "${CCOMMON_OPT} -DLAPACK_ILP64")
+endif ()
+set(TIMER "NONE")
```
mmuetzel: Nit: Intent block by two spaces?

mmuetzel (commented):
Just some comments on the existing changes. I hope this is helpful.

Makefile
martin-frbg: Drop the outer ifneq down there you mean ? Agree this should simplify things a bit.

cmake/f_check.cmake
martin-frbg: now fixed with the temporary exclusion of MSVC

cmake/system.cmake
martin-frbg: also fixed with the MSVC changes

mmuetzel:
Thanks for the recent changes regarding MSVC and C_LAPACK (or rather their mutual exclusivity for the time being).

I noticed the last couple of changes were related to an issue with clang-cl (MSVC) & flang on Windows. IIUC, that is probably because some tests are run now (after #3595 has landed) that have been skipped previously.
Are those failing tests a blocker for this PR?

martin-frbg:
not a blocker, i just noticed them in the context and wanted to make sure they are unrelated. different pr for that now, will clean up here later when i have time

lesteve:
> translations done with a modified version of the f2clapack script from PRIMME (which in turn based on the one from f2clapack)
and additional hand-editing to work around use of F90 idiom in current LAPACK.

@martin-frbg in case you still your modified f2clapack script easy accessible, I would be interested to take a look at it. I am guessing the PRIMME script you started from is https://github.com/primme/primme/blame/master/R/f2clapack.sh?

In a [Pyodide context](https://github.com/pyodide/pyodide) I have been trying to use OpenBLAS (see https://github.com/pyodide/pyodide/pull/3331 for more details). Being able to look at the f2c conversion script (or even run it if possible) you used for OpenBLAS, may help in some cases to reduce the discrepancy we are seeing between some f2c'ed scipy files and OpenBLAS.

martin-frbg:
yes that is what I started from, need to  clean up a few hacks that I hand-corrected in the converted files afterwards... mostly fixing some of the fortran function macros from the header part of the primme script that had bad variable types and adding implementations for some complex functions

lesteve:
OK so you changed mostly the header part of the primme script and then modified the result by hand?

I was wondering whether you changed the lex part too, in order to fix some signatures (full disclosure I don't understand what the lex part of the script does).

martin-frbg:
This should be it - sorry for the delay. Obviously some of the changes are OpenBLAS-specific (like the "blasint" typedef) and some of the (complex function) declarations will lead to "unused function" warnings during compilation, so there is certainly still room for improvement 
[f2clapack.sh.txt](https://github.com/xianyi/OpenBLAS/files/11246761/f2clapack.sh.txt)

lesteve:
Thanks a lot for the update!

PR #3541, opened 2022-02-23
Fix compilation for SkylakeX with gcc 6.x

martin-frbg:
fixes #3530 by falling back to the generic implementations of the small matrix kernels for old compilers that do not support the _mm512_reduce_add_pd intrinsic

PR #3544, opened 2022-02-23
Fix compilation of Skylake AVX512 kernels with GCC 6

giordano:
Ref: https://github.com/xianyi/OpenBLAS/pull/3541#discussion_r813384485.  I can
confirm with this PR I can build OpenBLAS with GCC 6 for x86_64.


martin-frbg:
thanks, i'll close mine then :)

giordano:
Thanks again! :slightly_smiling_face:

Issue #3557, opened 2022-03-04
Auto-detection of NO_AVX512 incorrect and potentially causing issues with x86_64 Builds 


e4t:
Commit 21eda8b5 introduced a detection feature for compiler support of AVX512 on x86_64 while patch 7a9a4dbc4 attempted to extend this to the 'native' CPU the compiler is running on.
These patches (and patches on top attempting to 'fix' these) are problematic for several reasons:
1. The check in is incorrect. To check for AVX512 support one would have to check whether `__AVX512F__` is defined to 1. The check for `__AVX2__` does just that: check if the compiler in its present settings will know of the AVX2 instruction set.
2. The check is orthogonal to multiple paradigms of OpenBLAS:
    1. It assumes that HOSTCC is the same as CC thus breaks cross compiling - which admittedly is not an important feature for x86_64.
    2. On top of 1. using `-march=native` assumes that the build machine has identical hardware to the machine the code will run on. Different build systems may provide different results.
    3. An `-march=native` test is not really valid in connection with DYNAMIC_ARCH.

There are several options to fix this. 
- The test could simply be removed. This would require the user to specify set the make variable NO_AVX512 explicitly. Not sure how relevant it is today to support a gcc version that doesn't support  AVX512 - especially when used on systems that provide this feature. Those users will lose some convenience.
- Since this test is limited to gcc meanwhile and we know which version of gcc has introduced AVX256 support, another easy solution would be to set NO_AVX512 depending on the gcc version used. This would allow to limit this features to gcc versions where support for AVX512 has stabilized.
- The automated detection could be retained by doing:
   `NO_AVX512=$(shell /bin/sh -c '{ echo | $(CC) -mavx512f -E - &>/dev/null; } && echo 1 || echo 0')`
  in Makefile.system.
I'd like to put this up for discussion - I'm happy to provide a patch for whatever solution is favored.

brada4:
It would work like that in a perfect world.
There are systems where AVX-es are in compiler but not in assembler and vice versa (think ubuntu16, rhel6, they are still alive for those willing to pay)
PR that does cleanup of related detection and ifdef-s is always welcome.

e4t:
Indeed. The current test would not have detected this either as it operates on the preprocessor level.
In theory, commit 7a9a4db (ie testing with `-march=native`) may have fixed an issue with CPU auto-detection. But the comment to it doesn't seem that this was the reason.
I will submit a patch to clean this up.

[...]

PR #3569, opened 2022-03-15
Fix Bug#3557 - Disable Attempt to detect Target Features on Build Host when building with DYNAMIC_ARCH

e4t:
This patch set contains of two patches:
1. The first patch changes the build system to generate separate getarch-<TARGET> binaries and - for non cmake builds - separate config_kernel_<TARGET>h and Makefile.<TARGET>.conf files when building with DYNAMIC_TARGET. This helps to better debug build problems.
2. The 2nd patch proposes a fix for issue #3557.


martin-frbg:
Thanks. However that is a lot of changes for a fairly localised bug, not sure I'd want the build system less compact by default (though it could be useful as a patchset specifically for debugging). My gut feeling is that I'd rather go for something that reverts the misguided test in getarch.c and forwards c_check's conclusion from running the actual compile test instead.

e4t:
> Thanks. However that is a lot of changes for a fairly localised bug, not sure I'd want the build system less compact by default (though it could be useful as a patchset specifically for debugging). 
My gut feeling is that I'd rather go for something that reverts the misguided test in getarch.c and forwards c_check's conclusion from running the actual compile test instead.

The two patches address different things each: 

- the first one tries to make the build a bit more transparent by not overwriting generated config files. This helped me debugging this issue as well as  #3570 and therefore seemed to be worthwhile sharing.
- The 2nd one simply attempts to disable the automatic detection for cases where the target is specified. 

Utilizing the detection of compiler features would definitely be the way to go, however, I'm afraid this will require a better separation of overall settings vs. target specific settings. 
One step towards this would be to include Makefile.prebuild and create real build targets - instead of running it as a 'side effect' of setting a dummy variable. The first patch may actually help with this.
I was looking into this, however, I was afraid of breaking things unintentionally. 
This patchset caused some of the tests to fail already - I'm not sure why the error shown is related. I'm still investigating.

e4t:
Most of this has been fixed with PR  #3580 et.al.

PR #3570, opened 2022-03-15
Fix preset value of gotoblas->sbgemm_r  in COOPERLAKE Kernel when using  DYNAMIC_ARCH

e4t:
This patch set contains of two patches:
1. Do not include symbols defined in driver/others/parameter.c when building with  DYNAMIC_ARCH - this helps to detect issues like in 2.
2. Use a fixed value for gotoblas->sbgemm_r in the COOPERLAKE kernel as sbgemm_r does not exist as a global parameter when building with DYNAMIC_ARCH.


martin-frbg:
This was already addressed in #3547

e4t:
> This was already addressed in #3547

Indeed. I believe you've made this submission right around the time I was investigating this.
However, there are two patches in this set. The other one may still be worthwhile as it will make symbols not available when building with DYNAMIC_ARCH invisible (by not declaring them). This should help to detect issues like this one more easily. 
I'm happy to submit this patch separately.

e4t:
> > This was already addressed in #3547
> 
> Indeed. I believe you've made this submission right around the time I was investigating this. However, there are two patches in this set. The other one may still be worthwhile as it will make symbols not available when building with DYNAMIC_ARCH invisible (by not declaring them). This should help to detect issues like this one more easily. I'm happy to submit this patch separately.

I've looked into this a bit deeper - I would claim that while commit d9894f4 does seem to fix this issue, it is incorrect: there is a:
```
#ifdef BUILD_BFLOAT16
...
TABLE_NAME.sbgemm_r = SBGEMM_DEFAULT_R;
...
#endif
```
around line 1242 of setparam-ref.c already. Thus, the patch above will only overwrite this while the first reference will still use the declared - but with DYNAMIC_ARCH never defined - sbgemm_r (to which SBGEMM_DEFAULT_R expands to).
Thus, a non-existing symbol will still be used and cause issues during linking. The only reason I can think of why this does not actually happen is that the compiler  optimizes out the first instance (the one quoted above). I've verified that this is indeed the case by building the code with -O0 like:

> /tmp/ccrMUjaB.ltrans0.ltrans.o: in function `init_parameter.lto_priv.12':
> /mounts/users-space/eich/sources/HPC/OpenBLAS/test/../kernel/setparam_COOPERLAKE.c:1242: undefined reference to `sbgemm_r'
> /usr/lib64/gcc/x86_64-suse-linux/11/../../../../x86_64-suse-linux/bin/ld: warning: creating DT_TEXTREL in a PIE
> collect2: error: ld returned 1 exit status
> make[1]: *** [Makefile:329: test_sbgemm] Error 1

 I've got a linking error which did not occur when building with -O2.

martin-frbg:
Oops, you are right. I should have followed the example of sgemm_r in all respects here...

e4t:
> Oops, you are right. I should have followed the example of sgemm_r in all respects here...

This would always pick the value from gotoblas_COOPERLAKE and ignore what's set in blas_set_parameter() - regardless if built with DYNAMIC_ARCH or not.
So this begs the question if the latter function should not be grandfathered - at least for x86_64. I'm not so sure about other arches. The problem seems to be that these parameters can be set in way too many ways. This ought to be cleaned up.

martin-frbg:
setparam_ref is only used in DYNAMIC_ARCH builds

e4t:
> setparam_ref is only used in DYNAMIC_ARCH builds

Right. Can you share why this is the case? 
It seems like blas_set_parameter() is doing the same what init_parameter() does - in the target specific gotoblas_t structure.
I don't seem to have grasped the subtle difference between the two, yet, that warrant two separate sets of init()/setup() functions.

martin-frbg:
Almost certainly one was derived from the other when DYNAMIC_ARCH was first implemented, but this is original GotoBLAS code for which no meaningful development history or documentation exists. I'd rather work within the existing framework than try to rewrite it for clarity (with a high probability of finding out why simplification was and is not possible)

e4t:
>  I'd rather work within the existing framework than try to rewrite it for clarity (with a high probability of finding out why simplification was and is not possible)

The unfortunate downside if this approach is that the code is harder to extend (add new kernels) as very similar changes need to happen at multiple places. This can easily be forgotten (which ideally would be detected by testing during CI) or be out of sync leading to slight, yet unintended differences in behavior (which is much harder to detect by testing). 
Here, we've already seen examples of this.

e4t:
The topics covered by this PR have been resolved with  PR #3547 & PR #3580 or resubmitted as PR #3589.

PR #3571, opened 2022-03-15
Fix malfunctioning AVX512 check in getarch

martin-frbg:
As explained in #3557, the additional test introduced in #1980 to ensure that the compiler is capable of handling AVX512 instructions is broken. Rather than checking a compiler property, as the largely redundant test in c_check does, it added an unwanted dependency on the compile host. As it turns out, the CMAKE build was already making the state of the NO_AVX512 variable from c_check available to the getarch build, this PR adds the equivalent for gmake by parsing the initial Makefile.conf output from c_check.

### DIFF
--- a/Makefile.prebuild
+++ b/Makefile.prebuild
@@ -71,7 +71,11 @@ endif
 
 
 getarch : getarch.c cpuid.S dummy $(CPUIDEMU)
-	$(HOSTCC) $(HOST_CFLAGS) $(EXFLAGS) -o $(@F) getarch.c cpuid.S $(CPUIDEMU)
+ifneq (,$(findstring NO_AVX512,$(file < $(TARGET_MAKE))))
+	$(HOSTCC) $(HOST_CFLAGS) $(EXFLAGS) -DNO_AVX512=1 -v -o $(@F) getarch.c cpuid.S $(CPUIDEMU) >getarchlog 2>&1
+else
+	$(HOSTCC) $(HOST_CFLAGS) $(EXFLAGS) -v -o $(@F) getarch.c cpuid.S $(CPUIDEMU) >getarchlog 2>&1
+endif
 
 getarch_2nd : getarch_2nd.c config.h dummy
 ifndef TARGET_CORE
--- a/Makefile.prebuild
+++ b/Makefile.prebuild
@@ -71,10 +71,10 @@ endif
 
 
 getarch : getarch.c cpuid.S dummy $(CPUIDEMU)
-ifneq (,$(findstring NO_AVX512,$(file < $(TARGET_MAKE))))
-	$(HOSTCC) $(HOST_CFLAGS) $(EXFLAGS) -DNO_AVX512=1 -v -o $(@F) getarch.c cpuid.S $(CPUIDEMU) >getarchlog 2>&1
+ifeq (1, $(shell perl ./avx512_check $(CC) $(TARGET_FLAGS)))
+	$(HOSTCC) $(HOST_CFLAGS) $(EXFLAGS) -DNO_AVX512=1 -o $(@F) getarch.c cpuid.S $(CPUIDEMU)
 else
-	$(HOSTCC) $(HOST_CFLAGS) $(EXFLAGS) -v -o $(@F) getarch.c cpuid.S $(CPUIDEMU) >getarchlog 2>&1
+	$(HOSTCC) $(HOST_CFLAGS) $(EXFLAGS) -o $(@F) getarch.c cpuid.S $(CPUIDEMU)
 endif
 
 getarch_2nd : getarch_2nd.c config.h dummy
--- a/avx512_check
+++ b/avx512_check
@@ -125,7 +125,7 @@ if (($architecture eq "x86") || ($architecture eq "x86_64")) {
     if ($@){ 
 	$no_avx512 = 0;
     } else {
-	($fh,$tmpf) = tempfile( SUFFIX => '.c' , UNLINK => 0 );
+	($fh,$tmpf) = tempfile( SUFFIX => '.c' , UNLINK => 1 );
 	$code = '"vbroadcastss -4 * 4(%rsi), %zmm2"';
 	print $fh "#include <immintrin.h>\n\nint main(void){ __asm__ volatile($code); }\n";
 	$args = " -march=skylake-avx512 -v -c -o $tmpf.o $tmpf";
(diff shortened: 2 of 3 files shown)


e4t (commented):
I'm not convinced that this will work: TARGET_MAKE will be created by Makefile.prebuild - it must be available at the time Makefile.prebuild is read, not when the rule executes (which would not be the case anyway). Therefore, it will not be available when Makefile.prebuild is run for the first time (with the same setting of TARGET_CORE).
If nothing else, this adds to the confusion around Makefile.prebuild. It would be nice if the goal was to make the build process more transparent.

martin-frbg:
~~Well, it appears to work in principle. CI exposed just two obstacles - (1) there seems to be some intrinsics-using code that is not guarded by appropriate ifdefs for compiler version (probably previously hidden by the wrong check in getarch.c), and (2) reading from a file via `file <` is not available in  gmake versions earlier than 4.2~~
Meh.

e4t (changes requested):
Why not simply reuse c_check and call it with:
`ifeq (1, $(shell /usr/bin/env perl c_check - - $(CC) $(TARGET_FLAGS) $(CFLAGS) | awk  '/NO_AVX512.*/{split($0,a,"="); print a[2];}'))`
in Makefile.prebuild.
Adding more scripts that may potentially get out of sync with one another seems to be a recipe for errors down the road.

Or even simpler:
```
getarch : getarch.c cpuid.S dummy $(CPUIDEMU)
         avx512=$$(perl c_check - - gcc | grep NO_AVX512); \
         $(HOSTCC) $(HOST_CFLAGS) $(EXFLAGS) $${avx512:+-D$${avx512}} -o $(@F) getarch.c cpuid.S $(CPUIDEMU)
```

martin-frbg:
Right, thanks, indeed I would want this to be contained within Makefile.prebuild and no extra files. Just not sure if awk and split can be expected to be available everywhere, but a simple grep would probably do and is already used in Makefile.system.

e4t:
Right - in fact, after I wrote the first version of the comment  it occurred to me that I could also have used perl (which we need already) - I just felt too lazy to hack up the command line.

martin-frbg:
Hehe, that's cursed.

e4t:
While we are looking at Makefile.prebuild anyway: 
Looking into this, the reason why things get rebuild every time `make -f Makefile.prebuild ...` is run is the 'dummy' target. Maybe a:
```
.PHONY: dummy
```
would be useful to let make know about the 'phony-ness' of this target.
Also, this file contains a build rule for config.h where the build target doesn't match what's build. Example:
```
config.h : c_check f_check getarch
        perl ./c_check $(TARGET_MAKE) $(TARGET_CONF) $(CC) $(TARGET_FLAGS) $(CFLAGS)
      ...
```
Here, the real build target is  $(TARGET_CONF) not config.h. While the rules seem to work - thanks to the 'dummy' target - I wonder if there is a reason not to specify the build target explicitly:
```
$(TARGET_CONF):  c_check f_check getarch
        perl ./c_check $(TARGET_MAKE) $(TARGET_CONF) $(CC) $(TARGET_FLAGS) $(CFLAGS)
       ...
```
Unless there is a reason for not doing this I do not currently see.

martin-frbg:
Superseding with #3579 now.
@e4t I think you are right about using TARGET_CONF instead of config.h (though I expect the change would be purely cosmetic).
Have not looked at implications of changing the properties of the dummy target yet - anyway Makefile.prebuild should not execute all that often. (Needs to rebuild getarch for every DYNAMIC_ARCH cpu though, as the FORCE_... to tell it to fetch a particular dataset instead of autodetecting is a compile-time option)

e4t:
> Superseding with #3579 now. @e4t I think you are right about using TARGET_CONF instead of config.h (though I expect the change would be purely cosmetic). Have not looked at implications of changing the properties of the dummy target yet - anyway Makefile.prebuild should not execute all that often. (Needs to rebuild getarch for every DYNAMIC_ARCH cpu though, as the FORCE_... to tell it to fetch a particular dataset instead of autodetecting is a compile-time option)

@martin-frbg Indeed, DYNAMIC_ARCH is what I'm most interested in. Thus, my other patch creating separate Makefile_<arch>.conf and config_kernel_<ARCH>.h for improving debugging. I've got a patch for separating TARGET_CONF and adding a .phony target, Will look into this again.

One further note: the test for avx512 support is a bit of a 'one off', that will not scale well: Cooperlake has introduced support for avx512bf16, which is only supported since gcc10. The older SkylakeX and the newer Sapphirerapids may also have introduced restrictions beyond avx512. Thus, a much more fine-grained separation would have to be in place with the appropriate fallbacks (like Cooperlake->SkylakeX->Haswell). I'm afraid this will not scale well with the present structure.

martin-frbg:
Correct, the AVX512 test is neither expected nor intended to scale beyond ensuring basic AVX512 capability of the toolchain. It dates back to when the first Skylake Xeon models became available and some distributions still shipped with compilers as old as gcc4. The separation and fallback code is in cpuid_x86.c  (for single target) and driver/others/dynamic.c (dynamic_arch case) where relevant CPUID is queried.

e4t:
> Superseding with #3579 now. @e4t I think you are right about using TARGET_CONF instead of config.h (though I expect the change would be purely cosmetic). ...

I've not tried to pull-request a simple patch addressing both issues - unfortunately, the test suite fails for a build using cmake  on OSX with the error message:

> /Applications/Xcode_12.4.app/Contents/Developer/Toolchains/XcodeDefault.xctoolch
f951: Warning: Reading file '<stdin>' as free form
ld: warning: option -noall_load is obsolete and being ignored
Undefined symbols for architecture x86_64:
  "_xerbla_", referenced from:
         _sgemv_ in libopenblas.a(sgemv.c.o)
         _sger_ in libopenblas.a(sger.c.o)
          ...
     (maybe you meant: _xerbla_array_)
ld: symbol(s) not found for architecture x86_64

The patched file (Makefile.prebuild) should not be used at all here. I wonder why these builds are so fragile.

martin-frbg:
Yes that's totally unrelated - a workaround for an OSX/CMAKE/gfortran linking issue with response files that is only working half of the time. Restarting the build will see it pass so I consider it an improvement over the original case of DYNAMIC_ARCH builds not working on OSX with CMAKE at all.

e4t:
@martin-frbg makes sense. After re-triggering, the build has succeeded, now. Thanks!

PR #3573, opened 2022-03-17
Create independent kernel Makfile & configuration when building DYNAM…

e4t:
…IC_ARCH

- For 'classic' builds, generate separate config_kernel_<TARGET>.h,
  Makfile_<TARGET>.conf and getarch-<TARGET> files/binaries
- For cmake builds, generate separate getarch-<TARGET> binaries
for better debugging.


e4t:
This needs to be addressed differently.

PR #3575, opened 2022-03-20
Fix LAPACKE with 64-bit indexing on Windows.

mmuetzel:
Windows 64-bit uses LLP64 data model.

Without this change, unexpected errors or segfaults occur when configuring with `-DBINARY=64 -DINTERFACE64=1` on Windows.


mmuetzel:
Some checks were failing. But I'm not sure how they can be related to the proposed change in this PR.

martin-frbg:
Thanks. Looks like I somehow missed https://github.com/Reference-LAPACK/lapack/pull/624 (which defines it as int64_t and at least to me looks to be a simpler fix (?))

martin-frbg:
BTW test failures are indeed unrelated, seems MS-hosted Ci is in transition between 2019 and 2022 hosts a bit earlier than anticipated.

mmuetzel:
I wasn't sure which is the minimum C standard version that is required to compile OpenBLAS. But if you require C99 or newer anyway, using `int64_t` would indeed simplify that part of the code.

Edit: As an example, `stdint.h` wasn't available in MSVC before version 2010 afaict. Not sure if you'd like to support versions older than that.

mmuetzel:
That `WeirdNEC` macro is - well - weird. Is that how CBLAS switches between 32-bit and 64-bit indices?
Should the CBLAS headers also be changed? I.e., basically cherry-pick the commit you pointed to?

martin-frbg:
Good point - we do require C99 on x86_64 but that is mainly due to feature creep in recent kernels. Pretty sure attempting to support VS2010 or earlier would be pointless (ISTR we need 2013 or newer for various reasons).
The entire lapack-netlib tree is imported from Reference-LAPACK (with some functions subsequently overridden by what is in lapack) so minimal difference would be a plus. (Perhaps long-time members of the LAPACK team recall what the "weird NEC"
macro was about - might well be specific to some particularly exotic 1990s hardware)

mmuetzel:
I cherry-picked that commit to the lapack-netlib folder.
I didn't know what to make out of the following hunk:
``` diff
--- lapack-netlib/CMakeLists.txt
+++ lapack-netlib/CMakeLists.txt
@@ -53,7 +53,7 @@ if(BUILD_INDEX64)
   set(LAPACKELIB "lapacke64")
   set(TMGLIB "tmglib64")
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DWeirdNEC -DLAPACK_ILP64 -DHAVE_LAPACK_CONFIG_H")
-  set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -fdefault-integer-8")
+  set(FORTRAN_ILP TRUE)
 else()
   set(BLASLIB "blas")
   set(CBLASLIB "cblas")
```
I omitted it for now. Is that ok?

martin-frbg:
Looks to be only tangentially related - `guess the intent is to replace the compiler directive with a flag that gets translated into the appropriate compiler-specific idiom later but I´d need to check where Fortran_ILP came in (possibly another recent PR in their develop branch)

martin-frbg:
Ok, the logic for Fortran_ILP was introduced in that same PR (but I notice Reference-LAPACK merged it with an unresolved change request...) Need to make sure it does the right thing when built as part of OpenBLAS, as we override some of the  CMakeLists

mmuetzel:
IIUC (and please correct me if that is wrong), OpenBLAS doesn't use their cmake files to build lapack-netlib.

The changes regarding `FORTRAN_ILP` in that changeset feels kind of orthogonal to the rest. 
It might make sense to adopt this part of their changeset (maybe in a different PR):
``` diff
diff --git a/CMAKE/CheckLAPACKCompilerFlags.cmake b/CMAKE/CheckLAPACKCompilerFlags.cmake
index d727a1592..609bd4115 100644
--- a/CMAKE/CheckLAPACKCompilerFlags.cmake
+++ b/CMAKE/CheckLAPACKCompilerFlags.cmake
@@ -14,6 +14,19 @@ macro( CheckLAPACKCompilerFlags )
 
 set( FPE_EXIT FALSE )
 
+# FORTRAN ILP default
+if ( FORTRAN_ILP )
+    if( CMAKE_Fortran_COMPILER_ID STREQUAL "Intel" )
+        if ( WIN32 )
+            set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} /integer-size:64")
+        else ()
+            set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -integer-size 64")
+        endif()
+    else()
+        set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -fdefault-integer-8")
+    endif()
+endif()
+
 # GNU Fortran
 if( CMAKE_Fortran_COMPILER_ID STREQUAL "GNU" )
   if( "${CMAKE_Fortran_FLAGS}" MATCHES "-ffpe-trap=[izoupd]")
```

But apply it to somewhere inside `cmake/fc.cmake` (and replace `FORTRAN_ILP` with `INTERFACE64`).

IIUC, that change would be necessary to support Intel compilers with 64-bit indexing builds. (Feels unrelated to this change here...)

martin-frbg:
Yes lapack and lapacke CMakeLists are replaced by respective scripts in ~/cmake but need to confirm their toplevel CMakeLists is also ignored. (And if that is the case - fairly sure but no time to check right now - it would make sense to import their complete PR instead of cherrypicking __within__ it, to limit divergence from upstream to what is necessary. And then check if OpenBLAS needs to copy the equivalent of this Fortran_ILP thing or if it is already handled correctly elsewhere (I think it is)

mmuetzel:
I couldn't find that line in `lapack-netlib/CMakeLists.txt`:
`set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -fdefault-integer-8")`

So, I don't know what to do other than omitting that part of their commit. I took the remainder (including the part that I showed in the previous comment).

Maybe OpenBLAS' copy of `lapack-netlib` is missing some previous changes?

Edit: It looks like the entire `BUILD_INDEX64` from netlib isn't in OpenBLAS yet. (But that is probably unrelated to the motivating issue of this PR.)

mmuetzel:
> And then check if OpenBLAS needs to copy the equivalent of this Fortran_ILP thing or if it is already handled correctly elsewhere (I think it is)

I think that change would need to go here:
https://github.com/xianyi/OpenBLAS/blob/develop/cmake/fc.cmake#L70

mmuetzel:
I also cherry-picked the patch those changes were based on and also ported the part about 64-bit integers with Intel Fortran compilers to OpenBLAS' build rules.
I hope this is better.

I built a version with 32-bit and another one with 64-bit indexing on mingw with gcc and gfortran (both version 11.2.0). Both versions pass `ctest` for me.

Edit: I didn't drop parts of the upstream commits while cherry-picking this time.

PR #3578, opened 2022-03-22
WIP: remove perl build dependency

ioraff:


brada4:
Did not know bash-isms do work on Windoze

martin-frbg:
> Did not know bash-isms do work on Windoze

Probably requires either WSL or a native installation of git (for gitbash) ? (Not sure how much of an improvement over requiring perl it would be, also some Unix systems may not come with bash installed. Doing it all in a minimal POSIX sh seemed impossible last I looked into this)

ioraff:
> Did not know bash-isms do work on Windoze

There aren't any bashisms here

martin-frbg:
Impressive then. As you can tell I have not had time to look at it in detail, but the crashes in CI suggest there may be a problem with detecting the right representation of complex variables on some platforms. (There are some errors in tests like SASUM that I do not understand yet, may make sense to compare the configuration files created by the perl and sh versions in each case. And f_check line 95 seems to have a problem with obtaining the version number of the fortran compiler in some cases)

martin-frbg:
for f_check around line 95 I suspect the pattern matching should start with `v=${data##*GCC: *\) }` to catch all the vendor/distributor compilers that call themselves "GCC: (vendorname-somepattern) X.Y.Z" rather than the canonic "GCC: (GNU) X.Y.Z"

c_check
```diff
@@ -1,427 +1,381 @@
...
-
-if ($compiler eq "CLANG") {
-    $openmp = "-fopenmp";
-}
-
-if ($compiler eq "GCC" || $compiler eq "LSB") {
-    $openmp = "-fopenmp";
+{
+    data=$($compiler_name $flags -E ctest.c)
+} || {
```
brada4, line 27: Please use if [ everywhere or shorthands everywhere.

brada4 (changes requested):

c_check
martin-frbg: brada4 I am not convinced that unification would improve readability, both variants have their uses

brada4: this looks well... obfuscated, more verbose version as everywhere else would be more handy

ioraff: To my knowledge, the way I've used both variants here is pretty standard practice in shell scripts. Rewriting [this](https://github.com/xianyi/OpenBLAS/blob/7c146da155d628de02f17c4ffafb05c2af93a868/f_check#L69), for example, with only if-else would be a mess.

martin-frbg: To be clear I am happy with merging this as-is. I only want to try it on some machines not available in the current CI setup first (and maybe keep the perl scripts around somewhere as a fallback, but that can be looked into later)

martin-frbg:
quick update, sorry for the delay - Solaris sh does not like the `f="${flag#\}"` near line 295 of c_check (gives "bad substitution")
and it also looks as if the contents of the "old_flags" list has been replaced with the verbatim "echo" command at this point, rather than its output

martin-frbg:
looks like all `$( ... )` command substitutions need to be replaced with backticks syntax as well

martin-frbg:
(this is with Solaris 10 to be precise - host gcc210 in the GCC compile farm.)

brada4:
It fails on "common linux" too, say dash on Debian:
```
exports/gensymbol: 3620: [: Illegal number: 
c_check: 21: shift: can't shift that many
```
I dont think it is practical to introduce new bash dependency on !linux platforms.

martin-frbg:
Nobody plans to replace the perl dependency with a bash one, the question is simply if it is possible to rewrite the script in an idiom that is understood by all shells (and apparently POSIX is asking too much here, seems the old Suns use the original Bourne shell).

brada4:
Indeed, probably worth documenting what config files get generated there so that one willing can get through without bash or perl.

ioraff:
> It fails on "common linux" too, say dash on Debian:
> 
> ```
> exports/gensymbol: 3620: [: Illegal number: 
> c_check: 21: shift: can't shift that many
> ```
> 
> I dont think it is practical to introduce new bash dependency on !linux platforms.

Again, there aren't any bashisms here. Perhaps that was a bug with an older dash, as I have no problems running this with dash 0.5.11.5.

martin-frbg:
Still a WIP, but this appears to work with both SunOS' bourne shell and Linux' bash
[shellfiles.zip](https://github.com/xianyi/OpenBLAS/files/8557721/shellfiles.zip)

martin-frbg:
Sorry, it seems I cannot resolve the merge conflict (that I probably created a week ago with the riscv additions to the perl c_check) from my side. Can you please try if you can override it with a force-push again ?

ioraff:
well, I've clearly broken this. not sure what to do now.

martin-frbg:
hmm. my low-tech idea would be to create a new branch from current `develop`, copy the changed files from the stalled branch over and do a new PR ?

martin-frbg:
(I'm not sure I understand what exactly went wrong, except that a force-push was probably a very bad suggestion). I may still have a .diff version of this PR from happier times

martin-frbg:
This is what it looked like on April 10 (which is what I based my Apr 24 comments on - https://github.com/xianyi/OpenBLAS/pull/3578#issuecomment-1107876155
[3578.diff.gz](https://github.com/xianyi/OpenBLAS/files/8681548/3578.diff.gz)
and it looks like the links to your individual changes since then ("readlink -f is not POSIX" et al.) stlll work (and so does the usual github magic whereby appending ".diff" to the URL of the commit view turns it into a raw context diff suitable for patching)

martin-frbg:
I have now checked that all individual commits preceding "use backticks" are already contained in the 3578.diff of April 10, so only this one and my attempt to update your c_check today (minus the typos) would need to be reapplied on top of it

ioraff:
A force-push was probably the right idea. I just borked the rebase.

The diff fails to apply to the current `develop` `HEAD` for me. Not sure how to go about debugging it since the changes are basically all in one hunk. Might just rewrite at this point.

martin-frbg:
Hmm, I'm seeing a patch failure in Makefile.prebuild, and the major one in c_check. Easiest solution for the latter would probably be to check out an earlier stage of `develop`, i.e. from April 10 or thereabouts, apply the patch there and copy the c_check file into the new PR branch

martin-frbg:
This here is from applying 3578.diff to an April 17 checkout of `develop` (18b19d13 , or right after merging #3605)
[c_check.txt](https://github.com/xianyi/OpenBLAS/files/8682280/c_check.txt)
(and the diff for Makefile.prebuild was only about removing the "perl" command from script invocations)

brada4:
> 0.5.11.5

It has to work on existing platforms, including RHEL6, SLES 12, DEBIAN 9, all BSD-s etc.

PR #3587, opened 2022-03-28
Use CC and full command line instead of hard-coding gcc for AVX512 ch…

e4t:
…ecking

Hard-coding gcc may not provide incorrect results when a different compiler
for the target build is used. To remain in sync with the main call to c_check,
pass the full command line.


martin-frbg:
Indeed, that shorthand was left over from local testing.

PR #3589, opened 2022-03-29
Do not include symbols defined in driver/others/parameter.c in DYNAMI…

e4t:
…C_ARCH

driver/others/parameter.c does not get build during DYNAMIC_ARCH, thus,
do not declare its symbols. This will make the build fail early and in
an obvious way if functions are trying to use these symbols.

This helps to catch issues like the one fixed in PR #3547 / PR #3580 earlier
as they become obvious when building the library already.

### DIFF
--- a/common_macro.h
+++ b/common_macro.h
@@ -2610,8 +2610,9 @@
 #endif
 
 #ifndef ASSEMBLER
-#if defined(ARCH_X86) || defined(ARCH_X86_64) || defined(ARCH_IA64) || defined(ARCH_MIPS64) || defined(ARCH_ARM64)\
-|| defined(ARCH_LOONGARCH64) || defined(ARCH_E2K)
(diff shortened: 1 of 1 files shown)


martin-frbg:
I like to think that that particular problem was a one-off (more related to the fairly recent addition of bfloat16) rather than a recurring problem with DYNAMIC_ARCH additions, but I guess it cannot hurt to have this guard. (And at some point we may want to add fp16 or other datatypes, creating a new opportunity to repeat the mistake)

e4t:
@martin-frbg - thanks for accepting this! I believe the problem mentioned was introduced with the introduction of the Intel Cooperlake support - at least this is where I had bisected it to. Thus I was assuming that the introduction of further optimized kernels for other arches may potentially risk the same.

PR #3590, opened 2022-03-31
Add action for MSYS2 builds.

mmuetzel:
This adds CI jobs that compile and test OpenBLAS in different build environments that are provided by msys2 on Windows.
It uses cmake/ninja for configuration and building. GNU autotools are notably slower compared to them on Windows for some reason.
It already adds the configuration switch `-DC_LAPACK=ON` for the jobs that run with the CLANG64 build environment. So, it should pick up the changes from #3539 when they are committed. That switch is ignored currently.

I hope this can be useful.


martin-frbg:
Thank you

PR #3600, opened 2022-04-08
Disable flang (over-)optimizations in BLAS tests

martin-frbg:
CMAKE's default of -O3 for "Release" builds causes spurious test failures in ?TRMM and ?TRMM that are not reflected in the corresponding CBLAS tests, and go away when just the BLAS3 tests themselves are compiled at -O2. Fixes test failures seen both locally and in Azure CI


mmuetzel:
I added a comment to the commit. But I don't know if that gets sent anywhere. Repeating here. (Sorry for the noise if you've already been noticed):

Should that be this?
``` diff
-	set(CMAKE_Fortran_FLAGS "$CMAKE_Fortran_FLAGS -O2")
+	set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -O2")
```

PR #3601, opened 2022-04-09
Consolidate actions on GitHub runners.

mmuetzel:
Re-organize build matrix for Ubuntu and MacOS runners.
Don't start runners that don't do anything.
Run tests.

This also adds a job that tests cmake with classic flang on Ubuntu. That wasn't exercised before (only GNU make with flang). It could be excluded if that test isn't necessary.

### DIFF
--- a/.github/workflows/dynamic_arch.yml
+++ b/.github/workflows/dynamic_arch.yml
@@ -5,27 +5,20 @@ on: [push, pull_request]
 jobs:
   build:
     runs-on: ${{ matrix.os }}
+
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
         fortran: [gfortran, flang]
         build: [cmake, make]
+        exclude:
+          - os: macos-latest
+            fortran: flang
+
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
-
-      - name: Compilation cache
-        uses: actions/cache@v2
-        with:
-          path: ~/.ccache
-          # We include the commit sha in the cache key, as new cache entries are
-          # only created if there is no existing entry for the key yet.
-          key: ${{ runner.os }}-ccache-${{ github.sha }}
-          # Restore any ccache cache entry, if none for
-          # ${{ runner.os }}-ccache-${{ github.sha }} exists
-          restore-keys: |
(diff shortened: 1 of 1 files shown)


mmuetzel:
ubuntu-latest,flang,cmake failed:
```
2022-04-09T17:32:30.3148258Z [ 91%] Building Fortran object test/CMakeFiles/sblat1.dir/sblat1.f.o
2022-04-09T17:32:30.3152940Z cd /home/runner/work/OpenBLAS/OpenBLAS/build/test && ccache /usr/bin/flang  -I/home/runner/work/OpenBLAS/OpenBLAS/lapack-netlib/LAPACKE/include -I/home/runner/work/OpenBLAS/OpenBLAS -I/home/runner/work/OpenBLAS/OpenBLAS/build MAKE_Fortran_FLAGS -O2 -O3 -DNDEBUG -c /home/runner/work/OpenBLAS/OpenBLAS/test/sblat1.f -o CMakeFiles/sblat1.dir/sblat1.f.o
2022-04-09T17:32:30.3213601Z flang: error: no such file or directory: 'MAKE_Fortran_FLAGS'
```
I didn't see that error in my fork...

martin-frbg:
Clearly my fault, though curious how/why it seems to have split the $CMAKE... at $C

martin-frbg:
Wondering why the flang/make job from this is failing now that C_LAPACK has landed - test programs from lapack-netlib/INSTALL appear to be getting linked by a ccache run of flang, instead of the C compiler, leading to a duplicate definition of "main". For this to happen, C_LAPACK would appear to have to be both defined and undefined at the same time. Spurious invocation of ccache with contents from preceding run ?

mmuetzel:
Hmmm...
Looking at this again I'm wondering if those `make` commands are even supposed to run the tests generally. 
I was working with the `cmake` build system of OpenBLAS mainly before this. And most of the autotools projects I've come across have a `configure` scripts that "store" configure options in their generated Makefile.
Do the `make` flags need to be repeated on each invocation?
Which commands should be run to test if the built version of OpenBLAS is working correctly? (Kind of as a drop-in for `ctest`.)

mmuetzel:
Might be avoided with #3604.

PR #3609, opened 2022-04-21
Update LAPACK/LAPACKE to Reference-LAPACK 3.10.1

### LINKED ISSUE #2154: Hidden ABI problems in calls from C to FORTRAN
As noted (by a gfortran developer) in https://github.com/Reference-LAPACK/lapack/issues/339, the de facto ABI for calling FORTRAN code from C requires the lengths of any string arguments to be passed as additional, hidden arguments at the end of the argument list. Changes in gcc from about gcc8 onwards (higher patchlevels of gcc7 might be equally affected) have apparently enforced this for single-character arguments as well, which affects most LAPACK(E) functions through their use of flags like N/T for (non)transpose or U/L for upper/lower triangular matrix. This has already led to misbehaviour in LAPACK calls from codes like Armadillo https://github.com/RcppCore/RcppArmadillo/issues/254 - as noted there, a workaround is to compile the FORTRAN code with -fno-optimize-sibling-calls and upcoming releases of gcc will incorporate a similar workaround for the time being.
This still leaves existing installations with unpatched gcc 7.x, 8.x, 9.x vulnerable to the problem, so while (most of) an eventual "proper" fix would have to happen upstream in the netlib LAPACKE we need to fix this locally as well. (It appears to be still unclear how compilers other than gfortran handle the length requirement for single-character strings, but it has been argued that current code is at best relying on undefined behaviour.)

### LINKED ISSUE #3495: Update LAPACK to 3.10
Just opening an issue to track since there doesn't seem to be one. Quoting from #2320:
> @martin-frbg: As usual, this is not simply a drop-in replacement here as OpenBLAS uses modified Makefiles (and own CMakelists) to replace some functions with optimized implementations. There are also some platform-specific changes and header rearrangements that need to be merged during an update.

martin-frbg:
fixes #3495 and fixes #2154


h-vetinari:
Hey @martin-frbg, do you think it would make sense to include this PR in 0.3.21?

martin-frbg:
I remember seeing a handful of new failures in the LAPACK testsuite with this, which is why it  is not merged yet. Seems some of the new algorithms are a lot more sensitive to numerical accuracy of the underlying BLAS than before.

martin-frbg:
Current state on Haswell for reference (used to be just 1 each in REAL/COMPLEX due to ?TFSM) - additional errors are in SGS/CGV/ZGV (with significant error values, in the 1.e6 and 1.e16 range). ARMV8 is clean but with a low test count.

```
                        -->   LAPACK TESTING SUMMARY  <--
SUMMARY                 nb test run     numerical error         other error  
================        ===========     =================       ================  
REAL                    1308531         2       (0.000%)        1       (0.000%)
DOUBLE PRECISION        1318689         0       (0.000%)        0       (0.000%)
COMPLEX                 768751          5       (0.001%)        0       (0.000%)
COMPLEX16               776246          21      (0.003%)        0       (0.000%)

--> ALL PRECISIONS      4172217         28      (0.001%)        1       (0.000%)
```

brada4:
Well FMA and parallel accumulators almost guarantee different result. I suppose those tests impose exact result.

martin-frbg:
Tests are not meant to impose exact results - in fact the input files define thresholds and the netlib faq explains that even further deviations in about the same order of magnitude can be expected to arise from harmless numerical differences. I guess there is a risk that more involved calculations, e.g. iterative ones, may be derailed to give a perturbed result. In any case, errors of 1e+16 will warrant a closer look.

brada4:
It is to look at each test individually why rounding particularities do not cancel out correctly.

PR #3611, opened 2022-04-23
Add OSX docker build to Azure CI

martin-frbg:


martin-frbg:
Not as useful as I had hoped - apparently the Microsoft-hosted Macs in Azure CI are all 2nd generation MacPros with IvyBridge Xeons, so AVX only.

PR #3612, opened 2022-04-24
build: minor fixes to build on windows with make

niyas-sait:
This patch contains the following fixes

- Fix to build without PIC flag. 

Add `PIC` flag only if `NEED_PIC` is defined as `1`

- Define LAPACK_COMPLEX_STRUCTURE for windows. 

CMake defines this for windows (system.cmake, line 576) but Makefile seems to skip it. Without this flag following error is observed

```
lapacke_make_complex_float.c(49,16): error: invalid operands to binary expression ('float' and '_Fcomplex'
      (aka 'struct _C_float_complex'))
    z = re + im*I;
```

The changes have been tested on Windows(Arm64) target.


mmuetzel:
I might be wrong here. Doesn't ~(modern)~ MinGW-w64 have support for C99 `_Complex`? Should `LAPACK_COMPLEX_STRUCTURE` be set only for MSVC (and not always on Windows)?
The cmake rules could also be changed probably to reflect that.

martin-frbg:
Interesting find, thanks. This inconsistency between gmake and cmake build settings appears to have been present for years,(probably not least because the Reference-LAPACK sources rely primarily on cmake and there is no provision for this setting in their make.inc. While the LAPACKE default _is_ C99, the original LAPACKE CMakelists.txt does set LAPACKE_COMPLEX_STRUCTURE for any "WIN32 and not UNIX" situation. I guess this is simpler and safer than trying to detect the C99 capabilities of the individual toolchain at build time.

mmuetzel:
IIUC, `_Complex` is supported in mingw-w64 since ~at least 15 years~ always. ~It is probably safe to assume that anyone compiling OpenBLAS currently is on a version newer than that.~

See:
https://github.com/Alexpux/mingw-w64/blame/master/mingw-w64-headers/crt/complex.h

I think the only modern compiler that doesn't implement C99 `_Complex` is MSVC.
I don't know if there are any performance considerations. But I'd guess it might be best to use the "native" complex type for all others.

PR #3613, opened 2022-04-26
Fix riscv64 detect

### LINKED ISSUE #3606: The compiler doesn't support lp64v, can I compile the riscv version with lp64?
The compiler doesn't support lp64v, can I compile the riscv version with lp64?

RevySR:
Fix https://github.com/xianyi/OpenBLAS/issues/3606

When CORE is empty, use -march=loongson3a. Fix it.

And Fix riscv64 detect

### DIFF
--- a/Makefile.system
+++ b/Makefile.system
@@ -847,7 +847,7 @@ CCOMMON_OPT += -mabi=32
 BINARY_DEFINED = 1
 endif
 
-ifeq ($(CORE), $(filter $(CORE),LOONGSON3R3 LOONGSON3R4))
+ifneq (, $(filter $(CORE),LOONGSON3R3 LOONGSON3R4))
 CCOMMON_OPT += -march=loongson3a
 FCOMMON_OPT += -march=loongson3a
 endif
--- a/cpuid_riscv64.c
+++ b/cpuid_riscv64.c
@@ -1,5 +1,5 @@
 /*****************************************************************************
-Copyright (c) 2011-2014, The OpenBLAS Project
+Copyright (c) 2011-2022, The OpenBLAS Project
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
@@ -13,9 +13,9 @@ met:
       notice, this list of conditions and the following disclaimer in
       the documentation and/or other materials provided with the
       distribution.
-   3. Neither the name of the OpenBLAS project nor the names of 
-      its contributors may be used to endorse or promote products 
-      derived from this software without specific prior written 
+   3. Neither the name of the OpenBLAS project nor the names of
+      its contributors may be used to endorse or promote products
+      derived from this software without specific prior written
       permission.
 
 THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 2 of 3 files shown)


RevySR:
build failed on gentoo
https://bugs.gentoo.org/837806

martin-frbg:
Thank you. (I am not sure if the change to Makefile.system is really necessary when your fix for getarch.c already ensures that CORE gets set correctly ?) CPU autodetection for RISCV simply is not implemented, and I assume most users will be cross-compiling rather than compiling on the device itself anyway (?)

RevySR:
> Thank you. (I am not sure if the change to Makefile.system is really necessary when your fix for getarch.c already ensures that CORE gets set correctly ?) CPU autodetection for RISCV simply is not implemented, and I assume most users will be cross-compiling rather than compiling on the device itself anyway (?)

I'm trying to fix a distribution recognition error. In multi-arch distribution, auto detect 'arch generic' is needed.

RevySR:
> Thank you. (I am not sure if the change to Makefile.system is really necessary when your fix for getarch.c already ensures that CORE gets set correctly ?) CPU autodetection for RISCV simply is not implemented, and I assume most users will be cross-compiling rather than compiling on the device itself anyway (?)

https://github.com/xianyi/OpenBLAS/blob/ee4244538cc8c2ec1ccb6d0cd6a799bb9bd75a63/getarch.c#L1145
Quote above, in FORCE_RISCV corename is RISCV64_GENERIC

https://github.com/xianyi/OpenBLAS/blob/ee4244538cc8c2ec1ccb6d0cd6a799bb9bd75a63/cpuid_riscv64.c#L76-L79
Unknown is typo.
Implementation continues in the wrong way

https://github.com/xianyi/OpenBLAS/blob/5d2f451d09516d71e9a454b87472de5ec7d5e92f/TargetList.txt#L119-L121

In 'TARGET=C910V' sense
https://github.com/xianyi/OpenBLAS/blob/ee4244538cc8c2ec1ccb6d0cd6a799bb9bd75a63/getarch.c#L1574-L1586

In 'TARGET=RISCV64_GENERIC' sense
https://github.com/xianyi/OpenBLAS/blob/ee4244538cc8c2ec1ccb6d0cd6a799bb9bd75a63/getarch.c#L1135-L1147

martin-frbg:
Yeah, the "UNKOWN" without the "N" is a historic typo dating back to the original GotoBLAS. I sort-of agree that an automatic fallback to RISCV64_GENERIC would be useful in your context. On the other hand, the drawback of always autodetecting as this would be that builds on C910V and compatibles would end up with the same unoptimized codebase as well. Unfortunately I am not familiar with the current situation regarding RISCV - is there some sort of universal CPUID instruction for determining
make/model, or even a higher-level interface like /proc/cpuinfo on Linux, or is everything still depending on human eyes looking at board specifications ?

RevySR:
> Yeah, the "UNKOWN" without the "N" is a historic typo dating back to the original GotoBLAS. I sort-of agree that an automatic fallback to RISCV64_GENERIC would be useful in your context. On the other hand, the drawback of always autodetecting as this would be that builds on C910V and compatibles would end up with the same unoptimized codebase as well. Unfortunately I am not familiar with the current situation regarding RISCV - is there some sort of universal CPUID instruction for determining make/model, or even a higher-level interface like /proc/cpuinfo on Linux, or is everything still depending on human eyes looking at board specifications ?

The C910V is not publicly available for sale, so I can't do the relevant tests

martin-frbg:
Oh, I was not aware of that. From what i could find, maybe it is possible to read the `isa` entry from /proc/cpuinfo and just return C910V when the string is rv64imafdv or rv64imafdcv (indicating vector support) or RISCV64_GENERIC when not.

brada4:
it is actually bitmask register from CPUID CSR, available in unprivileged modes.

PR #3624, opened 2022-05-12
rewrite perl scripts in universal shell

ioraff:


martin-frbg:
CI failure on the manylinux image looks as if the new c_check failed to set NO_AVX512, did you change anything compared to the previous iterations ?

martin-frbg:
Looks like something is broken in the Azure/docker environment and the "manylinux" job is no longer passing the DYNAMIC_ARCH=1 etc. arguments to make, resulting in a build for whatever host architecture it is running on (which on Azure alternates between Haswell and SkylakeX generation systems). I cannot reproduce this misbehaviour locally with docker, guess I will need to rerun this docker job on SkylakeX hardware to be completely sure it behaves as expected with AVX512  hardware and an ancient compiler

martin-frbg:
Right - the docker build would run to completion with DYNAMIC_ARCH=1 on SkylakeX, but it fails when this option does not get
passed. And the source of this problem is simply that your Makefile.prebuild still has two c_check calls that have not been adapted - they still use perl, and moreover they use "-" for the output filenames in order to parse what would be written to Makefile.conf - your c_check does not handle this yet and would create a regular file.
[fixedfiles.zip](https://github.com/xianyi/OpenBLAS/files/8721134/fixedfiles.zip)

PR #3626, opened 2022-05-13
Fix build on new-world LoongArch

xen0n:
This is fix for https://bugs.gentoo.org/844013. The LoongArch port is tested only on old-world systems, and fails to build on new-world distros such as Gentoo.

I can successfully build and test OpenBLAS with the changes applied:

```

  OS               ... Linux             
  Architecture     ... loongarch64               
  BINARY           ... 64bit                 
  C compiler       ... GCC  (cmd & version : cc (Gentoo 12.1.0 p5) 12.1.0)
  Fortran compiler ... GFORTRAN  (cmd & version : GNU Fortran (Gentoo 12.1.0 p5) 12.1.0)
  Library Name     ... libopenblas_loongson3r5p-r0.3.20.a (Multi-threading; Max num-threads is 4)
```

@gxw-loongson: Please test on the old world to make sure I don't regress your usecase.

### DIFF
--- a/c_check
+++ b/c_check
@@ -112,7 +112,7 @@ case "$architecture" in
     	defined=1
     	;;
     arm|arm64) defined=1 ;;
-    zarch|e2k|alpha|ia64|riscv64|loonarch64)
+    zarch|e2k|alpha|ia64|riscv64|loongarch64)
     	defined=1
     	BINARY=64
     	;;
@@ -240,6 +240,29 @@ if [ "$architecture" = "riscv64" ]; then
     rm -rf "$tmpd"
 fi
 
+no_lasx=0
+if [ "$architecture" = "loongarch64" ]; then
+    tmpd=`mktemp -d`
+    tmpf="$tmpd/a.c"
+    # Old-world assembly flavor: LASX registers named "$xrNN", different
+    # from the ISA manual which suggests "$xNN". This is the flavor we
+    # currently support.
+    #
+    # As the LASX ISA manual is not out yet, we cannot predict what the
+    # new-world flavor would look like, so do not probe flavor for now.
+    # The compiler flags are also unsuitable for new-world gcc.
+    code='"xvld $xr0, $a0, 0\n"'
+    printf "int main(void){ __asm__ volatile(%s); }\n" "$code" >> "$tmpf"
+    args=" -march=loongarch64 -mabi=lp64 -mlasx -c -o $tmpf.o $tmpf"
+    no_lasx=0
+    {
+        $compiler_name $flags $args >/dev/null 2>&1
+    } || {
+        no_lasx=1
+    }
+    rm -rf "$tmpd"
+fi
+
 c11_atomics=0
 case "$data" in
     *HAVE_C11*)
@@ -350,6 +373,7 @@ done
 
  [ "$makefile" = "-" ] && {
     [ "$no_rv64gv" -eq 1 ] && printf "NO_RV64GV=1\n"
+    [ "$no_lasx" -eq 1 ] && printf "NO_LASX=1\n"
     [ "$no_avx512" -eq 1 ] && printf "NO_AVX512=1\n"
     [ "$no_avx2" -eq 1 ] && printf "NO_AVX2=1\n"
     [ "$oldgcc" -eq 1 ] && printf "OLDGCC=1\n"
@@ -380,6 +404,7 @@ done
         printf "MSA_FLAGS=%s\n" "$msa_flags"
     }
     [ "$no_rv64gv" -eq 1 ] && printf "NO_RV64GV=1\n"
+    [ "$no_lasx" -eq 1 ] && printf "NO_LASX=1\n"
     [ "$no_avx512" -eq 1 ] && printf "NO_AVX512=1\n"
     [ "$no_avx2" -eq 1 ] && printf "NO_AVX2=1\n"
     [ "$oldgcc" -eq 1 ] && printf "OLDGCC=1\n"
--- a/c_check
+++ b/c_check
@@ -240,8 +240,19 @@ if [ "$architecture" = "riscv64" ]; then
     rm -rf "$tmpd"
 fi
 
+loongarch_cc_flavor=""
 no_lasx=0
 if [ "$architecture" = "loongarch64" ]; then
+    # Currently there are only 2 flavors: new-world and old-world.
+    # They differ in ABI names accepted, so check it to avoid creating temp
+    # files.
+    loongarch_cc_flavor="NEW"
+    {
+        $compiler_name -mabi=lp64d -E -o - /dev/null >/dev/null 2>&1
+    } || {
+        loongarch_cc_flavor="OLD"
+    }
+
     tmpd=`mktemp -d`
     tmpf="$tmpd/a.c"
     # Old-world assembly flavor: LASX registers named "$xrNN", different
@@ -260,6 +271,7 @@ if [ "$architecture" = "loongarch64" ]; then
     } || {
         no_lasx=1
     }
+
     rm -rf "$tmpd"
 fi
 
@@ -377,6 +389,7 @@ done
     [ "$no_avx512" -eq 1 ] && printf "NO_AVX512=1\n"
     [ "$no_avx2" -eq 1 ] && printf "NO_AVX2=1\n"
     [ "$oldgcc" -eq 1 ] && printf "OLDGCC=1\n"
+    [ -n "$loongarch_cc_flavor" ] && printf "LOONGARCH_CC_FLAVOR=%s\n" "$loongarch_cc_flavor"
     exit 0
 }
 
@@ -408,6 +421,7 @@ done
     [ "$no_avx512" -eq 1 ] && printf "NO_AVX512=1\n"
     [ "$no_avx2" -eq 1 ] && printf "NO_AVX2=1\n"
     [ "$oldgcc" -eq 1 ] && printf "OLDGCC=1\n"
+    [ -n "$loongarch_cc_flavor" ] && printf "LOONGARCH_CC_FLAVOR=%s\n" "$loongarch_cc_flavor"
 } >> "$makefile"
 
 os=`echo "$os" | tr '[[:lower:]]' '[[:upper:]]'/ `
--- a/kernel/loongarch64/KERNEL.LOONGSON3R5
+++ b/kernel/loongarch64/KERNEL.LOONGSON3R5
@@ -1,3 +1,7 @@
+ifeq ($(NO_LASX), 1)
+# No LASX support in compiler, unable to consume the optimized version
+# leave out DGEMMKERNEL to pull in the generic version
+else
 DGEMMKERNEL    = dgemm_kernel_16x4.S
 DGEMMINCOPY    = dgemm_ncopy_16.S
 DGEMMITCOPY    = dgemm_tcopy_16.S
@@ -7,6 +11,7 @@ DGEMMINCOPYOBJ = dgemm_incopy$(TSUFFIX).$(SUFFIX)
 DGEMMITCOPYOBJ = dgemm_itcopy$(TSUFFIX).$(SUFFIX)
 DGEMMONCOPYOBJ = dgemm_oncopy$(TSUFFIX).$(SUFFIX)
 DGEMMOTCOPYOBJ = dgemm_otcopy$(TSUFFIX).$(SUFFIX)
+endif
 
 DTRSMKERNEL_LN  = ../generic/trsm_kernel_LN.c
 DTRSMKERNEL_LT  = ../generic/trsm_kernel_LT.c
--- a/Makefile.loongarch64
+++ b/Makefile.loongarch64
@@ -1,4 +1,13 @@
 ifeq ($(CORE), LOONGSON3R5)
-CCOMMON_OPT += -march=loongarch64 -mabi=lp64
-FCOMMON_OPT += -march=loongarch64 -mabi=lp64
+CCOMMON_OPT += -march=loongarch64
+FCOMMON_OPT += -march=loongarch64
+
+ifeq ($(LOONGARCH_CC_FLAVOR), NEW)
+CCOMMON_OPT += -mabi=lp64d
+FCOMMON_OPT += -mabi=lp64d
+endif
+ifeq ($(LOONGARCH_CC_FLAVOR), OLD)
+CCOMMON_OPT += -mabi=lp64
+FCOMMON_OPT += -mabi=lp64
+endif
 endif
--- a/Makefile.loongarch64
+++ b/Makefile.loongarch64
@@ -1,3 +1,4 @@
-ifdef BINARY64
-else
+ifeq ($(CORE), LOONGSON3R5)
+CCOMMON_OPT += -march=loongarch64 -mabi=lp64
+FCOMMON_OPT += -march=loongarch64 -mabi=lp64
 endif
--- a/Makefile.system
+++ b/Makefile.system
@@ -1702,6 +1702,7 @@ export TARGET_CORE
 export NO_AVX512
 export NO_AVX2
 export BUILD_BFLOAT16
+export NO_LASX
 
 export SBGEMM_UNROLL_M
 export SBGEMM_UNROLL_N
--- a/Makefile.system
+++ b/Makefile.system
@@ -1702,6 +1702,7 @@ export TARGET_CORE
 export NO_AVX512
 export NO_AVX2
 export BUILD_BFLOAT16
+export LOONGARCH_CC_FLAVOR
 export NO_LASX
 
 export SBGEMM_UNROLL_M
--- a/Makefile.system
+++ b/Makefile.system
@@ -895,11 +895,6 @@ ifeq ($(OSNAME), AIX)
 BINARY_DEFINED = 1
 endif
 
-ifeq ($(ARCH), loongarch64)
-CCOMMON_OPT += -march=loongarch64 -mabi=lp64
-FCOMMON_OPT += -march=loongarch64 -mabi=lp64
-endif
-
 endif
 
 ifndef BINARY_DEFINED


martin-frbg:
I do not get the background of the "old world vs. new world" systems, is this something that is not simply tied to specific (vendor) gcc/toolchain versions ? (Otherwise at least the difference in `-mabi` arguments could be handled via the existing GCCVERSIONGE11 or equivalent variable, without adding compile tests to c_check)

martin-frbg:
So background appears to be https://sourceware.org/bugzilla/show_bug.cgi?id=29059 a change in the GNU gcc and assembler ABI argument naming that went live with gcc-12.1 and accompanying binutils 2.38

martin-frbg:
Testing now on the Loongson-sponsored machine in the gcc compile farm. The gcc version test I outlined does solve the build problem with gcc12, but there may be more amiss as the LAPACK testsuite appears to go into an endless loop. (Apparently inside the plain-C sgemm_otcopy called from sgetrf, and irrespective of optimization level or number of threads)
(The same machine also has a recent clang installed, but the build fails some of the utests at any level above `-O0` and eventually segfaults there.)

XiWeiGu:
@xen0n Thx, successfully build on old world.

kernel/loongarch64/KERNEL.LOONGSON3R5
```diff
@@ -1,3 +1,7 @@
+ifeq ($(NO_LASX), 1)
```
XiWeiGu, line 1: There's a slight problem. If `NO_LASX==1`,  We shoule resize `DGEMM_DEFAULT_UNROLL_M` from 16 to 2, `DGEMM_DEFAULT_UNROLL_N` from 4 to 8 in file param.h.

XiWeiGu:
> Testing now on the Loongson-sponsored machine in the gcc compile farm. The gcc version test I outlined does solve the build problem with gcc12, but there may be more amiss as the LAPACK testsuite appears to go into an endless loop. (Apparently inside the plain-C sgemm_otcopy called from sgetrf, and irrespective of optimization level or number of threads)
> (The same machine also has a recent clang installed, but the build fails some of the utests at any level above `-O0` and eventually segfaults there.)

Please privode more details about the Loongson-sponsored machine and the amiss, thx.

kernel/loongarch64/KERNEL.LOONGSON3R5
xen0n: > There's a slight problem. If `NO_LASX==1`,  We shoule resize `DGEMM_DEFAULT_UNROLL_M` from 16 to 2, `DGEMM_DEFAULT_UNROLL_N` from 4 to 8 in file param.h.

Okay, I'll try to take care of this in the next version.

martin-frbg:
> Please privode more details about the Loongson-sponsored machine and the amiss, thx.

host gcc400.fsffrance.org (see https://cfarm.tetaneutral.net/ for GCC compile farm overview) features a quad-core 3A5000
running "Cross-LinuxFromScratch" with gcc 12. I realize now that it is only a gcc development snapshot from early december (12.0.0 20211202) rather than the recent official release, so this may or may not have anything to do with the LAPACK test looping. (If you do `make && make lapack-test`, summary output stalls after "Testing REAL Constrained Linear Least Squares routines". If you run the stalling `LIN/xlintsts <stest.in` in lapack-netlib/TESTING directiy, it appears to be looping in gemm_tcopy_2.c - or may be just unexpectedly slow.)

c_check
```diff
@@ -293,6 +293,21 @@ if (($architecture eq "riscv64")) {
...
+    $args = " -mabi=lp64d -E -o - /dev/null";
+    my @cmd = ("$compiler_name $flags $args >/dev/null 2>/dev/null");
+    system(@cmd) == 0;
+    if ($? != 0) {
+        $loongarch_cc_flavor = "OLD";
+    } else {
+        $loongarch_cc_flavor = "NEW";
+    }
+}
+
```
yinshiyou, line 255: As '-mabi=lp64' is only supported by old binutils (2.38) and not supported by compiler gcc. 
This check may never set flavor to "OLD" unless someone is using a transitional version compiler inside Loongson.
I suggest changing it directly to '-mabi=lp64d' to avoid confusion.

martin-frbg:
I just noticed that cpuid_loongson64.c already queries the LASX flag (to decide between reporting LOONGSON3R5 or generic UNKNOWN). If this was printed by the get_cpuconfig() function there (like it is done e.g. for HAVE_AVX2 in cpuid_x86.c) and similarly exported from Makefile.system I believe we would not need the extra test in c_check.

c_check
```diff
@@ -293,6 +293,37 @@ if (($architecture eq "riscv64")) {
...
+        system(@cmd) == 0;
+        if ($? != 0) {
+            $no_lasx = 1;
+        } else {
+            $no_lasx = 0;
+        }
+        unlink("$tmpf.o");
+    }
+}
+
```
yinshiyou, line 255: Maybe we can use intrinsics here to avoid distinguising '$xNN' or '$xrNN', and so that we don't need to talk about the new world and the old world.

XiWeiGu:
> I just noticed that cpuid_loongson64.c already queries the LASX flag (to decide between reporting LOONGSON3R5 or generic UNKNOWN). If this was printed by the get_cpuconfig() function there (like it is done e.g. for HAVE_AVX2 in cpuid_x86.c) and similarly exported from Makefile.system I believe we would not need the extra test in c_check.

cpuid_loongson64.c only queries cpu support LASX or not. We also need extra test in c_check to query compiler support LASX or not(like AVX2).

martin-frbg:
> > I just noticed that cpuid_loongson64.c already queries the LASX flag (to decide between reporting LOONGSON3R5 or generic UNKNOWN). If this was printed by the get_cpuconfig() function there (like it is done e.g. for HAVE_AVX2 in cpuid_x86.c) and similarly exported from Makefile.system I believe we would not need the extra test in c_check.
> 
> cpuid_loongson64.c only queries cpu support LASX or not. We also need extra test in c_check to query compiler support LASX or not(like AVX2).

So this is uncoupled from compiler version ? My impression so far (and supported by yinshiyou's comments) was that this is a
question of "is this a mainstream gcc-12 or an older (and probably internal) version with Loongson patches"

yinshiyou:
> > > I just noticed that cpuid_loongson64.c already queries the LASX flag (to decide between reporting LOONGSON3R5 or generic UNKNOWN). If this was printed by the get_cpuconfig() function there (like it is done e.g. for HAVE_AVX2 in cpuid_x86.c) and similarly exported from Makefile.system I believe we would not need the extra test in c_check.
> > 
> > 
> > cpuid_loongson64.c only queries cpu support LASX or not. We also need extra test in c_check to query compiler support LASX or not(like AVX2).
> 
> So this is uncoupled from compiler version ? My impression so far (and supported by yinshiyou's comments) was that this is a question of "is this a mainstream gcc-12 or an older (and probably internal) version with Loongson patches"

Yes, I think the main purpose of this PR is to distinguish between different versions of the compilation toolchain (new world and old world xenOn said), then xenOn added a check in c_check (by the way) to test if the compiler supports LASX.

1. I agree with gxw-loongson, the check in c_check is also necessary and can be used to decide whether the compiler compiles the LASX code into binary.
2. About the problem of '-mabi=lp64/lp64d' , this option is passed to compiler cc in OpenBLAS, not directly to binutils, consider that the compiler of gcc upstream  support 'lp64d'(let's ignore the loongson internal transitional version which support 'lp64', it has changed to 'lp64d' too), so can we directly changing it to '-mabi=lp64d' .
3. About potential '$xNN' and '$xrNN' problem, maybe we can use intrinsics or directly postpone the submission of this part after the toolchain is stable.

Thanks xenOn and martin-frbg.

c_check
xen0n: You have `.S` files, so these must be taken care of too. C files making use of LASX via intrinsics are unaffected anyway.

c_check
xen0n: So the LoongArch gcc provided by Deepin/UOS/Loongnix *all* support `-mabi=lp64d` now?

martin-frbg:
my thought/question is just if any "old world" compiler would identify itself as gcc 12 (or higher), or if that is a clear identification of "new world". And are there "old" compilers that do not support old-world "xrNN" register naming ?

xen0n:
> my thought/question is just if any "old world" compiler would identify itself as gcc 12 (or higher), or if that is a clear identification of "new world". And are there "old" compilers that do not support old-world "xrNN" register naming ?

Ah, I see what you mean; @yinshiyou @gxw-loongson can you confirm that there's no plan to upgrade the old-world compiler to gcc 12? If so we could replace the feature detection with the existing version check.

(Personally I'd prefer feature detection, due to having done many front-end work in the past where version checking doesn't really work, but here things may be different, and I'll write comments of course.)

martin-frbg:
Thinking old-world compiler upgraded to gcc-12 would call itself version 12 ... In the long term, feature detection may be safer if/when other compilers start supporting Loongarch64 (LLVM will be at version by then 15 I guess, but no idea about others).
Anyway this is partly just to avoid adding to the perl-based c_check when it may transition to a sh-based one soon.

martin-frbg:
I have merged the conversion of c_check to posix sh now, but reinstated the perl original as c_check.pl - hopefully that minimises file conflicts for you

yinshiyou:
> > my thought/question is just if any "old world" compiler would identify itself as gcc 12 (or higher), or if that is a clear identification of "new world". And are there "old" compilers that do not support old-world "xrNN" register naming ?
> 
> Ah, I see what you mean; @yinshiyou @gxw-loongson can you confirm that there's no plan to upgrade the old-world compiler to gcc 12? If so we could replace the feature detection with the existing version check.
> 
> (Personally I'd prefer feature detection, due to having done many front-end work in the past where version checking doesn't really work, but here things may be different, and I'll write comments of course.)

Sorry to confirm, do you mean in the OS product version?  Old OS products version like loongnix20-1 and UOS integrates gcc-8 by default which support '-mabi=lp64', and they also integrates OpenBLAS which use '-mabi=lp64'.  The future OS products version will integrate a compiler which support '-mabi=lp64d'(no matter what version). I am not sure if someone will upgrade compiler to gcc-12 in old OS, I think both fixes for openblas work fine with gcc-12.

Why I suggest changing 'lp64' directly to 'lp64d' in OpenBLAS upstream is just don't want to make it too complicated in OpenBLAS. GCC upstream community only support 'lp64d', the first time gcc support LoongArch is from version 12 and it use '-mabi=lp64d'.  For option '-mabi', there is no new/old world in gcc upstream version. 

Thanks!

martin-frbg:
@gxw-loongson I have now built the release version of gcc 12.1.0 on "gcc400" but the problem with the LAPACK testsuite persists. (This does not happen with the pure Fortran Reference-LAPACK, so must be either a C-Fortran argument passing problem or a gcc bug)

XiWeiGu:
> @gxw-loongson I have now built the release version of gcc 12.1.0 on "gcc400" but the problem with the LAPACK testsuite persists. (This does not happen with the pure Fortran Reference-LAPACK, so must be either a C-Fortran argument passing problem or a gcc bug)

Strangely, the same problem occurs on Loongnix GNU/Linux 20 with internal gcc version 8.3.0.

martin-frbg:
Hangs in the "xGE" and "xS2" tests from stest.in, dtest.in and dstest.in, everything else appears to run perfectly. (Have not tried the complex/double complex tests but expect them to be similar). Happens with and without OpenMP, happens with a single-threaded build as well. Weird.

martin-frbg:
Looks like hangs can be worked around by changing P and R parameters in param.h to
```
#define SGEMM_DEFAULT_P 32
#define DGEMM_DEFAULT_P 32
#define QGEMM_DEFAULT_P qgemm_p
#define CGEMM_DEFAULT_P 64
#define ZGEMM_DEFAULT_P 64
#define XGEMM_DEFAULT_P xgemm_p
 
#define SGEMM_DEFAULT_R 512
#define DGEMM_DEFAULT_R 858
#define QGEMM_DEFAULT_R qgemm_r
#define CGEMM_DEFAULT_R 512
#define ZGEMM_DEFAULT_R 512
#define XGEMM_DEFAULT_R xgemm_r
```
but that was just a quick hack without serious analysis

martin-frbg:
Better fix for the hangs, now with much less restrictive P/Q/R (file also incorporates the NO_LASX conditional for the DGEMM M,N): 
[param.h.zip](https://github.com/xianyi/OpenBLAS/files/8809242/param.h.zip)

martin-frbg:
Plus the new-style (POSIX sh) c_check with the equivalent changes from your PERL version (which now needs renaming to c_check.pl)
[c_check.zip](https://github.com/xianyi/OpenBLAS/files/8809441/c_check.zip)
everything else (Makefile.system,Makefile.loongarch64, KERNEL.LOONGSON3R5) remains unchanged from your first draft

xen0n:
@martin-frbg: Thanks for the help! I'll do the rebase later (busy at `$DAY_JOB` and focusing on upstreaming of the Linux/LoongArch port lately). I hope to finish this in this week.

h-vetinari:
What's the status here? 🙃

xen0n:
Very much drowned by day job, I'll return to this later this week.

xen0n:
I've seen a fatal error but `make` completed nevertheless, interesting...

```

  OS               ... Linux
  Architecture     ... loongarch64
  BINARY           ... 64bit
  C compiler       ... GCC  (cmd & version : cc (Gentoo 12.1.1_p20220723 p9) 12.1.1 20220723)
  Fortran compiler ... GFORTRAN  (cmd & version : GNU Fortran (Gentoo 12.1.1_p20220723 p9) 12.1.1 20220723)
  Library Name     ... libopenblas_loongson3r5p-r0.3.20.a (Multi-threading; Max num-threads is 4)
```

The error:

```
         EXPECTED RESULT                    COMPUTED RESULT
      1       0.0805368       0.0805368
      2       0.0777824       0.0777824
      3        0.119519        0.119519
      4      0.00836127      0.00836127
      5        0.183431        0.183431
      6       -0.180621       -0.180621
      7      -0.0500419      -0.0500419
      8       0.0583612       0.0583612
      9       0.0138702       0.0138702
     10       -0.130621       -0.130621
     11       0.0111158       0.0111158
     12      -0.0583053      -0.0583053
     13       0.0638702       0.0638702
     14         0.13894         0.13894
     15       0.0611157       0.0611157
     16       -0.116708       -0.116708
     17      -0.0527964      -0.0527964
     18         0.18894         0.18894
     19      -0.0555508      -0.0555508
     20      -0.0667085      -0.0667085
     21       -0.177866       -0.177866
     22       0.0722734       0.0722734
     23     -0.00555089     -0.00555089
     24       -0.183375       -0.183375
     25       -0.127866       -0.127866
     26       -0.119463       -0.119463
     27       -0.122217       -0.122217
     28       -0.133375       -0.133375
     29        0.141694        0.141694
     30       -0.069463       -0.069463
     31      0.00560678      0.00560678
     32        0.136185        0.136185
     33       -0.194533       -0.123395
     34        0.200098        0.110756
     35       0.0556067      -0.0265489
      THESE ARE THE RESULTS FOR COLUMN 1
 ******* cblas_dgemm  FAILED ON CALL NUMBER:
 26734: cblas_dgemm   CblasColMajor   CblasNoTrans   CblasNoTrans
35 9 2  1.0 A, 36, B, 3,  0.0, C, 36.
 ******* cblas_dgemm  FAILED ON CALL NUMBER:
     1: cblas_dgemm   CblasRowMajor   CblasNoTrans   CblasNoTrans
1 1 1  0.0 A, 2, B, 2,  0.0, C, 2.

****** FATAL ERROR - TESTS ABANDONED ******
```

martin-frbg:
Yes, these accuracy errors are non-fatal as these tests (inherited from the reference BLAS in "netlib" Reference-LAPACK) are more geared towards being watched by human eyes (and "small" deviations "may" be tolerable depending on hardware etc).
(FYI the related issue is https://github.com/Reference-LAPACK/lapack/issues/497 - somehow I remember seeing a fix that redirected the test outputs to a file and then grepped that for "HALF ACCURATE", but it may have been some distribution build script or my imagination)
Do you see the same failure in the corresponding plain BLAS dgemm test that runs a bit earlier ?

xen0n:
Hmm, there are some more errors. I've attached the full `make` output (not the first invocation so the log is not flooded with gcc calls) below.

[make-20220729.log](https://github.com/xianyi/OpenBLAS/files/9218150/make-20220729.log)

martin-frbg:
No test failures with current `develop` (including XiWeiGu's PR from yesterday) on the Loongson-hosted 3A5000 in the gcc compile farm. (With just "lp64" changed to "lp64d" to get it to compile with gcc 12). Will try to apply the PR next, but I guess it may be missing the part of the changes to param.h where it needs different DGEMM_UNROLL when your changed KERNEL.LOONGSON3R5 selects a different DGEMM kernel for NO_LASX. Or maybe even that NO_LASX switch is outdated, as it seems the optimized one seems to build and run just fine in my test ?

martin-frbg:
param.h https://github.com/xianyi/OpenBLAS/pull/3626#issuecomment-1142610530 (probably needs updating now to match XiWeiGu's latest)
and we'd also need the old-style perl c_check.pl for completeness https://github.com/xianyi/OpenBLAS/pull/3626#issuecomment-1142642264 (I think/hope nothing changed there in between -- eh, the Fujitsu compiler bits, so let me just add that file later, not so relevant for this PR)

xen0n:
Okay I'll send an updated revision hopefully today. Thanks for reviewing.

xen0n:
Looking at the cpuid implementation for loongarch64, it is wrong, again because of the new-world/old-world split: availability of LSX/LASX in CPUCFG doesn't imply usability, because the kernel may not have the support compiled in (due to unavailability of public docs regarding the ISA extensions), so you could still get SIGILLs on LSX/LASX-capable CPUs if you're running upstream kernel as of July 2022.

I'll fix them together with the revisions to hopefully fix the tests.

xen0n:
Also, the model name 2K1000 is referring to the MIPS model. The LoongArch model is named 2K1000LA which I'll also drive-by-fix.

(At this point I'm actually not sure if the LOONGSON3R5 and LOONGSON2K1000 impls should be renamed to the more generic LASX and LSX instead, as I can't see any microarchitecture-specific thoughts infused in the code. But that would involve much more renaming so I'd rather do that in a later PR.)

martin-frbg:
I must admit I have absolutely no insight into LoongArch64 model naming and related issues, and no way (or desire) to find out if any contributor with a Loongson affiliation is "official" company staff, temporary student intern or whatever. I suspect XiWeiGu may be the former gxw-loongson account - can you two please coordinate on naming (e.g. there may be plans to add microarch-specific changes already)  ? Also renaming existing TARGETs after a few releases may be impractical for existing users/packagers...

xen0n:
> I must admit I have absolutely no insight into LoongArch64 model naming and related issues, and no way (or desire) to find out if any contributor with a Loongson affiliation is "official" company staff, temporary student intern or whatever. I suspect XiWeiGu may be the former gxw-loongson account - can you two please coordinate on naming (e.g. there may be plans to add microarch-specific changes already) ? Also renaming existing TARGETs after a few releases may be impractical for existing users/packagers...

Understood; I'll try communicating. (I have connections with many other Loongson employees, but not @XiWeiGu.) The quality of all Loongson contribution is not equal, and oftentimes the "official" staff is actually lacking in experience with the community-driven software stack, so this is something to pay attention to. Some other community members and I try to keep an eye on all things LoongArch though, so papercuts are hopefully solved after all.

And to clarify: I won't just go ahead and blindly rename things. Now you mentioned there may be downstream depending on the names (which I don't know of earlier), I think I would just keep the (unfortunate) current names then.

XiWeiGu:
> Also, the model name 2K1000 is referring to the MIPS model. The LoongArch model is named 2K1000LA which I'll also drive-by-fix.
> 
> (At this point I'm actually not sure if the LOONGSON3R5 and LOONGSON2K1000 impls should be renamed to the more generic LASX and LSX instead, as I can't see any microarchitecture-specific thoughts infused in the code. But that would involve much more renaming so I'd rather do that in a later PR.)

Thank you for your advice, the existing way of naming is really inappropriate (lack of microarchitecture-specific thoughts like you said). I will propose a more reasonable name.

yinshiyou:
I prefer not  to add distinctions for old and new wold to  openblas upstream. there are only one world in upstream of kernel and gcc(upstream only accepted the 'new world').

xen0n:
> I prefer not  to add distinctions for old and new wold to  openblas upstream. there are only one world in upstream of kernel and gcc(upstream only accepted the 'new world').

But this unfortunately means disabling/removing all LSX/LASX assembly, because there's no public documentation of LSX/LASX yet, nor any toolchain support. Hence I was forced to maintain compatibility with both worlds...

martin-frbg:
I'd prefer to retain compatibility unless it is absolutely clear that there is nobody with an earlier system (and vendor toolchain) unable to update. And maybe once the assembly syntax changes are documented they will turn out to be fairly trivial to implement in the old codebase.

martin-frbg:
BTW the "HALF ACCURATE" thingy is definitely from having the wrong DGEMM_UNROLL_N, DGEMM_UNROLL_M in param.h
(assembly kernel is 16/4, generic is 8/2) - I did not remember that I had also made these changes in my tree on the gccfarm machine yesterday when I tested  XiWeiGu's PR before merging)

yinshiyou:
> > I prefer not  to add distinctions for old and new wold to  openblas upstream. there are only one world in upstream of kernel and gcc(upstream only accepted the 'new world').
> 
> But this unfortunately means disabling/removing all LSX/LASX assembly, because there's no public documentation of LSX/LASX yet, nor any toolchain support. Hence I was forced to maintain compatibility with both worlds...

Thank you very much for your reply. Not only OpenBLAS project, loongson only uploaded 'new world' to upstream of other projects too, so I really recomeded we can keep it is. If someone got an 'old world' OS, he must have a way to get the old world openblas from UOS/Kylin/Loongson. If needed, I can help to  add a download URL of 'old world openblas' on loongnix.cn recently. Thanks again, and hope you can take my advice.

xen0n:
> > > I prefer not  to add distinctions for old and new wold to  openblas upstream. there are only one world in upstream of kernel and gcc(upstream only accepted the 'new world').
> > 
> > 
> > But this unfortunately means disabling/removing all LSX/LASX assembly, because there's no public documentation of LSX/LASX yet, nor any toolchain support. Hence I was forced to maintain compatibility with both worlds...
> 
> Thank you very much for your reply. Not only OpenBLAS project, loongson only uploaded 'new world' to upstream of other projects too, so I really recomeded we can keep it is. If someone got an 'old world' OS, he must have a way to get the old world openblas from UOS/Kylin/Loongson. If needed, I can help to add a download URL of 'old world openblas' on loongnix.cn recently. Thanks again, and hope you can take my advice.

Unfortunately, the new world toolchain doesn't recognize any of the LASX syntax yet, so if you want to support only the new world, all LASX code must be unconditionally disabled, at which point removing might be the better option.

But yes, the upstream community is pretty much unanimous in the ABI support -- all treat the old world as non-existent. Nobody noticed the problem when the LASX code got first posted so we're left with this unfortunate pile of code, but indeed it may be better to maintain the current LASX code in a vendor fork so the upstream stays "clean".

martin-frbg:
Most likely there simply was no upstream ecosystem at the time the loongarch64 kernel tree was dropped into OpenBLAS (IIRC it almost coincided with a similarly big PR for the linux kernel (and subsequent complaints that it was all just mips64 rebranded). Still I see no point in removing the "old world" code while there is even a slim chance that someone relies on it - it isn't like it would bite a casual user, or would it ?

yinshiyou:
> > > > I prefer not  to add distinctions for old and new wold to  openblas upstream. there are only one world in upstream of kernel and gcc(upstream only accepted the 'new world').
> > > 
> > > 
> > > But this unfortunately means disabling/removing all LSX/LASX assembly, because there's no public documentation of LSX/LASX yet, nor any toolchain support. Hence I was forced to maintain compatibility with both worlds...
> > 
> > 
> > Thank you very much for your reply. Not only OpenBLAS project, loongson only uploaded 'new world' to upstream of other projects too, so I really recomeded we can keep it is. If someone got an 'old world' OS, he must have a way to get the old world openblas from UOS/Kylin/Loongson. If needed, I can help to add a download URL of 'old world openblas' on loongnix.cn recently. Thanks again, and hope you can take my advice.
> 
> Unfortunately, the new world toolchain doesn't recognize any of the LASX syntax yet, so if you want to support only the new world, all LASX code must be unconditionally disabled, at which point removing might be the better option.
> 
> But yes, the upstream community is pretty much unanimous in the ABI support -- all treat the old world as non-existent. Nobody noticed the problem when the LASX code got first posted so we're left with this unfortunate pile of code, but indeed it may be better to maintain the current LASX code in a vendor fork so the upstream stays "clean".

Let's disable LA optimization by default temporarily, once ecosystem prepared, enable it.

yinshiyou:
> Most likely there simply was no upstream ecosystem at the time the loongarch64 kernel tree was dropped into OpenBLAS (IIRC it almost coincided with a similarly big PR for the linux kernel (and subsequent complaints that it was all just mips64 rebranded). Still I see no point in removing the "old world" code while there is even a slim chance that someone relies on it - it isn't like it would bite a casual user, or would it ?

Thanks for your reply. The progress of merging code in each upstream community is different, which leads to the current problem. 1) upstream openblas code with latest kernel will not recognize LASX instructions. 2) upstream openblas has compatibility problems when used in the 'old world' OS.
Disable LA optimization by default temporarily might be a more appropriate solution. When other upstream like kernel and gcc prepared, we can enable it by default again.

yinshiyou:
> Most likely there simply was no upstream ecosystem at the time the loongarch64 kernel tree was dropped into OpenBLAS (IIRC it almost coincided with a similarly big PR for the linux kernel (and subsequent complaints that it was all just mips64 rebranded). Still I see no point in removing the "old world" code while there is even a slim chance that someone relies on it - it isn't like it would bite a casual user, or would it ?

Thanks for your reply. The progress of merging code in each upstream community is different, which leads to the current problem that upstream openblas code with latest kernel will not recognize LASX instructions. 
Disable LA optimization by default temporarily might be a more appropriate solution. When other upstream like kernel and gcc prepared, we can enable it by default again.

xen0n:
Thanks for both your input @martin-frbg and @yinshiyou. I think I'll simply unconditionally disable the LASX assembly then, but not deleting them for now. We can always do so later if the LSX/LASX manuals and toolchain implementation fail to turn up eventually. Is this approach okay?

EDIT: And all that "world flavor" detection would be gone if we take this route. Commercial distros based on vendor forked toolchains should get their vendor-specific bits from, ahem, vendor repos of course.

martin-frbg:
Guess commenting out the entries in KERNEL.LOONGSON3R5 would do the least "damage". Then it looks to me as if the only difference between the available LoongArch64 targets will be L2 cache size (which only has one rather obscure use in the buffer allocation code for now)

xen0n:
> Guess commenting out the entries in KERNEL.LOONGSON3R5 would do the least "damage".

I think so. I'll try this later, today or tomorrow.

> Then it looks to me as if the only difference between the available LoongArch64 targets will be L2 cache size (which only has one rather obscure use in the buffer allocation code for now)

Actually the current code is messy and incorrect, as previously pointed out (either the flavors get renamed to LASX and LSX, or the CPU detection should be based on processor ID instead of feature bits). But the code works regardless of the exact LoongArch flavor selected so technically we could polish that part later.

XiWeiGu:
Using `LASX` and `LSX` as target name is not a  perfect fit.  In PR #3711, I have used `LA464`, `LA264`(read `PRID` through `cpucfg`) to replace the original targets. Based on the common review we have reached(disable LASX optimization by default), no additional kernel feature checking was done.

If there is no problem,  I will go ahead and fix the following issues:

1. Temporarily disable `LASX` optimization by default
2. Directly modify `ABI` to `lp64d`
3. Add compiler feature check

yinshiyou:
> Actually the current code is messy and incorrect, as previously pointed out (either the flavors get renamed to LASX and LSX, or the CPU detection should be based on processor ID instead of feature bits). But the code works regardless of the exact LoongArch flavor selected so technically we could polish that part later.

About CPU detection, using 'cpucfg' to detect feature bits in loongarch is the same with using 'cpuid' to detect feature bits in X86, just with different name in different ISA.

XiWeiGu:
All issues in this PR have been resolved and it can be closed.

PR #3642, opened 2022-06-02
Add ARM64 support for Windows

nursik:
This PR makes possible to build OpenBLAS for Windows on ARM using MSVC. Changes are made in order to bring arm64-windows support for vcpkg,

### DIFF
--- a/cmake/system_check.cmake
+++ b/cmake/system_check.cmake
@@ -31,7 +31,11 @@ endif()
 
 # Pretty thorough determination of arch. Add more if needed
 if(CMAKE_CL_64 OR MINGW64)
-  set(X86_64 1)
+  if (CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64.*|AARCH64.*|arm64.*|ARM64.*)")
+    set(ARM64 1)
+  else()
+    set(X86_64 1)
+  endif()
 elseif(MINGW OR (MSVC AND NOT CMAKE_CROSSCOMPILING))
   set(X86 1)
 elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "ppc.*|power.*|Power.*")
--- a/common_arm64.h
+++ b/common_arm64.h
@@ -35,9 +35,9 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #ifdef C_MSVC
 #include <intrin.h>
-#define MB do {} while (0)
-#define WMB do {} while (0)
-#define RMB
+#define MB __dmb(_ARM64_BARRIER_ISH)
+#define WMB __dmb(_ARM64_BARRIER_ISHST)
+#define RMB __dmb(_ARM64_BARRIER_ISHLD)
 #else
 #define MB   __asm__ __volatile__ ("dmb  ish" : : : "memory")
 #define WMB  __asm__ __volatile__ ("dmb  ishst" : : : "memory")
@@ -92,7 +92,12 @@ static void __inline blas_lock(volatile BLASULONG *address){
 #if !defined(OS_DARWIN) && !defined (OS_ANDROID)
 static __inline BLASULONG rpcc(void){
   #ifdef C_MSVC
-  return __rdtsc();
+    const int64_t pmccntr_el0 = (((3 & 1) << 14) |  // op0
+        ((3 & 7) << 11) |  // op1
+        ((9 & 15) << 7) |  // crn
+        ((13 & 15) << 3) | // crm
+        ((0 & 7) << 0));   // op2
+    return _ReadStatusReg(pmccntr_el0);
   #else
   BLASULONG ret = 0;
   blasint shift;
--- a/common_arm64.h
+++ b/common_arm64.h
@@ -33,9 +33,16 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #ifndef COMMON_ARM64
 #define COMMON_ARM64
 
+#ifdef C_MSVC
+#include <intrin.h>
+#define MB do {} while (0)
+#define WMB do {} while (0)
+#define RMB
+#else
 #define MB   __asm__ __volatile__ ("dmb  ish" : : : "memory")
 #define WMB  __asm__ __volatile__ ("dmb  ishst" : : : "memory")
 #define RMB  __asm__ __volatile__ ("dmb  ishld" : : : "memory")
+#endif
 
 #define INLINE inline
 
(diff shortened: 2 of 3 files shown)


martin-frbg:
Thanks. Now that I looked into this, I see https://docs.microsoft.com/en-us/cpp/intrinsics/arm64-intrinsics?view=msvc-170#BarrierRestrictions seems to document specific intrinsics to implement the memory barriers. Would it make sense to use them here ?

nursik:
Do you mean particularly in this PR? I assume the use of plain C functions (generic kernel) should be sufficient and is in line with x86_64 version.

martin-frbg:
Oh, that comment was only about the MB/WMB/RMB implementations in common_arm64.h where you appear to have put empty loops. From my cursory reading of the documentation page, it looked as if `_dmb(_ARM64_BARRIER_ISH)` etc. were
direct equivalents of the `dmb ish` etc. assembly used on the Unixoid platforms.
For the kernels, the generic C versions are certainly sufficient (though it is a pity that the existing assembly kernels are not usable with MSVC)

nursik:
Thank you for the comments👍. I do not have enough expertise to say that it can be helpful or right to use these intrinsics in MB/RMB/WMB. I will check that.

> though it is a pity that the existing assembly kernels are not usable with MSVC

The same feeling.

martin-frbg:
I know very little about the MS platform, but I know what happened with multithreaded code on other "weakly-ordered" systems when there were no proper barriers in place.

brada4:
will llvm for woa64 get detected?

martin-frbg:
> will llvm for woa64 get detected?

I see no reason why it shouldn't. This PR is specifically about MSVC, which seems to have the same restrictions w.r.t. assembly code as on x86

brada4:
I see , thanks.

nursik:
@martin-frbg Could you please help me with providing some details on how to test the changes (specifically saying rpcc, mb, lock)? As I understand these changes are included into the build only if `USE_OPENMP `is on. I wanted to run `cpp_thread_test`, but it seems that it requires Fortran and also changes in tests to be compilable by MSVC.

I will be glad to hear any other suggestions or your concerns about the code changes 👍

martin-frbg:
cpp_thread_test should not depend on Fortran at all, this is C++ code that only uses BLAS functions. Maybe there are some details e.g. in the use of the random number generator that are not compatible with MSVC.
The effects of rpcc and the memory barriers should be visible in multithreaded code even without OPENMP, in particular in BLAS3 functions like GEMM or SYMM where Goto's paralellization algorithm distributes the workload across multiple threads.
I agree that it is somewhat cursed that a C implementation of BLAS comes with tests (in the test and ctest folders) that are written in Fortran - otherwise ctest/xdcblat3 etc. should be sufficient to check the implementation (e.g. #2363 exposed a bug in the original placement of the barriers that rendered them useless)

nursik:
Thank you for your answer. I managed to build `cpp_thread_test` tests and run them multiple times. Everything OK. 
The comment about Fortran - [this line in CMakeLists.txt](https://github.com/xianyi/OpenBLAS/blob/7060ca50020e990f2e4cc7c3f340fcf1a887af7d/CMakeLists.txt#L314). It requires `NOFORTRAN `to be false in order to build the `cpp_thread_test `folder and [OPEN_MP to be set](https://github.com/xianyi/OpenBLAS/blob/7060ca50020e990f2e4cc7c3f340fcf1a887af7d/cpp_thread_test/CMakeLists.txt#L8).

martin-frbg:
Did you need to make any changes in the cpp_thread_test sources to get them to compile with MSVC ?

nursik:
Yes, of course. [std::getline to getline](https://github.com/xianyi/OpenBLAS/blob/7060ca50020e990f2e4cc7c3f340fcf1a887af7d/cpp_thread_test/cpp_thread_safety_common.h#L5) and [std::stoul to std::strtoul](https://github.com/xianyi/OpenBLAS/blob/7060ca50020e990f2e4cc7c3f340fcf1a887af7d/cpp_thread_test/dgemm_thread_safety.cpp#L33)

nursik:
Also, did you think of adding cross-compiling build for windows-arm64?

martin-frbg:
Have not tried as I do not have any windows-arm64 for testing, but I guess it might work "out of the box" at least with LLVM (which also does not have the limitations w.r.t assembly). Direct compilation with LLVM on windows-arm64 hosts is already supported, see https://github.com/xianyi/OpenBLAS/wiki/How-to-build-OpenBLAS-for-Windows-on-ARM64

nursik:
I have meant for CI. It would be the same as for `windows-cl` job, but you need to pass `-A arm64` and maybe forcing `-DCMAKE_SYSTEM_PROCESSOR=ARM64` as mentioned [here](https://linaro.atlassian.net/wiki/spaces/WOAR/pages/28716138610/CMake+Limitations).

brada4:
That does not do any corectness tests. Fact that artifact is generated does not mean new platfrm (combination) porting is complete

nursik:
Are there any more concerns about the PR?

martin-frbg:
Time constraints on my end only, sorry. Plan to merge this after the C versions of CTEST build on all platforms. (Minor nit: the KERNEL.generic could probably be simplified to just "include $(KERNELDIR)/../arm/KERNEL.ARMV5" but that can be done after the merge)

common_arm64.h
```diff
@@ -80,13 +91,22 @@ static void __inline blas_lock(volatile BLASULONG *address){
 
 #if !defined(OS_DARWIN) && !defined (OS_ANDROID)
 static __inline BLASULONG rpcc(void){
+  #ifdef C_MSVC
```
hmartinez82, line 94: Won't this also work on GCC or Clang's GCC frontend?

martin-frbg: As far as I know, this would be a problem if using MSC_VER in the ifdef here, but not C_MSVC. (or I may be misunderstanding your question, but if you're suggesting to use that implementaton of rpcc instead of the existing assembly one with all three compilers I do not see the advantage)

hmartinez82: I mean to say that `_ReadStatusReg()` could be available in other compilers too. But I see, the PR as is the most conversative in changes. It was just a suggestion ;-)

martin-frbg: Right, most conservative change to accomodate the least recommendable (as it cannot use the existing optimized assembly kernels) compiler. Thanks for your suggestion though :)

brada4: MSVC was not available on aarch64 before, so the required port is in this PR. clang, gcc, clang-cl are supposed to work unless somebody with enough time and test base complains on contrary.

cmake/system_check.cmake
```diff
@@ -31,7 +31,11 @@ endif()
 
 # Pretty thorough determination of arch. Add more if needed
 if(CMAKE_CL_64 OR MINGW64)
-  set(X86_64 1)
+  if (CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64.*|AARCH64.*|arm64.*|ARM64.*)")
```
hmartinez82, line 34: Won't this also require a change in line 27 to detect aarch64 so MINGW64 is set to 1?

hmartinez82: I mean when building with MSYS2's `clangarm64` environment

Issue #3383, opened 2021-09-17
dgemm crash on macOS 12.0 beta arm64


ilian8:
With macOS 12.0 (21A5506j) arm64 and OpenBLAS 0.3.17 (earlier versions as well) there is a crash in the utest target

```
* thread JuliaLang/julia#3, stop reason = EXC_BAD_ACCESS (code=2, address=0x10031c000)
    frame #0: 0x000000010005d394 openblas_utest`.Ldgemm_tcopy_L8_M8_20 + 116
openblas_utest`.Ldgemm_tcopy_L8_M8_20:
->  0x10005d394 <+116>: ldp    q30, q31, [x13], #0x20
    0x10005d398 <+120>: st1.2d { v28, v29, v30, v31 }, [x20]
    0x10005d39c <+124>: add    x20, x20, #0x40           ; =0x40 
    0x10005d3a0 <+128>: add    x14, x14, x5
Target 0: (openblas_utest) stopped.
```
See also https://github.com/JuliaLang/LinearAlgebra.jl/issues/866.

There is no problem on maOS 11.6 (but the library built on macOS 11 also crashes when run on macOS 12).

martin-frbg:
Unfortunately I do not have M1 hardware (beyond what is available in the gcc compile farm). The OpenBLAS dgemm_tcopy kernel is shared among all supported ARMV8 cpus however, so unlikely to be the culprit.

ilian8:
Thanks, it seems a bit mysterious how the same assembly can run differently on the same CPU.

Does the dgemm_tcopy kernel make any use of the x18 register by any chance? 

This is generally available on ARMv8 CPUs, but is reserved on the Apple M1 according to 

https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms.

martin-frbg:
Yes it does. Looks like it may be an easy fix as it is `#define`d to a variable I and there is an unused `#define TEMP2 x21` in dgemm_tcopy_8.S as well. Of course there may be additional instances in other kernels (and all have been there for years)

martin-frbg:
Tentative fix in JuliaLang/julia#3384 - have not checked yet if I broke anything with the naive reshuffle...

mryodo:
> Tentative fix in JuliaLang/julia#3384 - have not checked yet if I broke anything with the naive reshuffle...

if you want I can try building it on `m1`, although simple `make` results into undefined symbols for `arm64` architecture

[...]

PR #3652, opened 2022-06-16
DOC: Add tip about macOS SDK

larsoner:
On my new M1 machine I installed Xcode a couple of weeks ago, and today did a fresh `git clone` here. When I did `make` I got a bunch of warnings:
```
ld: warning: object file (/opt/homebrew/Cellar/gcc/11.3.0_1/lib/gcc/11/gcc/aarch64-apple-darwin21/11/libgcc.a(enable-execute-stack.o)) was built for newer macOS version (11.0) than being linked (10.8)
```
then eventually get errors like the one in this issue:
```
Undefined symbols for architecture arm64:
  "___chkstk_darwin", referenced from:
      _sgemv_ in libopenblas_vortexp-r0.3.20.a(sgemv.o)
      _sger_ in libopenblas_vortexp-r0.3.20.a(sger.o)
      _cblas_sgemv in libopenblas_vortexp-r0.3.20.a(cblas_sgemv.o)
      _cblas_sger in libopenblas_vortexp-r0.3.20.a(cblas_sger.o)
      _dgemv_ in libopenblas_vortexp-r0.3.20.a(dgemv.o)
      _dger_ in libopenblas_vortexp-r0.3.20.a(dger.o)
      _cblas_dgemv in libopenblas_vortexp-r0.3.20.a(cblas_dgemv.o)
      ...
```
This seems to be an Xcode SDK problem, as I'm on 13.4 since this is what I downloaded a couple of weeks ago when I got my new M1 machine. I then tried `MACOSX_DEPLOYMENT_TARGET=11.0 make` and things compiled fine.

Another option is probably to install an older Xcode, so I made a note about that option, too.

I'm happy to do something else if there is a better recommendation, but figured if there isn't, then this could be something useful to add to the README!

FYI I'm trying to build locally so that I can determine if a SVD failure on M1 using NumPy from conda-forge is a cross-compilation problem or a problem with OpenBLAS itself:

https://github.com/numpy/numpy/issues/21756

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,8 @@ If you compile this library with `USE_OPENMP=1`, you should set the `OMP_NUM_THR
 environment variable; OpenBLAS ignores `OPENBLAS_NUM_THREADS` and `GOTO_NUM_THREADS` when
 compiled with `USE_OPENMP=1`.
 
+On newer versions of Xcode and on arm64, you might need to compile with a newer macOS target (11.0) than the default (10.8) with `MACOSX_DEPLOYMENT_TARGET=11.0`, or switch your command-line tools to use an older SDK (e.g., [13.1](https://developer.apple.com/download/all/?q=Xcode%2013)).
+
 ### Setting the number of threads at runtime
 
 We provide the following functions to control the number of threads at runtime:


martin-frbg:
This sort of info usually ends up in the FAQ section of the Wiki (which may be suboptimal for offline installations I guess). Actually this particular issue was already seen in a later comment on issue #3383 but back then I did not find an explanation for why the SDK default was wrong.

larsoner:
Okay, feel free to close and make a new wiki page if that's preferable!

larsoner:
Done! https://github.com/xianyi/OpenBLAS/wiki/How-to-build-OpenBLAS-for-macOS-M1---arm64

martin-frbg:
Thanks. I'm about to move the default deployment target to 11.0 for ARM-based macs (same as Julialang does), but learned today that someone even had to use `MACOSX_DEPLOYMENT_TARGET=12.4`, 11 being not enough to make it work "on version 12.4".

PR #3664, opened 2022-06-22
(ignore)

### LINKED ISSUE #3659: Invalid Intel-host ARM64-target cross-compilation flag -m128bit-long-double
We (Numpy / Scipy) are compiling OpenBLAS binaries over at https://github.com/MacPython/openblas-libs .  Our arm64 builds are failing in Github Actions - e.g. https://github.com/MacPython/openblas-libs/runs/6832414065?step:7:1321

The error is:

```
arm64-apple-darwin20.0.0-gfortran: error: unrecognized command-line option ‘-m128bit-long-double’
```

I think this is a problem in OpenBLAS cross-compilation support.  The `-m128bit-long-double`  flag arises either around https://github.com/xianyi/OpenBLAS/blob/develop/Makefile.system#L536 or around https://github.com/xianyi/OpenBLAS/blob/develop/cmake/arch.cmake#L18

I believe the problem is that the `NO_EXPRECISION` Makefile variable is set to 0 (yes, there is extended precision) when compiling on an Intel *host*, but it should be set only when compiling for an Intel *target*.  

Is that analysis right? 

Should `NO_EXPRECISION` be set to 0 for `arm64` targets?

martin-frbg:
written for #3659 before proper analysis, ignore

### DIFF
--- a/Makefile.system
+++ b/Makefile.system
@@ -1463,6 +1463,9 @@ ifneq ($(ARCH), x86)
 NO_AFFINITY = 1
 endif
 endif
+ifdef ARMV8
+NO_AFFINITY = 1
+endif
 
 ifdef NO_AFFINITY
 ifeq ($(NO_AFFINITY), 0)


martin-frbg:
This cannot work, as ARMV8 is defined in C space (config.h) but not Makefile space. Need a proper definition of CROSSARCH that covers all architectures, perhaps derived from the cross-compiler triplet

martin-frbg:
Hmm. Bad dream. ARMV8 define does get put in Makefile.conf as well (but so does the "ARM64" architecture so the reported issue should not even arise)

Issue #3671, opened 2022-06-28
sbgemm support for ARM Neoverse N2


imzhuhl:
The previous issue has discussed how valuable BFloat16 is in the field of Deep Learning and Machine Learning, and the `develop`  branch also supports bf16 on a few Intel platform. However, there seems to be no implementation for the ARM platform.

Currently, I have a need for optimization of matrix multiplication of type BFloat16 on ARM Neoverse N2. I found that the original implementation used the `gemmkernel_2x2.c` file in the `generic` directory, which resulted in low efficiency. On ARM Neoverse N2, we can use BFMMLA instruction to accelerate calculation.

Currently, some cloud service vendors have launched ARM servers, such as AWS Graviton and Alibaba Cloud Yitian 710. Maybe there will be a lot of demand for ARM service platforms. Does OpenBLAS have an optimization plan for the ARM platform?

imzhuhl:
I've already done a preliminary implementation on the Neoverse N2 platform with a single thread of around 90GFLOPS compared to FP32 with a single thread of 40 GFLOPS

martin-frbg:
Unfortunately the core developer team of OpenBLAS is very small (mostly just me in my limited spare time, trying to keep the number of open issues from exploding), so patches/PRs are always welcome.

martin-frbg:
Closing as resolved by your PR #3672

snadampal:
Hi @imzhuhl , Please let me know how you tested this feature? 
When I tried the "test_sbgemm", to compare the sgemm vs sbgemm output, it was failing with the data mismatch.
Appreciate if you can share details on what and how it has been tested. Thank you!

martin-frbg:
That's weird, I'm pretty sure it passed on my Cortex-X2.,. reopening

imzhuhl:
> Hi @imzhuhl , Please let me know how you tested this feature? When I tried the "test_sbgemm", to compare the sgemm vs sbgemm output, it was failing with the data mismatch. Appreciate if you can share details on what and how it has been tested. Thank you!

I tried again and it passed `test_sbgemm`. In addition to `test_sbgemm` I have my own implementation of a test where I directly compare results between sbgemm and naive loop (triple for loop).

Could you run the test again on your machine with the following modifications:
1. force a single threaded run (make sure to exclude multi-thread errors)
2. change the test data to random integers (make sure to exclude errors due to precision)

snadampal:
[...]

PR #3672, opened 2022-06-28
sbgemm support for ARM Neoverse N2

imzhuhl:
#3671


Makefile.arm64
```diff
@@ -121,7 +121,7 @@ ifeq ($(CORE), NEOVERSEN2)
 ifeq (1, $(filter 1,$(GCCVERSIONGTEQ7) $(ISCLANG)))
 ifeq ($(GCCVERSIONGTEQ9), 1)
 ifeq (1, $(filter 1,$(GCCMINORVERSIONGTEQ4) $(GCCVERSIONGTEQ10)))
-CCOMMON_OPT += -march=armv8.5-a -mtune=neoverse-n2
+CCOMMON_OPT += -march=armv8.5-a+sve+sve2+bf16 -mtune=neoverse-n2
```
brada4, line 124: I suppose gfortran could gain from same parameters.

Makefile.arm64
```diff
@@ -121,7 +121,7 @@ ifeq ($(CORE), NEOVERSEN2)
 ifeq (1, $(filter 1,$(GCCVERSIONGTEQ7) $(ISCLANG)))
 ifeq ($(GCCVERSIONGTEQ9), 1)
 ifeq (1, $(filter 1,$(GCCMINORVERSIONGTEQ4) $(GCCVERSIONGTEQ10)))
```
brada4, line 123: Add gccversion>=11 too ;-)

Makefile.arm64
martin-frbg: I doubt gfortran does autovectorizing, anyway out of the scope of the PR I think

Makefile.arm64
martin-frbg: @brada4 11 was >=10  already last time I checked

Makefile.arm64
imzhuhl: OK, I've already added it

Makefile.arm64
brada4: I see in other places like 3 evaluations:
`GCCVERSIONCHECK := $(GCCVERSIONGTEQ5)$(GCCVERSIONGTEQ4)$(GCCMINORVERSIONGTEQ7)` (x86_64)

martin-frbg: That's basically the same, just on a single line with a more complicated evaluation instead of split into two nested ifeq.
The idea is that we want e.g. gcc 9.4 or better, but the "better" is any higher major number and not just 10.4, 11.4 etc. CMAKE does semantic version checks but gmake needs them spelled out.

martin-frbg:
Thank you very much for your contribution.
I notice that the ncopy and tcopy sources do not appear to have anything neoverse-specific in them, so they could probably be renamed to their unroll factor suffix and/or reused for other broadly similar cpus. (And sbgemm_beta is probably identical to the generic sgemm_beta - or at least I do not see why it could not be - so we could just reference that in the KERNEL file, but a bit
of redundancy does not hurt.)

PR #3675, opened 2022-07-02
workaround ThunderX2 DNRM2 fault with ssq=inf,scale=0

### LINKED ISSUE #3654: BUG: Regression in DGESDD / DLASCL on macOS Vortex M1
On 8c20ca345aad43c2f74a72b356afdbc1ec368e31 using OpenBLAS to build and compile SciPy I get problems when I try to take use DGESDD (or DGESVD) to take the SVD of a matrix of ones of shape `(41, 50)`:
```
OpenBLAS larsoner$ make clean; git reset --hard; git clean -xdf; MACOSX_DEPLOYMENT_TARGET=11.0 make && make install
scipy larsoner$ git clean -xdf && ln -s ../site.cfg && python setup.py develop && python -c "import numpy as np, scipy.linalg; scipy.linalg.svd(np.ones((41, 50)))"
 ** On entry to DLASCL parameter number  4 had an illegal value
 ** On entry to DLASCL parameter number  4 had an illegal value
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/larsoner/python/scipy/scipy/linalg/_decomp_svd.py", line 131, in svd
    raise LinAlgError("SVD did not converge")
numpy.linalg.LinAlgError: SVD did not converge
```
On the previous commit 8e4c209002d1a648094a1ac5ca02efbae69d253f, things seem to be okay (SVD decomp matches that on x86_64 Linux). Same thing with being on latest `develop` / 7060ca50020e990f2e4cc7c3f340fcf1a887af7d and doing `git revert 8c20ca3` to revert just one of the two commits (8c20ca3) from #3399 things are okay.

@martin-frbg looks like https://github.com/xianyi/OpenBLAS/pull/3399 this was your PR, can you replicate?

cross-ref https://github.com/numpy/numpy/issues/21756

martin-frbg:
fixes #3654

### DIFF
--- a/kernel/arm64/dznrm2_thunderx2t99.c
+++ b/kernel/arm64/dznrm2_thunderx2t99.c
@@ -404,6 +404,7 @@ FLOAT CNAME(BLASLONG n, FLOAT *x, BLASLONG inc_x)
 #else
 	nrm2_compute(n, x, inc_x, &ssq, &scale);
 #endif
+	if (fabs(scale) <1.e-300) return 0.;
 	ssq = sqrt(ssq) * scale;
 
 	return ssq;


larsoner:
Locally I did `MACOSX_DEPLOYMENT_TARGET=11.0 make TARGET=VORTEX` then rebuilt numpy+scipy and all seems well. Thanks for digging into this @martin-frbg !

Issue #3679, opened 2022-07-06
BUG: utest dnrm2_tiny failed


XiWeiGu:
Utest dnrm2_tiny failed on Loongson-3A5000

> TEST 30/35 rot:srot_inc_0 [OK]
> TEST 31/35 rot:csrot_inc_0 [OK]
> TEST 32/35 dnrm2:dnrm2_inf [OK]
> TEST 33/35 dnrm2:dnrm2_tiny [FAIL]
>   ERR: test_dnrm2.c:65  expected 0.000e+00, got inf (diff -inf, tol 1.000e-13)
> TEST 34/35 fork:safety [OK]
> TEST 35/35 fork:safety_after_fork_in_parent [OK]
> RESULTS: 35 tests (34 ok, 1 failed, 0 skipped) ran in 223 ms

Utest dnrm2_tiny failed on Loongson-3A4000

> TEST 30/35 rot:zdrot_inc_0 [OK]
> TEST 31/35 rot:drot_inc_0 [OK]
> TEST 32/35 dnrm2:dnrm2_tiny [FAIL]
>   ERR: test_dnrm2.c:65  expected 0.000e+00, got inf (diff -inf, tol 1.000e-13)
> TEST 33/35 dnrm2:dnrm2_inf [OK]
> TEST 34/35 fork:safety [OK]
> TEST 35/35 fork:safety_after_fork_in_parent [OK]
> RESULTS: 35 tests (34 ok, 1 failed, 0 skipped) ran in 583 ms

Loongson-3A4000 DNRM2 kernel: `kernel/mips64/dnrm2.S`，the following code causes the test to fail:

```
    div.d   ALPHA, ALPHA, s1
    MOV max, s1
```
`s1` is the maximum value of the input vector,  the initial value of `ALPHA` is 1.
In the test case of dnrm2_tiny, `s1=7.457008414e-310`, `ALPHA=1/s1` is `inf`, test fail.
Loongson-3A5000 which using DNRM2 kernel: `kernel/loongarch64/dnrm2.S` has the similar problem.

Similar code has found in the power DNRM2 kernel, perhaps other platforms will have similar problem.

martin-frbg:
[...]

PR #3682, opened 2022-07-07
Fix dnrm2_tiny testcase failure

XiWeiGu:
`s1` stores the maximum value of the input vector, I think it will be better to  make sure `s1` is greater than or equal to 1.0

### DIFF
--- a/kernel/loongarch64/dnrm2.S
+++ b/kernel/loongarch64/dnrm2.S
@@ -53,6 +53,7 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define s4     $f9
 #define ALPHA  $f4
 #define max    $f5
+#define INF    $f6
 
    PROLOGUE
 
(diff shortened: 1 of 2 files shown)


XiWeiGu:
#3679

martin-frbg:
I'm sorry, but that seems to me to be an unsafe assumption. And indeed the LAPACK testsuite (`make lapack-test`) gains about 10000 failures for DOUBLE with (what I assume to be) the arm64 and loongarch64 equivalents of this patch, unfortunately. Could you check this please - I do not have access to a MIPS64 with working fortran compiler at the moment ?

XiWeiGu:
> I'm sorry, but that seems to me to be an unsafe assumption. And indeed the LAPACK testsuite (`make lapack-test`) gains about 10000 failures for DOUBLE with (what I assume to be) the arm64 and loongarch64 equivalents of this patch, unfortunately. Could you check this please - I do not have access to a MIPS64 with working fortran compiler at the moment ?

Thank you very much, I will check.

XiWeiGu:
I'm sorry, the previous patch will cause lapack testsuit failures for DOUBLE on MIPS64. I have updated the patch to slove this issue.

martin-frbg:
Thank you - but wouldn't the new patch make it return the INF value, instead of zero ?

XiWeiGu:
If `ALPHA` is equal to `INF`, it will return the value in `s1` directly, not INF or zero.

martin-frbg:
Ah, sorry. Guess I must have made a mistake in my port to the Loongarch64 DNRM2 then.

martin-frbg:
Thank you very much - I was using the wrong kind of instructions for loading the INF on Loongarch64.

PR #3684, opened 2022-07-12
Neoverse N2: DYNAMIC_ARCH

imzhuhl:
Enable DYNAMIC_ARCH feature on Arm Neoverse N2


brada4:
Please use [this](https://github.com/xianyi/OpenBLAS/pull/3278) as a checklist.

martin-frbg:
@brada4 what are you on about ? The N2 is already a supported target.

martin-frbg:
Thank you - looks like this fixes an oversight in #3508 (and I'll see to fixing similar omissions with  V1 ~~and Cavium&Ampere~~. Embarassing to note that the list of kernels for DYNAMIC_ARCH has been longer than what can get actually detected at runtime )

PR #3709, opened 2022-08-02
Add TCORE Generic

nursik:
During the cross-compilation on x64 host with MSVC for ARM, the build fails as there is no define directives for Generic core.


martin-frbg:
Thank you - reminds me that this target list is still very sketchy, and all just to avoid running getarch like the gmake build does.

PR #3711, opened 2022-08-03
LoongArch64: Rename targets

XiWeiGu:
The existing targets naming convention for LoongArch64 is not appropriate(lack of microarchitecture-specific thoughts).
Now I use `LA464`, `LA264` to replace the original targets.

### DIFF
--- a/kernel/generic/lsame.c
+++ b/kernel/generic/lsame.c
@@ -38,7 +38,7 @@
 
 #include <ctype.h>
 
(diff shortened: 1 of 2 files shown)


xianyi:
I suggest provide some loongson build bots to http://ci.openblas.net/

XiWeiGu:
Thank you for your advice. Since I am not familiar with build bots, do you want me to provide compile options for LoongArch64?

PR #3718, opened 2022-08-08
Fix SBGEMM test to work with INTERFACE64

martin-frbg:


test/compare_sgemm_sbgemm.c
```diff
@@ -76,9 +76,9 @@ float16to32 (bfloat16_bits f16)
 int
 main (int argc, char *argv[])
 {
-  int m, n, k;
+  blasint m, n, k;
   int i, j, l;
-  int x;
+  blasint x;
```
RajalakshmiSR, line 81: Can you also please remove line#129 `for (l = 0; l < k; l++)` where we compare CC & DD.

martin-frbg: Oh, right, this loop does not really make sense - and same for line 115 I guess :)

PR #3720, opened 2022-08-08
Make it work on general MIPS64 processors

FlyGoat:
This PR addressed some build failures for ARCH_MIPS64, also provide a MIPS64_GENERIC target so we should be able to use OpenBLAS on Cavium processors.


cpuid_mips64.c
```diff
@@ -136,9 +138,11 @@ void get_subarchitecture(void){
     printf("P6600");
   }else if(detect()==CPU_I6500){
     printf("I6500");
-  }else{
+  }else if(detect()==CPU_SICORTEX){
     printf("SICORTEX");
-  }
+  }else{
+    printf("MIPS64_GENEIRC");
```
martin-frbg: small typo here :)

martin-frbg:
Thanks - I have an Octeon3 patch lying around that I had meant to add "soon", but it does not really have much if anything Cavium-specific in it. (It should be possible to build on MIPS64 with TARGET=GENERIC already, but I suspect that would currently produce a 32bit build, might need to re-check that)

cpuid_mips64.c
FlyGoat: Good catch :-)
Fixed

martin-frbg:
getarch still returns UNKNOWN cpu rather than your MIPS64_GENERIC type with this patch on Octeon

FlyGoat:
> getarch still returns UNKNOWN cpu rather than your MIPS64_GENERIC type with this patch on Octeon

Fixed. We can safely assume all Linux MIPS system is safe to run MIPS64_GENERIC.

martin-frbg:
now that I have merged this, I am wondering if there isn't considerable overlap with the venerable SICORTEX target, maybe that is the effectively generic target I remembered ?

FlyGoat:
> now that I have merged this, I am wondering if there isn't considerable overlap with the venerable SICORTEX target, maybe that is the effectively generic target I remembered ?

Yeah probably… Though I can’t spot fallback to SICORTEX in code.

martin-frbg:
cpuid_mips64 does it when there is a model name in /proc/cpuinfo that does not match any of the Loongson types. seems on your machine there is no model name entry in cpuinfo ? the octeon I used for testing is running BSD, I'll need to dig out my own (early octeon3 running linux) to check

PR #3722, opened 2022-08-08
GitHub Actions: Add cross compile tests

FlyGoat:
Add cross compile tests without running checks. Currently
only mips64el and riscv64 is wired up.
Just help us make sure those less popular CPUs are not
messed up by changes.

And it seems like riscv64 is already broken in tree :-(

### DIFF
--- a/.github/workflows/dynamic_arch.yml
+++ b/.github/workflows/dynamic_arch.yml
@@ -257,3 +257,53 @@ jobs:
       - name: Run tests
         timeout-minutes: 60
         run: cd build && ctest
+
+  cross_build:
+    runs-on: ubuntu-22.04
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - target: mips64el
+            triple: mips64el-linux-gnuabi64
+            opts: DYNAMIC_ARCH=1
+          - target: riscv64
+            triple: riscv64-linux-gnu
+            opts: TARGET=RISCV64_GENERIC
+          - target: mipsel
+            triple: mipsel-linux-gnu
+            opts: TARGET=MIPS1004K
+          - target: alpha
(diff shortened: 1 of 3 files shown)


martin-frbg:
You need to supply a valid TARGET (as enumerated in TargetList.txt) when cross-compiling, I do not see that in your PR?

FlyGoat:
Ah I see, in #3720 I added `DYNAMIC_ARCH` support for MIPS64 and it will be able to probe arch and target from toolchain. For riscv64 it is incapable to do so. Will fix.

Thanks.

martin-frbg:
You need either DYNAMIC_ARCH or a TARGET, there is no build-time probing of cpu in cross-compiling as the target cpu will in most cases be on a completely different computer. RISCV already has RISCV64_GENERIC but no DYNAMIC_ARCH support (would be a bit overkill when it is just a choice between two targets, the optimized C910V and the generic one).

brada4:
Debian has cross-toolchains for all of its own arches, should be like 1:1 to just convert distro to debian 11 now, then extend if some niche arch gets problems some day.

FlyGoat:
> Debian has cross-toolchains for all of its own arches, should be like 1:1 to just convert distro to debian 11 now, then extend if some niche arch gets problems some day.

GitHub Action itself doesn't provide Debian VM. Using Docker seems a little bit overkill. Ubuntu 22.04 provides almost all cross toolchains shipped with Debian11 :-)

FlyGoat:
Now mipsel and alpha are wired up as well.
Alpha build issues are addressed by #3725.

martin-frbg:
unrelated manylinux build failure was fixed by my #3726 (guess this only came up as the manylinux instance lost its fortran compiler somehow). sorry for the delay in merging

PR #3723, opened 2022-08-10
LoongArch64: Rename targets

XiWeiGu:
The existing targets naming convention for LoongArch64 is not appropriate(lack of microarchitecture-specific thoughts).
Now I use LA464, LA264 to replace the original targets.


XiWeiGu:
@yinshiyou @martin-frbg please help me review those patches.

PR #3725, opened 2022-08-11
Alpha build fixes

FlyGoat:
Just make it compile with alpha cross toolchain.
Not tested but should be fine as there is no functional change.


martin-frbg:
Hehe, I wonder if anybody is still running a DEC Alpha somewhere (let alone with GotoBLAS/OpenBLAS on it)

FlyGoat:
> Hehe, I wonder if anybody is still running a DEC Alpha somewhere (let alone with GotoBLAS/OpenBLAS on it)

Theoretically sunway processors used by some Chinese HPCs are Alpha derivative but I'm sure they've got their own port. It looks like Debian/Alpha and Gentoo/Alpha is still active so it's probably worthy to maintain it in OpenBLAS.

martin-frbg:
I dimly remember someone asking about alpha support for some Chinese cpu, may well have been Sunway, but nothing became of it (or at least that person never followed up).

Issue #3740, opened 2022-08-25
SIGSEGV when tested with gcc-12 on macos


devcrocod:
```
os: macOS 12.5.1
arch: x86_64
C compiler: gcc-12 (Homebrew GCC 12.1.0) 12.1.0
Fortran compiler: GNU Fortran (Homebrew GCC 12.1.0) 12.1.0
OpenBLAS version: 0.3.20/0.3.21/develop
```

make command:
```bash
make CC=gcc-12 FC=gfortran-12 HOSTCC=gcc BINARY=64 F_COMPILER=GFORTRAN FEXTRALIB=-lgfortran USE_OPENMP=0 NO_AVX512=1 DYNAMIC_ARCH=1 NUM_THREADS=64
```

When testing Complex BLAS, I get segmentation fault:
```
 Complex BLAS Test Program Results


 Test of subprogram number  1            ZDOTC 

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x10ef5b90e
#1  0x10ef5aaed
#2  0x7ff80b637dfc
#3  0x10e7980c0
#4  0x10e798361
#5  0x10be5ee06
#6  0x10be5f0d4
#7  0x10e924e8e
make[5]: *** [level1] Segmentation fault: 11
make[5]: *** Waiting for unfinished jobs....

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x1060d990e
#1  0x1060d8aed
#2  0x7ff80b637dfc
#3  0x1059025a0
#4  0x105902841
#5  0x102f9ac86
#6  0x102f991f6
#7  0x102f92cd1
#8  0x102f979b3
#9  0x105a8f13e
/bin/sh: line 1: 72622 Segmentation fault: 11  OPENBLAS_NUM_THREADS=1 OMP_NUM_THREADS=1 ./zblat2 < ./zblat2.dat
```

martin-frbg:
hmm, this used to work with earlier gcc, so should not be anything fundamental (like internal representation of complex numbers).I wonder if this could be a regression in gcc-12 itself. (Have not updated gcc in our CI yet and am currently travelling)

martin-frbg:
fails at -O2, passes at just -g (so -O0), looks like overenthusiastic optimization by version 12.1 to me

devcrocod:
[...]

PR #3748, opened 2022-08-30
cmake: Allow specifying optimization level in COMMON_OPT

mmuetzel:
When building with `make` a default optimization level of `-O2` is used.
See:
https://github.com/xianyi/OpenBLAS/blob/00534523ad999d89945d23b7df0eafc69c31f1b3/Makefile.system#L1551-L1557

By default, `cmake` uses `-O3` for `Release` builds currently:
https://github.com/Kitware/CMake/blob/v3.24.1/Modules/Compiler/GNU.cmake#L59

This means binaries are built with different optimization levels depending on whether `make` or `cmake` is used.

Additionally, it is currently not possible to override the optimization level with, e.g., `-DCOMMON_OPT=-O2` because it appears before the `CMAKE_${Language}_FLAGS_RELEASE` flags (which effectively overrides the flag in `COMMON_OPT`).

See also: #3740

The proposed change sets the default compiler optimization level that is used with `cmake` to `-O2` (i.e., the same that is used by `make`).
Additionally, it allows overriding the compiler optimization level with flags in `COMMON_OPT`, `CCOMMON_OPT`, and `FCOMMON_OPT`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,6 +87,41 @@ if(MSVC AND MSVC_STATIC_CRT)
   endforeach()
 endif()
 
+# Set default optimization level
+if (NOT MSVC)
+  # Check if user specified optimization level in COMMON_OPT flags.
+  string(REGEX MATCH "-O[0-9]|-Ofast|-Og|-Os" OPENBLAS_COMMON_OPT_OFLAGS
+         "${COMMON_OPT} ${CCOMMON_OPT} ${FCOMMON_OPT}")
+
+  string(TOUPPER "${CMAKE_BUILD_TYPE}" CMAKE_BUILD_TYPE)
+  if (NOT "${OPENBLAS_COMMON_OPT_OFLAGS}" STREQUAL "")
+    # Filter out optimization level from cmake's default flags.
+    # They would override the optimization level set in COMMON_OPTs.
+    set(CompilerLangs
+        C
+        CXX
+        Fortran
+        ASM
+        )
+    foreach(CompilerLang ${CompilerLangs})
+      string(REGEX REPLACE "-O[0-9]|-Ofast|-Og|-Os" ""
+             CMAKE_${CompilerLang}_FLAGS_${CMAKE_BUILD_TYPE}
+             "${CMAKE_${CompilerLang}_FLAGS_${CMAKE_BUILD_TYPE}}")
+    endforeach()
+  endif ()
+  # Prepend default optimization level to COMMON_OPTs.
+  # These flags might be overridden by cmake's default flags.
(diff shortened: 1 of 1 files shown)


brada4:
Maybe it is worth playing warning, my rationale is that -Os has a practical purpose always when used, but say -Ofast guarantees numeric faults.

mmuetzel:
> Maybe it is worth playing warning, my rationale is that -Os has a practical purpose always when used, but say -Ofast guarantees numeric faults.

I'm not sure I understand your comment.
The proposed change doesn't touch any flags supplied by the user in the `COMMON_OPT` variables. And I'd argue that this should stay as it is. The build system should assume that a user knows what they are doing when setting custom flags...

brada4:
PR is not wrong, just that probably needs some discussion/consideration.

mmuetzel:
> PR is not wrong, just that probably needs some discussion/consideration.

I'm happy to discuss. But I don't understand *which* changes you are proposing.

mmuetzel:
I'm not sure why one of the checks of this PR was failing:
https://github.com/xianyi/OpenBLAS/runs/8095576089?check_suite_focus=true

IIUC, that check uses `make` to build OpenBLAS. Afaict, this PR doesn't touch the Makefile rules.

martin-frbg:
> I'm not sure why one of the checks of this PR was failing:
Me neither, that cross-build check has been failing more or less randomly lately and I have not managed to reproduce this locally.

On the actual topic of your PR, I am unsure if it would be useful to change the well-defined CMAKE default behaviour for Release builds years after adding cmake support. Perhaps it might be more useful to just handle the case where the user declared no build type at all (as a novice user might), which currently would result in no `-O` option at all (I think) ?

mmuetzel:
Afaict, it is not completely uncommon that projects override the flags in `CMAKE_${Language}_FLAGS_${CMAKE_BUILD_TYPE}`. But I understand your point of view. 
We could just remove the block that modifies these flags from the PR.

At the same time, it might be surprising for users if optimization flags that they pass with the `COMMON_OPT` variables don't actually have any effect. (At least, it was for me.)

Should the build rules check if there are any `-O` flags in the `COMMON_OPT` variables and remove them from `cmake`'s default flags conditional on that? That would be possible. But it would be pretty involved (and maybe more error prone?).

mmuetzel:
I updated the PR to leave cmake's default optimization level alone unless the user explicitly specifies optimization flags in any of the `COMMON_OPT` variables.

brada4:
I think we set release_with_debuginfo for novice user case which roughly matches makefile behaviour.
-frecursive is a paramount, it is obscure name for thread safety in fortran....

PR #3750, opened 2022-09-01
CI (MSYS2): Configure with `-DCMAKE_BUILD_TYPE=Release`.

mmuetzel:
Explicitly set `CMAKE_BUILD_TYPE` to `Release` for the runners that use MSYS2. IIUC, `None` was used before because no build type was specified.

The self-tests with gcc 12 are likely to fail with this change because of #3740.

### DIFF
--- a/.github/workflows/dynamic_arch.yml
+++ b/.github/workflows/dynamic_arch.yml
@@ -150,6 +150,7 @@ jobs:
       matrix:
         msystem: [MINGW64, MINGW32, CLANG64]
         idx: [int32, int64]
+        build-type: [Release]
         include:
           - msystem: MINGW64
             idx: int32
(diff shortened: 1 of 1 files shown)


mmuetzel:
I erroneously assumed that some self-test would trigger the error reported in https://github.com/msys2/MINGW-packages/issues/12857. But that doesn't seem to be the case.

Should the reproducer from that downstream report be added to the test suite?

mmuetzel:
Like suggested by @martin-frbg in #3740, I added a runner with build type `None`. Instead of blowing up the entire build matrix, I opted for adding a single runner with that configuration. I hope that is also good enough.

martin-frbg:
Thank you. (I do not think it is necessary to add the reproducer from the msys issue, and the underlying problem appears to be in the specific compiler version rather than OpenBLAS anyway)

PR #3758, opened 2022-09-07
Remove excessive quoting of arguments in Makefile.prebuild again

### LINKED ISSUE #3755: Fortran detection causing issues with AVX512 building-Gnu fortran12
Platform: 

<details><summary>Neofetch: </summary>

OS: Pop!_OS 22.04 LTS x86_64 

Kernel: 5.19.0-local-fclc 

Shell: bash 5.1.16 

CPU: 12th Gen Intel i7-12700K (16) 

GPU: AMD ATI Radeon RX 5600 OEM/5600 XT / 5700/5700 XT 

GPU: Intel AlderLake-S GT1 

Memory: 13152MiB / 31801MiB 

Compilers: GCC/G++/Gfort-12
</details>

Whilst I work on PR #3754 I've come across a problem with detection and propagation of AVX512 which will subsequently cause a build to fail. 

Initially, any build target including `make, make TARGET=SKYLAKEX, make TARGET=COOPERLAKE` etc. (targets that enable AVX512) would fail.

I was able to track this down to my gfortran-12 being linked to `/usr/bin/fortcom` where as OpenBLAS expected `/usr/bin/fortc`

thinking this could be related I then tried soft linking gfortran-12 to fortc. 

This allowed AVX512 builds when using a specified target. 

Currently this only leaves builds using `make` with no target being problematic. 

Something else I've noticed is that, while building with --silent (to make catching errors easier to find) OpenBLAS still seems to be strugling with Gfortran-12, sending out the error message of
`OpenBLAS: Detecting fortran compiler failed. Can only compile BLAS and f2c-converted LAPACK.`

martin-frbg:
fixes #3755 , bug introduced in #3722

### DIFF
--- a/Makefile.prebuild
+++ b/Makefile.prebuild
@@ -60,9 +60,9 @@ all: getarch_2nd
 	./getarch_2nd  1 >> $(TARGET_CONF)
 
 $(TARGET_CONF): c_check$(SCRIPTSUFFIX) f_check$(SCRIPTSUFFIX) getarch
-	./c_check$(SCRIPTSUFFIX) $(TARGET_MAKE) $(TARGET_CONF) "$(CC)" "$(TARGET_FLAGS) $(CFLAGS)"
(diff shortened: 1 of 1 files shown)


FCLC:
tested and closes #3755

PR #3763, opened 2022-09-13
MIPS64: Using the macro MTC rather than MTC1

### LINKED ISSUE #3761: test failures on mips64 in dnrm2_tiny and dnrm2_inf
In OpenBLAS 0.3.21, on mips64, the Debian package fails to build because there are test failures in the utests for `dnrm2_tiny` and `dnrm2_inf`:
```
TEST 32/37 dnrm2:dnrm2_tiny [FAIL]
  ERR: test_dnrm2.c:65  expected 0.000e+00, got inf (diff -inf, tol 1.000e-13)
TEST 33/37 dnrm2:dnrm2_inf [FAIL]
  ERR: test_dnrm2.c:52  expected inf, got 0.000e+00 (diff inf, tol 1.000e-13)
```
(see https://buildd.debian.org/status/fetch.php?pkg=openblas&arch=mips64el&ver=0.3.21%2Bds-1&stamp=1662336716&raw=0 for the full build log).

I noticed that commit cce4b1d9562d60b875982ce9587c68fbac666ec8 by @XiWeiGu was supposed to fix `dnrm2_tiny`.

Interestingly, if I revert that very commit, then it fixes `dnrm2_inf`:
```
TEST 32/37 dnrm2:dnrm2_tiny [FAIL]
  ERR: test_dnrm2.c:65  expected 0.000e+00, got inf (diff -inf, tol 1.000e-13)                                                    
TEST 33/37 dnrm2:dnrm2_inf [OK]
```

So it seems that the logic of commit cce4b1d9562d60b875982ce9587c68fbac666ec8 is wrong, and that instead of fixing `dnrm2_tiny`, it broke `dnrm2_inf`.

XiWeiGu:
Fixed #3761

### DIFF
--- a/kernel/mips64/dnrm2.S
+++ b/kernel/mips64/dnrm2.S
@@ -90,7 +90,7 @@
 	//Init INF
 	lui     TEMP, 0x7FF0
 	dsll    TEMP, TEMP, 32
-	MTC1    TEMP, INF
+	MTC     TEMP, INF
 
 	LD	a1,  0 * SIZE(X)
 	daddiu	N, N, -1


martin-frbg:
Thank you very much, this does indeed fix the problem. (Not sure how this slipped through, perhaps I should add a qemu-based CI job so that we do not have to rely on my manual testing)

PR #3770, opened 2022-09-16
Update Makefile of ReLAPACK

amontoison:
We need this modification if we want to compile `ReLAPACK` and also not override the .o and symbols generated by LAPACK (`INCLUDE_ALL=0` in `relapack/config.h`).

### DIFF
--- a/relapack/Makefile
+++ b/relapack/Makefile
@@ -6,35 +6,35 @@ include $(TOPDIR)/Makefile.system
 SRC = $(wildcard src/*.c)
 
 SRC1 = \
-	src/slauum.c src/clauum.c src/dlauum.c src/zlauum.c \
-	src/strtri.c src/dtrtri.c src/ctrtri.c src/ztrtri.c \
-	src/spotrf.c src/dpotrf.c src/cpotrf.c src/zpotrf.c \
-	src/sgetrf.c src/dgetrf.c src/cgetrf.c src/zgetrf.c
+	slauum.c clauum.c dlauum.c zlauum.c \
+	strtri.c dtrtri.c ctrtri.c ztrtri.c \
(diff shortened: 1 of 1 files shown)


martin-frbg:
Ahem, wouldn't that break the default build by enforcing relapack prefixes even when not desired ?

amontoison:
You're right @martin-frbg, we should only add a prefix if `INCLUDE_ALL=0` in [relapack/config.h](https://github.com/xianyi/OpenBLAS/blob/develop/relapack/config.h#L48).
If we have two `.o` with different names but the same exported symbols, the library will use the routines of the first `.o` file added in the library.

Do you have an idea how can check the value of `INCLUDE_ALL` from `relapack/Makefile`?
We could also just define `RELAPACK_PREFIX` in the same spirit of `SUFFIX` in the `relapack/Makefile` for more flexibility.

martin-frbg:
It would probably make sense to have tighter integration with the OpenBLAS makefiles, i.e. moving the INCLUDE_ALL from config.h to a RELAPACK_INCLUDEALL in Makefile.rule (so that one does not have to hack around in the ReLAPACK internals) and then having the ReLAPACK makefile add `-DINCLUDE_ALL=$RELAPACK_INCLUDEALL` to the `%.o: %.c` build rule.
(Plus the equivalent for the cmake build system)

amontoison:
I saw that you opened #3796 and #3802, should I close this PR?

martin-frbg:
Sorry, I did not intend to deny you recognition but the changes got a bit out of hand, what with the preexisting bugs in the cmake files and the need to integrate with the hackish GEMMT I already had lying around

amontoison:
It's perfectly fine @martin-frbg! I don't know well CMake.
I also didn't know if I should update the symbols or RELAPACK in `OpenBLAS/exports` folder.

PR #3771, opened 2022-09-17
Add KERNEL file for MIPS64_GENERIC as a copy of GENERIC

martin-frbg:
(Hopefully temporary) fix for faulty DGEMM etc. with MIPS64_GENERIC, as without a KERNEL file the default selection of sources does not match the 2x2 MN defined in param.h. Ideally the MIPS64_GENERIC introduced with #3720 needs to be removed again, and the getarch code improved instead to properly distinguish between MIPS32 and MIPS64 on its own. (It has always been possible to build with TARGET=GENERIC, but autodetection of the "GENERIC" target never worked)

### DIFF
--- a/kernel/mips64/KERNEL.MIPS64_GENERIC
+++ b/kernel/mips64/KERNEL.MIPS64_GENERIC
@@ -1,0 +1,160 @@
+SGEMM_BETA = ../generic/gemm_beta.c
+DGEMM_BETA = ../generic/gemm_beta.c
+CGEMM_BETA = ../generic/zgemm_beta.c
+ZGEMM_BETA = ../generic/zgemm_beta.c
+
+STRMMKERNEL	= ../generic/trmmkernel_2x2.c
+DTRMMKERNEL	= ../generic/trmmkernel_2x2.c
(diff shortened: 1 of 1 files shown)


XiWeiGu:
There are many `GENERIC`. At the very top we have `TARGET=GENERIC`, for `RISC-V 64` we have `TARGET=RISCV64_GENERIC`, for `System Z` we have `TARGET=ZARCH_GENERIC` and so on. This might be complicated.
`GENERIC` is meant to  use the C kernel and turn off platform optimization. Can we just keep one `TARGET=GENERIC` and 
remove the others?
If something is wrong with my understanding,please pardon me.

martin-frbg:
Correct, this is just a quick fix to silence the CI jobs. But at least for mips this is complicated as autodetection (c_check, getarch etc) does not correctly determine 32 or 64bit ABI right now. i guess this is why FlyGoat added MIPS64_GENERIC to use in DYNAMIC_ARCH. The other NAMED_GENERIC _may_ have additional parameters compared to, err, generic GENERIC - need to look into that.

Issue #1237, opened 2017-07-13
Denormal floats handling


Arech:
Dear @xianyi and everyone!

Is there a way to forbid denormals in the OpenBLAS?

I tried to execute the following code (MSVC2015) before calling any OpenBLAS routines:

		unsigned int current_word = 0;
		_controlfp_s(&current_word, _DN_FLUSH, _MCW_DN);
		_MM_SET_DENORMALS_ZERO_MODE(_MM_DENORMALS_ZERO_ON);
		_MM_SET_FLUSH_ZERO_MODE(_MM_FLUSH_ZERO_ON);

and it probably worked for some time (about a year)... However, I've just noticed that denormals started to reappear in results and that extremely slows down computations. I'm heavily suspecting that it is the OpenBLAS, who is responsible for them. Probably, the worker threads might still have enabled denormals?...

How to get rid of them?

I'm using precompiled OpenBLAS-v0.2.19-Win64-int32.

Here is the complete list of used functions (if it helps):

- cblas_sgemm
- cblas_ssyrk
- cblas_ssymm
- LAPACKE_sgesvd

brada4:
Care to mention processor ID?
OpenBLAS is not compiled by MSVC, and unlikely MSVC macros will have any influence whatsoever on imported DLL file.
You may use CONSISTENT_FPCSR=1 build option to get consistent float handling between SSE and AVX

Arech:
> Care to mention processor ID?

AMD Phenom II X6 1090, if it matters... It doesn't have AVX, SSE only.

> OpenBLAS is not compiled by MSVC, and unlikely MSVC macros will have any influence whatsoever on imported DLL file.

These functions+macroses change processor state (some bits in processor's control registers), therefore they do affect how subsequent code is executed by a processor, don't they?

brada4:
I asked you if passing CONSISTENT_FPCSR compile option to OpenBLAS solves your problem.

Can you provide sample code that reproduces your problem?

Arech:
> I asked you if passing CONSISTENT_FPCSR compile option to OpenBLAS solves your problem.

I'm going to try it, however it'll require a plenty of time to setup build environment for OpenBLAS.

> Can you provide sample code that reproduces your problem?

Problem occurs during fairly complex neural network training using my [nntl project](https://github.com/Arech/nntl). I'll try to isolate the problem source and make a sample code out of it to reduce the number of unrelated code.

What I learned till this moment is this:

[...]

PR #3772, opened 2022-09-21
Support CONSISTENT_FPCSR on aarch64 systems

siko1056:
As stated in `Makefile.rule`, OpenBLAS supports status register synchronization x86/x86_64, if compiled with the `CONSISTENT_FPCSR` flag.  By default this feature is disabled and this PR will not affect most of OpenBLAS users.  However, this feature is of some importance for numerical research in verified computations.

This PR extends this feature with minimal code changes to the aarch64 platform (e.g. Apple M1), which is frequently used by researchers nowadays.

I would greatly appreciate the inclusion of this patch and thank you for your great work with the OpenBLAS library!

### DIFF
--- a/Makefile.rule
+++ b/Makefile.rule
@@ -207,7 +207,7 @@ NO_AFFINITY = 1
 # to the user space. If bigphysarea is enabled, it will use it.
 # DEVICEDRIVER_ALLOCATION = 1
 
-# If you need to synchronize FP CSR between threads (for x86/x86_64 only).
+# If you need to synchronize FP CSR between threads (for x86/x86_64 and aarch64 only).
 # CONSISTENT_FPCSR = 1
 
 # If any gemm argument m, n or k is less or equal this threshold, gemm will be execute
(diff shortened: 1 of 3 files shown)


martin-frbg:
Thanks, makes sense. BTW so far my impression has been that CSR inconsistency is mainly an issue on Windows systems, is that a misconception and you would rather see this option active by default on the relevant hardware ?

siko1056:
Thank you for accepting the patch @martin-frbg .

In my field of research, this CSR inconsistency exist on all "commonly used" platforms (x86_64 / aarch64) and operating systems (mac, Windows, Linux) and I always compile OpenBLAS having this flag set.

I cannot point to open OpenBLAS bugs that would be fixed by enabling this flag by default, but there are claims that setting `CONSISTENT_FPCSR = 1` by default might cause a performance loss.  I do not have benchmarks supporting or objecting this statement.   See for example your comment from 2017: https://github.com/xianyi/OpenBLAS/issues/1237#issuecomment-316336665

Personally, I never experience any drawbacks, but more predictable and correct results (for interval arithmetic, for example, the rounding mode is switched via the CSR and having control over it is crucial).  Maybe a representative benchmark should be done before making it a default enabled option.  I would greatly appreciate seeing this flag enabled by default.

PR #3773, opened 2022-09-22
Add `OPENBLAS_DEFAULT_NUM_THREADS`

staticfloat:
This allows Julia to set a default number of threads (usually `1`) to be used when no other thread counts are specified [0], to short-circuit the default OpenBLAS thread initialization routine that spins up a different number of threads than Julia would otherwise choose.

The reason to add a new environment variable is that we want to be able to configure OpenBLAS to avoid performing its initial memory allocation/thread startup, as that can consume significant amounts of memory, but we still want to be sensitive to legacy codebases that set things like `OMP_NUM_THREADS` or `GOTOBLAS_NUM_THREADS`.  Creating a new environment variable that is openblas-specific and is not already publicly used to control the overall number of threads of programs like Julia seems to be the best way forward.

[0] https://github.com/JuliaLang/julia/pull/46844

### DIFF
--- a/driver/others/init.c
+++ b/driver/others/init.c
@@ -823,6 +823,8 @@ void gotoblas_affinity_init(void) {
 
   if (numprocs == 0) numprocs = readenv_atoi("OMP_NUM_THREADS");
 
+  if (numprocs == 0) numprocs = readenv_atoi("OPENBLAS_DEFAULT_NUM_THREADS");
+
   numnodes = 1;
 
   if (numprocs == 1) {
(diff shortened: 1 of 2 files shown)


martin-frbg:
Sorry for the delay here - I _guess_ it would be nicer if the existing OPENBLAS_NUM_THREADS could be reworked to act the same with both pthreads and OpenMP, but a way to limit the thread count for just this library rather than the entire process does certainly make sense in any case.

PR #3793, opened 2022-10-19
New sbgemm implementation for Neoverse N2

imzhuhl:
This is a new implementation of sbgemm that use the BFMMLA instruction on Neoverse N2.

1. Use UZP instructions but not gather load and scatter store instructions to get lower latency.
2. Padding k to a power of 4.

Why a new implementation is needed? The main reason is performance. The new implementation achieves about 130 GFLOPS on a single-core neoverse n2, which is significantly higher than the old implementation (approximately 90 GFLOPS). The issue of the old implementations is here: #3671 

There are two main changes. 
First, I use the UZP instructions. This can converts gather load and scatter store to contiguous load and store with much lower latency.
Second, padding in the dimension k so that it is a multiple of 4. The input of the BFMMLA instruction are a 2x4 and 4x2 matrix, here m = n = 2, and k = 4(on Neoverse N2, vector length is 128bit). If the kernel function inputs a matrix whose k is not a multiple of 4, we then have to handle the different value of k in each set of iterations, like that:
```c
// kernel/arm64/sbgemm_kernel_8x4_neoversen2_impl.c
if (krest) {
    if (krest == 1) {
        LOAD_KREST_1(a, 0); LOAD_KREST_1(a, 1);
        LOAD_KREST_1(a, 2); LOAD_KREST_1(a, 3);
        LOAD_KREST_1(b, 0); LOAD_KREST_1(b, 1);
    } else if (krest == 2) {
        LOAD_KREST_2(a, 0); LOAD_KREST_2(a, 1);
        LOAD_KREST_2(a, 2); LOAD_KREST_2(a, 3);
        LOAD_KREST_2(b, 0); LOAD_KREST_2(b, 1);
    } else if (krest == 3) {
        LOAD_KREST_3(a, 0); LOAD_KREST_3(a, 1);
        LOAD_KREST_3(a, 2); LOAD_KREST_3(a, 3);
        LOAD_KREST_3(b, 0); LOAD_KREST_3(b, 1);
    }
    MATMUL(0, 0); MATMUL(0, 1);
    MATMUL(1, 0); MATMUL(1, 1);
    MATMUL(2, 0); MATMUL(2, 1);
    MATMUL(3, 0); MATMUL(3, 1);
}
```
This reduces performance and makes the code tedious. Now, I pad k in ncopy and tcopy, so that in the kernel function, k is calculated as a multiple of 4. This step will inevitably require changes to the code of `level3.c` and `thread_level3.c`, because the offset of the buffer needs to be modified.

This commit involves code changes to the driver part rather than simply modifying the kernel part, and I'm not sure if we should accepts this solution.

if it is acceptable, I need to do the following things (TODO):
1. Distinguish whether alpha is equal to 1.0 to further optimization.
2. Change the file name to match the norm.
3. Delete useless code and modify the naming of some codes.


martin-frbg:
The padding in the level3 driver is a bit problematic as that `ifdef` will not work in DYNAMIC_ARCH builds (where the common code, i.e. everything in interface/level2/level3/other is only compiled once)

imzhuhl:
> The padding in the level3 driver is a bit problematic as that `ifdef` will not work in DYNAMIC_ARCH builds (where the common code, i.e. everything in interface/level2/level3/other is only compiled once)

Yes, I add one line `align_k` in gotoblas_t, then level3 could get gotoblas->align_k in DYNAMIC_ARCH builds. the align_k will init at init_parameter func in setparam_$(TARGET_CORE).c file, so we could assign align_k different values for different architectures.

martin-frbg:
Hmm yes that would work, though it creates a weird duality where all the existing "_k" symbols in common_param.h are pointers to BLAS kernels. More importantly, "hiding away" cpu-specific parameters in individual source files is likely to create interesting problems later as new cpus get added. I wonder if this could be changed to a `#define SBGEMM_ALIGN_K` in param.h  that gets used in setparam_ref.c (and the level3 files) so that cpu-specific parametrization stays limited to param.h ?

imzhuhl:
> Hmm yes that would work, though it creates a weird duality where all the existing "_k" symbols in common_param.h are pointers to BLAS kernels. More importantly, "hiding away" cpu-specific parameters in individual source files is likely to create interesting problems later as new cpus get added. I wonder if this could be changed to a `#define SBGEMM_ALIGN_K` in param.h that gets used in setparam_ref.c (and the level3 files) so that cpu-specific parametrization stays limited to param.h ?

Good suggestion, I've modified the param.h file.

martin-frbg:
Great, thank you.

PR #3795, opened 2022-10-21
Use FMA for cscal and zscal Haswell microkernels.

bartoldeman:
This patch has two benefits:
1. Using vfmaddsub231p[sd] instead of vaddsubp[sd] eliminates a vmulp[sd] instruction, giving a ~10% speedup, measured from ~33 to ~36 Gflops for cscal with 4096 elements and from ~17 to ~19 Gflops for zscal on my Kaby Lake laptop, see e.g. OPENBLAS_LOOPS=10000 benchmark/cscal.goto 4096 4096.

2. Using it for both the main loop and the tail end makes sure the same FMA instruction is used for all loop iterations, which is not the case with the current situation where the tail loop is implemented in C, if the compiler is allowed to use FMA instructions. This is important for some LAPACK eigenvalue testcases that rely on bitwise identical results independent of how many loop iterations are used.

### DIFF
--- a/kernel/x86_64/cscal.c
+++ b/kernel/x86_64/cscal.c
@@ -337,6 +337,15 @@ int CNAME(BLASLONG n, BLASLONG dummy0, BLASLONG dummy1, FLOAT da_r, FLOAT da_i, 
 	}
 
 
+#ifdef HAVE_KERNEL
+	if ( da_r != 0.0 && da_i != 0.0 ) {
+		alpha[0] = da_r;
+		alpha[1] = da_i;
+		cscal_kernel(n , alpha, x);
+		return(0);
+	}
+#endif
+
 	BLASLONG n1 = n & -16;
 	if ( n1 > 0 )
 	{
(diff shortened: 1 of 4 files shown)


bartoldeman:
I put this to draft since I bumped into https://github.com/xianyi/OpenBLAS/commit/22aa81f3e587c
The alternative is then to force no FMAs in the C code, of course, using e.g.
`#pragma GCC optimize ("fp-contract=off")`
from that patch or some other technique.

martin-frbg:
oh right, I had a feeling of deja vu but was too tired to search yesterday. at this point I am no longer sure if we should even rely this much on the LAPACK testsuite or simply adjust its thresholds to allow "healthy" deviations.

bartoldeman:
I played a bit with the thresholds but there were still some hard cases that fail there.
There is some annoying behaviour with using FMAs to multiply complex numbers, in particular as 22aa81f3e587c noted, when the imaginary part of the result is close to 0, e.g. with `(a+bi) * (a-bi) = a**2 + b**2`; without FMAs the imaginary part of the result is always 0, but with them you can end up with e.g. `-5.55112e-17` for `a=3.0` and `b=1/3.0` due to the asymmetry in `a*b-a*b` with FMAs (of course we know that a*b isn't exactly 1.0, that's well known numerically, but subtraction of two things rounded the same way still gives 0).

There are a few relevant places in LAPACK where they compute a * conj(a), and that doesn't help for sure, but for now it simply seems best to turn off FMAs for complex multiplications to avoid surprising behaviour.

PR #3799, opened 2022-10-27
x86_64: prevent GCC and Clang from generating FMAs in cscal/zscal.

bartoldeman:
If e.g. -march=haswell is set in CFLAGS, GCC generates FMAs by default, which is inconsistent with the microkernels, none of which use FMAs. These inconsistencies cause a few failures in the LAPACK testcases, where eigenvalue results with/without eigenvectors are compared.

Moreover using FMAs for multiplication of complex numbers can give surprising results, see 22aa81f for more information.

This uses the same syntax as used in 22aa81f for zarch (s390x).


martin-frbg:
Thank you.

PR #3800, opened 2022-10-28
add raptor lake ids

thrasibule:
tested on a i7-13700k. I used parameters from here: https://en.wikichip.org/wiki/intel/microarchitectures/raptor_lake


martin-frbg:
Thanks.

Issue #3805, opened 2022-11-03
Building OpenBLAS v0.3.21 with spack stop working


fspiga:
See documented issue here: https://github.com/spack/spack/issues/33693

martin-frbg:
Looks like you are building with DYNAMIC_ARCH support, i.e. all supported arm64 targets, and your compiler is too old to understand the `-march` option for the more recently added NeoverseN2 cpus. Which is a bit weird as this is guarded against in Makefile.arm64, requiring a sufficiently recent gcc.  You appear to be using 11.2(?), while a comment in the Makefile claims that 
N2 support was added in gcc-9.4

martin-frbg:
Hmm, the "9.4 or newer" looks like copypasta from the N1 section, but gcc online docs suggest that 10.x onwards should support all of `+sve` `+sve2` `+bf16` feature modifiers as used in the `-march` option. @imzhuhl which gcc version(s) are you using to develop and test your contributions ?

imzhuhl:
```bash
root@62869467bcb3:/# gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/aarch64-linux-gnu/11/lto-wrapper
Target: aarch64-linux-gnu
[...]

PR #3811, opened 2022-11-09
Improve gcc arch option selecting for Neoverse cpus

martin-frbg:
related to #3805


Makefile.arm64
```diff
@@ -119,11 +119,11 @@ endif
 endif
 
 # Use a72 tunings because Neoverse-N2 is only available
-# in GCC>=9.4
+# in GCC>=10.4
```
giordano, line 122: This should have been 10.1?  It's only a comment, but it's confusing when you read the code below

martin-frbg: the code says 10.4 too - either a minor version of 4 or major 11 and above

PR #3819, opened 2022-11-13
Fix implicit conversions and unused variables in LAPACKE and TESTING (Reference-LAPACK PR 703)

martin-frbg:


martin-frbg:
(relies on functionality from lapack PR 651 (#3821)

PR #3842, opened 2022-11-25
Add SVE implementation for sdot/ddot

Mousius:
This adds an SVE implementation to sdot/ddot when available, falling back to the previous Advanced SIMD kernel where there's no SVE implementation for the kernel.

All the targets were essentially treating `dot_thunderx2t99.c` as the Advanced SIMD implementation so I've renamed it to better fit with the feature detection.

This patch does not intend to introduce further micro-architecture tuning, that would be a separate future patch if necessary.

### DIFF
--- a/kernel/arm64/dot.c
+++ b/kernel/arm64/dot.c
@@ -29,7 +29,17 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include "common.h"
 
+// Some compilers will report feature support for SVE without the appropriate
+// header available
 #ifdef HAVE_SVE
+#if defined __has_include 
+#if __has_include(<arm_sve.h>) && __ARM_FEATURE_SVE
+#define USE_SVE
+#endif 
+#endif
+#endif
+
+#ifdef USE_SVE
 #include "dot_kernel_sve.c"
 #endif
 #include "dot_kernel_asimd.c"
@@ -46,7 +56,7 @@ static RETURN_TYPE dot_compute(BLASLONG n, FLOAT *x, BLASLONG inc_x, FLOAT *y, B
 
 	if ( n <= 0 ) return dot;
 
-#ifdef HAVE_SVE
+#ifdef USE_SVE
 	if (inc_x == 1 && inc_y == 1) {
 		return dot_kernel_sve(n, x, y);
 	}
--- a/kernel/arm64/dot.c
+++ b/kernel/arm64/dot.c
@@ -1,0 +1,111 @@
+/***************************************************************************
+Copyright (c) 2017, The OpenBLAS Project
+Copyright (c) 2022, Arm Ltd
+All rights reserved.
+Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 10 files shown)


brada4:
If you examine remaining dot.c files ypu will figure more appropriate name

Mousius:
Hi @brada4, 

Can you elaborate on your comment? Are you in favour of renaming to match the architectural features in these other files or would you prefer it to remain targeting a specific micro-architecture which is different from the one being built for?

brada4:
Something as inspiring as axpy_sse3.c . the historic duplication of generic implementation is haunting around. I see your other ticket. It shows the generational development - maybe you add new file for N1 and N2 and additionally reconfigure thunderx? ?

Mousius:
The new file `dot_kernel_sve.c` is enabled by `HAVE_SVE`:
* [NEOVERSEV1](https://github.com/xianyi/OpenBLAS/blob/develop/cmake/prebuild.cmake#L967)
* [NEOVERSEN2](https://github.com/xianyi/OpenBLAS/blob/develop/cmake/prebuild.cmake#L995) 

Both micro-architectures support the architectural feature of SVE, they use the same kernel implementation by feature detection. This reduces the need for micro-architecture specific files in favour of the architectural features of AArch64. 

ThunderX doesn't support SVE, it doesn't get enabled with `HAVE_SVE`, the shared threading implementation and the Advanced SIMD kernel are used.

Mousius:
Also, this patch does not intend to introduce further micro-architecture tuning, that would be a separate patch if necessary.

kernel/arm64/dot.c
```diff
@@ -0,0 +1,121 @@
...
+OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
+USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*****************************************************************************/
+
+
+#include "common.h"
+
+// Some compilers will report feature support for SVE without the appropriate
+// header available
+#ifdef HAVE_SVE
```
Mousius, line 34: This can be moved to a header in a future patch to allow for re-use between kernels.

martin-frbg: Agreed - or alternatively the header presence could be checked by the c_check script (similar to how AVX512 is handled on x86_64)

martin-frbg:
I'm fine with the renaming of the file - it just so happens that thunderx2t99 was the most "modern" implementation we had, so I reused it for the more recent ARM cpus (while not yet confident enough to write new ones from scratch)

kernel/arm64/dot.c
Mousius: Ooh, I'll take a look into that :+1:

Mousius:
> I'm fine with the renaming of the file - it just so happens that thunderx2t99 was the most "modern" implementation we had, so I reused it for the more recent ARM cpus (while not yet confident enough to write new ones from scratch)

That sounds like a sensible approach to me 😸 and it works the same, ASIMD is still relevant

PR #3843, opened 2022-11-25
Propagate SWITCH_RATIO to DYNAMIC_ARCH builds

Mousius:
Previously dynamic builds were either using the default SWITCH_RATIO or one from the higher level architecture; this patch ensures the dynamic builds can use this parameter as well.

### DIFF
--- a/common_param.h
+++ b/common_param.h
@@ -1,5 +1,6 @@
 /*********************************************************************/
 /* Copyright 2009, 2010 The University of Texas at Austin.           */
+/* Copyright 2023 The OpenBLAS Project.                              */
 /* All rights reserved.                                              */
 /*                                                                   */
 /* Redistribution and use in source and binary forms, with or        */
@@ -45,6 +46,7 @@
 
 typedef struct {
   int dtb_entries;
+  int switch_ratio;
   int offsetA, offsetB, align;
 
 #ifdef BUILD_BFLOAT16
--- a/driver/level3/level3_gemm3m_thread.c
+++ b/driver/level3/level3_gemm3m_thread.c
@@ -1,5 +1,6 @@
 /*********************************************************************/
 /* Copyright 2009, 2010 The University of Texas at Austin.           */
+/* Copyright 2023 The OpenBLAS Project.                              */
 /* All rights reserved.                                              */
 /*                                                                   */
 /* Redistribution and use in source and binary forms, with or        */
@@ -42,10 +43,6 @@
 
 #ifndef DIVIDE_RATE
 #define DIVIDE_RATE 2
-#endif
-
-#ifndef SWITCH_RATIO
-#define SWITCH_RATIO 2
 #endif
 
 //The array of job_t may overflow the stack.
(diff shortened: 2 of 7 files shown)


Mousius:
@martin-frbg can we also get this to 0.3.22? it should be a significant speed boost for those who've tuned this parameter and `DYNAMIC_ARCH` is used by some core data science packages (i.e. `numpy`, `scipy`)

martin-frbg:
No fundamental objection, I'm just busy with other things. (BTW do you need the Copyright lines on everything, even where you just moved a line from one file to another ? As far as I can tell, no other contributor ever bothered with this)

Mousius:
No worries, I just noticed you added [0.3.22](https://github.com/xianyi/OpenBLAS/milestone/41) to my other branch and wanted to check this one could also make that release 😸 

And you're quite right on the copyright, I applied it incorrectly to all files, the move isn't valid - sorry about that!

martin-frbg:
Ah don't read too much into that - I am mainly tagging PRs to make it easier to write the eventual release message and changelog, so sometimes contributions get tagged right when I merge them. So not a lot of cunning planning behind my moves

martin-frbg:
@xianyi are you OK with the additional copyright notices ? This appears to be a first...

Mousius:
@martin-frbg any update on this?

martin-frbg:
Very intermittent communication... would still prefer if you added yourself to CONTRIBUTORS.md like everybody else for the time being. But can easily defer this to 0.3.23, I'm struggling this week to get 0.3.22 released amid other commitments. (Next release should not take that long again or I'll lose my mind anyway...)

Mousius:
I'm not really sure what the purpose of the CONTRIBUTORS.md is? It looks like it's a subset of the git log? There only appears to be copyright claims in the various files and then in the LICENSE - so potentially there? Fear not though, I'll ask where the best place would be, was still under the impression we could use the more explicit approach despite the hostility in the linting PR.

martin-frbg:
> I'm not really sure what the purpose of the CONTRIBUTORS.md is? It looks like it's a subset of the git log? 

git log without the granularity, I guess others may call such a thing CREDITS.txt  . Anyway this seems to have served the project well enough since its inception (can't count the other 10+ years of GotoBLAS before that as that seems to have been strictly a one-man project). 

> There only appears to be copyright claims in the various files and then in the LICENSE - so potentially there?

Not sure if it even makes sense to add copyright claims to individual files (where they would seem to cover only a handful of lines actually affected by the change - and what to do with attributions of all previous changes, some likely pre-git ?) but I'm mostly trying to avoid opening a can of worms. Some files have "Copyright ... The OpenBLAS project", not sure if that counts for anything (I recall FLTK for one using "Copyright Bill Spitzak and others" which I _guess_ it at least as unspecific.

> Fear not though, I'll ask where the best place would be, was still under the impression we could use the more explicit approach despite the hostility in the linting PR.

My impression was that the FSFE tool was too much geared towards GPL-licensed projects, and poor at recognizing Copyright/License statements that are not verbose copies of well-known GPL/BSD/MIT headers. By not recognizing
copyright statements that are actually present if a bit non-standard and/or buried in the comments at the head of the file it seemed to create more work than it would appear to save. Rearranging our handling of the imported LAPACK tree or even the files themselves just to cater to its stubbornness certainly seemed excessive - at that point IMHO it would make more sense to go through the actual OpenBLAS files by hand IFF a suitable Copyright header was deemed necessary and agreed on.

Mousius:
@martin-frbg I think this and #3855 are now acceptable? Let me know if I can do anything more 😸

PR #3844, opened 2022-11-25
Set SWITCH_RATIO for Arm(R) Neoverse(TM) V1 CPUs

Mousius:
From testing this yields better results than the default of `2`.


martin-frbg:
Thank you. I get the feeling SWITCH_RATIO may be an underappreciated parameter for other arm64 targets as well

Mousius:
> Thank you. I get the feeling SWITCH_RATIO may be an underappreciated parameter for other arm64 targets as well

Agree, I'll see if I can tweak a few more 😸

PR #3845, opened 2022-11-25
Remove unnecessary instructions from Advanced SIMD dot

Mousius:
The existing kernel was issuing extra instructions to organise the arguments into the same registers they would usually be in and similarly to put the result into the appropriate register.

This has an impact on smaller sized dots and seemed like a quick fix

### DIFF
--- a/kernel/arm64/dot_thunderx2t99.c
+++ b/kernel/arm64/dot_thunderx2t99.c
@@ -1,5 +1,6 @@
 /***************************************************************************
 Copyright (c) 2017, The OpenBLAS Project
+Copyright (c) 2022, Arm Ltd
 All rights reserved.
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are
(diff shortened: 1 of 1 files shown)


martin-frbg:
with code like this I am always unsure if it simply reflects the thought process of its development, or if it was (and still is) a necessity to work around a deficiency in some particular assembler version

Mousius:
> with code like this I am always unsure if it simply reflects the thought process of its development, or if it was (and still is) a necessity to work around a deficiency in some particular assembler version

I understand your concerns, in this case it's most likely to be the thought process of the developer, the [calling convention for AArch64](https://github.com/ARM-software/abi-aa/blob/main/aapcs64/aapcs64.rst#parameter-passing) is standardised to use the `x0-x7` registers for these parameters and the inline assembly is standard between GCC/Clang. 

Follow on question, is there a minimum compiler supported by OpenBLAS? I can see some guidance on https://github.com/xianyi/OpenBLAS#troubleshooting but many of those versions are now unsupported ([GCC 10 being the oldest supported by the GCC developers](https://gcc.gnu.org/gcc-10/)).

martin-frbg:
I personally prefer to keep it compilable with whatever old version a user may have, which may well include versions that are no longer officially supported. Some Linux distributions come with fairly old compilers by default, and users on shared systems may not have the priviledge to update them (nor the time and experience to get a local installation of a newer compiler working, especially within the constraints of an equally old system libc). So maybe no longer anything as historic as gcc4 nowadays, but compatibiltiy with gcc 7.x would definitely be desirable.

martin-frbg:
I have to admit I'm a bit confused about the relationship between this PR and #3842 - guess this is supposed to go in first ?

Mousius:
It can be done in either order, I just need to fix up after the other goes in - given #3842 is green I would lean towards merging that and I can rebase this one on top of it?

martin-frbg:
Whatever makes more sense - the single failure here is spurious, some of the CI jobs simply time out occasionally as Azure assigns either fast or slow hardware at random.

Mousius:
@martin-frbg happy to merge this one first then 😸

PR #3846, opened 2022-11-28
Improve the performance of sbgemm_tcopy on neoversen2

lilh9598:
Based the implementation of https://github.com/xianyi/OpenBLAS/pull/3793 by @imzhuhl , I use neon to improve the performance of sbgemm_tcopy in this PR.   

I benchmarked cblas_sbgemm (m=n=k=1000) on neoversen2(2.75GHz), and the performance improved from 134.52468 GFLOPS to 143.59852 GFLOPS.

### DIFF
--- a/kernel/arm64/KERNEL.NEOVERSEN2
+++ b/kernel/arm64/KERNEL.NEOVERSEN2
@@ -190,10 +190,10 @@ ZGEMMOTCOPYOBJ =  zgemm_otcopy$(TSUFFIX).$(SUFFIX)
 
 SBGEMM_BETA    =  sbgemm_beta_neoversen2.c
 SBGEMMKERNEL    = sbgemm_kernel_$(SBGEMM_UNROLL_M)x$(SBGEMM_UNROLL_N)_neoversen2.c
-SBGEMMINCOPY    = sbgemm_ncopy_$(SBGEMM_UNROLL_N)_neoversen2.c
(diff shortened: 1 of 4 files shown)


lilh9598:
I found that sbgemm_ncopy_8_neoversen2 and sbgemm_tcopy_4_neoversen2 are missing, which will lead to wrong results from calling cblas_sbgemm(CblasColMajor, **CblasTrans**, **CblasTrans**...). To fix this bug, I implement these two functions and add a new commit.

imzhuhl:
> Based the implementation of #3793 by @imzhuhl , I use neon to improve the performance of sbgemm_tcopy in this PR.
> 
> I benchmarked cblas_sbgemm (m=n=k=1000) on neoversen2(2.75GHz), and the performance improved from 134.52468 GFLOPS to 143.59852 GFLOPS.

Great, I've confirmed the performance improvement.

> I found that sbgemm_ncopy_8_neoversen2 and sbgemm_tcopy_4_neoversen2 are missing, which will lead to wrong results from calling cblas_sbgemm(CblasColMajor, **CblasTrans**, **CblasTrans**...). To fix this bug, I implement these two functions and add a new commit.

We should consider the case of Trans. Thank you for the correction.

But there may be a small mistake in sbgemm_ncopy_8_neoversen2.c:

image

The unroll number is 8 now and we may need to modify the code.

lilh9598:
> The unroll number is 8 now and we may need to modify the code.  

Fixed, thanks for your reminder!

martin-frbg:
Oops, sorry for almost forgetting to merge this

PR #3851, opened 2022-12-04
Allocate work array in LAPACKE ?TGSEN when ijob is zero (Reference-LAPACK PR 733)

martin-frbg:


martin-frbg:
this should have been "PR 773" rather than 733

PR #3853, opened 2022-12-06
Remove SVE from Arm(R) Neoverse(TM) N1 CPU in Makefile

Mousius:
I incorrectly added `+sve` to the Neoverse(TM) N1 CPU GCC parameters, which doesn't support SVE - this results in failed builds when using a compiler that doesn't support `-mtune=neoverse-n1` which appears to hide the mistake.


martin-frbg:
Oops. I even remember noticing that change but obviously forgot to ask.

Mousius:
Would it be possible to add GCC 8.x testing into CI @martin-frbg ? Unsure where/how we could do this but it would've caught this I think.

PR #3854, opened 2022-12-07
Travis CI: Add a DYNAMIC_ARCH build on Neoverse using older gcc8

martin-frbg:
for https://github.com/xianyi/OpenBLAS/pull/3853#issuecomment-1340727903


Mousius (approved):
Thanks @martin-frbg 😸 !

PR #3855, opened 2022-12-07
SWITCH_RATIO for Arm(R) Neoverse(TM) architecture

Mousius:
This seems like a good balance of values for reasonably sized matrices. With `SWITCH_RATIO=16` the DGEMM scales better to bigger sizes but the better solution would be some kind of thread throttling so I've gone with `SWITCH_RATIO=8`.


Mousius:
@martin-frbg as you suggested in #3844 I took a look at some of the other targets 😸

PR #3858, opened 2022-12-09
Patch for Intel Fortran compiler

glennpj:
This PR patches the f_check scripts to detect the ifort compiler. With out this, the value of F_COMPILER is G77 which causes the link phase to use icc rather than ifort, resulting in missing libifcore.

### DIFF
--- a/f_check
+++ b/f_check
@@ -98,7 +98,7 @@ else
                             vendor=FLANG
                             openmp='-fopenmp'
                             ;;
-                        *ifx*)
+                        *ifort*|*ifx*)
                             vendor=INTEL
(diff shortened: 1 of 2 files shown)


martin-frbg:
Thank you - looks like I forgot to check if ifort had also started to mention GCC in the .ident line of its generated assembly files. (This method of compiler identification appears to become more and more fragile with increasing "creativity" in the identifiers. By chance I just came across a bug report elsewhere that had a vendor gcc identifying as "GCC (version ... for Intel architecture)" which would probably trigger a wrong match as well.

PR #3868, opened 2022-12-16
Remove prefetches from SVE kernels

Mousius:
This is a precursor to enabling the SVE kernels for Arm(R) Neoverse(TM) V1 which has 256-bit SVE. Testing revealed that the SVE kernel was actually worse in some cases than the existing kernel which seemed odd - removing these prefetches the underlying architecture seems to do a better job :smile_cat:


Mousius:
You can see the "lumpy" performance in the charts below:

martin-frbg:
Thank you - I remember @binebrank remarking that these kernels were not yet optimized for speed, and I assume he included the prefetches as they are/were usually a good idea. If needed, we could always reinstate the prefetching versions on the A64FX target for which they were originally written.

PR #3886, opened 2023-01-12
fix typo

uniontech-lilinjie:
Signed-off-by: lilinjie <lilinjie@uniontech.com>
fix typo

### DIFF
--- a/GotoBLAS_01Readme.txt
+++ b/GotoBLAS_01Readme.txt
@@ -80,7 +80,7 @@
                       SUN
                       Fujitsu
 
-4. Suported precision
+4. Supported precision
 
  Now x86/x86_64 version support 80bit FP precision in addition to
 normal double presicion and single precision. Currently only


martin-frbg:
Thanks - this is actually "historic" documentation from the original GotoBLAS project, not sure we even want to touch it.

Mousius:
@martin-frbg what would be the best approach to tidying this up?

martin-frbg:
@Mousius I am questioning if this needs tidying up at all. Maybe these old docs need to be moved to their own directory, and a unified set of documentation created (e.g. integrating parts of the gh wiki)

Mousius:
> @Mousius I am questioning if this needs tidying up at all. Maybe these old docs need to be moved to their own directory, and a unified set of documentation created (e.g. integrating parts of the gh wiki)

That's what I was getting at, if they're not useful and we're not maintaining them we could just remove them? Git history will preserve them 😸

martin-frbg:
Removing  them is a bit much I think, especially as they serve as a reminder of the GotoBLAS heritage - at least it irks me when people assume or even claim that OpenBLAS was created from scratch in 2010, when it really is just a continuation of a much older work that had quite an impact in its time.

Mousius:
> Removing them is a bit much I think, especially as they serve as a reminder of the GotoBLAS heritage - at least it irks me when people assume or even claim that OpenBLAS was created from scratch in 2010, when it really is just a continuation of a much older work that had quite an impact in its time.

Isn't the initial line in the README enough to inform people? Could also add a `Heritage` section or similar? It worries me that this contribution is here because of the confusion introduced by retaining the older files.

PR #3891, opened 2023-02-01
update for riscv V extension 1.0 and arbitrary shape gemm kernels

sergei-lewis:
* all modules compile with latest tools and vector extension version
* easily reconfigure for different architecture preferences
* reduce LMUL to avoid register spills (note LMUL=8 uses a quarter of the register bank per variable!)
* multiple test fixes - corner cases (zero/negative inputs), nrm2 numeric stability
* added vectorised implementations for sum/zsum


martin-frbg:
Unfortunately this looks like it is conflicting with recent developments on the riscv branch, I'll leave this for xianyi
Changes to cpuid_riscv64.c look a bit spurious though ?

sergei-lewis:
> Unfortunately this looks like it is conflicting with recent developments on the riscv branch, I'll leave this for xianyi

Since this PR, there have been further breaking changes to the standard intrinsics, and the toolchains have now picked these up; e.g. https://github.com/riscv-non-isa/rvv-intrinsic-doc/pull/189

I have further changes to address these. Is it preferable if I add them to this PR, close this and make a brand new PR against develop, or merge+test+ make a PR against the riscv branch?

Thanks

martin-frbg:
I'd really like to see xianyi's take on that - but if rvv intrinsics are still so much in flux, maybe a PR against the riscv branch would be a better choice ? (Not sure where that leaves the current `develop` - does its riscv code still compile at least with some well-defined older toolchain  ? If not, I guess we might as well replace it directly without going through the riscv branch...)

sergei-lewis:
Since this PR is out of date now, I'm closing it and making a new one against the riscv branch as discussed: https://github.com/xianyi/OpenBLAS/pull/3919

I'm not aware of any toolchain that is able to build the code as currently committed - even the current T-head toolchain uses rvv1.0

PR #3896, opened 2023-02-02
Fix USE_PERL option usage

antonio-rojas:
Once you define the USE_PERL variable it is always set as DEFINED regardless of its value. One should chack for the actual variable value.


martin-frbg:
Thanks, one more for the "nobody will ever set this to zero instead of leaving it unmentioned" gallery

antonio-rojas:
It is already set to OFF a few lines above, so there is currently no way to reach the "no perl" code path

martin-frbg:
Ah, ok, I'd wondered how I missed it - but in that case it was more like an overcorrection

PR #3905, opened 2023-02-09
Add linting workflow to GitHub Actions

Mousius:
Currently this just checks for correct Copyright and License headers but in future we can expand to check other things.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,0 +1,14 @@
+# SPDX-FileCopyrightText: Copyright 2023 Arm Limited and/or its affiliates <open-source-office@arm.com>
+# SPDX-License-Identifier: BSD-3-Clause
+
+name: Linting Checks
+
+on: [push, pull_request]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: REUSE Compliance Check
+      uses: fsfe/reuse-action@v1


Mousius:
This also raises the question as to whether we should have https://github.com/jidicula/clang-format-action or similar to ensure consistency in our C files?

martin-frbg:
Not sure what to make of this, e.g. (most) files in lapack-netlib are imported from the (active) Reference-LAPACK project  and do contain authorship/origin information in some form or other, just no express Copyright statement. Some files from the original GotoBLAS, mostly its sketchy documentation, have "see license file" somewhere pointing to a BSD-style license with University of Texas copyrights

Mousius:
> Not sure what to make of this, e.g. (most) files in lapack-netlib are imported from the (active) Reference-LAPACK project and do contain authorship/origin information in some form or other, just no express Copyright statement. Some files from the original GotoBLAS, mostly its sketchy documentation, have "see license file" somewhere pointing to a BSD-style license with University of Texas copyrights

Ok dokie, I think my interpretation is the same, so here's some suggestions:
- Move lapack-netlib to a git submodule, I'm unsure if we have patches different but as it's a reference it would seem ok to keep it in its entirety? 
- Move the https://github.com/xianyi/OpenBLAS/blob/974acb39ff86121a5a94be4853f58bd728b56b81/GotoBLAS_00License.txt into LICENSES/University-of-Texas or similar and mark those files as using this variant - it's interesting this isn't explicitly BSD-3-Clause as that impacts the license of the rest of the project?

martin-frbg:
lapack-netlib has some patches applied over the Reference-LAPACK original, e.g. for better integration into the OpenBLAS build system. (Also occasionally patches go there first before being committed upstream). Not sure how having a submodule would influence this, can't the tool be told to ignore certain subdirectories ?
The original license is 20+ years old and I'd guess K.Goto just used what everybody was using at that university&time, IDK what versions of the BSD license were around then but for all these years nobody bothered

Mousius:
> lapack-netlib has some patches applied over the Reference-LAPACK original, e.g. for better integration into the OpenBLAS build system. (Also occasionally patches go there first before being committed upstream). Not sure how having a submodule would influence this, can't the tool be told to ignore certain subdirectories ?

Yip, that would be plan B, the benefit of the submodule is it's an explicit marker it's an external project with a compatible license - also eases the maintenance burden integrating patches from lapack, if I had a go would you be game to try it? We can also mark the entire folder with a blanket BSD-3-Clause at the top level, with the current copyrights, that does introduce maintenance to maintain the copyright but may be preferable? 

> The original license is 20+ years old and I'd guess K.Goto just used what everybody was using at that university&time, IDK what versions of the BSD license were around then but for all these years nobody bothered

I understand the history, licensing is like security, it doesn't become an issue until it becomes an issue... and then it's very much an issue 😸 From my naive point of view (read: not a lawyer) I don't see it being a big issue as it seems like an interestingly worded BSD License - I'll try and find someone to advise me. Do we know if we can contact the university if we wanted to move it fully to BSD-3-Clause?

martin-frbg:
Honestly I do not really want to go down this rabbit hole to "fix" something that (at least from my sketchy interpretation of the situation) would have been an issue only before the US became a late signatory to the Berne convention.

Mousius:
> Honestly I do not really want to go down this rabbit hole to "fix" something that (at least from my sketchy interpretation of the situation) would have been an issue only before the US became a late signatory to the Berne convention.

I do not understand what you mean by this?

brada4:
> 1.6.4        The SPDX file character set must support UTF-8 encoding.  

C source files are anachronism under this requirement and cannot be made "compliant with recommendation" independently of reuse tool/action saying they are. You know C compilers come in the way.

brada4:
https://github.com/xianyi/OpenBLAS/actions/runs/4133211660/jobs/7142841297#step:4:6932

Besides 10k faults regarding netlib lapack it requires licence file to have another licence. Do you volunteer with 30k line patch to fix all that?

Mousius:
Hi @brada4,

> > 1.6.4        The SPDX file character set must support UTF-8 encoding.
> 
> C source files are anachronism under this requirement and cannot be made "compliant with recommendation" independently of reuse tool/action saying they are. You know C compilers come in the way.

This doesn't appear to be an issue with our source files?

> https://github.com/xianyi/OpenBLAS/actions/runs/4133211660/jobs/7142841297#step:4:6932
> 
> Besides 10k faults regarding netlib lapack it requires licence file to have another licence. Do you volunteer with 30k line patch to fix all that?

Yip, I am volunteering to help fix this as it's a bit of a mess, I accept it's hard to pay down this much debt without a large amount of file changes. However, if there's limited interest (which I think is what's suggested by @martin-frbg's comment), hopefully we can at least include our explicit claims and I will invest my energies elsewhere 😸

brada4:
Pun was intended. Upstream LAPACK does not share your "standard" for last 50 years.

PR #3928, opened 2023-03-02
dot.c early bail fix

sergei-lewis:
The dot kernel should return 0 for N=0. An existing early bail checks for N<0 and returns 0 in this case. Adjust this to N<1 to simplify maintenance: with this change we don't have to reason through subsequent code to confirm it handles that corner case correctly, and future platform ports have a slightly more pleasant baseline to begin from.


martin-frbg:
ah, sorry, that would be against the main `develop` branch (not sure if/how I could redirect the merge there ?)

sergei-lewis:
Gotcha, I'll close this and make a new one. One sec

Issue #3899, opened 2023-02-02
parallel build with gnu make 4.4 fails


haampie:
I seem to get a build issue when using GNU make 4.4 with

```
'make' 'MAKE_NB_JOBS=0' 'ARCH=x86_64' 'TARGET=ZEN' 'USE_LOCKING=1' 'USE_OPENMP=1' 'USE_THREAD=1' 'libs' 'netlib' 'shared'
```

with `-j16` at the top-level make (openblas is built with a recursive make; not sure if relevant) which fails after some time with

```
     29      Warning: Possible change of value in conversion from REAL(4) to INTEGER(4) at (1) [-Wconversion]
     30      sgebak.f:253:19:
     31      
     32        253 |                K = SCALE( I )
     33            |                   1
     34      Warning: Possible change of value in conversion from REAL(4) to INTEGER(4) at (1) [-Wconversion]
  >> 35      make[2]: *** No rule to make target '../libopenblas_zenp-r0.3.21.a', needed by '../libopenblas_zenp-r0.3.21.so'.  Stop.
     36      make[2]: Leaving directory '/tmp/user/spack-stage/spack-stage-openblas-0.3.21-rqvfhzdue2wzonslpabua3qssw7t6mu2/spack-src/exports'
  >> 37      make[1]: *** [Makefile:125: shared] Error 2
     38      make[1]: *** Waiting for unfinished jobs....
     39      /home/user/spack/lib/spack/env/gcc/gfortran -O2 -Wall -frecursive -fno-optimize-sibling-calls -m64 -fopenmp -fPIC -msse3 -msse4.1 -mavx -mavx2 -mavx2 -c -o sgebd2.o sgebd2.f
...
```

Is there anything that jumps out of the changelog that could explain this? https://lists.gnu.org/archive/html/info-gnu/2022-10/msg00008.html. 

<details>
<summary>Maybe this?</summary>

>  Previously each target in a explicit grouped target rule was considered
>  individually: if the targets needed by the build were not out of date the
>  recipe was not run even if other targets in the group were out of date.  Now
>  if any of the grouped targets are needed by the build, then if any of the
>  grouped targets are out of date the recipe is run and all targets in the
>  group are considered updated.
</details>

With GNU make 4.3 everything looks fine.

martin-frbg:
[...]

PR #3934, opened 2023-03-07
Fix parallel make again

haampie:
`netlib` needs `lapacklib` which writes to `$(LAPACKLIB)`, concurrently with `libs`, causing a race.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ LAPACK_NOOPT := $(filter-out -O0 -O1 -O2 -O3 -Ofast -O -Og -Os,$(LAPACK_FFLAGS))
 SUBDIRS_ALL = $(SUBDIRS) test ctest utest exports benchmark ../laswp ../bench cpp_thread_test
 
 .PHONY : all libs netlib $(RELA) test ctest shared install
-.NOTPARALLEL : all libs $(RELA) prof lapack-test install blas-test
+.NOTPARALLEL : all libs netlib $(RELA) prof lapack-test install blas-test
 
 all :: libs netlib $(RELA) tests shared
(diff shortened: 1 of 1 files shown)


haampie:
Hm, this is not how to do it, you can only specify the target who's prereqs can't be run in parallel, not which targets should run mutually exclusive.

martin-frbg:
wonder which platform experiences a race here, or if something else could be going wrong (timestamps/filesystem) ?

haampie:
It's an issue with OpenBLAS's Makefiles, and is exposed for some reason more with recent gnu make.

If you want to build only a subset of `all`, e.g. not `tests`, and you do:

```
make -j8 libs netlib shared
```

you run into this

https://github.com/xianyi/OpenBLAS/blob/9feaaa3f395ef2ad654df9f34d2e335d2a1816c0/lapack-netlib/SRC/Makefile#L618-L622

and this included in OpenBLAS's own sources

https://github.com/xianyi/OpenBLAS/blob/9feaaa3f395ef2ad654df9f34d2e335d2a1816c0/Makefile.tail#L45-L46

running concurrently. The `.NOTPARALLEL` bits don't help

I think I'll just switch to the cmake build system, cause there's just too many issues.

martin-frbg:
so the added dependency from #3899 is not sufficient to make things compatible with 4.4 after all ?

haampie:
Yes, then I didn't realize `netlib` and `libs` targets updated the `libopenblas.a` archive concurrently.

martin-frbg:
not sure why I did not see this, but I am currently far away from the fast(er) 12-core hardware I last tested this with.

PR #3939, opened 2023-03-13
add riscv-v for nuclei UX900fd

surez-ok:
Nuclei released the core(UX900), and has rvv1.0 feature, 

I want to make some adaptations in this Openblas project


martin-frbg:
Thank you. May I ask why you go to the trouble of introducing an extra argument of `ARCH_EXT=v` instead of making that a fixed requirement in the compiler option string ? At least it looks to me as if compilation would fail anyway when this is not set, as there is no generic 16x8 GEMM kernel to fallback to. (And I assume one could just as well use TARGET=RISCV64_GENERIC with a compiler that does not offer RVV support)

surez-ok:
@martin-frbg   
Thank you for your review. 
Bring in `ARCH_EXT=v`  just for convenience of comparison between the version has no rvv support and has rvv support. 
your suggestion （use  TARGET=RISCV64_GENERIC ）is worth considering.  
I will redo my work

PR #3945, opened 2023-03-19
Wrap variables in cmake/prebuild in quotes to handle spaces in paths

ahojnnes:
Without the quotes, issues arise when the build path contains spaces.


martin-frbg:
Thank you. Feels like these things tend to grow back, maybe I should add a CI job specifically for spacy path names

PR #3971, opened 2023-03-28
Risc v

HellerZheng:
This PR adapts latest RISCV vector spec changes
- Added prefix (_riscv) for all riscv intrinsics
- Update some intrinsics' parameter, like vfredxxxx, vmerge

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,7 +186,7 @@ Please read `GotoBLAS_01Readme.txt` for older CPU models already supported by th
   ```
   (also known to work on C906)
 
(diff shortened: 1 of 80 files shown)


xianyi:
Is the vector intrinsic API freeze?

I fixed the intrinsics API issue at least two times.

HellerZheng:
> Is the vector intrinsic API freeze?
> 
> I fixed the intrinsics API issue at least two times.

Yes, it is painful, since the standard changed, have to live with it.

PR #3972, opened 2023-03-28
remove unqualified ifndef's for NO_LAPACK(E)

h-vetinari:
Came up in https://github.com/conda-forge/openblas-feedstock/pull/150. I did `git grep "ifndef NO_LAPACK"` and found/replaced these.

The ones in `exports/Makefile` are actually correct, and I didn't touch `kernel/setparam-ref.c` since it's a c-file, not a makefile.

### DIFF
--- a/Makefile.install
+++ b/Makefile.install
@@ -77,7 +77,7 @@ endif
 endif
 
 ifneq ($(OSNAME), AIX)
-ifndef NO_LAPACKE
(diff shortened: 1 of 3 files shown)


martin-frbg:
thanks, will rework my "competing" PR 3969 to address the other issue of  cspmv.o being on the "wrong" object list altogether

PR #3983, opened 2023-03-31
parallel build fixes

thrasibule:


martin-frbg:
Good catch, had not realized that getarch was effectively redefining make for the submakes.

haampie:
Thanks, this is a better fix than my attempt from a while back. I think the intention of `.NOTPARALLEL: x y z` was "don't run x, y, z in parallel", but the documentation says it really means "don't run the _prereqs_ of x, y, z in parallel".

PR #3994, opened 2023-04-07
Export `ssyconvf` symbol

rgommers:
This was apparently missed in commit a836fe8ec when adding the LAPACK 3.7.0 symbols. We noticed when adding wrappers for 3.7.0 routines in SciPy. For more details, see https://github.com/rgommers/scipy/issues/143

### DIFF
--- a/exports/gensymbol
+++ b/exports/gensymbol
@@ -716,6 +716,7 @@ lapackobjs2z="$lapackobjs2z
 # functions added for lapack-3.7.0
 lapackobjs2s="$lapackobjs2s
     slarfy
+    ssyconvf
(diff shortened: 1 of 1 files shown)


martin-frbg:
Thank you. Very interesting - one would think it would have been missed sometime in the six years since #1046 , but I notice even the LAPACK testsuite exercises the respective `rook` diagonal pivoting codepaths only

rgommers:
Indeed. They're pretty niche - we also didn't have a feature request that requires adding them, it was only a "expose all routines from LAPACK 3.7.0" exercise.

PR #3997, opened 2023-04-07
Add an M1-based OSX crossbuild and a NeoverseN1 build to CIRRUS CI

martin-frbg:


.cirrus.yml
```diff
@@ -2,9 +2,45 @@ macos_instance:
...
+#    config_artifacts:
+#      path: "*conf*"
+#      type: text/plain
+#  lib_artifacts:
+#    path: "libopenblas*"
+#    type: application/octet-streamm
+
+
+task:
+  name: Graviton2
```
Mousius: Graviton 2 is the AWS variant, according to https://cirrus-ci.org/guide/custom-vms/ Cirrus uses GCP which would be Ampere Altra?

Mousius (changes requested):

.cirrus.yml
martin-frbg: I guess so, probably should use NeoverseN1 for naming here but that part was only added as an afterthought (duplicates what is already in Travis). This entire PR is more of a scrapbook of things I had in my personal fork at the time when I decided to clean that up, did not think to mark it as draft as noone was watching anyway

PR #4002, opened 2023-04-13
Fix x86 detection error

imzhuhl:
Build error on intel sapphire rapids


martin-frbg:
Oops, not that silly copy-paste-error again... thanks for the fix

Issue #3995, opened 2023-04-07
GEMM numerical errors with complex single-precision (cgemm) on Apple M2


albestro:
On Apple M2 it seems there is a numerical error when using `CGEMM`, so just with single precision complex type. Actually, we found problems with `CHER2K` as well, but our speculation is that it probably shares code with `GEMM`.

Hereafter, a small reproducer and an excerpt of internal OpenBLAS tests where it is reported `CGEMM` failure.

## Reproducer
Here a reproducer of the GEMM problem.

```cpp
#include <complex>
#include <iostream>

extern "C" {
  void zgemm_(const char*, const char*,
      int*, int*, int*,
      std::complex<double>*,
      std::complex<double>*, int*,
      std::complex<double>*, int*,
      std::complex<double>*,
      std::complex<double>*, int*);

  void cgemm_(const char*, const char*,
      int*, int*, int*,
      std::complex<float>*,
      std::complex<float>*, int*,
      std::complex<float>*, int*,
      std::complex<float>*,
      std::complex<float>*, int*);
}

template <class T>
using gemm_func_t = void (*)(const char*, const char*,
    int*, int*, int*,
    std::complex<T>*,
    std::complex<T>*, int*,
    std::complex<T>*, int*,
    std::complex<T>*,
    std::complex<T>*, int*);

template <class T>
void test_gemm(gemm_func_t<T> gemm) {
  std::complex<T> a(1, 1);
  std::complex<T> b(1, 1);
  std::complex<T> c(1, 1);

  int m = 1, n = 1, k = 1;
  int ld = 1;

  std::complex<T> alpha{0.0, 1.0}, beta{0.0, 0.0};

  gemm("N", "N",
      &m, &n, &k,
      &alpha,
      &a, &ld,
      &b, &ld,
      &beta,
      &c, &ld);

  std::cout << c << "\n";
}

int main() {
  test_gemm<float>(&cgemm_);
  test_gemm<double>(&zgemm_);
}
```

Whose output on an Apple M2 is

```console
$ ./example
(0,0)
(-2,0)
```

when it is supposed to give exactly the same result, i.e. `(-2, 0)`.

## OpenBLAS test

Here an excerpt of the output of OpenBLAS tests, where a failure related to `CGEMM` is reported (it was the only failure).

```console
 ******* FATAL ERROR - COMPUTED RESULT IS LESS THAN HALF ACCURATE *******
                       EXPECTED RESULT                    COMPUTED RESULT
[...]

PR #4003, opened 2023-04-13
Fix instabilities in CGEMM/CTRMM/DNRM2 on Apple M1/M2 under OSX

martin-frbg:
for #3995 - M2 cpuid was still not matched correctly, the workaround for a corner case in DNRM2 could be optimized out by the compiler, and most importantly several ARM64 assembly kernels were still using (half of) register 18 that is reserved on OSX (as the previous round of fixes had only matched on x18, not w18)

### DIFF
--- a/kernel/arm64/sgemm_kernel_sve_v2x8.S
+++ b/kernel/arm64/sgemm_kernel_sve_v2x8.S
@@ -55,8 +55,8 @@ With this approach, we can reuse sgemm_n|tcopy_sve_v1.c packing functions. */
 #define lanes		x15
 #define pA1	    	x16
 #define pA2	    	x17
-#define alpha		w19
-#define vec_len		x20
+#define alpha		w18
+#define vec_len		x19
 #define vec_lenx2   x20
 
 #define alpha0		s10
--- a/kernel/arm64/sgemm_kernel_sve_v2x8.S
+++ b/kernel/arm64/sgemm_kernel_sve_v2x8.S
@@ -55,8 +55,8 @@ With this approach, we can reuse sgemm_n|tcopy_sve_v1.c packing functions. */
 #define lanes		x15
 #define pA1	    	x16
 #define pA2	    	x17
-#define alpha		w18
-#define vec_len		x19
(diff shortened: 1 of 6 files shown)


kernel/arm64/sgemm_kernel_sve_v2x8.S
```diff
@@ -55,8 +55,8 @@ With this approach, we can reuse sgemm_n|tcopy_sve_v1.c packing functions. */
 #define lanes		x15
 #define pA1	    	x16
 #define pA2	    	x17
-#define alpha		w18
-#define vec_len		x19
+#define alpha		w19
+#define vec_len		x20
```
Mousius: This overlaps with the below register, and I don't believe this kernel is in scope of this PR?

Mousius (changes requested):

kernel/arm64/sgemm_kernel_sve_v2x8.S
Mousius: Potentially we could run `ARMV8SVE` on one of the AWS nodes for testing this in CI?

martin-frbg: Indeed, I missed moving veclen2 to x21. And not exactly in scope, but same problem (once Apple releases a Vortex cpu with SVE - I was not sure about the exact capabilities of M2 at the time) - but I can drop this file from the PR if you prefer

Mousius: The future cores might not have the same requirements so may be better to hold off?

Also, let me know if I can help with CI configuration, [I have a history of missing things after all...](https://github.com/xianyi/OpenBLAS/pull/3853)

martin-frbg: reverted, thanks. And sadly none of our current CI providers offers SVE-capable hardware as far as I know - my blunder should eventually have been caught by running tests locally.

Mousius (approved):
LGTM!

PR #4007, opened 2023-04-17
Add Chris Sidebottom to CONTRIBUTORS.md

Mousius:


Mousius:
Sorry this took so long @martin-frbg, in response to #3843:

> > I'm not really sure what the purpose of the CONTRIBUTORS.md is? It looks like it's a subset of the git log?
> 
> git log without the granularity, I guess others may call such a thing CREDITS.txt . Anyway this seems to have served the project well enough since its inception (can't count the other 10+ years of GotoBLAS before that as that seems to have been strictly a one-man project).
> 

Done in this PR 😸 

> > There only appears to be copyright claims in the various files and then in the LICENSE - so potentially there?
> 
> Not sure if it even makes sense to add copyright claims to individual files (where they would seem to cover only a handful of lines actually affected by the change - and what to do with attributions of all previous changes, some likely pre-git ?) but I'm mostly trying to avoid opening a can of worms. Some files have "Copyright ... The OpenBLAS project", not sure if that counts for anything (I recall FLTK for one using "Copyright Bill Spitzak and others" which I _guess_ it at least as unspecific.

I'll update any "The OpenBLAS Project" copyrights and use that as needed, that should include me now hopefully 😸 

> > Fear not though, I'll ask where the best place would be, was still under the impression we could use the more explicit approach despite the hostility in the linting PR.
> 
> My impression was that the FSFE tool was too much geared towards GPL-licensed projects, and poor at recognizing Copyright/License statements that are not verbose copies of well-known GPL/BSD/MIT headers. By not recognizing copyright statements that are actually present if a bit non-standard and/or buried in the comments at the head of the file it seemed to create more work than it would appear to save. Rearranging our handling of the imported LAPACK tree or even the files themselves just to cater to its stubbornness certainly seemed excessive - at that point IMHO it would make more sense to go through the actual OpenBLAS files by hand IFF a suitable Copyright header was deemed necessary and agreed on.

`reuse` is definitely used beyond GPL projects, I've seen it used extensively. I had started going through and documenting the license headers for `reuse`, I think it'd be a one pass to clean it up and then we could maintain it from there. With LAPACK, we could've just marked the folder, I did mention that. The way LAPACK updates are handled seems quite cumbersome anyway so I didn't expect much the pushback in improving that.

Anyway, I got the impression there was a fairly strong resistance so I won't push it further, this should help unblock the other work and allow us to move forwards 😸

martin-frbg:
Great to see you in the Contributors file :)
As for the wider topic, it is just a mess I don't feel ready to tackle right now, and looking at other projects with a similarly involved history I believe I'm not alone there. Having Arm Ltd singled out as copyright holder for changes to a fraction of the lines in a file just "felt" wrong after merging contributions from industry for years without any such requirement.

PR #4022, opened 2023-04-27
fix cblas_?gemmt

martin-frbg:


BrannonKing:
Is this related to this issue? https://github.com/coin-or/Ipopt/issues/665

mmuetzel:
IIUC, this PR fixes a bug that was introduced in #4006.
That was pushed *after* the release of OpenBLAS 0.2.23. So unless you are using a version from the development branch, you are not affected by the bug fixed in the PR.

martin-frbg:
could be - I was not aware of that issue but gemmt in 0.3.23 happened to be fairly "limited in functionality" as the wrong source file got committed (see #4006 for that fix)

mmuetzel:
Fwiw, your changes look good to me.

PR #4028, opened 2023-04-29
Do not requires GNU mktemp

catap:
Historically the GNU mktemp was the first one which doesn't requires `-t` to create a directory.

Here I've introduced a fallback when `-t` is required.

For example MacPorts contains similar patch: https://github.com/macports/macports-ports/blob/bbe8abfe26651cc46d496bf48ecda506faa40a4a/math/OpenBLAS/files/patch-MacOSX-mktemp.diff

### DIFF
--- a/c_check
+++ b/c_check
@@ -168,7 +168,7 @@ fi
 
 no_msa=0
 if [ "$architecture" = "mips" ] || [ "$architecture" = "mips64" ]; then
-    tmpd="$(mktemp -d)"
+    tmpd=$(mktemp -d 2>/dev/null || mktemp -d -t 'OBC')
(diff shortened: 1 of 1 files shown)


martin-frbg:
Thanks - interesting, I had relied on BSD manpages to convince myself this would be portable, did not expect Apple to be that traditionalist (nor that packagers would fix this silently)

catap:
@martin-frbg just an example of issue inside MacPorts: https://trac.macports.org/ticket/67175

I've also with knowledge of that issue made some search and discovered a very nice and large explanation https://unix.stackexchange.com/a/84980 with the same conclusion that I made here.

Issue #4038, opened 2023-05-08
Bug/improvements in [z]imatcopy


klho:
First, a bug: the amount of workspace memory allocated for the two-pass omatcopy block may not be enough, leading to bad memory writes. Here is a simple test case that demonstrates this:

```c
#include <stdio.h>
#include "cblas.h"

int main() {
    const int m = 4, n = 2, N = m*n;
    float a[N];
    for (int i = 0; i < N; ++i) a[i] = i;
    cblas_simatcopy(CblasRowMajor, CblasNoTrans, m, n, 1, a, n, n);
    return 0;
}
```

The cause of the problem is:

https://github.com/xianyi/OpenBLAS/blob/c74ee11376600ecf546240a78c6bf6a9728b91e0/interface/imatcopy.c#L152-L155

In this case, because `*lda == *ldb` (`= n`), the workspace memory `msize = n*n = 4` only.

Some other suggested improvements:

1. `msize` above may also be an over-allocation; only `(*rows) * (*cols)` entries are needed.
2. The NEW_IMATCOPY block does not require `*rows == *cols` if no transpose is being done (e.g., just in-place conjugation).

martin-frbg:
Sorry for the late response, and thanks for looking into this. I have to admit that this function messed with my brain the last time I looked into it (when applying the fixes to the optimized in-place transform section).

martin-frbg:
Closing as fixed by your PR #4039

PR #4039, opened 2023-05-08
Bug fix and improvements for [z]imatcopy interface.

klho:
Fix/implementation for #4038.

### DIFF
--- a/interface/imatcopy.c
+++ b/interface/imatcopy.c
@@ -157,7 +157,7 @@ void CNAME( enum CBLAS_ORDER CORDER, enum CBLAS_TRANSPOSE CTRANS, blasint crows,
 	b = malloc(msize);
 	if ( b == NULL )
 	{
-		printf("Memory alloc failed\n");
+		printf("Memory alloc failed in imatcopy\n");
 		exit(1);
 	}
 
--- a/interface/imatcopy.c
+++ b/interface/imatcopy.c
@@ -120,17 +120,20 @@ void CNAME( enum CBLAS_ORDER CORDER, enum CBLAS_TRANSPOSE CTRANS, blasint crows,
     		BLASFUNC(xerbla)(ERROR_NAME, &info, sizeof(ERROR_NAME));
     		return;
   	}
+
 #ifdef NEW_IMATCOPY
-    if ( *lda == *ldb && *rows == *cols) {
+    if ( *lda == *ldb ) {
         if ( order == BlasColMajor )
         {
             if ( trans == BlasNoTrans )
             {
                 IMATCOPY_K_CN(*rows, *cols, *alpha, a, *lda );
+                return;
             }
-            else
+            else if ( *rows == *cols )
             {
                 IMATCOPY_K_CT(*rows, *cols, *alpha, a, *lda );
+                return;
             }
         }
         else
@@ -138,21 +141,18 @@ void CNAME( enum CBLAS_ORDER CORDER, enum CBLAS_TRANSPOSE CTRANS, blasint crows,
             if ( trans == BlasNoTrans )
             {
                 IMATCOPY_K_RN(*rows, *cols, *alpha, a, *lda );
+                return;
             }
-            else
+            else if ( *rows == *cols )
             {
                 IMATCOPY_K_RT(*rows, *cols, *alpha, a, *lda );
+                return;
             }
         }
-        return; 
     }
-
 #endif
 
-	if ( *lda >  *ldb )
-		msize = (size_t)(*lda) * (*ldb)  * sizeof(FLOAT);
-	else
-		msize = (size_t)(*ldb) * (*ldb)  * sizeof(FLOAT);
+	msize = (size_t)(*rows) * (*cols) * sizeof(FLOAT);
 
 	b = malloc(msize);
 	if ( b == NULL )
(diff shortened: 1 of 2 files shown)


brada4:
Sorry

martin-frbg:
> Why do you introduce memory leak claiming bug fix and memory savings?

@brada4 Why do you have to pop up here every now and again like a playground bully ? You have been warned several times that this behavior is not acceptable. If you cannot state your _technical_ objections in a civil manner, please refrain from alienating contributors here. Replying in your native language is perfectly fine if it is a language problem for you.

interface/zimatcopy.c
```diff
@@ -125,95 +125,93 @@ void CNAME( enum CBLAS_ORDER CORDER, enum CBLAS_TRANSPOSE CTRANS, blasint crows,
...
 
 	if ( order == BlasColMajor )
 	{
 
 		if ( trans == BlasNoTrans )
 		{
-	  		OMATCOPY_K_CN(*rows, *cols, alpha[0], alpha[1], a, *lda, b, *ldb );
-	  		OMATCOPY_K_CN(*rows, *cols, (FLOAT) 1.0, (FLOAT) 0.0 , b, *ldb, a, *ldb );
-			free(b);
-			return;
```
mmuetzel, line 195: The early `return` here might have been a micro-optimization.
Would it make sense to make this an `else if` ladder when removing the early `return` to safe a couple of comparison operations?

klho: Agreed, thanks -- will make the changes.

interface/zimatcopy.c
```diff
@@ -125,95 +125,93 @@ void CNAME( enum CBLAS_ORDER CORDER, enum CBLAS_TRANSPOSE CTRANS, blasint crows,
...
-        {
-                printf("Memory alloc failed in zimatcopy\n");
-                exit(1);
-        }
+	msize = (size_t)(*rows) * (*cols) * sizeof(FLOAT) * 2;
 
+	b = malloc(msize);
+	if ( b == NULL )
+	{
+		printf("Memory alloc failed\n");
```
mmuetzel: Should the original (more detailed) message be kept here?

klho: Sorry, I just copied this over from imatcopy.c (for consistency of tabs/spaces) and failed to notice that the message is materially different. I'll change it back to the more detailed message -- and also add the same more detailed print out to imatcopy.c as well.

interface/imatcopy.c
```diff
@@ -120,71 +120,71 @@ void CNAME( enum CBLAS_ORDER CORDER, enum CBLAS_TRANSPOSE CTRANS, blasint crows,
...
-        return; 
     }
-
 #endif
 
-	if ( *lda >  *ldb )
-		msize = (size_t)(*lda) * (*ldb)  * sizeof(FLOAT);
-	else
-		msize = (size_t)(*ldb) * (*ldb)  * sizeof(FLOAT);
+	msize = (size_t)(*rows) * (*cols) * sizeof(FLOAT);
```
mmuetzel, line 155: ~~I'm not sure about the background for the different permutations of the product here. Maybe, it was for numeric stability/overflow behavior/...?
Are you sure this is no longer needed? (Removing it seems unrelated to the other changes. But I could be wrong.)~~

Sorry. Misread the existing code.

mmuetzel: Also, what if `*ldb` is larger than the corresponding `*rows` or `*cols`? Wouldn't the allocated memory be too small with the new logic?

interface/zimatcopy.c
martin-frbg: Thanks

interface/imatcopy.c
martin-frbg: My reading of the MKL documentation (in lieu of an actual standard, given that it is a BLAS extension) also suggests that  either of `*lda,*ldb` may be greater than `*rows` or `*cols`. (I fear I may need to sketch a few cases on paper to see how this works out in practice though... but seeing that this survived in the codebase for almost a decade, the original assumptions must have been "mostly" correct )

mmuetzel: IIUC, the original code over-allocates memory if `*lda` is larger than the actual number of `*rows` or `*cols` it corresponds to.
The proposed change under-allocates if `*ldb` is larger than the actual number of `*rows` or `*cols` it corresponds to.

Edit: No. It's more complicated than that with the possible transposition.
Only the second part about the proposed code potentially under-allocating memory is true. ...I think...

mmuetzel: The following could be correct:
```
if ( (trans == BlasNoTrans) != (order == BlasColMajor) )
  // output is "input row" major
  msize = (size_t)(*rows) * (*ldb) * sizeof(FLOAT);
else
  // output is "input column" major
  msize = (size_t)(*cols) * (*ldb) * sizeof(FLOAT);
```

But please double check.

mmuetzel: Updated that snippet twice forth and back. That's a bigger head-scratcher than I'd like to admit.
But I think the last iteration is correct now. (As was the first. But I didn't believe in myself. 😉)

klho: I find it most helpful to think about how many entries are copied from `a` to `b` and then back again to `a`: the answer is simply `rows * cols`, regardless of the input (`lda`) and output (`ldb`) strides. (As a motivating example, consider transposing just a `10 x 10` block of a `10000 x 10000` matrix.)

Note, importantly, that `ldb` doesn't really have anything to do with `b` (maybe this is the cause of the confusion): `ldb` is the stride for the output matrix, which is stored in `a`, specifying where/how to write the data coming from `b`; while `b` just needs to hold that data, in principle with any stride it likes (as a purely internal implementation detail) -- but ideally stored as compactly as possible. From this point of view, it's clear that `msize` (= size of `b`) should not depend on `ldb` at all. (A more suggestive renaming of variables might read: `a` -> `ab` (as in the MKL documentation), `b` -> `tmp`, and `msize` -> `tmpsize`.)

The old code breaks precisely when `trans = BlasNoTrans` (or equivalent) because then `lda` and `ldb` both only give you information on the same dimension (either rows or columns, but not both), as in the opening test case in #4038. (It can be "fixed" by allocating `lda * ldb` entries instead, but this would not be optimal since `lda * ldb >= rows * cols`.)

mmuetzel: I might be misreading the MKL documentation. But afaict, `*ldb` is the stride that the function caller selects for the output matrix. `msize` is the memory in bytes that is allocated for that output.
As an example, assume a column-major output matrix `M` with 2×3 elements and a user-selected stride of 10. In that case, `M[10]` should hold the first value in the second column. If the memory allocated for `M` is only for 2*3=6 elements, that would be out of bounds.

If this is correct, please check if it is really possible to replace `ldb` in the function calls further down in this file.

mmuetzel: Oops. I was reading a page that was a bit misleading (I don't remember where I found it). The documentation here is much clearer:
https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-dpcpp/2023-0/imatcopy.html

You are completely right. `a` is the input *and* output buffer. `b` is just a temporary intermediate. Having it exactly match `*rows`*`*cols` is large enough.
Your proposed code looks good to me.

I'm sorry for having caused that confusion.

PR #4048, opened 2023-05-19
Sapphire Rapids sbgemm fix

imzhuhl:
I found the following issues while testing sbgemm in Sapphire Rapids:
1. Incorrect calculation results during sbgemm testing.
2. The OpenBLAS test_sbgemm encountered an illegal instruction error.
3. SPR cannot be built using DYNAMIC_ARCH.

**For the first point**, the error in the calculation results is due to two aspects:
1. `getarch.c` detect spr as cooperlake, resulting in `param.h`, `SBGEMM _ DEFAULT _ UNROLL _ N` is the setting of cooperlake (4), and spr needs to be greater than or equal to 16, so the kernel calculation error is caused after the matrix block. change code in `cpuid_x86.c` could fix this.
2. `sbgemm_kernel_16x16_tmpl.c` mentioned the problem of gcc11, the gcc 11.3 I used in ubuntu 22.04 did not reproduce this situation, so there is no need to multiply by 2 or 4, and the calculation result is normal after removing it. Considering the definitions of load and store, this problem should not exist in later gcc versions.

**For the second point**, The reason is that a syscall needs to be called before using amx, and then linux allows the use of amx, related links：
https://lwn.net/Articles/862541/
https://blogs.oracle.com/linux/post/intel-amx-with-uek

If OpenBLAS does not implement, then it needs to be implemented by the user programs or by the upper layer software itself, e.g. PyTorch + OpenBLAS software stack, syscall needs to be implemented in PyTorch, which involves platform, architecture detection, etc. I don't think it's suitable for all upper-level software to do this.

I read the implementation of [oneDNN](https://github.com/oneapi-src/oneDNN), and I find that it invoke syscall before calling amx related operations each time.
Therefore, I plan to make a syscall when the user calls sbgemm or cblas_sbgemm, and apply for the amx permission. Considering that a process may call sbgemm multiple times, it is a waste of time to make a syscall for each time, so an additional variable is used to record whether it has been obtained permission and avoid duplicate syscalls.

The specific design is as follows: if it is on the SPR platform, and the process calls sbgemm for the first time, it calls syscall, obtains the amx permission, and records whether the permission is obtained. If it takes, subsequent syscalls will not be invoked. Considering the situation of dynamic arch, a field is added to gotoblas to determine whether the runtime device needs amx permission.

**For the third point**, I just support DYANMIC_ARCH for SPR.


kernel/x86_64/KERNEL.SAPPHIRERAPIDS
```diff
@@ -1,6 +1,14 @@
...
-SBGEMM_SMALL_M_PERMIT = sbgemm_small_kernel_permit_spr.c
+undefine SBGEMM_SMALL_M_PERMIT
+undefine SBGEMM_SMALL_K_NN
+undefine SBGEMM_SMALL_K_B0_NN
+undefine SBGEMM_SMALL_K_NT
+undefine SBGEMM_SMALL_K_B0_NT
+undefine SBGEMM_SMALL_K_TN
+undefine SBGEMM_SMALL_K_B0_TN
+undefine SBGEMM_SMALL_K_TT
+undefine SBGEMM_SMALL_K_B0_TT
```
martin-frbg: "undefine" will not work with older versions of GNU make (and with the simple parser that retrieves these definitions for cmake builds),  but I think an empty definition should work, as in
"SBGEMM_SMALL_M_PERMIT ="

imzhuhl: OK

martin-frbg:
Looks like the one Windows job on Azure keeps failing as it uses the DYNAMIC_LIST option, can you please copy the DYN_COOPERLAKE section from around line 210 of dynamic.c for DYN_SAPPHIRERAPIDS ?
The OSX one is failing as the current Makefile in "kernel" erroneously assumes gcc10 to have native support for sapphirerapids (should be 11).

Also it seems to me that you still need to initialize your new variable need_amxtile_permission in the non-DYNAMIC_ARCH case, this is probably best done in driver/others/parameter.c

kernel/x86_64/KERNEL.SAPPHIRERAPIDS
martin-frbg: Thank you.

imzhuhl:
> The OSX one is failing as the current Makefile in "kernel" erroneously assumes gcc10 to have native support for sapphirerapids (should be 11).

Changed it, but it's a bit strange that the `-march` parameter was decided once in `Makefile` and again in `Makefile.x86_64`, which feels like a duplication

> Also it seems to me that you still need to initialize your new variable need_amxtile_permission in the non-DYNAMIC_ARCH case, this is probably best done in driver/others/parameter.c

A bit confused, under non-DYNAMIC_ARCH, there is no need to pass a variable like need_amxtile_permission to decide whether to make a syscall now, why do we need to initialize a new variable?

martin-frbg:
> Changed it, but it's a bit strange that the `-march` parameter was decided once in `Makefile` and again in `Makefile.x86_64`, which feels like a duplication

It is, but the entry in the kernel Makefile is only used for the specific subset of
SapphireRapids kernels during DYNAMIC_ARCH compilation, while the one from Makefile.x86_64 is applied for all files when (and only when) compiling
only for the SAPPHIRERAPIDS target.

> A bit confused, under non-DYNAMIC_ARCH, there is no need to pass a variable like need_amxtile_permission to decide whether to make a syscall now, why do we need to initialize a new variable?

You are right, that was a stupid suggestion, sorry for that.

Now what I am wondering about is whether it would be possible to move the AMX tile activation instruction to some place where we could still fall back to
a different SBGEMM kernel instead of aborting on failure. But this could certainly be the subject of a separate PR

imzhuhl:
> Now what I am wondering about is whether it would be possible to move the AMX tile activation instruction to some place where we could still fall back to a different SBGEMM kernel instead of aborting on failure. 

Agree. But I don't have a good solution yet.

In addition, I meet a build error on aarch64.
Because SYS_arch_prctl is related to the architecture, [https://man7.org/linux/man-pages/man2/arch_prctl.2.html](https://man7.org/linux/man-pages/man2/arch_prctl.2.html)
It says: "arch_prctl() is a Linux/x86-64 extension and should not be used in programs intended to be portable."
So I add `defined(__x86_64__)` to avoid error.

martin-frbg:
> Agree. But I don't have a good solution yet.

Me neither. I guess the SBGEMM kernel itself would need to check if AMX is usable (not sure if there even is a (hopefully cheap) call for that), and branch
to a non-AMX implementation if not. Hopefully that situation is extremely unlikely to occur in practice.

PR #4062, opened 2023-05-26
Improve input argument checks of gemmt

angsch:
* Fix return value for invalid info: Looks like `uplo` was moved in the function signature at some point and the return values were not adjusted accordingly
* Add missing checks for ldA, ldB
* Use reference-LAPACK-like checks: Instead of `ld < n` use `ld < MAX(1, n)` . In other words, `ld = 0 && n = 0` is now invalid


martin-frbg:
Thank you

PR #4080, opened 2023-06-10
Release OpenMP resources in blas_thread_shutdown

aitap:
OpenMP 5.0 gained an [`omp_pause_resource_all`](https://www.openmp.org/spec-html/5.0/openmpsu153.html) function designed to release the locks before `fork`ing the process. The parameters are described in the [`omp_pause_resource`](https://www.openmp.org/spec-html/5.0/openmpsu152.html) function. Using this function makes it possible for OpenMP builds of OpenBLAS to pass `fork` safety tests.

An important detail is that there exist OpenMP implementations that only claim OpenMP 4.5 compatibility (i.e. they `#define _OPENMP 201511`) while they do have the `omp_pause_resource_all`. We currently don't have a way to use `omp_pause_resource_all` with such implementations (e.g. GCC 10 on Debian 11).

The [first commit](43e81c3891cfb52a798dc9209b364c7f21c7f99a) adds the `omp_pause_resource_all` call to `blas_thread_shutdown`. You may want not to merge this if you'd like `blas_thread_shutdown` to be safe to call while other OpenMP operations are in progress.

The [second commit](570d6ba11700504202c3ca395e2e2b8c5cb7a272) adds a convoluted way to test `fork` safety of the resulting build on OpenMP ≥ 5.0 in addition to non-OpenMP builds. I'd be glad to see a better way of testing for OpenMP ≥ 5.0 in a Makefile.

### DIFF
--- a/c_check
+++ b/c_check
@@ -412,6 +412,14 @@ fi
 
 [ "$USE_OPENMP" != 1 ] && openmp=''
 
+have_omp_pause_resource_all=0
+if [ "$USE_OPENMP" = 1 ]; then
+    if $compiler_name $flags $openmp -o ctest3 ctest3.c; then
+        have_omp_pause_resource_all=1
+    fi
+    rm -f ctest3.o ctest3 ctest3.exe
+fi
+
 linker_L=""
 linker_l=""
 linker_a=""
@@ -501,6 +509,7 @@ done
     [ "$oldgcc" -eq 1 ] && printf "OLDGCC=1\n"
     [ "$no_lsx" -eq 1 ] && printf "NO_LSX=1\n"
     [ "$no_lasx" -eq 1 ] && printf "NO_LASX=1\n"
+    [ "$have_omp_pause_resource_all" -eq 1 ] && printf "HAVE_OMP_PAUSE_RESOURCE_ALL=1\n"
 } >> "$makefile"
 
 os=`echo "$os" | tr '[[:lower:]]' '[[:upper:]]'/ `
@@ -518,6 +527,7 @@ compiler=`echo "$compiler" | tr '[[:lower:]]' '[[:upper:]]' `
     [ "$c11_atomics" -eq 1 ] && printf "#define HAVE_C11\t1\n"
     [ "$no_lsx" -eq 1 ] && printf "#define NO_LSX\t1\n"
     [ "$no_lasx" -eq 1 ] && printf "#define NO_LASX\t1\n"
+    [ "$have_omp_pause_resource_all" -eq 1 ] && printf "#define HAVE_OMP_PAUSE_RESOURCE_ALL\t1\n"
 } >> "$config"
 
 
--- a/c_check.pl
+++ b/c_check.pl
@@ -394,6 +394,14 @@ $cross = 0 if (($os eq "Android") && ($hostos eq "Linux") && ($ENV{TERMUX_APP_PI
 
 $openmp = "" if $ENV{USE_OPENMP} != 1;
 
+$have_omp_pause_resource_all=0;
+if ($ENV{USE_OPENMP} == 1) {
+    if (system("$compiler_name $flags $openmp -o ctest3 ctest3.c") == 0) {
+        $have_omp_pause_resource_all=1;
+    }
+    unlink "ctest3.o", "ctest3", "ctest3.exe";
+}
+
 $linker_L = "";
 $linker_l = "";
 $linker_a = "";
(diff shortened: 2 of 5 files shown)


martin-frbg:
Thanks - I have not had time to review in detail, but perhaps it might be better to put the OpenMP version test in the c_check script rather than getarch

aitap:
Thanks for your reply!

What do you think about skipping the test using the same mechanism that
currently skips it on builds without double precision support? It seems
to make for a smaller diff.

I could definitely use c_check if you prefer it for architectural reasons
and the omp_pause_resource_all() idea turns out to be sound. (I didn't
originally notice that you have a mechanism for checking preprocessor
symbols.)

martin-frbg:
Sorry, must admit that I lost track of this... I'm not entirely sure of the binding region for this operation, what happens if this gets called while OpenBLAS is used by a program that uses multiple OpenMP threads itself (would there be a need for an `if (!omp_in_parallel)` guard or something like that) ?

aitap:
Good question. The binding region for `omp_pause_resource_all` is indeed the whole program, _and_ unspecified behaviour is allowed to occur if calling this when an explicit task (in a different thread) has not finalized execution. If another thread is running an OpenMP computation, `omp_pause_resource_all()` called from the `pthread_atfork()` handler is allowed to break it.

Unfortunately, OpenMP provides no other API to stop and restart the worker threads, and this function must be called before `fork()`, not afterwards (unlike the [`libgomp` patch](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=60035) that detects being forked post-factum and only restarts the threads in the child process).

Experimentally, the following program seems to behave fine with `omp_pause_resource_all` and break without it, but that's not enough for an argument for enabling this everywhere:

<details>
<summary><code>ex.c</code></summary>

```c
#ifndef _OPENMP
#error Please compile me with OpenMP enabled
#endif

#include <omp.h>
#include <pthread.h>
#include <stdbool.h>
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>

enum { OMP_THREADS=3 };

pthread_barrier_t barrier1, barrier2;

static void * do_fork(void * unused) {
	int sum = 0;
	puts("In child, before OpenMP loop1");
	#pragma omp parallel for num_threads(OMP_THREADS) reduction(+:sum)
	for (int i = 0; i < 100; ++i) {
		sum += i;
	}
	printf("In child, OpenMP loop1 done (%d)\n", sum);

	pthread_barrier_wait(&barrier1);
#ifndef OMIT_PAUSE_RESOURCE_ALL
	omp_pause_resource_all(omp_pause_hard);
#endif
	pid_t ret = fork();
	if (ret < 0) abort();
	if (ret > 0) {
		pthread_barrier_wait(&barrier2);
		return NULL;
	}
	puts("In child, before OpenMP loop2");
	sum = 0;
	#pragma omp parallel for num_threads(OMP_THREADS) reduction(+:sum)
	for (int i = 0; i < 100; ++i) {
		sum += i;
	}
	printf("In child, OpenMP loop2 done (%d)\n", sum);
	exit(0);
}


int main() {
	if (pthread_barrier_init(&barrier1, NULL, OMP_THREADS+1)) abort();
	if (pthread_barrier_init(&barrier2, NULL, OMP_THREADS+1)) abort();

	pthread_t t;
	int ret = pthread_create(&t, NULL, do_fork, NULL);
	if (ret) abort();

	puts("Starting parent loop");

	bool waited1=false, waited2=false;
	int sum = 0;
	#pragma omp parallel for num_threads(OMP_THREADS) private(waited1, waited2) reduction(+:sum)
	for (size_t i = 0; i < OMP_THREADS*10; ++i) {
		//do_fork(NULL);
		if (!waited1) {
			pthread_barrier_wait(&barrier1);
			waited1=true;
		}
		sum += i;
		if (!waited2) {
			pthread_barrier_wait(&barrier2);
			waited2=true;
		}
	}

	printf("Parent loop1 done (%d), attaching child thread...\n", sum);
	ret = pthread_join(t, NULL);
	printf("...ret=%d\n", ret);

	sum = 0;
	#pragma omp parallel for num_threads(OMP_THREADS) private(waited1, waited2) reduction(+:sum)
	for (size_t i = 0; i < OMP_THREADS*10; ++i) {
		sum += i;
	}

	printf("Parent loop2 done (%d)\n", sum);
}
```

</details>

```sh
gcc -g -fopenmp -o ex ex.c && ./ex
# Starting parent loop
# In child, before OpenMP loop1
# In child, OpenMP loop1 done (4950)
# Parent loop1 done (435), attaching child thread...
# ...ret=0
# Parent loop2 done (435)
# In child, before OpenMP loop2
# In child, OpenMP loop2 done (4950) # <-- child process succeeds using OpenMP
gcc -g -DOMIT_PAUSE_RESOURCE_ALL -fopenmp -o ex ex.c && ./ex
# Starting parent loop
# In child, before OpenMP loop1
# In child, OpenMP loop1 done (4950)
# Parent loop1 done (435), attaching child thread...
# ...ret=0
# Parent loop2 done (435)
# In child, before OpenMP loop2
# # <-- child process hung in libgomp initializing for the second loop
```

Back to the drawing board? Detect being a `fork()` child around here:
https://github.com/OpenMathLib/OpenBLAS/blob/ceb8f1e34b878b223776eff181f5e5cd8770ce3a/driver/others/blas_server_omp.c#L407
...and fall back to sequential operation instead of calling into OpenMP runtime and potentially hanging?

PR #4102, opened 2023-06-23
fix #4101, bounds issue in goto_set_num_threads on Windows

### LINKED ISSUE #4101: Bug: bounds issue in driver/others/blas_server_win32.c
In re-working the multi-threading for Windows, I believe I've come across a bounds issue. I discovered this when debugging my threading changes - a global static array was being overwritten. Debugging this showed the following:

In `goto_set_num_threads()` starting at line 571, there is the following code:

```c
for(i = blas_num_threads - 1; i < num_threads - 1; i++) {
    blas_threads[i] = CreateThread(NULL, 0, blas_thread_server, (void*)i, 0 &blas_threads_id[i]);
}
```

The problem being that when `goto_set_num_threads()` is called at the start of the program, as the documentation recommends, `blas_num_threads` is 0. So the first thread ID is written to `blas_threads[-1]`, causing that thread id to be lost and overwriting `static blas_pool_t pool`. 

The issue doesn't cause an issue normally because the pool variable normally is not yet initialized when this happens. However, if you called `goto_set_num_threads()` or `openblas_set_num_threads()` after the server was initialized you would overwrite the pool.

I discovered this because the code I wrote for threading improvements added a free-list array of cached `Event` objects just before `blas_threads` to avoid excessive creation and destruction of `Event` objects. The end of this array was overwritten causing a race condition as threads were created, which usually resulted in a lost Event object.

mseminatore:


mseminatore:
Failed builds look related to PR #4092

martin-frbg:
Thanks. (Still need to convince myself that the "blas_num_threads-1" is also incorrect when threads have been initialized already and we do not just want to add a special case for n== 0 here).

mseminatore:
@martin-frbg Just an FYI that this change accidentally rode along with the threading one. My sincere apologies for missing that in my branch hygiene.

I can fix up this PR if you've decided you would prefer to address the issue in a different way. Just LMK.

martin-frbg:
Oops. Missed that too - my primary concern with this part of the PR is/was that the previous "blas_num_threads -1" _may_ have been "more" correct for the (more likely) case where a non-zero number of threads had already been set up. But we can close this PR I guess- it might have been more practical to keep issues separated just in case either part needed to be reverted later, but it is not much of a problem.

PR #4103, opened 2023-06-24
Improve Windows threading performance scaling

mseminatore:

PR #4112, opened 2023-06-27
restore lock in exec_blas_async to prevent race on queue

mseminatore:


martin-frbg:
closing as this apparently did not solve the problem observed with the now reverted #4103

PR #4119, opened 2023-07-02
[WIP,Testing] Add experimental build with flang-new to Cirrus CI

martin-frbg:


mmuetzel:
For the logic that tries to determine the vendor from the file name of the compiler: Keep in mind that the LLVM Flang front-end executable is to be renamed from `flang-new` to `flang` at some point in the (near?) future. After that, it won't be possible to distinguish Classic Flang from LLVM Flang by the file name alone.
See also: https://reviews.llvm.org/D143592

martin-frbg:
Yes I know, but not yet relevant right now. Using the name of the executable is a kludge anyway, the original concept in f_check had it look for unambigous version strings in the intermediate files created by the compiler.

PR #4125, opened 2023-07-05
Fixes RVV masked intrinsics for iamax/iamin/imax/imin kernels

OMaghiarIMG:
Noticed that utests 9 to 12(ismin and ismax) are failing on x280 target when running with VLEN=128.
Since the code didn't seem to rely on x280's 512-bit VLEN, I tracked down the issue to the [commit](https://github.com/xianyi/OpenBLAS/commit/1374a2d08b078451dcfaf723614ea13e441e1d06) that updated the RVV intrinsics to v0.11.
After the update the functionality of the masked intrinsics has changed, __m_ translates to mask agnostic and no longer has maskedoff argument, previous functionality seems to be maintained by __mu_ intrinsics.
PR changes masked intrinsics from __m_ to __mu_ and reintroduces maskedoff argument.
All utests now pass using QEMU for x280 build and VLEN=128.

### DIFF
--- a/kernel/riscv64/iamax_rvv.c
+++ b/kernel/riscv64/iamax_rvv.c
@@ -45,9 +45,9 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define VFMAXVV_FLOAT           __riscv_vfmax_vv_f64m8
 #define VFIRSTM                 __riscv_vfirst_m_b8
 #define UINT_V_T                vuint64m8_t
-#define VIDV_MASK_UINT          __riscv_vid_v_u64m8_m
+#define VIDV_MASK_UINT          __riscv_vid_v_u64m8_mu
 #define VIDV_UINT               __riscv_vid_v_u64m8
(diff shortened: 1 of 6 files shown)


ken-unger:
Thanks for the note @OMaghiarIMG.   Just curious what LLVM version you are using for your testing?  We have been using SiFive's 15.9 (2023.03) which is based from LLVM 16 if I understand correctly.

OMaghiarIMG:
> Thanks for the note @OMaghiarIMG. Just curious what LLVM version you are using for your testing? We have been using SiFive's 15.9 (2023.03) which is based from LLVM 16 if I understand correctly.

I've used a version based on LLVM 16 as well (available in Imagination's [Catapult SDK](https://developer.imaginationtech.com/catapult-sdk/)).

OMaghiarIMG:
Added additional commit with same changes for izamax/izamin kernels.

OMaghiarIMG:
I built the BLAS-Tester tests, before changes multiple failures noted for AMIN/AMAX on vlen=512, agnostic policy = undisturbed:
xsl1blastst 118/132
xdl1blastst 96/112
xcl1blastst 103/121
xzl1blastst 102/121

After changes all Level 1 tests pass.

PR #4127, opened 2023-07-06
LoongArch64 CI

XiWeiGu:


martin-frbg:
This does a bit more than just add "LoongArch64 CI" - is this supposed to supersede or complement the old #3626 by @xen0n where ISTR there was some inconclusive discussion about naming and what of LASX would work with old vs new world compilers and systems ? A bit more context than "no description provided" would be appreciated here...

xen0n:
Ah. Last year I lost some of my free time so OpenBLAS work got preempted by other higher-priority bring-up work in the LoongArch ecosystem. I plan to check this PR this week (hopefully in a couple of days though) and try to revive the #3626 work...

XiWeiGu:
> This does a bit more than just add "LoongArch64 CI" - is this supposed to supersede or complement the old #3626 by @xen0n where ISTR there was some inconclusive discussion about naming and what of LASX would work with old vs new world compilers and systems ? A bit more context than "no description provided" would be appreciated here...

My apologies, I forgot to provide the necessary information for my question。
From my perspective, PR #3626  did the following things (please correct me if I'm mistaken):

1. Compatibility with both the ”old-world" and “new world” （-mabi=lp64 or -mabi=lp64d).
2. Compiler feature checks to avoid illegal instructions during compilation.

The first item has already been fixed and merged in PR #4077.
The functionality of the first commit in this PR is related to the second item mentioned above. However, in its implementation, it didn't take into account the differences in LASX and LSX register naming between the so-called “new world” and “old world”. Instead, it directly used $xrN and $vrN.(Currently, I haven't obtained any official information that the register naming convention $vrN and $xrN will be abandoned in the ”new world“. If any issues arise later, I will update it promptly.)
The second commit replaces the ”cpucfg“ instruction with ”getauxval“ for runtime detection when DYNAMIC_ARCH is enabled. The ”getauxval“ instruction can check whether the kernel supports LSX and LASX instructions, making it safer. The issue was raised [here](https://github.com/xianyi/OpenBLAS/pull/3626#issuecomment-1199046295).
The third commit adds CI testing for the LoongArch64 architecture using the provided Loongson cross-compilation toolchain which currently does not support LSX/LASX instructions.
So, this PR is intended to replace the old #3626, If you feel that this PR contains too much content, I can split it into multiple ones.

PR #4136, opened 2023-07-11
Fix gmake build not always picking the right ARM64 arch options for clang

martin-frbg:
fixes build failures particularly with SVE kernels for Neoverse V1


martin-frbg:
waiting to rerun checks after 4138 goes in

martin-frbg:
remaining CI errors on Cirrus are still from using the Fortran sources for C/ZRSCL in a NOFORTRAN build (already fixed on develop)

PR #4137, opened 2023-07-11
Fix riscv64 detection in system_check.cmake

felixonmars:


felixonmars:
The test failures look irrelevant to me.

Issue #4130, opened 2023-07-07
drotg implementation is not up to date with the reference BLAS


vladimir-ch:
The Reference BLAS changed their `DROTG` implementation in https://github.com/Reference-LAPACK/lapack/pull/527 to use a safe scaling. In Gonum we updated our implementation and tests accordingly in https://github.com/gonum/gonum/issues/1623. Unfortunately, the updated tests with extreme values fail in our CBLAS interface package (https://github.com/gonum/netlib/pull/92) where we use OpenBLAS as a reference. It would be nice if OpenBLAS used the same implementation with safe scaling as the reference BLAS.

angsch:
The same problem will likely trigger for `NRM2`. As far as I am aware, the assembly implementations in OpenBLAS have not been updated. I think that changing all assembly implementations is a challenge, so one option may be to add an additional symbol, one for the netlib implementation and one for the old optimized implementation.

martin-frbg:
Pretty much the only somewhat short-term solution I see is to update the "generic" reimplementations and "temporarily" use them for all architectures/cpus that have hand-crafted assembly now

martin-frbg:
Small clarification - ?ROTG is easy as it is all done in trivial C code in the interface routine, no cpu-specific kernels involved. NRM2 will require a bit more work as there are around 50 assembly kernels in all (and still about 10 if only looking at the most important ones)

PR #4143, opened 2023-07-12
Update to use safe scaling algorithm from Reference-LAPACK PR 527

martin-frbg:
fixes original topic of #4130


interface/rotg.c
```diff
@@ -65,32 +85,40 @@ void CNAME(FLOAT *DA, FLOAT *DB, FLOAT *C, FLOAT *S){
...
 
-  roe = db;
-  if (ada > adb) roe = da;
 
-  if (scale == ZERO) {
+  if (adb == ZERO) {
     *C = ONE;
     *S = ZERO;
-    *DA = ZERO;
+    DA = ZERO;
```
vladimir-ch: I think this is unintentional.

martin-frbg: Certainly hope so...but this is just a dump of my first draft, too much work too little brain right now

interface/rotg.c
```diff
@@ -112,7 +112,7 @@ void CNAME(FLOAT *DA, FLOAT *DB, FLOAT *C, FLOAT *S){
   if (adb == ZERO) {
     *C = ONE;
     *S = ZERO;
-    DA = ZERO;
+    *DA = ZERO;
```
vladimir-ch: I'm not sure about this change. If the input vector is [a,0], then the rotation is the identity and the rotated vector [a,0]. Maybe your initial intention was to delete the whole line as in the other `#if` branch?

martin-frbg: yeah, looks like a cut-n-paste gone wrong. thanks for your vigilance - still haven't run any tests with this

PR #4150, opened 2023-07-19
Fix runtime detection in ARMV8 DYNAMIC_ARCH to check SVE capability

martin-frbg:
Detection of Neoverse V1/N2 was based on cpu part code only, but SVE can apparently be unavailable in certain container or cloud environments. Tentative fix for https://github.com/numpy/numpy/issues/24028

### DIFF
--- a/driver/others/dynamic_arm64.c
+++ b/driver/others/dynamic_arm64.c
@@ -137,6 +137,8 @@ extern gotoblas_t  gotoblas_CORTEXA55;
 #endif
 
 extern void openblas_warning(int verbose, const char * msg);
+#define FALLBACK_VERBOSE 1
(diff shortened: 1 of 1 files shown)


Mousius (changes requested):
Just seeing this, it's weird behaviour for sure.

We should be relying less on CPU sniffing and more on feature detection anyway, so we don't need a warning here.

martin-frbg:
I definitely want the warning here, so that it is clear to the user why performance is lower than expected. Agreed that comparing cpu part numbers may not be the best way of selecting the appropriate kernels - unless there are known peculiarities of a particular cpu. That's just how GotoBLAS/OpenBLAS has always done it.

PR #4160, opened 2023-07-25
Add ARMV8SVE to AArch64 Dynamic Dispatch

Mousius:
In order to enable support for future cores which have similar tunings (in this case I'm doing this for the Arm(R) Neoverse(TM) V2 core), this generically detects SVE support and enables it. This should better manage the size and complexity of dynamic dispatch rather than just copy pasting the same parameters.

To make `ARMV8SVE` more representive of the common 128-bit SVE case, I've split it and similar parameters from A64FX which has the wider 512-bit SVE.


martin-frbg:
the armv7 test failure is spurious of course, but there are more ambiguities (for AppleClang) in ztrmm_lncopy_sve

Mousius:
> the armv7 test failure is spurious of course, but there are more ambiguities (for AppleClang) in ztrmm_lncopy_sve

Yip, sorry about missing those 😿 I blame my editor but getting all of them is definitely fun 😸

martin-frbg:
my fault for not fixing them all when I encountered the first of them...

PR #4165, opened 2023-07-26
Add documentation on redistributing OpenBLAS

### LINKED ISSUE #3798: Provide documentation for software distributions
As a distro packager without much insight into openblas it's not very clear to me how a software distribution (like a Linux distro, homebrew, MSYS2 etc.) should ideally build openblas and package it, with the goal of providing a generic package, or a set of packages, that can be re-used by other packages. They all seem to handle it slightly differently:

* Arch Linux: https://github.com/archlinux/svntogit-community/blob/packages/openblas/trunk/PKGBUILD
* MSYS2: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-openblas/PKGBUILD
* homebrew: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/openblas.rb
* Debian: https://salsa.debian.org/science-team/openblas/-/blob/master/debian/rules

There is probably no easy answer for this, but maybe there could be some recommendations/tips/tricks in the wiki? Maybe it's already there and I missed it somehow.

rgommers:
This touches on the following:

- build configurations
- naming of symbols, shared/static libraries and other build outputs like pkg-config and CMake files
- (in more detail) guidance on ILP64 builds

It tries to explain that, while this is only guidance and there may be reasons to deviate from that, for some build options there are best practices, and for some others there are choices to make.

It also links to a number of well-maintained build recipes in order to help packagers of other distros make choices.

Closes gh-3798

[skip ci]

### DIFF
--- a/docs/distributing.md
+++ b/docs/distributing.md
@@ -1,0 +1,270 @@
+# Guidance for redistributing OpenBLAS
+
+*We note that this document contains recommendations only - packagers and other
+redistributors are in charge of how OpenBLAS is built and distributed in their
+systems, and may have good reasons to deviate from the guidance given on this
+page. These recommendations are aimed at general packaging systems, with a user
+base that typically is large, open source (or freely available at least), and
+doesn't behave uniformly or that the packager is directly connected with.*
+
+OpenBLAS has a large number of build-time options which can be used to change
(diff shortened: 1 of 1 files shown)


lazka:
Late response, but this is very helpful, thanks!

PR #4167, opened 2023-07-27
Fix ZHEMM copy for SVE

Mousius:
Whilst disambiguating whilelt, I inadvertantly used the wrong datatype for offsets, which can be negative. This rectifies that.

### DIFF
--- a/kernel/arm64/zhemm_ltcopy_sve.c
+++ b/kernel/arm64/zhemm_ltcopy_sve.c
@@ -80,7 +80,7 @@ int CNAME(BLASLONG m, BLASLONG n, FLOAT *a, BLASLONG lda, BLASLONG posX, BLASLON
         gat_ind = svadd_m(cmp, gat_ind, lda_vec);
         gat_ind = svadd_m(svnot_z(pg, cmp) , gat_ind, 2);
         if (offset <= 0) {
-            svbool_t off_g = svwhilelt_b64((uint64_t)offset, (uint64_t)0LL);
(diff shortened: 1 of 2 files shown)


Mousius:
@martin-frbg this is where it'd be great to get https://github.com/xianyi/OpenBLAS/pull/4157 working, as that'd test SVE on Graviton 3. Can I help with the CI configuration here?

martin-frbg:
Oops. I saw the uint but naturally assumed I could learn from you there :)
Yes already having the CI job on SVE hardware running in this repo would have caught this, but unfortunately as with all
administrative issues it is still only Xianyi who can take the required action - I do not (co-)own the gh project, all I have is write access to the repository under it.

Mousius:
> Oops. I saw the uint but naturally assumed I could learn from you there :) Yes already having the CI job on SVE hardware running in this repo would have caught this, but unfortunately as with all administrative issues it is still only Xianyi who can take the required action - I do not (co-)own the gh project, all I have is write access to the repository under it.

Do you own the Cirrus configuration? I think that was added later? That would also work to hook in the AWS account.

Mousius:
> Oops. I saw the uint but naturally assumed I could learn from you there :)

Haha, trust but verify 😸

martin-frbg:
Wot to do when I'm too smol... if I had Graviton3 here I wouldn't need AWS. (Had hoped that we could do it with the existing cirrus setup too, but it seems credentials management would be much more complex than with cirun)

PR #4170, opened 2023-07-28
Fix 64-bit fortran options for riscv64

felixonmars:
64-bit builds are currently broken without this flag.

Makefiles have done this already: https://github.com/xianyi/OpenBLAS/blob/5720fa02c58562c7d3e6a3e97b053598548e98d9/Makefile.system#L831


martin-frbg:
you appear to be mixing up BINARY64 (64bit platform) and INTERFACE64 (ILP64 data model, integers are long by default) here, or the PR is missing an important conditional in its Makefile counterpart

felixonmars:
> you appear to be mixing up BINARY64 (64bit platform) and INTERFACE64 (ILP64 data model, integers are long by default) here, or the PR is missing an important conditional in its Makefile counterpart

Indeed, you are correct. I have updated to apply these flags only on INTERFACE64.

PR #4171, opened 2023-07-28
Fix build with some clang installations when openmp is enabled

boomanaiden154:
There are two instances when building the tests where OpenBLAS fails to build with OpenMP and clang due to library paths getting reset as flags are set rather than appended. This seems to only affect certain clang/libomp installations, but if it's already grabbing the correct library paths we might as well use them.

### DIFF
--- a/ctest/Makefile
+++ b/ctest/Makefile
@@ -208,7 +208,7 @@ FLDFLAGS = $(FFLAGS:-fPIC=) $(LDFLAGS)
 ifeq ($(USE_OPENMP), 1)
 ifeq ($(F_COMPILER), GFORTRAN)
 ifeq ($(C_COMPILER), CLANG)
-CEXTRALIB = -lomp
+CEXTRALIB += -lomp
 endif
 endif
 ifeq ($(F_COMPILER), NAG)
--- a/test/Makefile
+++ b/test/Makefile
@@ -265,7 +265,7 @@ FLDFLAGS = $(FFLAGS:-fPIC=) $(LDFLAGS)
 ifeq ($(USE_OPENMP), 1)
 ifeq ($(F_COMPILER), GFORTRAN)
 ifeq ($(C_COMPILER), CLANG)
-CEXTRALIB = -lomp
+CEXTRALIB += -lomp
 endif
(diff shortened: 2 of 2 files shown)


martin-frbg:
Curious what else is in your CEXTRALIB that needs to be preserved for the build to work ? Or do you just have libomp in an unusual location that makes it necessary to pass a `-L ...` as well ?

boomanaiden154:
Yes. The [llvm apt repos](https://apt.llvm.org) puts the shared objects at `/usr/lib/llvm-<major version number>/lib/libomp.so` which doesn't get picked up by default by the compiler. There are a couple extra `-L...` directives in `CEXTRALIB` before it is overwritten that contain the correct include paths (although I'm not sure exactly where they're coming from).

The compilation command for `zblat1.o` looks like the following before this patch:
```shell
gfortran -O2 -Wall -frecursive -fno-optimize-sibling-calls -m64 -fopenmp  -msse3 -mssse3 -msse4.1 -mavx -mavx2 -mavx2 -fno-tree-vectorize  -o zblat1 zblat1.o ../libopenblas_zenp-r0.3.23.dev.a -lm -lpthread -lgfortran -lm -lpthread -lgfortran -lomp
```

and then like this after the patch:
```shell
gfortran -O2 -Wall -frecursive -fno-optimize-sibling-calls -m64 -fopenmp  -msse3 -mssse3 -msse4.1 -mavx -mavx2 -mavx2 -fno-tree-vectorize  -o zblat1 zblat1.o ../libopenblas_zenp-r0.3.23.dev.a -lm -lpthread -lgfortran -lm -lpthread -lgfortran -L/usr/bin/../lib/gcc/x86_64-linux-gnu/11 -L/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../lib64 -L/lib/x86_64-linux-gnu -L/lib/../lib64 -L/usr/lib/x86_64-linux-gnu -L/usr/lib/../lib64 -L/lib -L/usr/lib -L/usr/lib/llvm-16/lib  -lpthread -lc -lomp
```

Others file show the exact same changes.

PR #4185, opened 2023-08-07
MIPS: Enable MSA

XiWeiGu:
In PR #3880, NO_MSA is utilized for compiler feature detection. Corresponding modifications are required in KERNEL.* and param.h to prevent the inability to enable MSA optimization on the MIPS platform.


martin-frbg:
I appear to have missed this, sorry for the late merge

PR #4191, opened 2023-08-13
Correct handling of NaN values in I[S/D]AMIN and I[S/D]AMAX to "return position of first NaN"

martin-frbg:


brada4:
Some seem just duplicate generic kernels?

martin-frbg:
Yes they are (and have always been), but I do not intend to change this now.

PR #4196, opened 2023-08-15
Modernize obsolete inline order

TiborGY:
A handful of functions are declared like this: `static void __inline .....`
Apparently this is a non-standard, obsolete format, as the C standard demands that `static`, `inline` and similar must be specified before the return type. If extra warnings are enabled during build, GCC 11 issues this:
`warning: ‘__inline’ is not at beginning of declaration [-Wold-style-declaration]`

I am not aware that this causes any issues, but it is better to conform to the standard, unless there is reason not to.
This PR switches the placement of inline wherever applicable.

### DIFF
--- a/common.h
+++ b/common.h
@@ -525,7 +525,7 @@ static inline unsigned long long rpcc(void){
 #endif // !RPCC_DEFINED
 
 #if !defined(BLAS_LOCK_DEFINED) && defined(__GNUC__)
-static void __inline blas_lock(volatile BLASULONG *address){
+static __inline void blas_lock(volatile BLASULONG *address){
 
   do {
     while (*address) {YIELDING;};
--- a/common_alpha.h
+++ b/common_alpha.h
@@ -45,7 +45,7 @@
 #define WMB asm("wmb")
 #define RMB asm("mb")
 
-static void __inline blas_lock(unsigned long *address){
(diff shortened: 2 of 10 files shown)


TiborGY:
The one test failure is `continuous-integration/appveyor/pr` getting killed due to the 1 hour time limit. That test is flaky, as it is a matter of luck whether or not it completes in 1 hour.

martin-frbg:
yes, may be same issue as sometimes seen in Azure - the CI jobs simply get scheduled on varying generations of hardware

brada4:
It is in -Wextra since gcc3. Is it becoming a problem recently?

TiborGY:
> It is in -Wextra since gcc3. Is it becoming a problem recently?

It is not an immediate problem, but I can see no reason why it should not be fixed. AFAIK the last time this syntax was legal was K&R C, before the C89 standard. Compilers are still accepting it, but this should not be relied upon, unless there is a good reason for it.

Standard violations have a tendency to haunt you after a long period of no problems. The CBLAS hidden argument problems were also a case where violating the standard was OK for a long time, until one day it was not.

brada4:
Good job, its really continuous maintenance, like 2-3 nonsensical warning appearing with each new compiler release, like wrong front spacing or double vertical spacing....

PR #4202, opened 2023-08-24
Remove some unused inline macro definitions

TiborGY:
This PR removes some unused inline macro definitions.
Draft until #4196 is merged.


martin-frbg:
looks as if this PR actually contains your previous one. and I do no see what would be gained by removing a handful of defines just because they are currently not used ?

TiborGY:
The plan was to rebase this after the previous got merged.
What would be gained is a reduction of clutter in the source files, and fewer useless lines means improved readability, especially for new devs.

These macro definitions are likely remnants of history that are unlikely to be useful for anything in the future.

TiborGY:
Rebase done, ready for merge unless there is an objection.

PR #4235, opened 2023-09-20
Fix division by zero in [z]rotg

angsch:
<details>
<summary> Fix division by zero.</summary>

With that `if` condition, the division by `da_r` corresponds to a division by zero.
https://github.com/OpenMathLib/OpenBLAS/blob/bb2f1ec3b01d76c44017bb65cdbf03b05e644cb4/interface/zrotg.c#L63-L68

Fixed to such that `S = conjg(db)/|db|`
</details>

<details>
<summary>  Remove redundant branch. </summary>

With the algorithm change in reference LAPACK, the case `da == ZERO` is a special branch.
https://github.com/OpenMathLib/OpenBLAS/blob/bb2f1ec3b01d76c44017bb65cdbf03b05e644cb4/interface/rotg.c#L34
https://github.com/OpenMathLib/OpenBLAS/blob/bb2f1ec3b01d76c44017bb65cdbf03b05e644cb4/interface/rotg.c#L51-L56
Consequently, there is no longer the need to check. 
https://github.com/OpenMathLib/OpenBLAS/blob/bb2f1ec3b01d76c44017bb65cdbf03b05e644cb4/interface/rotg.c#L69
The commit removes the branch.
</details>

<details>
<summary>Add missing check for zero. </summary>

https://github.com/OpenMathLib/OpenBLAS/blob/bb2f1ec3b01d76c44017bb65cdbf03b05e644cb4/interface/rotg.c#L72-L73
does not have a check `c != 0` in the long double code. The float/double version does have this check. The new code is essentially identical to the float/double version

</details>

### DIFF
--- a/interface/rotg.c
+++ b/interface/rotg.c
@@ -66,13 +66,8 @@ void CNAME(FLOAT *DA, FLOAT *DB, FLOAT *C, FLOAT *S){
     c = da / r;
     s = db / r;
     z = ONE;
-    if (da != ZERO) {
-      if (ada > adb){
-	z = s;
-      } else {
-	z = ONE / c;
-      }
-    }
+    if (ada > adb) z = s;
+    if ((ada <= adb) && (c != ZERO)) z = ONE / c;
 
     *C = c;
     *S = s;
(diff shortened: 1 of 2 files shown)


angsch:
@vladimir-ch You have commented on rotg before and may want to have a look

vladimir-ch:
Thanks for the notification @angsch. I've checked and our Gonum BLAS already has this fix.

PR #4238, opened 2023-09-21
Fix building on POWER with IBM XLF 16.1.1

### LINKED ISSUE #4237: Fails to Compile with IBM XLF 16.1.1
I noticed that some XLF support was mentioned in an older change log. After examining `Makefile.power`, I noticed that the version being called out is XLF 11.1 (paths reference `/opt/ibmcmp/xlf/11.1`).

I first compiled with the default settings and ran into linking errors. I then adjusted the paths in `Makefile.power` to `/opt/ibm/xlf/16.1.1` and ran into the same linking errors.

I'm attaching the compressed output for `make VERBOSE=1 2>&1 | tee ~/oblas.log` as the original was 4 MB.

[oblas.log.gz](https://github.com/OpenMathLib/OpenBLAS/files/12686730/oblas.log.gz)

martin-frbg:
fixes #4237


ChipKerchner:
@martin Why did you add the underscore to all the Fortran functions with `-qextname`.  Any why only XLF and not gfortran?

martin-frbg:
I'm used to gfortran adding the underscore by default, cannot check right now if it is different on the Power systems I have access to but pretty sure I tested back then. My comments in the linked issue seem to agree - if there is a more elegant solution to the xlf/gcc interoperability problem I'm probably not sufficiently versed in xlf to know

PR #4245, opened 2023-09-29
powerpc: Fix build errors with xlf

RajalakshmiSR:
This patch fixes errors when using xlf as fortran compiler on Linux.
Tested with gcc/xlf and clang/xlf compiler combinations.


martin-frbg:
thank you

PR #4249, opened 2023-10-04
powerpc: Fix build errors with Open XL C

RajalakshmiSR:
This patch fixes errors when using Open XL C compiler on AIX. 
Tested with gcc/xlf and ibm-clang/xlf compiler combinations.

### DIFF
--- a/Makefile.power
+++ b/Makefile.power
@@ -98,6 +98,9 @@ endif
 endif
 endif
 
+ifeq ($(C_COMPILER), CLANG)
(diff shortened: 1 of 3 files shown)


RajalakshmiSR:
@martin-frbg I am working on a few more followup patches  to fix  compilation issues with openXL C. Do you have a planned date for 0.3.25 release?

martin-frbg:
early november would probably make sense in order to return to a sane 2 or 3 months between releases. milestone date currently set to november 5 but flexible

PR #4252, opened 2023-10-08
build: Allow simple fixed library names (as used in pkgsrc)

drhpc:
This is the current state of OpenBLAS library naming for the pkgsrc packages (NetBSD and cross-platform).

The idea is to be able to

1. be able to have a stable library name to expect from the build
2. install multiple variants next to each other (single-threaded, parallel)

For this, the new switch FIXED_LIBNAME is introduced. The build uses

make FIXED_LIBNAME=1 LIBNAMESUFFIX=openmp

In the case of 64 bit indices, it uses

make FIXED_LIBNAME=1 LIBNAMESUFFIX=openmp  INTERFACE64=1 LIBSONAMEBASE=openblas64

This results in what I assumed to be the intention of the name variables to begin with, namely primary predictable library names like

libopenblas_openmp.so
libopenblas64_openmp.so

(As pkgsrc keeps a list of expected file names resulting from a build on the user machine, too funky names with CPU model in there and symlinks are too funky.)

I see that the naming scheme of the 64 bit indices lib is still under discussion, along with symbol suffix. So far, we use the names without renamed symbols. They are all installable at the same time and user software can choose which one to use, also using the pkg-config file (which could be handled more elegantly). Without an established convention, I didn't think like we should make the call. Depends on what downstream packages use at some point.

I don't say that I do not cause more mess than necessary, or that I fully understand the already pesent mess. Just throwing this suggestion out there. Maybe my the FIXED_LIBNAME switch or something along that can be included. Or we work out something better that fulfills the demands of a pkgsrc installation.

See the github mirror of pgksrc for the patches in use.

- https://github.com/NetBSD/pkgsrc/tree/trunk/math/openblas
- https://github.com/NetBSD/pkgsrc/tree/trunk/math/openblas64
- https://github.com/NetBSD/pkgsrc/tree/trunk/math/openblas_openmp
- https://github.com/NetBSD/pkgsrc/tree/trunk/math/openblas64_openmp
- https://github.com/NetBSD/pkgsrc/tree/trunk/math/openblas_pthread
- https://github.com/NetBSD/pkgsrc/tree/trunk/math/openblas64_pthread

Also, I hope I didn't break any other configuration.

And, lastly, one idea: If we settle on OpenMP always and the ILP64 symbols with suffix, we could just get away with building _one_ libopenmp.so (with libtool versioning, perhaps, libopenblas.so.0.4.33 and symlinks …). That would also be great and lots of Makefile machinery could go. Until then, please consider my hack on top;-)


rgommers:
Thanks for sharing the pkgsrc needs and details @drhpc, interesting. 

I tried this PR with the given instructions above - the build works but the install doesn't; the static library name isn't yet handled correctly it looks like: 
```
$ make FIXED_LIBNAME=1 LIBNAMESUFFIX=openmp
$ make PREFIX=$PWD/install-dir install
make -f Makefile.install install
make[1]: Entering directory '/home/rgommers/code/tmp/OpenBLAS'
Generating openblas_config.h in /home/rgommers/code/tmp/OpenBLAS/install-dir/include
Generating f77blas.h in /home/rgommers/code/tmp/OpenBLAS/install-dir/include
Generating cblas.h in /home/rgommers/code/tmp/OpenBLAS/install-dir/include
Copying LAPACKE header files to /home/rgommers/code/tmp/OpenBLAS/install-dir/include
Copying the static library to /home/rgommers/code/tmp/OpenBLAS/install-dir/lib
install: cannot stat 'libopenblas_skylakexp-r0.3.24.dev.a': No such file or directory
make[1]: *** [Makefile.install:66: install] Error 1
make[1]: Leaving directory '/home/rgommers/code/tmp/OpenBLAS'
make: *** [Makefile:411: install] Error 2
```

> `LIBSOBASENAME=openblas64`

This should be `LIBSONAMEBASE` rather than `LIBSOBASENAME` it looks like (maybe edit your comment?).


>  be able to have a stable library name to expect from the build

You can already achieve this by just renaming the library after it's built, right? From my point of view, having to support a lot of ad-hoc naming conventions that different distros already make (I need to support auto-detection and linking against OpenBLAS in NumPy, SciPy and Meson), a free-form `set-name-to-xxx` switch is not great. I'd much rather extend the current naming scheme with fixed conventions and a boolean flag to say "give me those".

The current convention is `openblas.so` for LP64 and `openblas64.so` for ILP64. And now you want to extend that with:

> install multiple variants next to each other (single-threaded, parallel)

That seems very reasonable. I think any fixed naming works. MKL has a very structured scheme that I think is clear; doing something similar but using `pthread` instead of MKL's `sequential` (and `-seq` in .pc files) makes sense. For OpenMP, one question that comes to mind is if it should distinguish between GNU/LLVM/Intel OpenMP?


> If we settle on OpenMP always and the ILP64 symbols with suffix

OpenMP isn't always an option - for example on PyPI, where it's not possible to distribute an OpenMP runtime as a separate package, and vendoring it is not pretty (so we prefer pthreads instead for PyPI).

martin-frbg:
Actually I'd prefer that distributions rename or softlink the current, limited set of library names according to their needs, rather than add another option. And I assume the current situation must be bad enough for third parties like flexiblas

drhpc:
@rgommers About static libs: You missed giving the setttings also to the install make process. This is transient variables — that is why you could write it into Makefile.rule to persist.
```
~/src/OpenBLAS$ make FIXED_LIBNAME=1 LIBNAMESUFFIX=openmp -j6
[…]

  OS               ... Linux             
  Architecture     ... x86_64               
  BINARY           ... 64bit                 
  C compiler       ... GCC  (cmd & version : cc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0)
  Fortran compiler ... GFORTRAN  (cmd & version : GNU Fortran (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0)
  Library Name     ... libopenblas_openmp.a (Multi-threading; Max num-threads is 12)

To install the library, you can run "make PREFIX=/path/to/your/installation install".

Note that any flags passed to make during build should also be passed to make install
to circumvent any install errors.
~/src/OpenBLAS$ make FIXED_LIBNAME=1 LIBNAMESUFFIX=openmp PREFIX=/dev/shm/openblas install
make -f Makefile.install install
[…]
Copying the static library to /dev/shm/openblas/lib
Copying the shared library to /dev/shm/openblas/lib
Install OK!
[…]
~/src/OpenBLAS$ find /dev/shm/openblas/
/dev/shm/openblas/
/dev/shm/openblas/bin
/dev/shm/openblas/lib
/dev/shm/openblas/lib/libopenblas_openmp.so
/dev/shm/openblas/lib/libopenblas_openmp.so.0
/dev/shm/openblas/lib/libopenblas_openmp.a
/dev/shm/openblas/lib/pkgconfig
/dev/shm/openblas/lib/pkgconfig/openblas_openmp.pc
/dev/shm/openblas/lib/cmake
/dev/shm/openblas/lib/cmake/openblas
/dev/shm/openblas/lib/cmake/openblas/OpenBLASConfigVersion.cmake
/dev/shm/openblas/lib/cmake/openblas/OpenBLASConfig.cmake
/dev/shm/openblas/include
/dev/shm/openblas/include/lapacke_utils.h
/dev/shm/openblas/include/lapacke_mangling.h
/dev/shm/openblas/include/lapacke_config.h
/dev/shm/openblas/include/lapacke.h
/dev/shm/openblas/include/lapack.h
/dev/shm/openblas/include/cblas.h
/dev/shm/openblas/include/f77blas.h
/dev/shm/openblas/include/openblas_config.h
```
This obviously misses `OPENBLAS_INCLUDE_DIR` and `OPENBLAS_CMAKE_DIR` settings, but both libs are there (sans soversioning like usual system libs). I'll reply to other points in another comment.

drhpc:
Compiler-specific variants next to each other: Oh, wow. I can even be more complex! @rgommers are you referring only to the OpenMP runtime itself or also having OpenBLAS built with a differing compiler? In any case, I think that's too far. I don't see other projects branching out installs for differing OpenMP runtimes. It is already a rare nuisance that a library has serial and parallel variants (right now, I can only think of fftw as another example, with similar naming issues … threads, MPI, single, double quad precision). I'd rather have that space reduced than enlarged.

For a packager, it is weird to re-build a library for a differing API/ABI subset and have use cases for both variants in the same system. I know where it comes from. My own atmospheric model code _of course_ knows a switch to build with single or double precision. That's what scientists play with. But once such code enters mainstream system installs, it's weird. Offering the same API and ABI for differing data types and other conditions is a nightmare for keeping track of who is crashing because of which code.

That's why a symbol suffix is discussed/supported for ILP64 but not a standard from the beginning, sadly. With the suffix there, one could easily roll the 64 bit interface into the same library binary and just offer libopenblas without confusion. One could imagine adding openmp and pthread API into the same binary, for the somewhat ugly sideeffect that a serial or pthread application now links with openmp runtime for the shared lib.

(PyPI cannot ship an OpenMP runtime!? I did not think about that angle before. I must admit that openmp runtime as separate thing only entered my mind with pkgsrc, which caters for NetBSD installs that indeed miss libgomp. My world starts with a compiler toolchain, MPI library on top … and naturally includes an OpenMP lib. But of course, the world is wonderfully messy and colourful.)

Now, if there is a standard switch for non-conflicting library names and I can predict, I'd use it. It will be a annoying to switch to names that differ from what we settled on, but most pkgsrc installs use plain netlib where the user doesn't care, and people dealing with scientific code are used to some pain.

@martin-frbg The problem I had with the current names is that they are not (easily) predictable. If building with `DYNAMIC_ARCH=0`, I get
```
  Library Name     ... libopenblas_haswellp-r0.3.24.dev.a (Multi-threading; Max num-threads is 12)
[..]
~/src/OpenBLAS$ cat /proc/cpuinfo 
[…]
model name	: Intel(R) Core(TM) i7-10710U CPU @ 1.10GHz
[…]
```
I can derive the `r$VERSION` part, I can even derive the `p` part (though pthread or openmp?). But how should a non-trivial build script know that a native build on a Core i7-10710 will cause the library to be named after Haswell?

If there were a fixed scheme that

1. encodes ilp64 and parallelization differences to unique names,
2. does _not_ encode CPU variant, and
3. has or has not library version (release tarball version) in the name,

I could use it for a build in pkgsrc that wants to know the resulting file names in advance. I won't die on the hill about the exact naming scheme. I don't like introducing additional switches, but I saw all that complicated naming machinery and saw how I could bend it to my will. It would have been fine would it always just be libopenblas.so.0 for me to rename.

But thinking about `libopenblas.so.0` …
```
$ LANG=C readelf -d libopenblas_haswellp-r0.3.24.dev.so|grep SONAME
 0x000000000000000e (SONAME)             Library soname: [libopenblas.so.0]
$ LANG=C readelf -d /dev/shm/openblas/lib/libopenblas_openmp.so | grep SONAME
 0x000000000000000e (SONAME)             Library soname: [libopenblas_openmp.so.0]
```
If all variants have  SONAME `libopenblas.so.0`  by default, simple renaming and symlinking the library just won't do. There can be only one libopenblas.so.0 that is found at runtime, regardless of which variant the binary was linked to! I'm glad that my hack at least gets that right. An application wanting `libopenblas_openmp.so` does get the respective library at runtime.

So I add to my list above: The standard build please have an SONAME that matches the primary library name. Binary distros that offer one choice of BLAS at a time via alternatives symlinks already hack the builds to always provide SONAME=libblas.so, right?
```
$ for blas in blas blas64; do l=$(readlink $(readlink /usr/lib/x86_64-linux-gnu/lib$blas.so)); echo "$blas: $l $(LANG= readelf -d $l|grep SONAME|awk '{print $5}')"; done
blas: /usr/lib/x86_64-linux-gnu/blas/libblas.so [libblas.so.3]
blas64: /usr/lib/x86_64-linux-gnu/openblas64-openmp/libblas64.so [libblas64.so.3]
```
Ah, libblas.so.3 it is … to match what current netlib does. That number is a mess in itself, having to ensure that the ABI really matches. But this is not the way we choose in pkgsrc, and more detailed in my installs of it in HPC environments. I focus on a build environment where people can choose which toolchain to use and which library to link into their applications. Each user does that by choosing environments with environment modules, but also via specific BLAS library names. Providing a separate environment for all possible choices wouldn't scale. As long as not everybody is shipping their own binary container, these two approaches to installing BLAS should prevail. One wants to clearly tell variants apart, but offer them at the same time, the other wants to blur the distinction to be able to fling binary symlinks around on a per-user (single-user Linux system) install.

martin-frbg:
superseded by my #4485 , I think, though my approach may be less elegant ?

drhpc:
@martin-frbg Sorry for the late answer. I tried the latest OpenBLAS release. I can control the library name now like

```
$ variant=_openmp; for t in clean '' install; do make \
  FIXED_LIBNAME=1 LIBNAMESUFFIX=$variant INTERFACE64=0 USE_OPENMP=1 \
  PREFIX=/dev/shm/openblas$variant/ \
  OPENBLAS_INCLUDE_DIR=/dev/shm/openblas$variant/include/openblas$variant \
  OPENBLAS_CMAKE_DIR=/dev/shm/openblas$variant/lib/cmake/openblas$variant -j5 \
  $t; done
```

and

```
 variant=64_openmp; for t in clean '' install; do make \
  FIXED_LIBNAME=1 LIBNAMESUFFIX=$variant INTERFACE64=1 USE_OPENMP=1 \
  PREFIX=/dev/shm/openblas$variant/ \
  OPENBLAS_INCLUDE_DIR=/dev/shm/openblas$variant/include/openblas$variant \
  OPENBLAS_CMAKE_DIR=/dev/shm/openblas$variant/lib/cmake/openblas$variant -j5 \
  $t; done
```

This results in

```
/dev/shm/openblas64_openmp/lib/libopenblas64_openmp.so
/dev/shm/openblas64_openmp/lib/libopenblas64_openmp.a
/dev/shm/openblas64_openmp/lib/pkgconfig/openblas64.pc
/dev/shm/openblas64_openmp/lib/cmake/openblas64_openmp/OpenBLASConfigVersion.cmake
/dev/shm/openblas64_openmp/lib/cmake/openblas64_openmp/OpenBLASConfig.cmake
/dev/shm/openblas64_openmp/include/openblas64_openmp/lapacke_utils.h
/dev/shm/openblas64_openmp/include/openblas64_openmp/lapacke_mangling.h
/dev/shm/openblas64_openmp/include/openblas64_openmp/lapacke_config.h
/dev/shm/openblas64_openmp/include/openblas64_openmp/lapacke.h
/dev/shm/openblas64_openmp/include/openblas64_openmp/lapack.h
/dev/shm/openblas64_openmp/include/openblas64_openmp/cblas.h
/dev/shm/openblas64_openmp/include/openblas64_openmp/f77blas.h
/dev/shm/openblas64_openmp/include/openblas64_openmp/openblas_config.h
/dev/shm/openblas_openmp/lib/libopenblas_openmp.so
/dev/shm/openblas_openmp/lib/libopenblas_openmp.a
/dev/shm/openblas_openmp/lib/pkgconfig/openblas.pc
/dev/shm/openblas_openmp/lib/cmake/openblas_openmp/OpenBLASConfigVersion.cmake
/dev/shm/openblas_openmp/lib/cmake/openblas_openmp/OpenBLASConfig.cmake
/dev/shm/openblas_openmp/include/openblas_openmp/lapacke_utils.h
/dev/shm/openblas_openmp/include/openblas_openmp/lapacke_mangling.h
/dev/shm/openblas_openmp/include/openblas_openmp/lapacke_config.h
/dev/shm/openblas_openmp/include/openblas_openmp/lapacke.h
/dev/shm/openblas_openmp/include/openblas_openmp/lapack.h
/dev/shm/openblas_openmp/include/openblas_openmp/cblas.h
/dev/shm/openblas_openmp/include/openblas_openmp/f77blas.h
/dev/shm/openblas_openmp/include/openblas_openmp/openblas_config.h
```

SONAME matches the library (without any version number):

```
 $ LANG=C readelf -d /dev/shm/openblas64_openmp/lib/libopenblas64_openmp.so|grep SONAME                                                                        
 0x000000000000000e (SONAME)             Library soname: [libopenblas64_openmp.so]
 $ LANG=C readelf -d /dev/shm/openblas_openmp/lib/libopenblas_openmp.so|grep SONAME                                                                            
 0x000000000000000e (SONAME)             Library soname: [libopenblas_openmp.so]

```

So far, this is close to what I had. I expect (with `OPENBLAS_VARIANT` being the full name, like `openblas64_openmp`):
```
@comment $NetBSD: PLIST,v 1.2 2021/06/15 04:41:52 thor Exp $
include/${OPENBLAS_VARIANT}/cblas.h
include/${OPENBLAS_VARIANT}/f77blas.h
include/${OPENBLAS_VARIANT}/lapack.h
include/${OPENBLAS_VARIANT}/lapacke.h
include/${OPENBLAS_VARIANT}/lapacke_config.h
include/${OPENBLAS_VARIANT}/lapacke_mangling.h
include/${OPENBLAS_VARIANT}/lapacke_utils.h
include/${OPENBLAS_VARIANT}/openblas_config.h
lib/cmake/${OPENBLAS_VARIANT}/OpenBLASConfig.cmake
lib/cmake/${OPENBLAS_VARIANT}/OpenBLASConfigVersion.cmake
lib/lib${OPENBLAS_VARIANT}.a
lib/lib${OPENBLAS_VARIANT}.so
lib/lib${OPENBLAS_VARIANT}.so.0
lib/pkgconfig/${OPENBLAS_VARIANT}.pc
```

The name of the pkg-config file does not follow that. Can we get that, too? Seems like a simple omission.

I realize that there is the same question for the CMake files. Should the naming there also change from the generic OpenBLAS? Or is some game with the PATH-like variable for CMake to locate them the right thing to do? Like with the C headers. In my world, I will point users to the pkg-config module names, and perhaps just link blas.pc to my preferred variant.

I think having a major version at the end is also OK, libopenblas.so.5 instead of just libopenblas.so, as is customary. But in the case of BLAS we do have the expectation of extensive binary compatibility so that you can drop in the latest libopenblas.so for a program expecting some libblas from ages ago, so soname versioning seems less relevant.

Reading the CMake part of your changes … seems to me like the 64 bit suffix is hardcoded at the end, so that I cannot do `libopenblas64_openmp.so`. Is that right? Would be good to have the logic in sync once/if CMake takes over.

Btw. … How close are we to a build that combines 32 bit and 64 bit interfaces into one library with standardized 64 bit symbol suffix? At some point, we need to port all users to using those suffixes, I suppose. I'd fancy the option of just installing one build with OpenMP and 32 and 64 bit indices and be done with it. Then I can point

PR #4256, opened 2023-10-11
Fix bfloat16_bits union so that it always the sizeof unsigned short for AIX.

ChipKerchner:
Fix bfloat16_bits union so that it always the sizeof unsigned short for AIX.

### DIFF
--- a/test/compare_sgemm_sbgemm.c
+++ b/test/compare_sgemm_sbgemm.c
@@ -32,7 +32,7 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 typedef union
 {
   unsigned short v;
-  struct
+  struct __attribute__((packed))
   {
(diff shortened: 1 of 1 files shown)


ChipKerchner:
Unless we packed the bfloat16_bits structure, it can be something besides the sizeof unsigned short (AKA 2).  On AIX with OpenXLC, without this change the size is 4.

ChipKerchner:
Also ensure that the sizeof float32_bits is the sizeof float (AKA 4)

martin-frbg:
I wonder how portable this attribute is ?

RajalakshmiSR:
> I wonder how portable this attribute is ?

This is is tested on powerpc Linux and AIX with gcc and Open XL C. We can protect this with ifdef AIX, if needed.

martin-frbg:
Thanks. Either ifdef AIX or (I guess) ifndef MSVC would do then - bet Microsoft wants some #pragma  here instead

Issue #4263, opened 2023-10-13
undefined reference to `__imp__cprintf' during build x86_64-w64-mingw32 on Windows


SuperKogito:
I am trying to build OpenBlas so I can use `blas_sgemm()` on Windows 10 using x86_64-w64-mingw32. Unfortunately, I keep running across this error, when running the `make` command. I did not specify any extra arguments.
```
[...]

PR #4266, opened 2023-10-18
GH Workflows: Switch MINGW-W64 jobs to UCRT

martin-frbg:
related to #4263


mmuetzel:
Ah. You already made that switch. Looks good to me. 👍

PR #4272, opened 2023-10-19
POWER: AIX: Make use of power10 optimization

RajalakshmiSR:
POWER10 optimizations are disabled when using default AIX assembler. 
As we have fixed many issues recently, enabling optimization path for default assembler.


RajalakshmiSR:
@kavanabhat FYI

PR #4280, opened 2023-11-01
Add dynamic dispatch to AIX and clang for Power

ChipKerchner:
Known issue (to be done as a separate PR):  compiling with gcc/xlf in AIX.
Works with Linux - gcc & clang,  AIX - gcc/gfortran, xlc/xlf.

This adds dynamic dispatch to AIX.  It also adds dynamic dispatch for clang/LLVM for Linux.  All for PowerPC


driver/others/dynamic_power.c
```diff
@@ -36,13 +36,38 @@ char *gotoblas_corename(void) {
 	return corename[0];
 }
 
-#if defined(__clang__)
-static int __builtin_cpu_supports(char* arg) 
+#define CPU_UNKNOWN  0
+#define CPU_POWER5   5
+#define CPU_POWER6   6
+#define CPU_POWER7   7
```
martin-frbg: POWER7 currently does not exist as an actual, unique target - it is an alias for POWER6. Do you plan to change that ?

ChipKerchner: Fixed

martin-frbg (commented):
thanks - I guess arch==power7 would still be valid and return CPU_POWER6 though?

ChipKerchner:
Fixed.  We would like these changes in the next release (Nov 5th)

PR #4298, opened 2023-11-08
Only define __builtin_cpu_is and __builtin_cpu_supports if not present.

ChipKerchner:
Fixes instances where __builtin_cpu_is and __builtin_cpu_supports are already defined.

[4297](https://github.com/OpenMathLib/OpenBLAS/issues/4297)

### DIFF
--- a/driver/others/dynamic_power.c
+++ b/driver/others/dynamic_power.c
@@ -202,7 +202,7 @@ static int cpuid(void)
 #ifndef __BUILTIN_CPU_SUPPORTS__
 #include <string.h>
 
-#if defined(__has_builtin) && !__has_builtin(__builtin_cpu_is)
+#if defined(_AIX) || (defined(__has_builtin) && !__has_builtin(__builtin_cpu_is))
 static int __builtin_cpu_is(const char *arg)
 {
     static int ipinfo = -1;
@@ -227,7 +227,7 @@ static int __builtin_cpu_is(const char *arg)
 }
 #endif
 
-#if defined(__has_builtin) && !__has_builtin(__builtin_cpu_supports)
+#if defined(_AIX) || (defined(__has_builtin) && !__has_builtin(__builtin_cpu_supports))
 static int __builtin_cpu_supports(const char *arg)
 {
     return 0;
--- a/driver/others/dynamic_power.c
+++ b/driver/others/dynamic_power.c
@@ -202,6 +202,7 @@ static int cpuid(void)
 #ifndef __BUILTIN_CPU_SUPPORTS__
 #include <string.h>
 
+#if defined(__has_builtin) && !__has_builtin(__builtin_cpu_is)
 static int __builtin_cpu_is(const char *arg)
 {
     static int ipinfo = -1;
@@ -224,11 +225,14 @@ static int __builtin_cpu_is(const char *arg)
     }
     return 0;
 }
+#endif
 
+#if defined(__has_builtin) && !__has_builtin(__builtin_cpu_supports)
 static int __builtin_cpu_supports(const char *arg)
 {
     return 0;
 }
+#endif
 #endif
 
 static gotoblas_t *get_coretype(void) {


ChipKerchner:
This is good to go now.  We have one more fix for the Makefiles in a PR after this one.  So it is important to merge this ASAP.

imciner2:
This change appears to break the build using older GCC versions, specifically GCC 6 (which is used by Julia right now), e.g. https://buildkite.com/julialang/yggdrasil/builds/6594#018bcac6-5ab5-4ef8-ad1b-710d3d36d968 shows the build failure 

```
[22:49:16] dynamic_power.c:205:63: error: missing binary operator before token "("
[22:49:16]  #if defined(_AIX) || (defined(__has_builtin) && !__has_builtin(__builtin_cpu_is))
[22:49:16]                                                                ^
[22:49:16] dynamic_power.c:230:63: error: missing binary operator before token "("
[22:49:16]  #if defined(_AIX) || (defined(__has_builtin) && !__has_builtin(__builtin_cpu_supports))
[22:49:16]                                                                ^
[22:49:16] make[1]: *** [<builtin>: dynamic_power.o] Error 1
```

I believe what is happening is that the check
```
#if defined(_AIX) || (defined(__has_builtin) && !__has_builtin(__builtin_cpu_is))
```
is not valid in GCC 9 and before, because GCC 10 is the first one to add `__has_builtin`. The GCC docs seem to suggest that testing for `defined(__has_builtin)` should be done in a separate `#if` from the actual `__has_builtin` check, since I am guessing that since `__has_builtin` doesn't exist, it is getting an unsupported statement in the last condition.

martin-frbg:
Good to know that some high-profile users are still using a compiler from six years ago
My evil intention is to just
```
#if !defined(__has_builtin)
#define __has_builtin
#endif
```
before the first of the two occurences of this addition. (Next question is going to be if `cpuid()` is going to be supported in such an "old" context)

ChipKerchner:
I'm working on a patch with a similar fix for __has_builtin.  It should fix other issues with older versions of gcc too.

martin-frbg:
Ok, thanks, I'll hold off trying to fix this then - one warning that came up in my tests is that `__builtin_cpu_is` requires at least glibc 2.23 on non-AIX systems. (Seen on cfarm135 in the gcc compile farm, a power9 hosted by OSUOSL running CentOS 7.9 - this has glibc 2.17 and gcc 4.8.5 as the system compiler)

ChipKerchner:
That last one is a problem with __BUILTIN_CPU_SUPPORTS__ not fully supported due to glibc.  It would require an update of these libraries to resolve that.

giordano:
Yeah, in the Julia build linked above we target powerpc64le with glibc 2.17 (see https://github.com/JuliaPackaging/Yggdrasil/blob/4ab7f9990d0e0a2a2ddad02924a118936b32bc6e/RootFS.md#glibc-versions), because that's what RedHat 7 was using.

ChipKerchner:
You can try this branch if you'd like

```
https://github.com/ChipKerchner/OpenBLAS/tree/fixOldGCCPower
```

ChipKerchner:
Try this.  See if it fixes your issues.

[4320](https://github.com/OpenMathLib/OpenBLAS/pull/4320)

martin-frbg:
Makes me wonder if the PGI kludge would work for the "`if !defined(__BUILTIN_CPU_SUPPORTS__)`" case too, instead of chickening out ?

ChipKerchner:
I have no way to test that (PGI) setup.

martin-frbg:
Doesn't have to be PGI, my thought was simply to tack the `|| !defined(__BUILTIN_CPU_SUPPORTS)` to the "if PGI or CLANG" - thinking that should work on oldish Linux (which I can test) but not sure if the assembly hack would also work  on AIX _if it ever ends up in that branch_

ChipKerchner:
It "could" work but I'm not going to make that type of change without being able to test the PGI environment.

martin-frbg:
Tried it and it worked, but in the end it would only benefit gcc versions that are too old to compile anything newer than POWER8 code anyway.

ChipKerchner:
A reason for PGI to update its support for the Power gcc compiler.

martin-frbg:
AFAIK the NVIDIA HPC compiler is LLVM-based nowadays, PGI is history

PR #4320, opened 2023-11-14
Fix older versions of gcc - missing __has_builtin, cpuid and no support of P10.

ChipKerchner:
Fix older versions of gcc - missing __has_builtin, cpuid and no support of P10.


ChipKerchner:
Tested with gcc 4.8.5

giordano:
I can confirm compilation is successful for powerpc (and all the other platforms we support) is successful with this PR: https://buildkite.com/julialang/yggdrasil/builds/6623

PR #4326, opened 2023-11-17
Fix casum fallback kernel for x86_64

bartoldeman:
This kernel is only used on Skylake+ if the kernel with AVX512 intrinsics can't be used, but used the variable x1 incorrectly in the tail end of the loop, as it is still at the initial value instead of where x points to.

This caused 55 "other error"s in the LAPACK tests
(https://github.com/OpenMathLib/OpenBLAS/issues/4282)

This change makes casum.c as similar as possible as zasum.c, because zasum.c does this correctly.

### DIFF
--- a/kernel/x86_64/casum.c
+++ b/kernel/x86_64/casum.c
@@ -9,12 +9,12 @@
 #endif
 
 #ifndef HAVE_CASUM_KERNEL
-static FLOAT casum_kernel(BLASLONG n, FLOAT *x1)
+static FLOAT casum_kernel(BLASLONG n, FLOAT *x)
 {
 
     BLASLONG i=0;
     BLASLONG n_8 = n & -8;
-    FLOAT *x = x1;
+    FLOAT *x1 = x;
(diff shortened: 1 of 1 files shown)


bartoldeman:
Note: this is purely a bugfix. Performance wise I did not benchmark the backup kernel vs. `zasum_sse.S`, which is currently used on Haswell and Zen. Probably a strategy like the one used for `[sd]asum` would be best to cover all modern x86 (avx2+).

I also looked at `sasum.c` and `dasum.c`, they use `x` and `x1` similarly to the broken`casum.c`, except index `x1[i]` correctly instead of `x1[0]` in the tail.

martin-frbg:
Thanks, beat me to it - though I am not at all sure I would have been able to find it in due time... Not sure if performance should be an issue in what is supposed to be an increasingly unlikely case, although having an AVX2 microkernel to fall back to like in the non-complex cases would of course benefit Haswell-generation cpus too.

PR #4330, opened 2023-11-19
Use _mm_set1_epi{32,64x} to init mask in x86-64 [cz]asum

bartoldeman:
for skylake kernels. This is the same method as used in [sd]asum. _mm_set1_epi64x was commented out for zasum, but has the advantage of avoiding possible undefined behaviour (using an uninitialized variable), optimized out by NVHPC and icx. The new code works fine with those compilers.

For GCC 12.3 the generated code is identical; no matter what method you use, the compiler optimizes the code into a compile-time constant, there is no performance benefit using mm_cmpeq_epi8 since the corresponding instruction (VPCMPEQB) isn't actually generated!

### DIFF
--- a/kernel/x86_64/casum_microk_skylakex-2.c
+++ b/kernel/x86_64/casum_microk_skylakex-2.c
@@ -2,10 +2,9 @@
 #ifdef __NVCOMPILER
 #define NVCOMPVERS ( __NVCOMPILER_MAJOR__ * 100 + __NVCOMPILER_MINOR__ )
 #endif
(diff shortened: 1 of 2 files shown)


bartoldeman:
Some context about why this was dodgy code:
https://stackoverflow.com/questions/11962457/why-is-using-an-uninitialized-variable-undefined-behavior

bartoldeman:
@xiegengxin If you are still around, do you remember why you commented out `// abs_mask1 = (__m128d)_mm_set1_epi64x(0x7fffffffffffffff);` ?

martin-frbg:
Eek, thanks. Learn something new every bug...

PR #4333, opened 2023-11-20
Update the list of default dynamic targets for x86_64 in the README to be consistent with the Makefile

codeworm96:
This PR updates the list of default dynamic targets for x86_64 in the README to be consistent with the Makefile.
Makefile reference: https://github.com/OpenMathLib/OpenBLAS/blob/develop/Makefile.system#L663


martin-frbg:
Thanks. Somehow keeping the README up to date is its own challenge :)

PR #4340, opened 2023-11-27
Add some refines and optimizations for LoongArch.

yinshiyou:
1) Fixed a output format problem for benchmark.
2) Refine LA build check process.
3) Add s/ddot optimizations.

### DIFF
--- a/benchmark/trsv.c
+++ b/benchmark/trsv.c
@@ -127,7 +127,7 @@ int main(int argc, char *argv[]){
       long long muls = n*(n+1)/2.0;
       long long adds = (n - 1.0)*n/2.0;
 
-      fprintf(stderr, "%10d   %10.2f MFlops %10.6f sec\n", n,(muls+adds) / timeg * 1.e-6, timeg);
+      fprintf(stderr, "%10d :   %10.2f MFlops %10.6f sec\n", n,(muls+adds) / timeg * 1.e-6, timeg);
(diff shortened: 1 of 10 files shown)


martin-frbg:
Thanks. So we don't need to check if the compiler comes with the lsx intrinsics header anymore ?

yinshiyou:
> Thanks. So we don't need to check if the compiler comes with the lsx intrinsics header anymore ?

Thanks!  I think c_check has checked the inline assembler,  and the code in kernel/loongarch use assembler too.
So it should not disable assembler optimization for compiler which unsupport intrinsics but support raw assembler.

martin-frbg:
I'm just confused as it is only a few months ago that this test was added (#4127) - I do not know the LoongArch64 ecosystem well, so I can't tell if presence of the intrinsics header is a necessary sign that the toolchain "really" knows how to handle LSX instructions. (Or maybe @XiWeiGu is/was preparing the ground for intrinsics-based kernels in the future ?)

XiWeiGu:
On LoongArch64, if an intrinsics-based kernel is not used, there is no need for intrinsics header checks. Currently, LoongArch64 does not have an intrinsics-based kernel, and there are no plans to add one in the future. Therefore, intrinsics header checks can be removed.

martin-frbg:
Thank you.

PR #4344, opened 2023-11-27
Enable overstep of too long args without DYNAMIC_ARCH

catap:


martin-frbg:
Thanks. It is probably due to recent growth in (Reference-)LAPACK sources that this has become an issue for single-model builds as well.

PR #4351, opened 2023-11-30
Use 64bit build on `CMAKE_SYSTEM_PROCESSOR=i386` on Darwin

catap:
Here a bit tricky things.

A value `CMAKE_SYSTEM_PROCESSOR` is came from output of `uname -m` which migth be 32bit with 64bit building applicaiton.

So, for that case use `CMAKE_SIZEOF_VOID_P` to detect the target.

See https://trac.macports.org/ticket/68488


catap:
@martin-frbg it passed MacPorts build bot

Issue #4347, opened 2023-11-28
Undefined symbols when using cmake based build on macOS


catap:
Hey,

I've discovered a wired issue when I've switched to cmake based build:
```
:info:build Undefined symbols for architecture x86_64:
:info:build   "_xerbla_", referenced from:
:info:build       _sgemv_ in libopenblas.a(sgemv.c.o)
:info:build       _sger_ in libopenblas.a(sger.c.o)
:info:build       _strsv_ in libopenblas.a(strsv.c.o)
:info:build       _strmv_ in libopenblas.a(strmv.c.o)
:info:build       _ssyr2_ in libopenblas.a(ssyr2.c.o)
:info:build       _sgbmv_ in libopenblas.a(sgbmv.c.o)
:info:build       _ssbmv_ in libopenblas.a(ssbmv.c.o)
:info:build       ...
:info:build      (maybe you meant: _xerbla_array_)
:info:build ld: symbol(s) not found for architecture x86_64
```

It appears on macOS 10.5 and 10.6 only when build calls `gfortran` with parameters `-fpic -shared -Wl,-all_load -Wl,-force_load,libopenblas.a -Wl,-noall_load`.

Since 10.7 this issue do not reproduced, and the major differences that since 10.7 it has libc++ and clang based "world". I not sure how it related, but it seems some condition inside cmake files aren't hold.

Does it ring any bell?

martin-frbg:
Could be due to some tricks that are played so that xerbla can be overloaded with a user-provided error handler (which is e.g. what some tests use). I'm pretty sure that your libopenblas.a will have a `_xerbla_` (even if as a weak symbol), but could you check with `nm` or equivalent just in case ?

catap:
Well, it has only `U` symbols and a lot of them:
```
leopard:build catap$ nm libopenblas.a | grep _xerbla_ | grep -v 'U'
warning: nm: no name list
warning: nm: no name list
00000000 T _xerbla_array_
leopard:build catap$ nm libopenblas.a | grep _xerbla_ | grep 'U' | wc -l
warning: nm: no name list
warning: nm: no name list
    1785
leopard:build catap$ 
```

But the same behaviour on macOS 12 as well where it works:
```
√ build % nm libopenblas.a | grep _xerbla_ | grep -v 'U'
libopenblas.a:xerbla.c.o: no symbols
libopenblas.a:la_constants.f90.o: no symbols
0000000000000000 T _xerbla_array_
[...]

PR #4353, opened 2023-12-03
Fix _xerbla_ linking on older Mac versions

erikbs:
See issue #4347 for problem description. The symbol `_xerbla_` was missing from the `libopenblas.a` archive file, so weak linking was introduced for Darwin < 20 in PR #4348. This allowed OpenBLAS to compile on older Mac versions, but it did not actually fix the problem. The error would instead show up when trying to load the library, for example through NumPy:
```py
>>> import numpy
Traceback (most recent call last):
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/numpy/core/__init__.py", line 24, in <module>
    from . import multiarray
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/numpy/core/multiarray.py", line 10, in <module>
    from . import overrides
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/numpy/core/overrides.py", line 8, in <module>
    from numpy.core._multiarray_umath import (
ImportError: dlopen(/opt/local/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/numpy/core/_multiarray_umath.cpython-311-darwin.so, 2): Symbol not found: _xerbla_
  Referenced from: /opt/local/lib/libopenblas.0.dylib
  Expected in: flat namespace
 in /opt/local/lib/libopenblas.0.dylib
```

I found out that OpenBLAS builds two `xerbla.c.o` files, only one of which contains symbols, and links in the one with symbols explicitly on Darwin < 20. However, its timestamp is older than the object file without symbols, so `ar -ru` does nothing. Thus, the `_xerbla_` symbol will never be present in the final library.

My proposal is to write `xerbla.c.o` into the archive using `ar -rs` instead of allowing weak linking. With that option, the `xerbla.c.o` with symbols is written into the archive regardless of compilation order.

(Link to my bug report to MacPorts: https://trac.macports.org/ticket/68810)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -256,14 +256,14 @@ if (APPLE AND BUILD_SHARED_LIBS AND CMAKE_HOST_SYSTEM_VERSION VERSION_LESS 20)
   set (CMAKE_Fortran_USE_RESPONSE_FILE_FOR_OBJECTS 1)
   set (CMAKE_Fortran_CREATE_SHARED_LIBRARY
  "sh -c 'cat ${CMAKE_BINARY_DIR}/CMakeFiles/openblas_shared.dir/objects*.rsp | xargs -n 1024 ${CMAKE_AR} -ru libopenblas.a && exit 0' "
- "sh -c '${CMAKE_AR} -ru libopenblas.a ${CMAKE_BINARY_DIR}/driver/others/CMakeFiles/driver_others.dir/xerbla.c.o && exit 0' "
+ "sh -c '${CMAKE_AR} -rs libopenblas.a ${CMAKE_BINARY_DIR}/driver/others/CMakeFiles/driver_others.dir/xerbla.c.o && exit 0' "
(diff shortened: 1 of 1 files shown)


erikbs:
Info: patch accepted into MacPorts

martin-frbg:
I need to trawl through reference-lapack tickets but ISTR they solved it by enforcing flat namespace (if that makes sense?)

ryandesign:
I don't think you want to use the flat namespace. That's a relic from the days of Mac OS X 10.0 and 10.1. It was superseded by the two-level namespace in Mac OS X 10.2.

martin-frbg:
Thanks for the explanation. I wonder if this still allows overloading XERBLA with a user-supplied version though (not sure who uses this outside some of the included tests, but it is part of established BLAS/LAPACK behaviour)

PR #4355, opened 2023-12-04
[RISC-V] Add RISC-V Vector 128-bit target

OMaghiarIMG:
Current RVV x280 target depends on vlen=512-bits for Level 3 operations.
Commit adds generic target that supports vlen=128-bits.
New target uses the same scalable kernels as x280 for Level 1&2 operations, and autogenerated kernels for Level 3 operations.
Functional correctness of Level 3 operations tested with vlen=128-bits on user-mode QEMU v8.1.1 using ctests and BLAS-Tester.


martin-frbg:
Ah, sorry,merging sergei-lewis' recent 4439 to fix some issues with C910V and x280 appears to have created a conflict, probably only in the enumeration of cpu models in cpuid_riscv64.c - could you please check and update ? Also I think you may need to add your target to the `ifdef` in common_riscv64.h where it decides if to include riscv_vector.h (line 94)

OMaghiarIMG:
> Ah, sorry,merging sergei-lewis' recent 4439 to fix some issues with C910V and x280 appears to have created a conflict, probably only in the enumeration of cpu models in cpuid_riscv64.c - could you please check and update ? Also I think you may need to add your target to the `ifdef` in common_riscv64.h where it decides if to include riscv_vector.h (line 94)

Hello @martin-frbg, PR should be updated now.

martin-frbg:
Great, thank you very much.

Issue #4063, opened 2023-05-26
Threading performance issue on Windows


mseminatore:
Using 0.3.23 pre-built x64 binary DLL. Also tested against locally built (via cmake/ninja/clang) DLL and static lib sync'd both to head of `develop` branch and to the `v0.3.23` label. Code runs very slowly on Windows 10/11 tested against Sandybridge, Haswell and Zen2/threadripper. Tested against 4 cores, 8 cores, 24 cores and 128 cores.

Noted that on several machines with more than 8 cores only 1 core appears to be busy. On the Sandybridge machine 8 core all cores seem to be taking work.

The same code (both mine and OpenBLAS) runs dramatically (10x) faster on both M1 iMac as well as Intel MacPro. Could there be an issue with either the Windows build config or threading?

Also, if it set OPENBLAS_NUM_THREAD=1 on the Windows machines I see up to a 10x speedup.

My workload is a vectorized C NN library with the same NN structure and dataset as the TensorFlow MNIST Fashion tutorial. Primary CBLAS functions used are: SGER and SGEMV (over 65% execution time spent in those functions). Also used with much lower frequency:  SAXPY and SAXPBY.

Following two runs were on the same machine. In this case a Windows 11 Surface Laptop but similar results seen across many different Windows 10/11 desktop machines.

OpenBLAS 0.3.23 DYNAMIC_ARCH NO_AFFINITY Haswell MAX_THREADS=64
      CPU uArch: Haswell
  Cores/Threads: 8/8

Training ANN
------------
  Network shape: 784-128-10
      Optimizer: Adaptive SGD
  Loss function: Categorical cross-entropy
Mini-batch size: 8
  Training size: 60000 rows

Epoch 1/5
[====================] - loss: 0.31 - LR: 0.05
Epoch 2/5
[====================] - loss: 0.68 - LR: 0.047
Epoch 3/5
[====================] - loss: 0.12 - LR: 0.011
Epoch 4/5
[====================] - loss: 0.38 - LR: 0.016
Epoch 5/5
[====================] - loss: 0.27 - LR: 0.0037

Training time: 140.000000 seconds, 0.466667 ms/step
Test accuracy: 86.97%

C:\dev\NN>set OPENBLAS_NUM_THREADS=1

C:\dev\NN>ann
OpenBLAS 0.3.23 DYNAMIC_ARCH NO_AFFINITY Haswell MAX_THREADS=64
      CPU uArch: Haswell
[...]

PR #4359, opened 2023-12-06
Improve Windows threading performance scaling

mseminatore:
I am making another attempt at improving the threading model for the Windows server, (see #4063). This PR passes the CTEST collateral included in the repo. Unfortunately, with changes to the GitHub workflows I can no longer test against the range of runners without submitting a PR.

### DIFF
--- a/driver/others/blas_server_win32.c
+++ b/driver/others/blas_server_win32.c
@@ -51,15 +51,19 @@
 /* This is a thread implementation for Win32 lazy implementation */
 
 /* Thread server common information */
-typedef struct{
-  CRITICAL_SECTION lock;
-  HANDLE filled;
-  HANDLE killed;
+//typedef struct{
+//  CRITICAL_SECTION lock;
+//  HANDLE filled;
+//  HANDLE killed;
+//
+//  blas_queue_t	*queue;    /* Parameter Pointer */
+//  int		shutdown;  /* server shutdown flag */
+//
+//} blas_pool_t;
 
-  blas_queue_t	*queue;    /* Parameter Pointer */
-  int		shutdown;  /* server shutdown flag */
-
-} blas_pool_t;
+static blas_queue_t *work_queue = NULL;
+static HANDLE kickoff_event = NULL;
+static CRITICAL_SECTION queue_lock;
 
 /* We need this global for checking if initialization is finished.   */
 int blas_server_avail = 0;
@@ -67,7 +71,7 @@ int blas_server_avail = 0;
 /* Local Variables */
 static BLASULONG server_lock       = 0;
 
-static blas_pool_t   pool;
+//static blas_pool_t   pool;
 static HANDLE	    blas_threads   [MAX_CPU_NUMBER];
 static DWORD	    blas_threads_id[MAX_CPU_NUMBER];
 
@@ -209,7 +213,7 @@ static DWORD WINAPI blas_thread_server(void *arg){
   void *buffer, *sa, *sb;
   blas_queue_t	*queue;
   DWORD action;
-  HANDLE handles[] = {pool.filled, pool.killed};
+  //HANDLE handles[] = {pool.filled, pool.killed};
 
   /* Each server needs each buffer */
   buffer   = blas_memory_alloc(2);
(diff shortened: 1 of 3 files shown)


martin-frbg:
Thank you, looking great so far - including the msys2 runners with the LAPACK testsuite that was giving you grief the last time.

mseminatore:
Thank you, I've spent a fair amount of time working to better understand the code, both the win32 and *nix servers. I wrote my own test bed so that I could both iterate more quickly and test under stress. I am hopeful that I've got it sorted properly this time.

mseminatore:
Please see #4361

driver/others/blas_server_win32.c
```diff
@@ -225,29 +229,38 @@ static DWORD WINAPI blas_thread_server(void *arg){
...
-    if (action == WAIT_OBJECT_0 + 1) break;
+	// event raised when work is added to the queue
+	WaitForSingleObject(kickoff_event, INFINITE);
 
 #ifdef SMP_DEBUG
     fprintf(STDERR, "Server[%2ld] Got it.\n", cpu);
 #endif
 
-    EnterCriticalSection(&pool.lock);
+#if 1
```
martin-frbg, line 240: I wonder about this #if , which appears to disable newly introduced code in favor of something quite similar to the original ?

mseminatore: Certainly, let me try to explain the approach that I took. Given the previous difficulties in passing the full set of runners, and that I can no longer run them on my fork, I wanted to take care to be somewhat conservative in the changes that I made, focusing on correctness while still improving performance. The code guarded by the #if 1 is very similar to what was there previously with the simplification of removing the pool structure below.

```
/* Thread server common information */
typedef struct{
  CRITICAL_SECTION lock;
  HANDLE filled;
  HANDLE killed;

  blas_queue_t	*queue;    /* Parameter Pointer */
  int		shutdown;  /* server shutdown flag */

} blas_pool_t;
 ```
The block of code with the lock is mostly house-keeping. The alternate path is a lock-free version that I am exploring, but it seemed too aggressive for a first PR. I will continue to test that an possibly suggest it in the future if we accept this PR and I find the lock-free version to be an additional improvement.

The important change in this PR is the removal of the CRITICAL_SECTION and EVENT from each blas_queue_t entry.These objects were created and destroyed in the processing of each queue element. That led to the unnecessasry creation/deletion of **millions** of Windows kernel objects over the course of a heavy workload like training a NN. That logic was replaced by use of an int volatile finished.

```
typedef struct blas_queue {

  void *routine;
  BLASLONG position;
  BLASLONG assigned;

  blas_arg_t *args;
  void *range_m;
  void *range_n;
  void *sa, *sb;

  struct blas_queue *next;

#if defined( __WIN32__) || defined(__CYGWIN32__) || defined(_WIN32) || defined(__CYGWIN__)
  // CRITICAL_SECTION lock;
  // HANDLE finish;
  volatile int finished;
#else
  pthread_mutex_t	 lock;
  pthread_cond_t	 finished;
#endif
```

martin-frbg: I see , thank you for the detailed explanation - I was primarily wondering if it would make sense to re-run CI with that conditional inverted to exercise the lockless implementation

mseminatore: We could, but so far in my testing it has not conclusively shown itself to be faster. I wanted to explore that further before making the more significant change. Probably because that lock is held only very briefly and so is not in as much contention as I had suspected.

I can remove the disabled CAS code path for now if that is problematic.

Thinking longer term I'd like to help **blas_server_win32.c** and **blas_server.c** converge towards a more common code base for easier maintainability. But that would be a more involved change and I'm reluctant to attempt it without more success and experience with the codebase.

martin-frbg: I don't think removing it is necessary, but perhaps adding a comment would be good (just in case it stays that way and someone does not have the luxury of tracking down this github exchange in the future for whatever reason). 
And yes, converging the "server" implementations would be nice but a bit challenging, given that it is legacy code with little to no documentation and all kinds of software depend on it,often without the end user even knowing.

mseminatore: I just added the missing WIN_CAS definition, but did not add a comment. I also added a small bit of code to dynamically scale the number of threads so we don't have a ton of idle threads that can spuriously wake up and consume resources. I will wait to see how the build clears before I add a comment.

mseminatore: It looks like the 2 builds that failed were OSX and Linux builds so should not be affected by the code in my PR. Does that seem correct?

martin-frbg: Yes, seems they just fell over trying to install dependencies for the build... happens occasionally

mattip:
@mseminatore could you take a look at scipy/scipy#20294, where another library is also working with threads (maybe OpenMP?) and there seems to be a regression due to this PR

PR #4367, opened 2023-12-11
Fix arch detection with CMake build for PowerPC

barracuda156:
Currently CMake build forces `-m64` for 32-bit `ppc`, which unsurprisingly breaks configure. The first commit fixes that.
I use `CMAKE_SYSTEM_ARCHITECTURES` to accommodate Rosetta case, where a physical cpu will be Intel.

Second commit adds support for `CPU_SUBTYPE_POWERPC_7400`, which Apple headers also have. I have nothing to verify it is recognized, but I see no reason why it won’t be, since the code itself works.

P. S. This PR is not enough to fix the build, since there is another error left unfixed: https://github.com/OpenMathLib/OpenBLAS/issues/4366
However this does fix detection of `ppc`.


cmake/system_check.cmake
```diff
@@ -38,7 +38,7 @@ if(CMAKE_CL_64 OR MINGW64)
   endif()
 elseif(MINGW OR (MSVC AND NOT CMAKE_CROSSCOMPILING))
   set(X86 1)
-elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "ppc.*|power.*|Power.*")
+elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "ppc.*|power.*|Power.*" OR (CMAKE_SYSTEM_NAME MATCHES "Darwin" AND CMAKE_OSX_ARCHITECTURES MATCHES "ppc.*"))
```
martin-frbg, line 41: huh, what does cmake_system_processor return on Darwin/ppc if it is not some variation of ppc or power ? is this undefined on Darwin, or is that change just for the sake of completeness ?

cmake/system_check.cmake
```diff
@@ -109,7 +109,7 @@ else()
 endif ()
 
 if (NOT BINARY)
-  if (X86_64 OR ARM64 OR POWER OR MIPS64 OR LOONGARCH64 OR RISCV64)
+  if (X86_64 OR ARM64 OR MIPS64 OR LOONGARCH64 OR RISCV64 OR (POWER AND NOT (CMAKE_OSX_ARCHITECTURES STREQUAL "ppc")))
```
martin-frbg, line 112: I wonder if the CMAKE_SIZEOF_VOID_P trick would work on power, to define BINARY right when power is detected instead of adding such an involved construct here ?

cmake/system_check.cmake
barracuda156: `powerpc` on a native system, but if we build for `ppc` on Rosetta, arch detection based on `CMAKE_SYSTEM_PROCESSOR` will not work correctly.

cmake/system_check.cmake
barracuda156: Do you know what exactly it does? Here is the issue:

1. Physical PowerPC cpus may or may not support 64-bit binaries (G4 does not, G5 does). On G5 we may build for `ppc` or `ppc64`.
2. Kernel is always 32-bit on macOS ppc.
3. `uname` output will be `powerpc` or `Power Macintosh`, so it is useless to differentiate between `ppc` and `ppc64`.
4. On Rosetta physical cpu is Intel (32- or 64-bit), while building is for `ppc` (only 32-bit, no support for `ppc64` in Rosetta).

martin-frbg: Was not aware of this, that's why I'm asking - as far as I know, the powermac platform has been obsolete for more than 15 years, and cmake support in OpenBLAS is more recent - no wonder this was overlooked for so long

barracuda156: I think using CMAKE_OSX_ARCHITECTURES is the right thing to go with – it works reliably both natively and in Rosetta. (Macports uses that to set architecture as well.)
Unless there are some specific concerns about it.

martin-frbg: Alright, thanks.

barracuda156:
@martin-frbg Thank you for merging!

PR #4381, opened 2023-12-19
Update GEMM param for NEOVERSEV1

### LINKED ISSUE #4323: dgemm performance degradation on ARM NEOVERSEV1 with lower P*Q
I see about 2-10% perf degradation in OpenBLAS/gemm.c benchmark on single core ( also seen on multicore) of graviton3 machine. 
This is the issue due to P*Q used in NEOVERSEV1 is not comparable to what it’s  L2 cache size (which is 1MB).
from the faq of OpenBLAS  : A general rule of thumb for selecting a starting point seems to be that PxQ is about half the size of L2 cache.

image

So may be we should update the NEOVERSEV1 param P*Q as per the  L2 cache size.

Or can anyone help in understanding why NEOVERSEV1 P*Q is kept that low.

darshanp4:
For issue [dgemm performance degradation on ARM NEOVERSEV1 with lower P*Q #4323](https://github.com/OpenMathLib/OpenBLAS/issues/4323) updated the SGEMM and DGEMM PQ param as per the configuration of NEOVERSEV1 cache size. Not keeping too big but enough to get better performance and close enough to half of the cache size.
Below is OpenBLAS/gemm.c benchmark on single core ( also seen on multicore) of graviton3 machine. Performance for SGEMM improved by ~ 2-5% and **DGEMM improved by ~2-12%**

image

[@martin-frbg](https://github.com/martin-frbg) please help to get it review and merge.

### DIFF
--- a/param.h
+++ b/param.h
@@ -3396,13 +3396,13 @@ is a big desktop or server with abundant cache rather than a phone or embedded d
 #define ZGEMM_DEFAULT_UNROLL_N  4
 #define ZGEMM_DEFAULT_UNROLL_MN 16
 
-#define SGEMM_DEFAULT_P 128
-#define DGEMM_DEFAULT_P 160
+#define SGEMM_DEFAULT_P 240
+#define DGEMM_DEFAULT_P 240
 #define CGEMM_DEFAULT_P 128
 #define ZGEMM_DEFAULT_P 128
 
-#define SGEMM_DEFAULT_Q 352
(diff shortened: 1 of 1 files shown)


Mousius (approved):
These seem sensible to me! Thanks @darshanp4 😸

darshanp4:
@Mousius thank you for review , much appreciated! 👍

martin-frbg:
Great, thank you very much.

PR #4389, opened 2023-12-23
Use functionally equivalent dynamic targets

Mousius:
Similar to `drivers/other/dynamic.c`, I've looked for functionally equivalent targets and mapped them in the default DYNAMIC_ARCH build. Users can still build specific cores using DYNAMIC_LIST.


Mousius:
This reduces my `.so` from 25MB to 22MB fairly safely, I'm considering whether we can get the default DYNAMIC_ARCH build down to just ARMV8/ARMV8SVE/ARMV9 - any thoughts @martin-frbg ?

martin-frbg:
I guess we could loose some performance if we stop supporting cache size dependent GEMM parameters (though in practice most are probably just copypasted guesses from early Cortex designs). And offhand I do not remember seeing complaints about library size on arm64 lately

martin-frbg:
So on second look you're right that it does not matter for the targets you removed.

Mousius:
> I guess we could loose some performance if we stop supporting cache size dependent GEMM parameters (though in practice most are probably just copypasted guesses from early Cortex designs). And offhand I do not remember seeing complaints about library size on arm64 lately

Yip, should've mentioned that we also need to look at cores which can have different cache sizes - my current thought is to query sysconf and scale the parameters with the cache size dynamically. Unless there's a way to build a target with just parameters and not replicating the kernels?

PR #4390, opened 2023-12-23
Reduce duplication in kernel definitions

Mousius:
These files are exactly the same, so I believe we can reduce these files down. Other files require a slightly more complex unpicking.


Mousius:
I noticed you did a lot of this already @martin-frbg , thanks for picking that up in my absence 😸 It's getting a lot easier to understand and maintain.

martin-frbg:
Thanks. Much of the duplication is from earlier days when it was not quite clear if cpu-specific versions of all kernels would be necessary, rather than just suitable parameters. I had only done a minimal cleanup as some KERNEL files were no longer up to date with your recent SVE work.

PR #4397, opened 2023-12-27
Scale P and Q with L2 cache size for SVE

Mousius:
The defaults in param.h now reflect an L2 size of 128KB, and that is scaled based on the actual size.


Mousius:
@martin-frbg , this is closer to what I was thinking previously, what do you think? I can see others have done similar in `setparam-ref.c` but I'm assuming that doesn't work when building for a specific core outside of `DYNAMIC_ARCH`?

martin-frbg:
yes, for a specific cpu TARGET build I _think_ the factor would have to be applied in common_param.h but I have limited brain capacity for that right now

Mousius:
Thanks @martin-frbg, I'll look into it 😸 !

param.h
```diff
@@ -3517,13 +3517,13 @@ Until then, just keep it different than DGEMM_DEFAULT_UNROLL_N to keep copy rout
 #define ZGEMM_DEFAULT_UNROLL_N  4
 #define ZGEMM_DEFAULT_UNROLL_MN  16
 
-#define SGEMM_DEFAULT_P 128
-#define DGEMM_DEFAULT_P 160
+#define SGEMM_DEFAULT_P 30
```
DhanusML, line 3520: How were the default P and Q chosen for 128KB cache?

Mousius: https://github.com/OpenMathLib/OpenBLAS/pull/4381 demonstrated values that worked well for a 1MB L2 cache, so I divided that by 8.

If you have a more scientific approach, I'd be happy to hear it 😸

PR #4398, opened 2023-12-29
Add Optimizations for LoongArch.

yinshiyou:
After this PR, optimization has covered almost all API.


martin-frbg:
Thanks. Unfortunately it looks like the cross-compiler toolchain (along with the entire cross-build repo) has been removed from Loongson's github presence, causing all related CI jobs to fail. This happening just now is a bit unfortunate...

yinshiyou:
> Thanks. Unfortunately it looks like the cross-compiler toolchain (along with the entire cross-build repo) has been removed from Loongson's github presence, causing all related CI jobs to fail. This happening just now is a bit unfortunate...

Sorry for the confusion, I have feedback to the repository maintainer, and it has been recovered.

martin-frbg:
Great, thank you very much. I had updated our CI job to use a different gh source in the meantime, but the loongson repo is what is documented in qemu and probably countless other places.

yinshiyou:
> Great, thank you very much. I had updated our CI job to use a different gh source in the meantime, but the loongson repo is what is documented in qemu and probably countless other places.

As I know, it's a misoperation, 'build-tools' was setted to private so that CI can't download compiler. Now it has been recovered.
The repo 'sunhaiyong1978' you used is maintained by a loongson employee. Thank you for your timely handling.

Issue #1192, opened 2017-05-31
How the control the number of threads of blas functions inside OMP parallel regions


bachml:
I am working on something like the following code:
```
omp_set_num_threads(nth);
#pragma omp parallel for private(g) schedule(static)
for(g = 0; g < size; g++) {
    cblas_sgemm( ..., matrix_A + g * offset , matrix_B + g * offset, ...);
}
```
Due to matrix_A and matrix_B have small size, it won't get much acceleration with multithread sgemm funtion. So I want to use all threads of the processor in this pragma, while each thread compute a single sgemm function indepentently, e.g., 1 thread per sgemm funtion. 

But experiments shows that the performance has nothing difference compare to when I use multithread sgemm funtion without pragma. I'm wondering how could I control the number of threads of blas inside OPM parallel regions.

Thanks in advance.

martin-frbg:
There is openblas_set_num_threads() but offhand I am not sure if it will do what you need. Perhaps the trivial solution in your case is to compile OpenBLAS itself single-threaded as the threading appears to be managed by the code that calls it.

bachml:
@martin-frbg  
The trivial solution does not work, because other parts still need multi-thread blas operation.

P.S.   Actually what I am doing is implementing a caffe version of MobileNet[1704.04861]. which is based on parallel depth-wise convolution operation with BLAS backend. That's why I need multi-thread sgemm in some part while other parts still need multi-thread sgemm.

martin-frbg:
I see. At least in theory, the xGEMM functions should refrain from using multiple threads if the matrix size is small, but this is governed by compile-time constants that may not be appropriate for your case. (Or perhaps they are, and you are already comparing two cases that have essentially the same single-threaded sgemm behaviour, with the bottleneck elsewhere ?)
Which version of OpenBLAS are you using by the way, and on what platform ? Hopefully others here will be able to provide more informed comments.

brada4:
OpenBLAS calls omp_get_num_threads() just once per program's lifetime and has no instrumentation for OpenMP nesting.
You could patch num_cpu_avail to ifdef-ly set value to omp_get_num_threads() and try nested hierarchy calls.
Also there are thresholds that reduce small samples to one thread in interface/*c which may need improvements if you dare to share your calibration results....

DrTimothyAldenDavis:
[...]

PR #4425, opened 2024-01-13
Add BLAS extension openblas_set_num_threads_local()

### LINKED ISSUE #2392: finer control over the # of threads used by any given call to OpenBLAS
I need finer control of the # of threads that each call to the BLAS uses. My packages are themselves multithreaded. Each of my threads can make its own calls to the BLAS, in parallel. Some of those calls will be for small matrices, or for other cases where I know I need to use one thread. Other calls will want to use, say, 2 threads and no more (because I'm using threads elsewhere). Sometimes I want to use all the threads available. However, I cannot set a global setting, such as with openblas_set_num_threads, since that would affect all of my calls to the BLAS. My packages themselves are used in other parallel packages. So I need a thread-local way to set the # of threads that OpenBLAS uses, much like this function: https://software.intel.com/en-us/mkl-developer-reference-c-mkl-set-num-threads-local so that I can exactly control how many threads OpenBLAS uses, for each call to the BLAS.

The OpenBLAS cannot always assume it has all the threads available to it, since there are other things going on.

See also this discussion: DrTimothyAldenDavis/SuiteSparse#1

Is this possible with OpenBLAS? Is there an OpenBLAS equivalent to mkl_set_num_threads_local?

martin-frbg:
this complements the existing openblas_set_num_threads by
- returning the previously used number of threads
- overriding OpenMP environment variables and taking effect in OpenMP parallel regions as well
fixes #2392 (and related #1192)

### DIFF
--- a/cblas.h
+++ b/cblas.h
@@ -12,6 +12,7 @@ extern "C" {
 /*Set the number of threads on runtime.*/
 void openblas_set_num_threads(int num_threads);
 void goto_set_num_threads(int num_threads);
+int openblas_set_num_threads_local(int num_threads);
 
 /*Get the number of threads on runtime.*/
 int openblas_get_num_threads(void);
--- a/common_thread.h
+++ b/common_thread.h
@@ -137,19 +137,20 @@ typedef struct blas_queue {
 
 extern int blas_server_avail;
 extern int blas_omp_number_max;
+extern int blas_omp_threads_local;
 
 static __inline int num_cpu_avail(int level) {
 
 #ifdef USE_OPENMP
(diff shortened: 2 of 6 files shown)


nh2:
It would be absolutely fantastic if when new functions are added, a 1-sentence comment could be added what their purpose is.

That does not take a lot of time, but it makes the project a lot better, and saves the 100 people that will read it later some time. Or those don't know that this will solve their problems.

Nobody can understand what "local" means just from the naming. Only when one does all the archeology to find the original issue #1192.

PR #4434, opened 2024-01-15
Only use mtune=native in ARM64 fallback paths when not cross-compiling

### LINKED ISSUE #4433: SVE kernels use `-mtune=native` breaking cross-compilation
Following up from #4431, in https://github.com/JuliaPackaging/Yggdrasil/pull/7952 I'm trying to compile OpenBLAS v0.3.26 for libgfortran5 with GCC 10, but we cross-compile on an x86-64 machine and the compiler flags for the SVE kernels contain `-mtune=native`, which break cross-compilation:
https://github.com/OpenMathLib/OpenBLAS/blob/2527afaaa2a42350d6cb99f69b0f310f5cdac818/Makefile.arm64#L107
For reference the build log is available at https://buildkite.com/julialang/yggdrasil/builds/7771#018d0eae-525e-45dd-926d-848476008e0d/6-2898, but the problem is basically using `-mtune=native`.

martin-frbg:
fixes #4433

### DIFF
--- a/Makefile.arm64
+++ b/Makefile.arm64
@@ -104,9 +104,15 @@ ifneq ($(F_COMPILER), NAG)
 FCOMMON_OPT += -march=armv8.4-a -mtune=neoverse-v1
 endif
 else
-CCOMMON_OPT += -march=armv8.4-a+sve -mtune=native
+CCOMMON_OPT += -march=armv8.4-a+sve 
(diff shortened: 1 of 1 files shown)


Mousius (approved):
LGTM! Thanks for doing this so quickly @martin-frbg 😸

PR #4435, opened 2024-01-15
Fix Clang sapphire rapids march flag

imciner2:
A copy-paste error slipped in during the changes to how the Sapphire Rapids architecture flags were guarded in https://github.com/OpenMathLib/OpenBLAS/pull/4192, so it turns out Clang was never being told to use the Sapphire Rapids, and instead was just using Cooperlake instead.


martin-frbg:
Thanks, totally missed that - seems sometimes the mind reads what it expects to read rather than what the actual code/text is

PR #4437, opened 2024-01-16
Add CSKY support

Dirreke:
Add support for arch CSKY


Makefile.system
```diff
@@ -873,6 +873,16 @@ endif
 endif
 endif
 
+ifeq ($(ARCH), csky)
+NO_BINARY_MODE	= 1
+BINARY_DEFINED	= 1
+endif
+
+ifeq ($(ARCH), ck860v)
```
martin-frbg, line 881: the ifeq (ARCH,ck860v) part looks a bit redundant here, assuming csky is going to be the ARCH and ck860v the TARGET (or eventually, one of several cpu targets under the csky architecture) ?

Dirreke: Thanks. My bad, I will delete it.

martin-frbg:
Thank you - can you please also update TargetList.txt ? If I'm reading your PR correctly, you are adding CK860FV and a CSKY where I assume the latter will remain a "generic" target which will keep using the C kernels while CK860FV may get optimized code later ?

Dirreke:
Thanks, I pull the wrong branch, I will create a new PR to solve it. #4438

PR #4438, opened 2024-01-16
Add CSKY support

Dirreke:
Add support for arch CSKY


getarch.c
```diff
@@ -1766,6 +1795,10 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define OPENBLAS_SUPPORTED
 #endif
 
+#ifdef __csky__
+#include "cpuid_csky.c"
```
martin-frbg: cpuid_csky.c appears to be missing from the PR ?

Dirreke: I will delete it too. There's few cpus in this arch. Therefore, I don't want to add it for now. Thanks.

PR #4439, opened 2024-01-16
Fix builds with t-head toolchains that use old intrinsics spec

sergei-lewis:
* macro to control whether `__riscv_` prefix is added to riscv intrinsic names, as used in release candidate intrinsic spec; prefix is disabled if building with t-head toolchain while targeting rvv 0.7; check is performed in one place to make it easy to extend to other toolchains if we want
* individual fixes for obsolete extra arguments in reduction functions, also driven by the result of the above check
* also fix missing x280 entry in cpuid_riscv64.c caused by earlier mismerge


martin-frbg:
Thank you very much

PR #4452, opened 2024-01-24
Fix BLAS, BLAS-like functions and Generic RISC-V kernels

kseniyazaytseva:
* Fixed gemmt:  Added explicit declarations of gemmt functions. Fixed problems associated with incorrect processing of complex conjugate matrices. Error output has also been corrected.
* Fixed  CNC case of zimatcopy functions. In the case of alpha_r = 1.0 alpha_i = 0.0, need to perform the usual complex matrix conjugation
* Fixed imatcopy function memory allocation 
* Fixed calling cblas_cscal testing in ctest
* Removed rotmg unreacheble code
* Added zero size checks
* Added axpby kernels for GENERIC RISC-V target


martin-frbg:
Thanks - I believe that several of these were already fixed in the main `develop` branch in the rather (too) long time since the risc-v branch was created. Did you check ?

kseniyazaytseva:
Yes, the files `riscv64/dot.c` and `interface/imatcopy.c` have been fixed in the `develop` branch, as I see

For gemmt there are still no explicit declarations. The file `interface/gemmt.c`  in `develop` was modified several times, these changes are merged to the PR. But the main problem with incorrect processing of complex conjugate matrices wasn't fixed.

martin-frbg:
Ah thanks, that's unfortunate - seems I never commited the changes to cblas.h - and there's no question that the implementation of the complex conjugate case is/was broken.

martin-frbg:
BTW the CSCAL/CSCALTEST confusion in c_cblat1.f appears to be imported from https://github.com/Reference-LAPACK/lapack/blame/master/CBLAS/testing/c_cblat1.f - I wonder if you would like to open a PR there as well ?

kseniyazaytseva:
I agree that it should probably be fixed in Reference-LAPACK. But I'm just focused on OpenBLAS

martin-frbg:
I can create an issue/PR in Reference-LAPACK no problem - I just wanted to make sure you do not want to do this yourself.

PR #4460, opened 2024-01-26
Fix complex conjugate case in GEMMT, add its prototypes; remove overoptimization in zimatcopy_cnc (from #4452)

martin-frbg:
cherry-picked from kseniyazaytseva's #4452 on the RISC-V branch , just in case merging gets delayed further


martin-frbg:
obsoleted by merging the risc-v branch

PR #4461, opened 2024-01-26
Fix two issues with cpuid_riscv64.c

markdryan:
The code that extracts the model name is incorrect.  It searches in the wrong buffer for the ':' and doesn't check to see whether the character is found before incrementing the pointer to skip the colon, potentially resulting in a pointer value of 1.  There's also a mismatch between the LIBNAME used when the target is forced at build time and when it is dynamically detected.  Presumably, these two values should match.  They do on other architectures.

### DIFF
--- a/cpuid_riscv64.c
+++ b/cpuid_riscv64.c
@@ -86,23 +86,29 @@ int detect(void){
   char *pmodel = NULL, *pisa = NULL;
 
   infile = fopen("/proc/cpuinfo", "r");
+  if (!infile)
+    return CPU_GENERIC;
   while (fgets(buffer, sizeof(buffer), infile)){
     if(!strncmp(buffer, "model name", 10)){
       strcpy(model_buffer, buffer);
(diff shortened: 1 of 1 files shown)


markdryan:
Note that I don't have a C910V to test this patch on.

Also, note that this patch will alter the names of the libraries created when the target is not forced at build time.  I believe the name change is needed in order to have consistent library names, regardless of whether the target is forced or detected at runtime, but just wanted to highlight this.  For example, without this patch, the generated library name will be libopenblas64__riscv64_genericp-r0.3.26.dev.a when building with TARGET=RISCV64_GENERIC and libopenblas64__riscv64p-r0.3.26.dev.a without, regardless of whether getarch detected a riscv64_generic CPU or not.  With this patch the library will be called libopenblas64__riscv64_genericp-r0.3.26.dev.a in both cases (assuming a a riscv64_generic CPU is detected, i.e., detect() returns CPU_GENERIC).

PR #4465, opened 2024-01-30
utest: Add tests for zscal

XiWeiGu:
When Target is `HASWELL`, The following error messages will appear:

> TEST 36/43 zscal:i_nan [OK]
> TEST 37/43 zscal:i_nan_inc_2 [FAIL]
>   ERR: test_zscal.c:29  should be true
> TEST 38/43 zscal:nan_i [OK]
> TEST 39/43 zscal:nan_i_inc_2 [OK]
> TEST 40/43 zscal:i_inf [OK]
> TEST 41/43 zscal:i_inf_inc_2 [FAIL]
>   ERR: test_zscal.c:75  should be true
> TEST 42/43 zscal:inf_i [OK]
> TEST 43/43 zscal:inf_i_inc_2 [OK]
> RESULTS: 43 tests (41 ok, 2 failed, 0 skipped) ran in 2 ms

The PR #4419  has been fixed, but there are issues observed in testing when inc is greater than 1.
It's possible that other platforms may also have this issue

### DIFF
--- a/kernel/x86_64/zscal.c
+++ b/kernel/x86_64/zscal.c
@@ -69,16 +69,16 @@ static void zscal_kernel_8( BLASLONG n, FLOAT *alpha , FLOAT *x )
 
 	for( i=0; i<n; i+=4 )
 	{
-		t0 = da_r *x[0] - da_i *x[1];	
-		t1 = da_r *x[2] - da_i *x[3];	
-		t2 = da_r *x[4] - da_i *x[5];	
-		t3 = da_r *x[6] - da_i *x[7];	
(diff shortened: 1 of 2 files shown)


XiWeiGu:
In `kernel/x86_64/zscal.c`, handling the case when `inc > 1`.
Without access to an ATOM machine, it's unclear whether to hand of NAN and INF arguments.(Based on the assembly code, it seems unnecessary to handle it, but actual validation is still required.)

martin-frbg:
Thank you - I had simply forgotten to add the fix to the other branch of the conditional as well.

PR #4467, opened 2024-01-30
Add CBLAS interfaces for BLAS extensions ?AMIN/?AMAX and C/ZAXPYC

### LINKED ISSUE #4455: LoongArch64: AXPBY Interface Correctness Assurance
AXPBY is introduced as BLAS extensions, but I haven't found corresponding correctness tests. LoongArch64 has added corresponding optimizations, but correctness cannot be guaranteed. What should be done in this case?

martin-frbg:
CBLAS_?AMIN/CBLAS_?AMAX were already created by the CMAKE build while unavailable with gmake, on the other hand CAXPYC/ZAXPYC were missing in CMAKE builds - and none of them were exposed in the cblas header.
fixes #4455 (the part not already addressed by #4457)


martin-frbg:
redo this while a bit more awake

PR #4469, opened 2024-01-31
CI: Add github workflow using Apple M

martin-frbg:


martin-frbg:
going to split this into cmake fixes and the originally intended gh workflow

PR #4472, opened 2024-02-01
Merge risc-v branch to develop

sergei-lewis:


martin-frbg:
hey, I know people are getting impatient but my plan was to merge #4457 first (which got opened against the branch) but also make sure that its new tests do not blow up all the other platforms... would you prefer to defer that ?

sergei-lewis:
Makes perfect sense, completely happy to wait with this PR until everything is in place and update this PR as necessary. I've marked the PR as draft, will unmark once everyone is happy with what is happening (or close it if we decide to go for a different strategy).

I've tested this on x86, RISC64_GENERIC, RISCV64_ZVL128B and RISCV64_ZVL256B locally. A big part of my reason for kicking a PR is, given the size of the changeset and the fact that it includes many changes outside riscv64-specific modules I am super keen to kick off discussion early and also take advantage of OpenBLAS CI to exercise all the other configurations to get visibility on any potential problems.

martin-frbg:
Must admit that the size and range of this changeset is worrying me a bit as well, it used to look a lot more harmless when it was just a trivial update of kernel/riscv64... maybe you're right and rebasing the new utests after the merge makes more sense...

sergei-lewis:
An alternative strategy might be to rebase risc-v on current develop, deal with the fallout on the branch, then kick a new PR based on that. It might lead to a smaller changeset, and/or make it feasible to split the merge into several smaller PRs rather than one giant one; but it's no magic bullet, and it's not immediately clear that things would be better that way instead of worse.

martin-frbg:
Yes, lets go with this one unless anyone complains in the next few hours. The utest extensions are going to open their own can of worms that is totally unrelated to risc-v, and merging develop into risc-v would appear to create at least as many merge conflicts and chances for fallout than doing it here. Add to that that we do not have any CI job that runs on actual RISC-V hardware yet, so the chances of detecting fallout might be smaller. 
It is just that I was fixated on getting other PRs merged before tackling this one.

martin-frbg:
Looking good so far, only two timeouts due to slow CI hardware

martin-frbg (approved):
LGTM

sl236:
I’m happy if you’re happy!

martin-frbg:
Hmm... now that I have merged this, the hang in the C910V qemu job appears to be persistent :( @sergei-lewis  can you check if you see it on actual hardware too ?

sergei-lewis:
Apologies, was travelling. Will take a look ASAP.

martin-frbg:
I've switched AXPY back to generic kernels to get around the hang, but I suspect this may be just another thing that only happens in qemu.

sergei-lewis:
PR 4497 fixes the hang and reenables the vectorised AXPY kernels on C910V.

Any reason we shouldn't add CI testing for the new ZVL128B / ZVL256B targets, btw? I'll put together a PR if not.

martin-frbg:
thanks. no reason to not include them in CI, it isn't that long since the branch got merged, is it ? :)

PR #4474, opened 2024-02-01
Vectorize in-copy packing/copying for SGEMM - up to 4X faster.

ChipKerchner:
Vectorize in-copy packing/copying for SGEMM - up to 4X faster.

Contributors:
N=16 - Rajalakshmi Srinivasaraghavan
N=8 and 4 - Chip Kerchner


martin-frbg:
Great, thanks.

PR #4485, opened 2024-02-05
[WIP] Add a build option to suppress versioning and softlinking of the library

### LINKED ISSUE #4468: [Feature request] Allow linking a single non-versioned shared object
Over a few PRs to MacPython/openblas-libs, I changed the end result to be a Python wheel. This required packaging a single shared object with no symbolic links. I needed to
- use a prefix so the shared object is unique and can be globally loaded along side a different OpenBLAS build
- changed the Makefiles to allow building a single shared object, with no versioning and no symlinks
- changed the pkg-config file to reflect this

[Here](https://github.com/MacPython/openblas-libs/blob/main/patches/0001-create-a-single-shared-object.patch) is the patch to OpenBLAS 0.3.26 to remove all the symlinking and versioning. Note the prefixing requires [this line](https://github.com/MacPython/openblas-libs/blob/7bc05d17b4ef4aaa0112a66a83a8d1cceb60fffc/patches/0001-create-a-single-shared-object.patch#L131) and [this change](https://github.com/MacPython/openblas-libs/blob/7bc05d17b4ef4aaa0112a66a83a8d1cceb60fffc/patches/0001-create-a-single-shared-object.patch#L179-L180) in order to support prefixes in the pkg-config file. The rest of the patch is "just" removing things. 

Would you be open to merging these changes, toggled with a build option, to reduce openblas-lib's maintenance burden?

martin-frbg:
fixes #4468 
(simplified approach compared to #4252 and the MacPython patch presented in #4468 )


mattip:
Thanks!

martin-frbg:
Hope it matches your expectations...

PR #4490, opened 2024-02-07
Add missing CPU ID definitions for old versions of AIX.

ChipKerchner:
Add missing CPU ID definitions for old versions of AIX.  Without these, the dynamic arch will return an unknown CPU ID.

### DIFF
--- a/driver/others/dynamic_power.c
+++ b/driver/others/dynamic_power.c
@@ -43,6 +43,13 @@ char *gotoblas_corename(void) {
 #define CPU_POWER9   9
 #define CPU_POWER10 10
 
+#ifndef POWER_9
+#define POWER_9         0x20000         /* 9 class CPU */
+#endif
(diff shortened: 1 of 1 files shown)


ChipKerchner:
@martin-frbg When is 0.3.27 scheduled for release?

martin-frbg:
tentatively scheduled for end of March (I usually put a date that's 2 to 3 months out in the next "Milestones" entry of the project) but flexible if there is an important bugfix or major code change. With the merging of the RISC-V branch _and_ now the discovery of a fairly serious omission in the CMAKE build, it may make sense to plan for end of February instead. 
May I ask why you're asking - do you hope to see your cpuid fixes in a speedy release, or do you have something else in the pipeline that you need time to work on ?

ChipKerchner:
We had a group ask when the next OpenBLAS branch release is going to be.  That's all.

PR #4498, opened 2024-02-09
blas_server_win32.c pass to clean up code

mseminatore:
Per Discussion #4417 there was a collection of conflicting styles, indentation, etc. that has made it difficult to parse the code. Taking care not to change functionality in this PR, nor touch every single line of code. Also, the goal is to respect what appear to be overall coding sensibilities as expressed in blas_server.c et al.

The goal is that it becomes easier to compare blas_server.c and blas_server_win32.c and to keep the two in sync. And perhaps to find a way to merge the two using common low-level abstractions.


martin-frbg:
Thank you.

PR #4501, opened 2024-02-10
Adding basic support of benchmarks into CMake

dmikushin:
Adding basic support of benchmarks into CMake for single, double, single complex and double complex cases. Each benchmarking target has a suffix to identify the data type, for example `./benchmark_gemm3m_COMPLEX_DOUBLE` is a `gemm3m.c` source compiled with `COMPLEX` and `DOUBLE` macros defined.

This PR is done under assumption that the `benchmarks` folder is not yet wired into CMake compilation.

This PR is not trying to re-create the `benchmarks/Makefile` behavior. Instead, the purpose of this PR is to compile as many tests as possible from the existing material.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -452,6 +452,34 @@ if (BUILD_SHARED_LIBS AND NOT ${SYMBOLPREFIX}${SYMBOLSUFFIX} STREQUAL "")
   endif()
 endif()
 
+if (BUILD_TESTING)
+  find_package(OpenMP REQUIRED)
+  file(GLOB SOURCES "benchmark/*.c")
+  foreach(source ${SOURCES})
+    get_filename_component(name ${source} NAME_WE)
+    if ((NOT ${name} STREQUAL "zdot-intel") AND (NOT ${name} STREQUAL "cula_wrapper"))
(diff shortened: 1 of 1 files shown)


martin-frbg:
right, the benchmarks are currently not available in the cmake build, but the benchmark suite (which mostly dates back to GotoBLAS) is due for an overhaul anyway.

dmikushin:
We may still wish to enable it in its current state to provide a lot more code coverage. Otherwise, many functions are not tested at all. For example, cgemm3m/zgemm3m are completely untested, unless we accept these tests for now as a compromise. Of course, someone may work on bringing them to a much better state later.

I'm also using this PR to demonstrate the defect in https://github.com/OpenMathLib/OpenBLAS/issues/4500

martin-frbg:
The benchmarks make no claim to test for accuracy. The GEMM3M functions are supported on a handful of architectures only, and tests for them are included in test/ctest, though it turns out they are not getting built by default currently.

dmikushin:
I understand, still having more coverage is better than nothing. I mean, "Perfect is the enemy of good." (c) Voltaire

martin-frbg:
will follow this with a PR to make OpenMP optional (only required in the smallscaling benchmark) and probably make building the benchmarks optional as a whole (the gmake build does not include them by default either)

PR #4502, opened 2024-02-12
Add missing USE_GEMM3M macro into CMake

### LINKED ISSUE #4500: ZGEMM3M is broken due to cgemm3m_unroll_n never assigned
I tried ZGEMM3M on Haswell, it does not work, gives SIGFPE (division by zero). It happens because the divider is `cgemm3m_unroll_n`, which is never assigned with any value.

@wjc404 , could it be that the `cgemm3m_unroll_n` assignment was accidentally removed?

dmikushin:
Adding `USE_GEMM3M` macro into CMake kernel targets, so that the *gemm3m functions and parameters can be included into the `gotoblas` structure. Fixes #4500


martin-frbg:
Oh right, thanks. That's almost as bad as way back when the CMAKE build did not know about USE_TRMM

PR #4513, opened 2024-02-21
Fix get_num_cores for AIX.

ChipKerchner:
Fix get_num_cores for AIX.


martin-frbg:
Thank you - actually not quite sure how I did this, probably incomplete copying of a diff from one of the gccfarm machines

PR #4515, opened 2024-02-21
resolve second_ conflict which breaks xlf timef

frjohnst:
I am building openblas on an IBM Power AIX system using openxlf/openxlc.
I tried testing my build with a program which calls xlf's timef utility and found timef was incorrectly returning zero instead of the time in milliseconds. This problem happens because xlf's timef routine internally calls a second_ routine, and it should be
calling the second_ in xlf's runtime library, but instead it is calling a second_ in OpenBLAS (which only returns zero). 
The fix is to remove second_ from openblas builds done with xlf. For consistency, dsecnd_ was also removed, since the
dsecnd_ created by OpenBLAS is also only returning zero.

### DIFF
--- a/lapack-netlib/SRC/Makefile
+++ b/lapack-netlib/SRC/Makefile
@@ -101,8 +101,10 @@ SCLAUX = la_constants.o \
    slaset.o slasq1.o slasq2.o slasq3.o slasq4.o slasq5.o slasq6.o \
    slasr.o  slasrt.o slassq.o slasv2.o spttrf.o sstebz.o sstedc.o \
    ssteqr.o ssterf.o slaisnan.o sisnan.o \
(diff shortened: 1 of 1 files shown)


ChipKerchner:
I've verified that Frank's changes do work with xlf.

RajalakshmiSR:
Does it work for both Linux and AIX?

frjohnst:
The changes in openblas/lapack-netlib/SRC/Makefile prevent second_ or dsecnd_ from being built when xlf/xlc are used on AIX or Linux.

frjohnst:
As expected, this Makefile change did not affect behavior with gcc/gfortran on Power/Linux.

martin-frbg:
Thanks - that's due to a kludge from the LAPACK integration.. actually we should be using something more sophisticated in the lapack-netlib/INSTALL parts (or the make.inc we generate) to pick the appropriate implementation for each system.

PR #4517, opened 2024-02-22
Shared library support for AIX

ayappanec:
Through this PR, we are adding shared library support for AIX. 
AIX native linker doesn't have any option (like "--whole-archive") to create a shared library from a static archive file. The strategy employed here is to use the AIX native nm command to create an export file and then use that to create a shared library.


martin-frbg:
Thanks - interesting use of nm...

ayappanec:
Thanks @martin-frbg for the support.

PR #4520, opened 2024-02-23
fix conlict between PR 4515 and AIX shared obj support

frjohnst:
PR 4515 (removing second_ and dsecnd_ from IBM xlf builds) and the changes to enable AIX shared object support were delivered on the same day and were not tested together.  second_ and dsecnd_ must also be conditionally removed from linktest.c, used to verify the shared object build, and that is the purpose of this change.  The change was made in exports/Makefile because the F_COMPILER make variable is available there.


ChipKerchner:
@martin-frbg This needs to be included in 0.3.27

martin-frbg:
right, forgot the milestone tag but it is primarily for drafting the changelog anyway - anything in the develop branch at the time of release creation will normally end up in the release, can't recall this ever having been different

PR #4524, opened 2024-02-25
Enable GEMM3M tests when the architecture supports GEMM3M

martin-frbg:
related to #4500 , also solves some portability problems in the new utests from #4499


abhishek-iitmadras:
@martin-frbg 

1. is aarch64 support GEMM3M ? Currently i don't see any dedicated 3M kernels for aarch64.
But as per [link](https://www.openblas.net/Changelog.txt?utm_source=chatgpt.com#:~:text=%2D%20Changed%20the%20unfinished%20implementation%20of%20GEMM3M%20for%20the%20generic%0A%20%20target%20on%20all%20architectures%20to%20at%20least%20forward%20to%20regular%20GEMM) the generic implementation of ?gemm3m was changed to fall back to regular 4‑multiply GEMM when no architecture‑specific kernel is present 
Am i right ?

martin-frbg:
There are currently no optimized GEMM3M kernels for aarch64 or any other "non-Intel" architecture, unlike x86/x86_64 (and even there the majority dates back to GotoBLAS). 
And there are also no (functioning) generic kernels for GEMM3M, only a dummy put there by Xianyi (years ago with no followup) that would print out the arguments. Which is why I added the crude fallback to regular GEMM to avoid having "generic x86" builds of OpenBLAS crash in GEMM3M calls.

PR #4532, opened 2024-02-27
Cgemm zgemm c code

austinpagan:
This modification is specific to the Power10, and cannot affect anyone else.
We're just changing our CGEMM and ZGEMM implementations from assembler code to C code,
to match work already done on other similar routines.  
Two added files, and one modified file to point to them.  That's it.
I will be happy to address any concerns this raises for people...


austinpagan:
I doubt that any of my changes would have caused a build to last more than 60 minutes...?

martin-frbg:
I need to find someplace with faster hardware for those Intel-Apple jobs - they have a greater than 50 percent chance to get scheduled on some tired old box in the Azure cloud

PR #4534, opened 2024-02-27
Omit redundant prefixes or suffixes in library naming

### LINKED ISSUE #4528: Problems with windows and LIBNAMEPREFIX
#4485 went a long way to adding a prefix and suffix to the lib name, and allowing `FIXED_LIBNAME=1`, thanks.

But ... there is a difference between windows and posix. I used HEAD in https://github.com/MacPython/openblas-libs/pull/141. On non-windows, I need to define both `SYMBOLPREFIX=scipy_` and `LIBNAMEPREFIX=scipy_` (and correspondingly, `SYMBOLSUFFIX=64_` and `LIBNAMESUFFIX=64_` when using `INTERFACE64=1`). But on windows, I must only define `SYMBOLPREFIX=scipy_`. Defining `LIBNAMEPREFIX=scipy_` as well results in `libscipy_scipy_openblas...`

I currently worked around this since the CI runs are separate, but I wonder if this is intentional? The "offending" uses are in `Makefile.system`.

martin-frbg:
fixes #4528


mattip:
Thanks. This seems to be working for the openblas-libs wheels now.

PR #4537, opened 2024-02-28
utest: Add library dependencies to the target executable file

XiWeiGu:


martin-frbg:
I wonder if you actually came across a build failure due to this missing dependency, as it has "always" been this way ?

XiWeiGu:
Thank you for your prompt response. This is not a build failure issue, but rather a minor improvement. When a test case fails in 'utest', fix the bug and then compile again. 'openblas_utest' and 'openblas_utest_ext' may not update promptly, hindering the verification of correctness. It is necessary to manually enter the 'utest' directory, execute 'make clean', and then compile again. This issue does not occur when building with CMake.

martin-frbg:
I see - though given the involved nature of the makefile build, it is usually safer to do a complete `make clean` and full rebuild after making any changes anyway. (I know it probably should not be that way...)

XiWeiGu:
Alright, thank you for your reply. I understand now, and I will close this PR.

martin-frbg:
not saying we shouldn't do this, just that there's more in the build system that could hinder a rebuild-without-clean

PR #4543, opened 2024-03-08
Makefile: add NO_UTEST

ffontaine:
Add `NO_UTEST` to allow the user to disable building utest

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -168,8 +168,10 @@ ifndef NO_FBLAS
 	$(MAKE) -C test all
 endif
 endif
+ifneq ($(NO_UTEST), 1)
 ifneq ($(ONLY_CBLAS), 1)
 	$(MAKE) -C utest all
+endif
 endif
 ifneq ($(NO_CBLAS), 1)
 ifneq ($(ONLY_CBLAS), 1)


martin-frbg:
Thanks - I wonder why you would want to do this ?

ffontaine:
Because utest unconditonally uses `fork` resulting in the following build failure on toolchain without MMU:
```
test_fork.c:(.text+0x120): undefined reference to `fork'
```

Full build log: http://autobuild.buildroot.org/results/b9d/b9d097ae077f2cb40fc60de94bf25423636af005/build-end.log

martin-frbg:
I see, so it is actually this one test that should be excluded as not supported on your platform. Does the uclibc toolchain define anything that could be used to shortcut the fork test ? (Assuming that everything else needed by the OpenBLAS library itself is implemented)

ffontaine:
When MMU is disabled on an uclibc-ng toolchain, `usr/include/bits/uClibc_config.h` will contain:

```
#define __ARCH_HAS_MMU__ 1
#undef __ARCH_USE_MMU__
```

However, I still think that adding `NO_UTEST` is a better solution that adding uclibc-ng specific code to openblas. `NO_UTEST` could be used for example to speed up build time. 
An other solution would be to add MMU dependency to openblas in buildroot.

martin-frbg:
build should be fixed by #4544

ffontaine:
Thanks, however, `test_post_fork.c` shall also be patched to fully fix the build failure:
```
test_post_fork.c:(.text+0xb4): undefined reference to `fork'
```

martin-frbg:
sorry for that, now addressed in #4546

ffontaine:
Thanks, build failure is fixed, closing this PR.

PR #4552, opened 2024-03-11
Allow negative INC_X in arm64 ThunderX2T99 and Apple M ?NRM2

### LINKED ISSUE #4551: BUG: dnrm2 basic failures on a subset of platforms (observed in SciPy)
We recently experimented with bumping SciPy OpenBLAS version used in wheels from a pretty old version (`v0.3.20-571-g3dec11c6`) to `0.3.26` stable release in https://github.com/scipy/scipy/pull/20215, in the hopes of fixing the reproducer in https://github.com/scipy/scipy/issues/16930, which can be summarized as the following test, which we should perhaps adjust to be a tractable C-level test in this project:

```python
def test_gh_169309():
    x = np.repeat(10, 9)
    actual = scipy.linalg.blas.dnrm2(x, 5, 3, -1)
    expected = math.sqrt(500)
    assert_allclose(actual, expected)
```

Bumping to `0.3.26` appeared to fix the issue on x86_64 Linux, but not on some other platforms, including MacOS ARM and Linux `aarch64`. In both cases, looks like `actual` ends up as `0.`.

I could probably help draft in the regression test if you tell me where to put it, but I'm unlikely to have the expertise to provide the fix.

martin-frbg:
probably fixes #4551

### DIFF
--- a/kernel/arm64/dznrm2_thunderx2t99.c
+++ b/kernel/arm64/dznrm2_thunderx2t99.c
@@ -345,7 +345,7 @@ FLOAT CNAME(BLASLONG n, FLOAT *x, BLASLONG inc_x)
 #endif
 	FLOAT ssq, scale;
 
-	if (n <= 0 || inc_x <= 0) return 0.0;
+	if (n <= 0 || inc_x == 0) return 0.0;
 
 #if defined(SMP)
 	if (n <= 10000)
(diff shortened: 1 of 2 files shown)


mattip:
I can create a new openblas-libs tarball (used by scipy) to test this once it is merged.

PR #4562, opened 2024-03-14
Fold wiki contents into formal documentation, build-able with `mkdocs`

honno:
This PR essentially ports the contents of the [wiki](https://github.com/OpenMathLib/OpenBLAS/wiki) into the repo, which can be built into HTML docs via [`mkdocs`](https://www.mkdocs.org/).

I made what I believe are fairly uncontroversial changes to make these docs read better, and although there is room for improvement anything more opinionated seems best served in follow-up PRs. Any content additions like improved tutorials or project history seem best served in follow-up PRs too.

A live preview of this PR is available at https://honno.dev/OpenBLAS/, subject to being out-of-date. To serve the docs locally, you'll need Python installed, and then:

```
$ pip install mkdocs mkdocs-material
$ mkdocs serve  # check http://127.0.0.1:8000/
```

A GitHub action is introduced to build the docs and deploy on the `gh-pages` branch. I believe as-is this PR would update `openmathlib.github.io/OpenBLAS` to these docs if merged. Possibly its best if these docs just supersede the existing `openblas.net`, or live at `openblas.net/docs`—either way this would probably require a repo maintainer changing some repo settings related to GitHub Pages.

Note I've left the wiki refs in the [README.md](https://github.com/OpenMathLib/OpenBLAS/blob/develop/README.md) for now, as these would depend on the location of the wiki (unless we wanted to reference the doc markdown files e.g. `docs/install.md`). I'd suggest this PR could go in first, then a future PR can update references to a live documentation, and then we delete the GitHub wiki.

cc @martin-frbg @rgommers


martin-frbg:
thanks - not sure if the deletion of Usage.md is intentional, at least I do not think I saw its contents integrated in the new files ?

honno:
> thanks - not sure if the deletion of Usage.md is intentional, at least I do not think I saw its contents integrated in the new files ?

I forgot about this but just checked—the first two points on program termination and multi-threading exist in the FAQ, and the rest of the content exists in the user manual.

So yeah did this to remove duplication, although maybe in this instance its worth preserving? Possibly the README could link to the relevant docs, although as mentioned something we might need to leave for a future PR.

martin-frbg:
> I forgot about this but just checked—the first two points on program termination and multi-threading exist in the FAQ, and the rest of the content exists in the user manual.

Fair enough - though I would prefer to keep the old Usage.md file around for now. I think there needs to be a good bit of updating and de-duplicating between the FAQ and the other parts of the Wiki documentation before this can really replace the other documentation.

Unfortunately I am a bit unclear about the operation of mkdocs vs the gh-pages - I do not want to replace the entirety of openblas.net just yet as that page has different information on it, ~~but I do not see anything in the project settings that would let me reconfigure it to inject the mkdocs-generated files into a docs subdirectory while keeping the static content.~~ it seems that adding `site_url https://openblas.net/docs/` to the mkdocs.yml would be sufficient to achieve this (? If true, I would simply replace the current link to the wiki on the static page and we could discuss a potential merging of the page contents later. For one, I envisage presenting updated performance graphs for various architectures on the starting page again at some point)

docs/index.md
```diff
@@ -0,0 +1,11 @@
+## Introduction
+
+OpenBLAS is an optimized Basic Linear Algebra Subprograms (BLAS) library based on [GotoBLAS2](https://www.tacc.utexas.edu/research-development/tacc-software/gotoblas2) 1.13 BSD version.
+
+OpenBLAS implements low-level routines for performing linear algebra operations such as vector addition, scalar multiplication, dot products, linear combinations, and matrix multiplicationon. OpenBLAS makes these routines available on multiple popular platforms, covering desktop (Windows, Linux, Mac)  and mobile (Android, iOS) operating systems, as well as different architectures (namely x86 and ARM).
```
martin-frbg: "matrix multiplication." (typo). And I'd like to leave out the "popular" here, plus make it "server, desktop and mobile" and change the "namely x86 and ARM" to something like "including ARM, MIPS, PPC, RISC-V, x86 and zarch"

honno:
> though I would prefer to keep the old Usage.md file around for now.

Restored :+1:

> I do not want to replace the entirety of openblas.net just yet as that page has different information on it

Pushed a change to this effect. Yep `mkdocs` is limited here, hence some updates to the `docs.yml` workflow.

---

Also I see that https://www.openblas.net/ is not linked to the [`gh-pages`](https://github.com/OpenMathLib/OpenBLAS/tree/gh-pages) repo here, can't figure whether that site is sourced from a GitHub repo or something else :thinking: Assumedly we'd want that site to deploy from this repo's `gh-pages` branch.

martin-frbg:
www.openblas.net lives in its own repo currently, not sure if I want to change that while we are still experimenting. Assuming that the new docs tree is not going to change all that often, maybe we can live with some manual or semiautomatic copypasta on release days.

honno:
> [www.openblas.net](http://www.openblas.net) lives in its own repo currently, not sure if I want to change that while we are still experimenting.

In that case, I feel we could revert my changes so the docs build to the top-level of the `gh-pages` branch, and in this repo's settings it can be linked to `http://www.openblas.net/docs`? It's my assumption that's possible at least.

martin-frbg:
I'm trying to find out how to do this - currently it seems the gh-pages would deploy to www.openmathlib.org (due to the change from xianyi's project to an organisation). This probably won't hurt as long as there is nothing else under the OpenMathLib umbrella - but I have no idea if this would conflict with the automatic redirect from the former xianyi/OpenBLAS, or if www.openblas.net is actually seperated now.

honno:
> I'm trying to find out how to do this - currently it seems the gh-pages would deploy to [www.openmathlib.org](http://www.openmathlib.org) (due to the change from xianyi's project to an organisation). This probably won't hurt as long as there is nothing else under the OpenMathLib umbrella - but I have no idea if this would conflict with the automatic redirect from the former xianyi/OpenBLAS, or if [www.openblas.net](http://www.openblas.net) is actually seperated now.

@martin-frbg any updates on this? If there's something I can help with let me know!

martin-frbg:
Let me get the release out of the way first, then I'll try what happens to the openblas.net page when this is merged

martin-frbg:
OK, so this appears to have - mostly - worked as expected. One unexpected side effect (for me at least) was that deployment (and associated breaking of the automatic forwarding between openmathlib.org and openblas.net) uncovered a historical landing page from twelve years ago that I have now "manually" overwritten with the index.html and stylesheets from openblas.net, and added a link to the new docs tree.

rgommers:
Maybe it would make sense to fold the landing page into the docs build? So the front page would be more like the old landing page, and that would prominently link to docs (also in navbar)? Or merge the two?

martin-frbg:
so far I see no benefit in having the landing page generated as well, though that would be one way to solve the current problem of having no link back to it from the docs. may also need to tweak the style a bit - right now both old and new look a bit drab each in their own way.

honno:
So I see https://github.com/OpenMathLib/openmathlib.github.io is responsible for the top-level contents of `http://www.openmathlib.org/`—could that repo be deleted and this repo instead be responsible? That would mean everything in the `gh-pages` branch wouldn't be living in `/OpenBlas`, and we could go from there.

martin-frbg:
Not sure about that yet, as OpenMathLib was supposed to become an umbrella for various loosely related libraries in the (unspecified) future.

Issue #4066, opened 2023-05-28
Build openBLAS with oneTBB parallelism support


vineel96:
Hello,
I have following doubts:
1. how to build openBLAS with oneTBB parallelism support instead of openmp?
2. which combination is better to build openBLAS with? openBLAS +openMP or  openBLAS +oneTBB or openBLAS+pthreads?

martin-frbg:
There is currently no support for TBB in OpenBLAS, what people appear to be doing (with varying success if you search for earlier issues mentioning TBB) is use a single-threaded build of OpenBLAS with their TBB-parallelized program.
The choice of pthreads or OpenMP depends largely on what your main program uses. At least in theory, OpenMP would offer better thread safety and easier thread affinity handling, but may incur some overhead. On many/most platforms, OpenMP is implemented on top of pthreads anyway, and if your main
program uses OpenMP you will want OpenBLAS to be either built with OpenMP, or built-single threaded, to avoid having to sets of threads that do not know of each other and compete for resources.

vineel96:
Hi @martin-frbg ,
Thanks for the reply.
1. is there a way to modify makefile to build openBLAS with oneTBB similar to how openmp is given as option while building openBLAS? or only way is manually write oneTBB based code and use single threaded openBLAS?
2. I am trying to build openBLAS with oneTBB so that it is **algorithm/code independent**. which means whatever algorithm that runs on top of openBLAS should make use of oneTBB **without actually modifying the algorithm**. is this right way to try? and can we see any performance gain by doing this?

martin-frbg:
there is currently no support for this - only an unfinished concept from four years ago in PR #2255

brada4:
1/ TBB equivalents of OMP PARALLEL pragmas
2/ TBB mutexes
3/ TBB malloc in place of current sbrk/mmap allocator

vineel96:
Hi @martin-frbg and @brada4,
Thanks for the comments. I have some questions:
1. The file changes that are proposed in https://github.com/xianyi/OpenBLAS/pull/2255 pull request isnt enough to merge it to openblas? Why its not merged or further taken to implement if any if it boosts performance?
2. Also, can you suggest any room for improvement in openblas in threading part which helps improve parallelism and improve speed? Like any Gsoc project on this or i want to know if there is any scope in improving the threading part?
[...]

PR #4577, opened 2024-03-25
Introduced callback to Pthread, Win32 and OpenMP backend

shivammonaka:
**Motivation:**  This pull request enables OpenBLAS threading to be composable with caller multithreading. It allows OpenBLAS threading to be composable with callers using Intel’s Thread Building Blocks (TBB), etc.  

**Brief description:** 
1.	This work is a completion of work-in-progress pull request (PR) https://github.com/OpenMathLib/OpenBLAS/pull/2255, aimed at achieving thread composability in OpenBLAS with all threading backends(OpenMP, Pthreads and win32).
2.	Caller is required to register a callback to dynamically change the threading backend. Instead of creating its own threads, OpenBLAS will now pass on the work (blas_queue_t) to the registered callback.
3.	The integration of TBB on the caller side for executing the callback has provided us with significant performance improvement, particularly in nested parallelism scenarios. 
4.	A typical prototypical user callback would look like:
```
void myfunction_ (int sync, openblas_dojob_callback dojob, int numjobs, size_t jobdata_elsize, void *jobdata, int dojob_data)
{
    #pragma omp parallel for
    for(int i = 0; i < numjobs; i++)
    {
        void *element_adrr = (void *) (((char *)jobdata) + ((unsigned) i)*jobdata_elsize);
        dojob(i, element_adrr, dojob_data);
     }
    return;
}
```

5.	A typical callback registration would look like:
       `openblas_set_threads_callback_function(myfunction_);`

**Kindly note:** Extending this feature to Pthreads and Win32 threading backends required us to make additional changes, as outlined below:
1.	We have introduced exec_threads() function in pthreads and win32, which helps in setting the synchronization environment and calling the inner_thread routine. This change aligns with the existing OpenMP flow.
2.	We have introduced adjust_thread_buffers() function, similar to OpenMP, for initializing global thread buffers instead of the existing local buffers initialized in blas_thread_server.


goplanid:
@ [stevengj](https://github.com/stevengj) @martin-frbg

martin-frbg:
Thanks. CI suggests the changes currently make Windows/MSYS builds prone to hanging in/after a fork() for some reason - at least that is what the apparently hanging "test_kernel_regress" in utest does. 
The failed jobs on Azure appear to be down to older tools not accepting C99 initializers, and wanting a dedicated build rule for the new blas_server_callback file in the Makefile.
I'll try to squeeze in some local testing...

martin-frbg:
Sorry, I'm not sure I understand the latest update here, and which state of the respective files you are restoring with it - would you prefer to convert this to a draft for now ?

shivammonaka:
Hi @martin-frbg,
Sorry about the confusion, I was doing some local experiments and accidently pushed it on main branch. I will revert the changes and try to fix failing checks.

martin-frbg:
Ok, thanks, it was indeed a bit confusing to see seemingly unrelated blocks copied around without explanation, and then removed again in the same style. Unfortunately I hadn't gotten further than checking that the code still works on Linux/Unix when the new callback is not used, and that it compiles and works locally on Windows with gcc (but no explanation for the segfaults with clang)

shivammonaka:
Hi @martin-frbg ,
There is 1 check still failing. Is it because of some resource constraint or Is my code hanging somewhere.

martin-frbg:
these OSX jobs on Azure have been randomly timing out for a while - certainly unrelated to your PR. I still need to investigate if it is simply due to old hardware in the Azure cloud becoming too slow for dynamic_arch builds, or if some test started to hang on Nehalem

rageshhajela16:
> these OSX jobs on Azure have been randomly timing out for a while - certainly unrelated to your PR. I still need to investigate if it is simply due to old hardware in the Azure cloud becoming too slow for dynamic_arch builds, or if some test started to hang on Nehalem

@martin-frbg Thanks for the review. Are there any other comments for us to incorporate?

martin-frbg:
No concerns, I've just been very busy lately, sorry. I hope to merge this tomorrow after trying out and documenting what integration with TBB would involve.

martin-frbg:
fixes #4066

ogrisel:
@shivammonaka is your OpenBLAS+TBB integration packaged somewhere?

shivammonaka:
@ogrisel, this is a callback feature — the user can choose any threading backend to execute the OpenBLAS load. However, they must respect OpenBLAS’s internal threading constraints: each load requires a specific number of threads to complete execution. If not, the process can hang indefinitely due to two synchronization barriers within OpenBLAS’s matrix multiplication algorithm.

PR #4579, opened 2024-03-25
Fix global (static) constructor priority so that OpenBLAS gets initialized before other libraries.  Other unit test AIX fix.

ChipKerchner:
Fix global (static) constructor priority so that OpenBLAS gets initialized before other libraries.  Other unit test AIX fix.


ChipKerchner:
Please include this is 0.3.27

RajalakshmiSR:
@martin-frbg Do you have any tentative date for 0.3.27 release?

martin-frbg:
March 31 (see Milestones tab on the Issues page)

PR #4617, opened 2024-04-06
[Doc]Update user_manual.md for static linker

cyk2018:
when I use static link method to compile , result is undefined with pthread_create, so we should add -lpthread


cyk2018:
#4619

PR #4618, opened 2024-04-06
[Doc]Update user_manual.md

cyk2018:
update user_manual to fix runtime error with -lopenblas.


cyk2018:
#4619

martin-frbg:
Thank you - while mentioning LD_LIBRARY_PATH in this context may be a good idea, the documentation should not contain unnecessary external links (if only to other github-hosted projects), and several options for avoiding the linker/loader error are already mentioned under "Link shared library" right above the example.

cyk2018:
> Thank you - while mentioning LD_LIBRARY_PATH in this context may be a good idea, the documentation should not contain unnecessary external links (if only to other github-hosted projects), and several options for avoiding the linker/loader error are already mentioned under "Link shared library" right above the example.

Ok, Thanks for your reply.

PR #4624, opened 2024-04-10
Remove -openmp flag from XLF (since it doesn't support it).

ChipKerchner:
Remove -openmp flag from XLF (since it doesn't support it).


martin-frbg:
I wonder if it did in the (fairly distant) past, as this snippet appears to be a remnant from GotoBLAS

PR #4627, opened 2024-04-11
remove extraneous suffix from shared object SONAME

mattip:
Before, this created a shared object with SONAME `libscipy_openblas64_64_.so`. Now it is correctly `libscipy_openblas64_.so`. See MacPython/openblas-libs#151 and numpy/numpy#26258


mattip:
The OSX_OpenMP_Clang_cmake job ran for more than 60 minutes. Should the timeout be made longer?

martin-frbg:
weird, I thought I had covered all possible combinations.
regarding Azure, I do not think it is possible to extend the timeout in a "free" account. The behaviour of that job varies depending on which generation of hardware it gets scheduled, I started looking into reducing its dynamic_list but still need to check if one of the tests actually hangs when run on a certain cpu.

PR #4628, opened 2024-04-12
CI: add openblas_utest_ext for c910v, mips64 and loongarch64

XiWeiGu:


martin-frbg:
thanks. the dsum failures appear to be due to yet another uninitialized variable in the test code. will fix that first before merging

PR #4639, opened 2024-04-14
mips64: Fixed scalar optimization bug in dzsum

XiWeiGu:


martin-frbg:
Oops, sorry, did not realize we were both working on this after my blundering around last night.

PR #4642, opened 2024-04-15
CI: Add clang test for loongarch64

XiWeiGu:


martin-frbg:
AFAICT, github-hosted runners do not provide Debian 10 (or any version of Debian), I guess you could try if `runs-on: ubuntu-20.04` is close enough to support installing those packages

XiWeiGu:
> AFAICT, github-hosted runners do not provide Debian 10 (or any version of Debian), I guess you could try if `runs-on: ubuntu-20.04` is close enough to support installing those packages

Thank you for the reminder, I will revise promptly.

XiWeiGu:
By the way, I also wanted to ask how to enable large page memory allocation. Currently, I haven't found any available options, so I had to manually modify the code to achieve it.

martin-frbg:
There is an (almost) undocumented compile parameter named HUGETLBFILE_ALLOCATION that is to be set to the mount point of the hugetlbfs (for creation of a backing pseudofile named gotoblas.<pid>), or you can define HUGETLB_ALLOCATION=1 for accesses through shared memory instructions. Both are not used by default on any architecture, so there is a small risk that they may land you in code whose performance has not been tested since the early days of GotoBLAS

martin-frbg:
(that said, a build with `make HUGETLB_ALLOCATION=1` completes and tests without any issues on x86_64)

Issue #4644, opened 2024-04-15
Multi-threaded DGEMM becomes less efficient on many-core CPUs


yamazakimitsufumi:
OpenBLAS DGEMM achieves high efficiency, for example, over 90% of peak performance with 1 thread on Graviton3E, but the efficiency drops to about 73% when running DGEMM with 64 threads.
As is known, it is becoming difficult to keep high efficiency for multi-thread execution on recent many-core CPUs,  even if high-performance kernels are implemented for single-thread execution.

I am considering to adjust the shape of the submatrix handled by each thread by modifying [2D thread distribution](https://github.com/OpenMathLib/OpenBLAS/pull/1320).
I would appreciate it if you could let me know if you have any suggestions.

![graph](https://github.com/OpenMathLib/OpenBLAS/assets/166666378/12597c18-6746-4dcb-995f-d55c40bc4906)

brada4:
8mb matrices are probably less than caches by magnitude.

martin-frbg:
Closing as fixed by your #4655 , thank you again

PR #4655, opened 2024-04-18
Expanding the scope of 2D thread distribution to improve multi-threaded DGEMM performance

yamazakimitsufumi:
This pull request proposes a fix to the issue #4644 .

The currently implemented 2D thread distribution in level3_thread.c works well for small matrices, however, it falls into a simple one-dimensional distribution in the M direction as the size of matirix becomes larger.​ This pull request improves thread parallell performance for large matrices by expanding the scope of 2D thread distribution.​

Performance improved by about 10% on Graviton3E (64 cores) and more than 20% on Xeon Platinum 8375C (32 cores x 2 sockets).​

Graviton3E

Xeon Platinum 8375C

The calculations are distributed so that each thread handles about the same size of the range in the M and N directions, even when the input matrices are rectangle.​ Although not confirmed on all platforms, this relatively simple fix is expected to be generally effective on modern manycore CPUs.​

### DIFF
--- a/driver/level3/level3_thread.c
+++ b/driver/level3/level3_thread.c
@@ -826,6 +826,16 @@ int CNAME(blas_arg_t *args, BLASLONG *range_m, BLASLONG *range_n, IFLOAT *sa, IF
     if (nthreads_m * nthreads_n > args -> nthreads) {
       nthreads_n = blas_quickdivide(args -> nthreads, nthreads_m);
     }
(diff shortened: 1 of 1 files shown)


martin-frbg:
That's impressively elegant and effective.

rageshhajela16:
@martin-frbg  Thanks for the review. Are there any other comments for us to incorporate? CI failures seem to be not related to the code changes, might be environment issues, if you can please help to review and confirm. Thanks
cc: @yamazakimitsufumi

PR #4650, opened 2024-04-17
Fix a build issue for PRESCOTT on x86_64

zboszor:
Define everything the same way for PRESCOTT as for PENTIUM4.

This is an RFC pull request. I have not actually tested running on a Prescott machine, only that 0.3.27 builds for x86_64.

PR #4656, opened 2024-04-19
Add forgotten conditional uses of PREFETCH

zboszor:
This fixes a (cross-)compilation/linker error for PRESCOTT on Yocto.

Alternative to #4650


martin-frbg:
Thank you. Picking this one as the minimal necessary fix since I still suspect the problem lies with the Yocto-provided SDK.

PR #4664, opened 2024-04-28
Fixed the HUGETLB_ALLOCATION option

XiWeiGu:
Attempt to fix the HUGETLB_ALLOCATION option


martin-frbg:
thanks - I'm still testing various combinations but I (currently) believe the ALLOC_SHM and HUGETLB should not depend on each other (as claimed by the comment added in the questionable PR from 2015). Both options use shm calls but they seem to do so independently from each other, it appear to be just the preprocessor conditionals that create a dependency

XiWeiGu:
The modifications have been submitted in the PR #4662

PR #4680, opened 2024-05-07
Expose whether locking is enabled in get_config

theAeon:
Hello! I ran into an issue w/ single-threaded openblas and OpenMP which turned out to be the result of not having locking enabled. I then noticed that there's really no way to find out at runtime if it is enabled or not ([mgcv's blast.thread.test quite literally brute forces it as an alternative](https://github.com/cran/mgcv/blob/master/R/misc.r#L29C1-L48C22)). I figured it would make sense for this to be exposed by openblas_get_config instead.


martin-frbg:
Thanks, makes perfect sense to include this information.

PR #4690, opened 2024-05-12
use blasint instead of int to quiet warnings

mattip:


mattip:
When blasint is long, this emitted a warning

PR #4708, opened 2024-05-20
[WIP] forward GEMM workloads to GEMV when one argument is actually a vector

### LINKED ISSUE #528: GEMM slower than GEMV slower than AXPY equivalent on Intel i5 CPU
Consider the following Julia code (using Julia 0.4-dev):

```jl
let
       B = randn(1000, 1000)
       v = randn(1000)
       y = randn(1000); sB = [slice(B, :, j) for j = 1:size(B, 2)]

       @time for i = 1:1000;BLAS.gemm!('N', 'N', 1.0, B, v, 1.0, y);end;

       @time for i = 1:1000;BLAS.gemv!('N', 1.0, B, v, 1.0, y);end;

       @time for i = 1:1000;
           for j = 1:size(B,2)
             BLAS.axpy!(v[j], sB[j], y)
           end
       end
end
```

On @andreasnoack's machine, a Macbook Pro with i7-4870HQ CPU, GEMM is 4 times slower than GEMV:

```
elapsed time: 1.084909686 seconds (0 bytes allocated)
elapsed time: 0.2644927 seconds (0 bytes allocated)
elapsed time: 0.321705553 seconds (0 bytes allocated)
```

On my machine, a Macbook Pro with i5-4258U, I get similar behavior, but also that the AXPY equivalent is the fastest of the 3 computations:

```
elapsed time: 1.693223657 seconds (0 bytes allocated)
elapsed time: 0.818590556 seconds (0 bytes allocated)
elapsed time: 0.715702898 seconds (0 bytes allocated)
```

I find the relative performance behaviors surprising.

### LINKED ISSUE #4580: Openblas sgemm is slower for small size matrices in aarch64
I have built openblas in graviton3E with make USE_OPENMP=1 NUM_THREADS=256 TARGET=NEOVERSEV1.
mkl is built in icelake machine.

I have used openblas sgemm as 
`cblas_sgemm(CblasRowMajor, CblasNoTrans, CblasNoTrans, M, N, K, 1.0, A, K, B, N, 0.0, C, N);`

When performance timings are compared with intel mkl for the the smaller size matmuls,  aarch64 is slower . 

openblasvsmkl

These are the different shapes I have checked and their timings.

martin-frbg:
fixes #4580 and fixes #528

### DIFF
--- a/interface/gemm.c
+++ b/interface/gemm.c
@@ -47,22 +47,29 @@
 #define SMP_THRESHOLD_MIN 65536.0
 #ifdef XDOUBLE
 #define ERROR_NAME "QGEMM "
+#define GEMV BLASFUNC(qgemv)
 #elif defined(DOUBLE)
 #define ERROR_NAME "DGEMM "
+#define GEMV BLASFUNC(dgemv)
 #elif defined(BFLOAT16)
 #define ERROR_NAME "SBGEMM "
+#define GEMV BLASFUNC(sbgemv)
 #else
 #define ERROR_NAME "SGEMM "
+#define GEMV BLASFUNC(sgemv)
 #endif
 #else
 #define SMP_THRESHOLD_MIN 8192.0
 #ifndef GEMM3M
 #ifdef XDOUBLE
 #define ERROR_NAME "XGEMM "
+#define GEMV BLASFUNC(xgemv)
 #elif defined(DOUBLE)
 #define ERROR_NAME "ZGEMM "
+#define GEMV BLASFUNC(zgemv)
 #else
 #define ERROR_NAME "CGEMM "
+#define GEMV BLASFUNC(cgemv)
 #endif
 #else
 #ifdef XDOUBLE
@@ -485,9 +492,38 @@ void CNAME(enum CBLAS_ORDER order, enum CBLAS_TRANSPOSE TransA, enum CBLAS_TRANS
   }
 #endif
 #endif  // defined(__linux__) && defined(__x86_64__) && defined(BFLOAT16)
+ //         fprintf(stderr,"G E M M interface m n k %d %d %d\n",args.m,args.n,args.k);
 
   if ((args.m == 0) || (args.n == 0)) return;
 
+#if 1
+#ifndef GEMM3M
+  if (args.m == 1) {
+	  char *NT=(char*)malloc(2*sizeof(char));
+	  if (transb&1)strcpy(NT,"T");
+	  else NT="N"; 
+//	  fprintf(stderr,"G E M V\n");
+	  GEMV(NT, &args.n ,&args.k, args.alpha, args.b, &args.ldb, args.a, &args.m, args.beta, args.c, &args.m);
+//SUBROUTINE SGEMV(TRANS,M,N,ALPHA,A,LDA,X,INCX,BETA,Y,INCY)
+return;
+  } else {
+  if (args.n == 1) {
+#ifndef CBLAS
+	  char *NT=(char*)malloc(2*sizeof(char));
+	  strcpy(NT,"N"); 
+#else
+	  char *NT=(char*)malloc(2*sizeof(char));
+	  if (transb&1)strcpy(NT,"T");
+	  else strcpy(NT,"N"); 
+#endif
+//	  fprintf(stderr,"G E M V ! ! ! lda=%d ldb=%d ldc=%d\n",args.lda,args.ldb,args.ldc);
+	  GEMV(NT, &args.m ,&args.k, args.alpha, args.a, &args.lda, args.b, &args.n, args.beta, args.c, &args.n);
+//SUBROUTINE SGEMV(TRANS,M,N,ALPHA,A,LDA,X,INCX,BETA,Y,INCY)
+	return;
+    }
+  }
+#endif
+#endif
 #if 0
   fprintf(stderr, "m = %4d  n = %d  k = %d  lda = %4d  ldb = %4d  ldc = %4d\n",
 	 args.m, args.n, args.k, args.lda, args.ldb, args.ldc);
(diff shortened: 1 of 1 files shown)


martin-frbg:
obviously I don't really intend to kick out the recent Loongson patch here - this first draft was thrown together off-grid in an outdated fork

interface/gemm.c
```diff
@@ -485,9 +492,38 @@ void CNAME(enum CBLAS_ORDER order, enum CBLAS_TRANSPOSE TransA, enum CBLAS_TRANS
...
+ //         fprintf(stderr,"G E M M interface m n k %d %d %d\n",args.m,args.n,args.k);
 
   if ((args.m == 0) || (args.n == 0)) return;
 
+#if 1
+#ifndef GEMM3M
+  if (args.m == 1) {
+	  char *NT=(char*)malloc(2*sizeof(char));
+	  if (transb&1)strcpy(NT,"T");
+	  else NT="N";
```
lrbison, line 504: I think we can skip the malloc and go straight to the `NT="N"` or `NT="T"` assignment yes?
Actually, we don't need null termination so it should just be `char NT; NT='N'; ... &NT`.

interface/gemm.c
```diff
@@ -485,9 +492,38 @@ void CNAME(enum CBLAS_ORDER order, enum CBLAS_TRANSPOSE TransA, enum CBLAS_TRANS
...
+	  char *NT=(char*)malloc(2*sizeof(char));
+	  strcpy(NT,"N"); 
+#else
+	  char *NT=(char*)malloc(2*sizeof(char));
+	  if (transb&1)strcpy(NT,"T");
+	  else strcpy(NT,"N"); 
+#endif
+//	  fprintf(stderr,"G E M V ! ! ! lda=%d ldb=%d ldc=%d\n",args.lda,args.ldb,args.ldc);
+	  GEMV(NT, &args.m ,&args.k, args.alpha, args.a, &args.lda, args.b, &args.n, args.beta, args.c, &args.n);
+//SUBROUTINE SGEMV(TRANS,M,N,ALPHA,A,LDA,X,INCX,BETA,Y,INCY)
```
lrbison, line 521: I was thinking about this a little more.  For cases when n = 1, seems like this should be pretty straightforward, and parameters should just directly translate from GEMM to GEMV.

For cases when m = 1 I think there is a challenge.  in gemm (C:=AB) A is the vector, but GEMV (y:=Bx). so I think we need to re-arrange the equation by transposing (or, un-transposing) B.

lrbison (commented):
Thank you for posting this WIP PR.  I was able to confirm the failures CI sees:

```
       1       0.00000         -0.252747
 ******* SGEMM  FAILED ON CALL NUMBER:
   3403: SGEMM ('N','N',  1,  1,  0, 0.0, A,  2, B,  1, 0.0, C,  2).
```

I haven't had time to track it down, but I appreciate the starting point!

akote123:
@martin-frbg , 
Thank you for the PR. Just to be sure on sending data to GEMV: For example, when A is matrix 1xn and B is of nxk, then are we flattening A ( i.e to convert matrix to vector) to make it compatible with GEMV.

martin-frbg:
> @martin-frbg , Thank you for the PR. Just to be sure on sending data to GEMV: For example, when A is matrix 1xn and B is of nxk, then are we flattening A ( i.e to convert matrix to vector) to make it compatible with GEMV.

Yes in principle, but I am not convinced we actually have to transform the storage of A for that. (Note that the rough draft I posted here may not even compile. I need to update it and flesh it out when I have time)

interface/gemm.c
conradsnicta: @lrbison @martin-frbg As a general word of caution, be careful with introducing any extraneous transposes for gemm variants dealing with complex numbers (`cgemm` and `zgemm`).  

For complex vectors and matrices, the conjugate transpose is typically used, where the sign of the imaginary components is flipped.  As an example, treating a column vector as a row vector (and vice versa) for complex numbers may not be simply a matter of interpreting the dimensions differently.   
https://en.wikipedia.org/wiki/Conjugate_transpose

To reduce risk, I'd suggest to completely leave out `cgemm` and `zgemm` from the proposed optimisation path (ie. forwarding gemm calls to gemv).

martin-frbg: @conradsnicta right, likely the performance gains would be much smaller for the complex cases anyway do to the extra coding necessary. So far I haven't even found time to get back to this issue.

interface/gemm.c
```diff
@@ -485,9 +492,38 @@ void CNAME(enum CBLAS_ORDER order, enum CBLAS_TRANSPOSE TransA, enum CBLAS_TRANS
...
+#ifndef CBLAS
+	  char *NT=(char*)malloc(2*sizeof(char));
+	  strcpy(NT,"N"); 
+#else
+	  char *NT=(char*)malloc(2*sizeof(char));
+	  if (transb&1)strcpy(NT,"T");
+	  else strcpy(NT,"N"); 
+#endif
+//	  fprintf(stderr,"G E M V ! ! ! lda=%d ldb=%d ldc=%d\n",args.lda,args.ldb,args.ldc);
+	  GEMV(NT, &args.m ,&args.k, args.alpha, args.a, &args.lda, args.b, &args.n, args.beta, args.c, &args.n);
```
ChipKerchner, line 520: This causes an error when INTERFACE64=0.  The parameters passed by address need to be `blasint` and not `blaslong`.  Same for the other call to GEMV

ChipKerchner:
> Thank you for posting this WIP PR. I was able to confirm the failures CI sees:
> 
> ```
>  ******* FATAL ERROR - COMPUTED RESULT IS LESS THAN HALF ACCURATE *******
>            EXPECTED RESULT   COMPUTED RESULT
>        1       0.00000         -0.252747
>  ******* SGEMM  FAILED ON CALL NUMBER:
>    3403: SGEMM ('N','N',  1,  1,  0, 0.0, A,  2, B,  1, 0.0, C,  2).
> ```
> 
> I haven't had time to track it down, but I appreciate the starting point!

This is causes by a mismatch between sizeof(blasint) and sizeof(int)

martin-frbg:
Thanks... I have an equally unfinished newer version lying around somewhere but got caught up in other things.  Let me get the fixups for the SCAL fallout out of the way... but if anybody beats me to it on this here topic it's fine of course. (probably need to remove this from the 0.3.28 milestone anyway so that the release does not get delayed all summer)

ChipKerchner:
@martin-frbg This is an important PR since some project cases have a large portion of GEMM in which N=1.  In these cases we are spending significant time packing buffer(s) which is not necessarily needed if GEMV was called instead.

martin-frbg:
I am aware of that, but this has been an important issue for roughly 20 years (i.e. since inception of GotoBLAS), last discussed here sometime in 2015/16 IIRC. We're almost 2 weeks past the tentative release date for 0.3.28, it bundles an excessive number of changes already, and I still need to come up with assembly code fixes for the SCAL issue in a number of architectures (where assembly isn't my strongest skill anyway).

Mousius:
@martin-frbg I took a look into this in #4814

martin-frbg:
closing as superseded by #4814

PR #4711, opened 2024-05-21
Loongarch64: Small matrix opt

XiWeiGu:
Add small matrix multiplication optimization


martin-frbg:
I guess you could check the transa,transb settings in the permit function when you have "only" the NN kernel ready. (And don't let the somewhat experimental codspeed report discourage you - it only runs a select few tests on x86_64)

XiWeiGu:
Thank you very much for your suggestion. The current progress is that I am still adding assembly optimizations for the `TN`, `TT`, and `NT` cases.

PR #4712, opened 2024-05-22
POWER: Fix issues in zscal to address lapack failures

RajalakshmiSR:
This patch fixes following lapack failures with clang compiler on POWER.
zed.out: ZVX:   18 out of  5190 tests failed to pass the threshold
zgd.out: ZGV drivers:     25 out of   1092 tests failed to pass the threshold
zgd.out: ZGV drivers:      6 out of   1092 tests failed to pass the threshold

### DIFF
--- a/kernel/power/zscal.c
+++ b/kernel/power/zscal.c
@@ -38,6 +38,10 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #pragma GCC optimize "O1"
 
+#if defined(__clang__)
+#pragma clang fp contract(off)
+#endif
+
 #if defined(__VEC__) || defined(__ALTIVEC__)
 #if defined(POWER8) || defined(POWER9)
 #if defined(DOUBLE)


kernel/power/zscal.c
```diff
@@ -38,6 +38,10 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #pragma GCC optimize "O1"
 
+#if defined(__clang__)
```
ChipKerchner, line 41: Is your intention to disable FMA for clang for both SINGLE and DOUBLE versions of this algorithm?

martin-frbg:
At some point it will probably make more sense to replace the Reference-LAPACK testsuite, or at least relax its criteria. I only wish I knew how to define that point, and creating a replacement would be a huge task...

ChipKerchner:
How often are changes/fixes to LAPACK pulled in from?

```
https://github.com/Reference-LAPACK
```

martin-frbg:
If it is not a documentation-only fix (and not a huge change either, like having both 32 and 64bit integer versions of all functions in the library), usually within days of the PR getting merged (or sometimes as soon as a plausible PR gets posted).
Do you think I missed something lately ?

ChipKerchner:
Just wondering if I had to manually merge LAPACK changes.  Thanks for the clarification.

martin-frbg:
I realize we already have the same fix/workaround on x86_64, so merging this after all. Can still discuss adding another `ifdef` to exclude single-precision COMPLEX later...

PR #4725, opened 2024-05-30
Fix CMake warning

Neumann-A:
```
CMake Warning (dev) at cmake/prebuild.cmake:1346 (if):
  Policy CMP0054 is not set: Only interpret if() arguments as variables or
  keywords when unquoted.  Run "cmake --help-policy CMP0054" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  Quoted variables like "MSVC" will no longer be dereferenced when the policy
  is set to NEW.  Since the policy is not set the OLD behavior will be used.
Call Stack (most recent call first):
  cmake/system.cmake:166 (include)
  CMakeLists.txt:104 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

### DIFF
--- a/cmake/prebuild.cmake
+++ b/cmake/prebuild.cmake
@@ -1342,7 +1342,7 @@ else(NOT CMAKE_CROSSCOMPILING)
 
   if ("${CMAKE_C_COMPILER_ID}" STREQUAL "MSVC")
     #Use generic for MSVC now
-    message("MSVC")
+    message(STATUS "MSVC")
     set(GETARCH_FLAGS ${GETARCH_FLAGS} -DFORCE_GENERIC)
   else()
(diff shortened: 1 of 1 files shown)


martin-frbg:
Thank you - actually this looks like it might be a forgotten line from debugging, but I guess it can't hurt to keep it (in the proper syntax)

Neumann-A:
Modern CMake debugging is done with `--trace-expand` or directly attaching a debugger to CMake.
Quick question: You know that your `Windows_cl_flang` pipeline is broken? It errored but the pipeline is green.

martin-frbg:
yeah - not quite sure why it did not mark that job as failed (probably the usage message from ctest does not count as fatal, but why did it even proceed to that step after cmake errorred on the missing flang ?)

PR #4763, opened 2024-06-24
BENCH: sync codspeed-benchmarks with BLAS-benchmarks

ev-br:
Sync the benchmark suite we run on codspeed with http://www.openmathlib.org/BLAS-Benchmarks/#summarylist?sort=0&dir=asc:

- run s-/d-/c-/z- variants
- run more matrix sizes

### DIFF
--- a/benchmark/pybench/benchmarks/bench_blas.py
+++ b/benchmark/pybench/benchmarks/bench_blas.py
@@ -1,66 +1,76 @@
 import pytest
 import numpy as np
-from openblas_wrap import (
-    # level 1
-    dnrm2, ddot, daxpy,
-    # level 3
-    dgemm, dsyrk,
-    # lapack
-    dgesv,                   # linalg.solve
-    dgesdd, dgesdd_lwork,    # linalg.svd
-    dsyev, dsyev_lwork,      # linalg.eigh
-)
+import openblas_wrap as ow
+
+dtype_map = {
+    's': np.float32,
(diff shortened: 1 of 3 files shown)


ev-br:
Expanding the benchmark suite of course takes longer. Especially given that codspeed runs things under callgrind, it is not entirely unexpected to be slow. I'm not sure what are their free usage limits, 

Previously: the codspeed run was under 5 minutes (e.g. https://github.com/OpenMathLib/OpenBLAS/actions/runs/9632137639/job/26564891418)
This PR: an hour and countin (https://github.com/OpenMathLib/OpenBLAS/actions/runs/9644610499/job/26597037247?pr=4763)

I wonder if we should keep the a small set of system sizes and run s-/d-/c-/z- variants?

martin-frbg:
Yes I don't think it makes sense to expand the codspeed benchmarks to the same granularity and complexity as the standalone benchmark suite - apart from (potential direct monetary) cost, having a CI job run for hours can be a major nuisance when waiting to commit something. Small set of sizes and either of the two complex types in addition to the existing double precision real is probably good enough, and we are checking one architecture only anyway

ev-br:
Okay, down to 35 mins for the benchmarks themselves + some 15 mins build time with 
https://github.com/OpenMathLib/OpenBLAS/actions/runs/9646547071/job/26603220881?pr=4763

PR #4769, opened 2024-06-28
openblas: fix `BUFFERSIZE` value

drupol:
Hi,

I created this patch to facilitate its use in Nix.


martin-frbg:
(this came up while trying to debug #4766)

Issue #4166, opened 2023-07-27
Adding html docs for the project


rgommers:
I'd like to propose and work on adding html documentation for the project. This would make it easier for users - both direct C/Fortran API users and downstream users who experience OpenBLAS's runtime behavior through Python, Julia, R or other such languages - as well as packagers, contributors and other potentially interested folks to find what they are looking for. And it will help when trying to contribute to the docs (right now it's on wiki pages that cannot be edited by external contributors).

This would involve:
- incorporating most of the info now on the wiki,
- updating that wiki info to ensure it's all up-to-date and reads well (in separate PRs or commits, so it's easy to review what was changed)
- adding some new docs where appropriate,
- a CI job to auto-build the docs on PRs and build + deploy them on PR merges
- tweaks to other CI jobs to not run for docs-only PRs

This will require choosing a doc tool and deployment method. I'd propose [MkDocs](https://www.mkdocs.org/) with the [Material for MkDocs](https://squidfunk.github.io/mkdocs-material/) theme, as that is quite easy to use, well-documented/maintained, and has a nice look-and-feel. And deploying to GitHub Pages. @martin-frbg if you'd prefer another setup (e.g., [Docusaurus](https://docusaurus.io/) is good too), then of course I'd be happy to use whatever you would like to see here.

martin-frbg:
MkDocs looks nice enough (though I'd prefer to not tie this to one particular tool exclusively, which could create a barrier for future contributors. But if all it does is pretty-print markdown, there are probably ways around that).

Another option might be to set up a runner that auto-converts the wiki (which IIRC is just another repo) to html and possibly pdf. This would offer the advantage of having only a single source to update, though I guess the style options may be a bit limited, possibly leading to a less visually pleasing result.  
(Sad that we had to lock down the wiki due to repeated vandalism - it felt a lot more friendly when users could just add their build recipes and special use cases directly)

rgommers:
> But if all it does is pretty-print markdown,

Pretty much. The docs are written in pure Markdown, so it's about as easy as it gets for working on them. The structure, navbar, etc. are all derived from the titles and subfolders. Here's an example, it's basically one ~75 LoC `mkdocs.yml` config file and then plain `.md` files with docs: https://github.com/pypackaging-native/pypackaging-native, turns into [this website](https://pypackaging-native.github.io/).

> This would offer the advantage of having only a single source to update

[...]

PR #4774, opened 2024-06-30
Improve documention content, formatting, and html theme

rgommers:
This is a thorough update of the existing docs pages. I went through everything (except "Home", which has almost no content, and "Install OpenBLAS", which may need a bit of discussion), updated all the existing content, streamlined the language, added some context and more up-to-date info as I saw fit.

It follows up on gh-4562, and together with that PR addresses gh-4166.

Main changes:
- Improve the html theme (add a dark mode, nicer admonitions, footnotes, an auto-hiding top bar)
- Fix all links, code markup, and grammar.
- "About" page: add all funding sources I know about, and missing major users like NumPy and R
- "Developer manual": improve testing and benchmarking descriptions
- "User manual": start with context about what is covered and link to Netlib/MKL docs; add more info on linking

### DIFF
--- a/docs/about.md
+++ b/docs/about.md
@@ -8,6 +8,7 @@ This work was or is partially supported by the following grants, contracts and i
 
 * Research and Development of Compiler System and Toolchain for Domestic CPU, National S&T Major Projects: Core Electronic Devices, High-end General Chips and Fundamental Software (No.2009ZX01036-001-002)
 * National High-tech R&D Program of China (Grant No.2012AA010903)
+* [PerfXLab](http://www.perfxlab.com/)
 * Chan Zuckerberg Initiative's Essential Open Source Software for Science program:
     * Cycle 1 grant: [Strengthening NumPy's foundations - growing beyond code](https://figshare.com/articles/journal_contribution/Proposal_NumPy_OpenBLAS_for_Chan_Zuckerberg_Initiative_EOSS_2019_round_1/10302167) (2019-2020)
     * Cycle 3 grant: [Improving usability and sustainability for NumPy and OpenBLAS](https://chanzuckerberg.com/eoss/proposals/improving-usability-and-sustainability-for-numpy-and-openblas/) (2020-2021)
@@ -23,7 +24,7 @@ A list of backers is available [in BACKERS.md](https://github.com/OpenMathLib/Op
 We welcome hardware donations, including the latest CPUs and motherboards.
 
 
-## Users of OpenBLAS
+## Open source users of OpenBLAS
 
 Prominent open source users of OpenBLAS include:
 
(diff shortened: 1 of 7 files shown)


docs/about.md
```diff
@@ -2,25 +2,44 @@
...
+You can read OpenBLAS's statement of receipts and disbursement and cash balance in
+[this Google doc](https://docs.google.com/spreadsheet/ccc?key=0AghkTjXe2lDndE1UZml0dGpaUzJmZGhvenBZd1F2R1E&usp=sharing) (covers 2013-2016).
+A list of backers is available [in BACKERS.md](https://github.com/OpenMathLib/OpenBLAS/blob/develop/BACKERS.md) in the main repo.
+
+### Donations
+
+We welcome hardware donations, including the latest CPUs and motherboards.
+
 
 ## Users of OpenBLAS
```
martin-frbg: I believe the (short and mixed) original list here is the result of a user survey. Not sure if we want to kick out less-well known projects, and not sure if we want to group by prominence (which might vary with the profession of the user reading this). 
In that spirit, "Known open source users..." might be a more appropriate heading (and as far as I know R may still not be using OpenBLAS by default, only through a popular add-in) ?

docs/user_manual.md
```diff
@@ -1,70 +1,174 @@
...
+tips. Compiling OpenBLAS is optional, since you may be able to install with a
+package manager.
+
+!!! Note BLAS API reference documentation
+
+    The OpenBLAS documentation does not contain API reference documentation for
+    BLAS or LAPACK, since these are standardized APIs, the documentation for
+    which can be found in other places. If you want to understand every BLAS
+    function and definition, we recommend reading the
+    [Intel MKL reference manual](https://software.intel.com/en-us/intel-mkl/documentation)
```
martin-frbg: Not sure if we want MKL documentation here (which has a number of extensions that are neither in the Reference nor in OpenBLAS), AFAIK it has no advantages over the netlib one except perhaps nicer presentation ?

rgommers: Agreed. I kept it because it was already there, but there is no reason to link to it. Netlib docs are the authoritative ones, so I'll change to linking only those.

docs/about.md
rgommers: The two links I removed are for projects that are dead, which seemed like a good reason to remove them. And Ceemple turned into OpenCV Ceemple, and then the separate name went away so whatever that was in the past is now simply part of OpenCV (which explains why I added OpenCV).

> In that spirit, "Known open source users..." might be a more appropriate heading (and as far as I know R may still not be using OpenBLAS by default, only through a popular add-in) ?

Sure will change. And yes, not sure - may also depend on the R install method, there's more than one way.

martin-frbg: ok, if they're dead that does seem like a perfect reason for removal :)

docs/user_manual.md
martin-frbg: sorry,had not realized that this was straight from the wiki - guess xianyi must have been in a hurry when he put together that "user manual". netlib docs usually have the added advantage of providing a nice call graph for each function

rgommers:
Thanks for merging @martin-frbg. I broke the docs deployment CI job here. I'll fix that up, and will make sure that it also runs on PRs so for future docs PRs any issues are more easily visible.

PR #4777, opened 2024-07-02
ignore the gesdd failure on codspeed

ev-br:
<s>An experiment to fish out the source of https://github.com/OpenMathLib/OpenBLAS/issues/4776 : cannot repro the error locally, let's see what the GH runners do</s>

Temp ignore the `sgesdd` failure on codspeed, so that benchmark results upload.

### DIFF
--- a/.github/workflows/codspeed-bench.yml
+++ b/.github/workflows/codspeed-bench.yml
@@ -139,6 +139,13 @@ jobs:
           cd build/openblas_wrap
           python -c'import _flapack; print(dir(_flapack))'
 
+      - name: Run benchmarks under pytest-benchmark
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/codspeed-bench.yml
```diff
@@ -139,6 +139,12 @@ jobs:
           cd build/openblas_wrap
           python -c'import _flapack; print(dir(_flapack))'
 
+          # run the sdgesdd benchmark w/o codspeed
+          pip install pytest-benchmark
+          cd benchmark/pybench
```
ev-br, line 144:
```suggestion
          cd ../../benchmark/pybench
```

PR #4778, opened 2024-07-02
Add support for RISCV64_GENERIC in cmake

JAicewizard:
Support for RISCV64_GENERIC was missing in cmake, so I added support for it.
I am not sure where these values come from, but I copied them from the config.h when using Make.

### DIFF
--- a/cmake/prebuild.cmake
+++ b/cmake/prebuild.cmake
@@ -1309,6 +1309,15 @@ endif ()
       "#define DTB_DEFAULT_ENTRIES 128\n"
       "#define DTB_SIZE 4096\n"
       "#define L2_ASSOCIATIVE 8\n")
+  elseif ("${TCORE}" STREQUAL "RISCV64_GENERIC")
+    file(APPEND ${TARGET_CONF_TEMP}
+        "#define L1_DATA_SIZE 32768\n"
+      "#define L1_DATA_LINESIZE 32\n"
+      "#define L2_SIZE 1048576\n"
+      "#define L2_LINESIZE 32 \n"
+      "#define DTB_DEFAULT_ENTRIES 128\n"
+      "#define DTB_SIZE 4096\n"
+      "#define L2_ASSOCIATIVE 4\n")
   endif()
   set(SBGEMM_UNROLL_M 8)
   set(SBGEMM_UNROLL_N 4)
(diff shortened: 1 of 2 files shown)


martin-frbg:
Thanks - the values used in the Makefile build are an educated(?) guess for what should work on any RISCV hardware when the build system cannot simply compile and run getarch on the target. (Note to self- eventually we'll want the other RISCV targets in that ridiculously long table too. This PR is an outflow of https://github.com/OpenMathLib/OpenBLAS/issues/3640#issuecomment-2203798846 - getting a cross-compile to WEBM working with CMAKE as well).

JAicewizard:
Thanks! This must have happened in the period where gmail send all github mail to spam, so thats why im so late :)

NickelWenzel:
@JAicewizard Could you maybe provide the cmake command with which you manage to compile succesfully? I can't get it to work and am not sure if `-DCORE=RISCV64_GENERIC` is enough or if it's actually `-DTARGET=RISCV64 -DCORE=GENERIC` or something else.

I'm also trying to compile OpenBLAS for emscripten using vcpkg, i already had a custom port with extra patches which was using `-DCORE=GENERIC -DTARGET=X86` based on v0.3.25 but i'd rather omit the patches and go the already established way with `RISCV64_GENERIC`.

Thank you :)

JAicewizard:
Actually I also have a fork of vcpkg where I tried to get emscriptem to work (sadly openmp is still a blocker). That was the whole reason I wanted the riscv64 core to work with cmake in the first place. You can find it here: https://github.com/JAicewizard/vcpkg/blob/fix_openblas_macos/ports/openblas/portfile.cmake. I see that it is still using my fork of openblas, but I think thats just an outdated version of upstream atm.

martin-frbg:
Not sure where openmp comes in when talking about compiling openblas with/for emscripten ? In my earlier proof-of-concept, I had built a single-threaded version to avoid all the issues with locking and thread management (which IIRC were unsupported back then). TARGET=RISCV64_GENERIC is (ab)used for the purpose of having a clean C source without any os-specific ifdefs or assembly code hidden in header files. (As would be the case with the "generic x86" target). Eventually this should morph into a separate TARGET, perhaps even supporting optimization through whatever asm or intrinsics a future emscripten may support

JAicewizard:
It is not, it is just a blocker for my project

martin-frbg:
I see, thanks for the clarification. I'm checking emscripten compilation again now, seems one new hurdle is that the RISCV64_GENERIC target has since added `-march=` and `-mabi=` options that emcc obviously does not like. For now these are only an obstacle in the gmake build, but they should end up in CMake too once we claim CMake support for RISCV64 targets

PR #4787, opened 2024-07-05
Update cross compile info

vlad0x00:
When cross compiling OpenBLAS using `make` (with `TARGET`, `HOSTCC`, `CC`, and `FC` specified), I found that if I don't specify `CROSS=1` flag, the build process attempts to run the tests which may crash with `SIGILL: Illegal instruction` if the build machine does not have all the instructions of the target machine. The README doesn't state this clearly enough.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,17 +67,17 @@ build options you plan to set.
 ### Cross compile
 
 Set `CC` and `FC` to point to the cross toolchains, and set `HOSTCC` to your host C compiler.
-The target must be specified explicitly when cross compiling.
+The target must be specified explicitly when cross compiling. The `CROSS=1` flag should be specified.
 
 Examples:
 
 * On an x86 box, compile this library for a loongson3a CPU:
   ```sh
-  make BINARY=64 CC=mips64el-unknown-linux-gnu-gcc FC=mips64el-unknown-linux-gnu-gfortran HOSTCC=gcc TARGET=LOONGSON3A
+  make BINARY=64 CC=mips64el-unknown-linux-gnu-gcc FC=mips64el-unknown-linux-gnu-gfortran HOSTCC=gcc TARGET=LOONGSON3A CROSS=1
   ```
   or same with the newer mips-crosscompiler put out by Loongson that defaults to the 32bit ABI:
   ```sh
-  make HOSTCC=gcc CC='/opt/mips-loongson-gcc7.3-linux-gnu/2019.06-29/bin/mips-linux-gnu-gcc -mabi=64' FC='/opt/mips-loongson-gcc7.3-linux-gnu/2019.06-29/bin/mips-linux-gnu-gfortran -mabi=64' TARGET=LOONGSON3A
+  make HOSTCC=gcc CC='/opt/mips-loongson-gcc7.3-linux-gnu/2019.06-29/bin/mips-linux-gnu-gcc -mabi=64' FC='/opt/mips-loongson-gcc7.3-linux-gnu/2019.06-29/bin/mips-linux-gnu-gfortran -mabi=64' TARGET=LOONGSON3A CROSS=1
   ```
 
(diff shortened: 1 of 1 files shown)


martin-frbg:
I wonder what kind of machine you were cross-compiling on ? Normally/ideally the build system should detect and set CROSS=1 automatically.

vlad0x00:
> I wonder what kind of machine you were cross-compiling on ? Normally/ideally the build system should detect and set CROSS=1 automatically.

I was compiling on a Zen 3 machine (3rd generation AMD EPYC Milan) for a Sapphire Rapids (4th generation Intel Xeon) machine. Regardless, though, I don't see any mechanism in the `Makefile` (https://github.com/OpenMathLib/OpenBLAS/blob/develop/Makefile) that would set the `CROSS` flag automatically. There is only one line (401) where it's being used.

EDIT: Ah, I guess there might be some included `Makefile`s that also reference the `CROSS` flag? I didn't dig for that. But regardless, on my end, if I don't explicitly specify `CROSS=1`, the build crashes with illegal instruction error during tests.

vlad0x00:
> > I wonder what kind of machine you were cross-compiling on ? Normally/ideally the build system should detect and set CROSS=1 automatically.
> 
> I was compiling on a Zen 3 machine (3rd generation AMD EPYC Milan) for a Sapphire Rapids (4th generation Intel Xeon) machine. Regardless, though, I don't see any mechanism in the `Makefile` (https://github.com/OpenMathLib/OpenBLAS/blob/develop/Makefile) that would set the `CROSS` flag automatically. There is only one line (401) where it's being used.
> 
> EDIT: Ah, I guess there might be some included `Makefile`s that also reference the `CROSS` flag? I didn't dig for that. But regardless, on my end, if I don't explicitly specify `CROSS=1`, the build crashes with illegal instruction error.

When I was compiling, my command was `make USE_OPENMP=1 HOSTCC=gcc CC=gcc FC=gfortran TARGET=SAPPHIRERAPIDS` if that helps.

martin-frbg:
The problem is specific to "cross-compiling" for a more powerful cpu (instruction set) of the same architecture, which arguably is not cross-compiling in the usual sense (where the architectures are completely different so there is not even a common subset of instructions between the build host and target system). In a "regular" cross-(architecture) compilation, the CROSS option is added automatically by the c_check script, so I do not think all of the proposed changes to the documentation are necessary. 
That said, I have not come up with a convincing solution for your case (except perhaps forcing a second run of `getarch` to detect the host cpu when the given TARGET happens to match the host architecture and then comparing capabilities via a fixed list).  
So perhaps a more appropriate addition to the documentation would be something like `When compiling for a more modern cpu TARGET of the same architecture, e.g. TARGET=SKYLAKEX on a HASWELL host, option "CROSS=1" can be used to suppress the automatic invocation of the tests at the end of the build` ?

vlad0x00:
> The problem is specific to "cross-compiling" for a more powerful cpu (instruction set) of the same architecture, which arguably is not cross-compiling in the usual sense (where the architectures are completely different so there is not even a common subset of instructions between the build host and target system). In a "regular" cross-(architecture) compilation, the CROSS option is added automatically by the c_check script, so I do not think all of the proposed changes to the documentation are necessary. That said, I have not come up with a convincing solution for your case (except perhaps forcing a second run of `getarch` to detect the host cpu when the given TARGET happens to match the host architecture and then comparing capabilities via a fixed list). So perhaps a more appropriate addition to the documentation would be something like `When compiling for a more modern cpu TARGET of the same architecture, e.g. TARGET=SKYLAKEX on a HASWELL host, option "CROSS=1" can be used to suppress the automatic invocation of the tests at the end of the build` ?

Ah, interesting, I didn't know that's not considered cross compiling. I can modify the PR to reflect what you said

vlad0x00:
Updated the PR.

martin-frbg:
Thank you

Issue #4728, opened 2024-05-31
dscal: `0 * nan` ≠ `nan * 0` and `0 * inf` ≠ `inf * 0`


davidcl:
Here another issue related to #4726 and #4413, impacting `dscal` this time.

With OpenBLAS 0.3.27:
```
$ LD_LIBRARY_PATH=lib/thirdparty:lib/thirdparty/redist ./openblas_dscal
result of  nan *  inf =  nan
result of  inf *  nan =  nan
result of    0 *  nan =    0
result of  nan *    0 =  nan
result of    0 *  inf =    0
result of  inf *    0 = -nan
```

With refBLAS 3.12.0 :
```
$ LD_LIBRARY_PATH=$HOME/work/dependencies/lapack/build/lib ./openblas_dscal
result of  nan *  inf =  nan
result of  inf *  nan =  nan
result of    0 *  nan =  nan
result of  nan *    0 =  nan
result of    0 *  inf = -nan
result of  inf *    0 = -nan
```

<details><summary>test code</summary>

```c

#include <stdio.h>
#include <math.h>

extern void dscal_(int*, double*, double*, int*);

int main(int argc, char ** argv)
{
    int ONE = 1;

    double scalar = NAN;
    double inout = INFINITY;

    printf("result of %4.5g * %4.5g = ", scalar, inout);
    dscal_(&ONE, &scalar, &inout, &ONE);
    printf("%4.5g\n", inout);

    scalar = INFINITY;
    inout = NAN;

    printf("result of %4.5g * %4.5g = ", scalar, inout);
    dscal_(&ONE, &scalar, &inout, &ONE);
    printf("%4.5g\n", inout);

    scalar = 0;
    inout = NAN;

    printf("result of %4.5g * %4.5g = ", scalar, inout);
    dscal_(&ONE, &scalar, &inout, &ONE);
    printf("%4.5g\n", inout);

    scalar = NAN;
    inout = 0;

    printf("result of %4.5g * %4.5g = ", scalar, inout);
    dscal_(&ONE, &scalar, &inout, &ONE);
    printf("%4.5g\n", inout);

    scalar = 0;
    inout = INFINITY;

    printf("result of %4.5g * %4.5g = ", scalar, inout);
    dscal_(&ONE, &scalar, &inout, &ONE);
    printf("%4.5g\n", inout);

    scalar = INFINITY;
    inout = 0;

    printf("result of %4.5g * %4.5g = ", scalar, inout);
[...]

Issue #4776, opened 2024-07-02
`bench_blas.py::test_gesdd` in Codspeed CI job


rgommers:
This failure just showed up in CI, from [this log](https://github.com/OpenMathLib/OpenBLAS/actions/runs/9760654144/job/26939985244):
```
 benchmarks/bench_blas.py ............................................... [ 75%]
  .........F.....                                                          [100%]
  ================================ 62 benchmarked ================================
  
  =================================== FAILURES ===================================
  ______________________________ test_gesdd[mn1-s] _______________________________
  
  benchmark = <pytest_codspeed.plugin.BenchmarkFixture object at 0x152f7980>
  mn = (1000, 222), variant = 's'
  
      @pytest.mark.parametrize('variant', ['s', 'd'])
      @pytest.mark.parametrize('mn', gesdd_sizes)
      def test_gesdd(benchmark, mn, variant):
          m, n = mn
          rndm = np.random.RandomState(1234)
          dtyp = dtype_map[variant]
      
          a = np.array(rndm.uniform(size=(m, n)), dtype=dtyp, order='F')
      
          gesdd_lwork = ow.get_func('gesdd_lwork', variant)
      
          lwork, info = gesdd_lwork(m, n)
          lwork = int(lwork)
          assert info == 0
      
          gesdd = ow.get_func('gesdd', variant)
          u, s, vt, info = benchmark(run_gesdd, a, lwork, gesdd)
      
  >       assert info == 0
  E       assert 12 == 0
  
  benchmarks/bench_blas.py:237: AssertionError
  =========================== short test summary info ============================
  FAILED benchmarks/bench_blas.py::test_gesdd[mn1-s] - assert 12 == 0
  ================== 1 failed, 61 passed in 2080.74s (0:34:40) ===================
   ** On entry to SLASCL parameter number  4 had an illegal value
   ** On entry to SLASCL parameter number  4 had an illegal value
   ** On entry to SLASCL parameter number  4 had an illegal value
   ** On entry to SLASCL parameter number  4 had an illegal value
```

This benchmark was introduced in gh-4678 one and a half months ago. I'm not sure it failed before like this. @ev-br you may want to look into this?

ev-br:
Hm... Actually, it did fail like this before: https://github.com/OpenMathLib/OpenBLAS/actions/workflows/codspeed-bench.yml
[...]

PR #4794, opened 2024-07-10
Try to fixed numpy ci test failures

XiWeiGu:
Besides the S/D/C/ZSCAL interfaces in BLAS, there are many other interfaces that also call SCAL_K. When fixing issue #4728 , the behavior of those interfaces that call SCAL_K was also changed, which caused some NUMPY CI test failures. I made revisions on an AMD Ryzen 2600 to ensure that the behavior of those interfaces calling SCAL_K remains consistent with before. The local NUMPY (v1.26.0)CI test results are as follows:

> FAILED numpy/linalg/tests/test_linalg.py::TestEigvals::test_generalized_sq_cases - AssertionError: In test case: <LinalgCase: 8x8_tile213>
> ==================================== 1 failed, 37721 passed, 1990 skipped, 32 xfailed, 2 xpassed in 414.31s (0:06:54) ====================================

The number of failed test cases will be reduced to one (there is still one that needs further investigation).

### DIFF
--- a/interface/scal.c
+++ b/interface/scal.c
@@ -85,7 +85,7 @@ void CNAME(blasint n, FLOAT alpha, FLOAT *x, blasint incx){
   if (nthreads == 1) {
 #endif
 
-  SCAL_K(n, 0, 0, alpha, x, incx, NULL, 0, NULL, 0);
+  SCAL_K(n, 0, 0, alpha, x, incx, NULL, 0, NULL, 1);
 
 #ifdef SMP
   } else {
@@ -102,7 +102,7 @@ void CNAME(blasint n, FLOAT alpha, FLOAT *x, blasint incx){
 #else
 		       &alpha,
 #endif
-		       x, incx, NULL, 0, NULL, 0,  (int (*)(void))SCAL_K, nthreads);
+		       x, incx, NULL, 0, NULL, 1,  (int (*)(void))SCAL_K, nthreads);
 
   }
 #endif
--- a/kernel/loongarch64/scal.S
+++ b/kernel/loongarch64/scal.S
@@ -53,9 +53,86 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
    PROLOGUE
 
    li.d  TEMP, SIZE
+   ld.d XX, $sp, 0 // Load dummy2
+   slli.d XX, XX, BASE_SHIFT
    MTC  a1, $r0
    slli.d INCX, INCX, BASE_SHIFT
    bge $r0,    N, .L999
+   CMPEQ   $fcc0, ALPHA, a1
+   bceqz   $fcc0, .L50
+   beq     XX,    TEMP,  .L50  // if dummp2 == 1, do not directly copy 0
+   srai.d I, N, 3
+   bne INCX, TEMP, .L20
+   bge $r0,    I, .L15
+   .align 3
+
+.L12:
+   ST a1,  X,   0 * SIZE
+   ST a1,  X,   1 * SIZE
+   ST a1,  X,   2 * SIZE
+   ST a1,  X,   3 * SIZE
+   ST a1,  X,   4 * SIZE
+   ST a1,  X,   5 * SIZE
+   ST a1,  X,   6 * SIZE
+   ST a1,  X,   7 * SIZE
+   addi.w  I, I, -1
+   addi.d X, X, 8 * SIZE
+   blt $r0,    I, .L12
+   .align 3
+
+.L15:
+   andi    I,  N, 7
+   bge $r0,    I, .L999
+   .align  3
+.L16:
+   ST a1,  X,   0 * SIZE
+   addi.d  I, I, -1
+   addi.d X, X, SIZE
+   blt $r0,    I, .L16
+   move $r4, $r17
+   fmov.d $f0, $f22
+   jirl    $r0, $r1, 0x0
+   .align 3
+
+.L20:
+   srai.d  I, N, 3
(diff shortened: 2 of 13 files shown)


martin-frbg:
Thanks - I've come to the same conclusion (except that my version uses a changed dummy2 in the calls that do not originate from interface - which is a bit less practical than your suggestion but comes from me starting with the last failure in the numpy log). Guess it would make more sense to use dummy2=1 as the flag rather than 12, or is there a significance that I do not realize ?

XiWeiGu:
Thanks for your reply. Using dummy=12 has no special meaning (I was concerned that dummy2=1 might already be used as a flag). I will revise it to dummy=1 and test it.
Optimization code for SCAL_K on other platforms also needs to be revised, which is quite a workload.

XiWeiGu:
I also have an idea of identifying certain specific interfaces that shouldn't call SCAL_K and providing separate optimizations for them. The current approach is making the code less maintainable.

martin-frbg:
Yes, getting dummy2 into all the assembly code is not going to be fun, but the only alternative I can imagine is creating a second set of SCAL kernels so that one keeps the previous behaviour - this would make the code even less maintainable. (On the other hand, it might be better for performance if we can avoid a lot of new conditionals on the strictly internal code paths that way)

XiWeiGu:
Currently, the call to the revised SCAL_K in interface/gemv and driver/level3/syrk_k.c has caused the NUMPY CI test to fail. The test code for the cblas_dgemv interface is as follows:
```
#include <stdio.h>
#include <math.h>
#include <cblas.h>

int main() {
    int m = 3;
    int n = 2;

    double A[6] = {1.0, 2.0, 3.0, 4.0, 5.0, 6.0};
    double x[2] = {1.0, 1.0};
    double y[3] = {NAN, INFINITY, 0.0};

    cblas_dgemv(CblasRowMajor, CblasNoTrans, m, n, alpha, A, n, x, 1, beta, y, 1);

    printf("Result vector y:\n");
    for(int i = 0; i < m; i++) {
        printf("%f\n", y[i]);
    }

    return 0;
}
```
With `ATLAS 3.10.3`:

> Result vector y:
> 3.000000
> 7.000000
> 11.000000 

With `OpenBLAS` latest code：

> Result vector y:
> nan
> nan
> 11.000000

The syrk interface is likely a general issue as well. For these two interfaces, when beta=0.0, the output needs to be all zeros.

XiWeiGu:
One solution is to check if beta is 0.0 before calling SCAL_K in `gemv` and `syrk` and implement a bypass if it is. (Should we only need to check for `gemv` and `syrk`? Actually, there are other interfaces that call SCAL_K. Should we investigate if their expected results when beta=0.0 are the same?)

martin-frbg:
the one remaining error with your fix is due to a small oversight in your revision of dscal.c - in the loop on "n1" the code should call  dscal_kernel_8_zero when da=0&&dummy2=0

XiWeiGu:
> the one remaining error with your fix is due to a small oversight in your revision of dscal.c - in the loop on "n1" the code should call dscal_kernel_8_zero when da=0&&dummy2=0

Thank you, I've updated the code.

XiWeiGu:
It seems that the performance of gesadd has declined; further adjustments are needed

martin-frbg:
gesdd has scal on its call graph https://netlib.org/lapack/explore-html-3.6.1/d4/dca/group__real_g_esing_gaf60b27e77bfbeffe1ec63e0f360c4564_gaf60b27e77bfbeffe1ec63e0f360c4564_cgraph_org.svg but there is also some codspeed-specific problem with this particular benchmark (#4776). Maybe your changes happened to fix the premature exit (at least until the assembly kernels are updated) and the "regression" is actually the benchmark running to completion instead of erroring out early.

martin-frbg:
I have fixes for x86_64 lined up (still need to test on Windows to check I read the correct argument there). will try to do arm64, ppc and riscv today

XiWeiGu:
> I have fixes for x86_64 lined up (still need to test on Windows to check I read the correct argument there). will try to do arm64, ppc and riscv today

Okay, I can fix mips/mips64. I also plan to add unit tests for s/dgemv to ensure it can follow the correct branch of the SCAL_K interface.

mattip:
Trying to use latest HEAD over at [openblas-libs](https://github.com/MacPython/openblas-libs/pull/168), I am seeing a failure in i686
```
 TEST 77/107 sscal:0_inf_inc_2 [FAIL]
   ERR: test_zscal.c:77  should be true
 TEST 78/107 sscal:0_inf [FAIL]
   ERR: test_zscal.c:65  should be true
 TEST 79/107 sscal:nan_0_inc_2 [OK]
 TEST 80/107 sscal:nan_0 [OK]
 TEST 81/107 sscal:0_nan_inc_2 [FAIL]
   ERR: test_zscal.c:31  should be true
 TEST 82/107 sscal:0_nan [FAIL]
   ERR: test_zscal.c:19  should be true
```

This is using 
```
CFLAGS=' -fvisibility=protected -Wno-uninitialized'
make BUFFERSIZE=20 DYNAMIC_ARCH=1 USE_OPENMP=0 NUM_THREADS=64 BINARY=32 \
   OBJCONV=/io/objconv/objconv SYMBOLPREFIX=scipy_ LIBNAMEPREFIX=scipy_ \
   FIXED_LIBNAME=1 TARGET=PRESCOTT
```

inside a quay.io/pypa/manylinux2014_i686 docker image on a linux x86_64 host.

martin-frbg:
hopefully fixed by #4817 - I seem to have gotten the stack offset to the flag value wrong in the single-precision case

PR #4798, opened 2024-07-14
[WIP] re-evaluate building for AVX512 targets with flang-18 on Windows

martin-frbg:


martin-frbg:
Pointless to test in CI when the relevant jobs now get consistently scheduled on Zen3 EPYC...

PR #4800, opened 2024-07-15
Add missing parentheses

vlad0x00:
The two lines had missing / unmatched parentheses.


vlad0x00:
I'm not sure why jenkins is not passing for CI. It simply says "This commit cannot be built" and I get "This page may not exist, or you may not have permission to see it." when I open `Details`.

martin-frbg:
Right, thank you very much. OSUOSL Jenkins is a bit peculiar in that one can only see the build log when logged in on their site - but the error is obviously unrelated, for some reason it could not start the container for this job.

PR #4801, opened 2024-07-16
Add autodetection for riscv64

markdryan:
Implement DYNAMIC_ARCH support for riscv64.  Three cpu types are supported, riscv64_generic, riscv64_zvl256b, riscv64_zvl128b. The two non-generic kernels require CPU support for RVV 1.0 to function correctly.  Detecting that a riscv64 device supports RVV 1.0 is a little complicated as there are some boards on the market that advertise support for V via hwcap but only support RVV 0.7.1, which is not binary compatible with RVV 1.0.  The approach taken is to first try hwprobe.  If hwprobe is not available, we fall back to hwcap + an additional check to distinguish between RVV 1.0 and RVV 0.7.1.

Tested on a VM with VLEN=256, a CanMV K230 with VLEN=128 (with only the big core enabled), a Lichee Pi with RVV 0.7.1 and a VF2 with no vector.

A compiler with RVV 1.0 support must be used to build OpenBLAS for riscv64 when DYNAMIC_ARCH=1.


martin-frbg:
Thanks. (Not sure if the detect code has to be in its own file, but OTOH this could probably be leveraged for the cpu detection at build time). I guess there is very little chance to find a compiler that would allow bundling rvv0.71 and rvv1.0 kernels...

markdryan:
> Thanks. (Not sure if the detect code has to be in its own file, but OTOH this could probably be leveraged for the cpu detection at build time).

The issue here is that the code in detect_riscv64.c needs to be built with 'v' enabled, otherwise it won't compile, and the code in dynamic_riscv64.c cannot be built with 'v' enabled as it has to run without crashing on machines which don't support vector.  Originally, I tried to place everything in dynamic_riscv64.c using ```__attribute__((target("arch+=v")))``` to decorate the functions that needed vector but I ran into a [gcc bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=113739) that crashed the compiler, so I ended up splitting the code into two files.

>  I guess there is very little chance to find a compiler that would allow bundling rvv0.71 and rvv1.0 kernels...

The goal here would be to enable the C910 kernels via DYNAMIC_ARCH=1 as well?  I must admit I didn't try this.  It's not clear to me what the runtime detection code for the C910 kernels should look like.

martin-frbg:
> The issue here is that the code in detect_riscv64.c needs to be built with 'v' enabled, otherwise it won't compile, and the code in dynamic_riscv64.c cannot be built with 'v' enabled as it has to run without crashing on machines which don't support vector. Originally, I tried to place everything in dynamic_riscv64.c using `__attribute__((target("arch+=v")))` to decorate the functions that needed vector but I ran into a [gcc bug](https://gcc.gnu.org/bugzilla

I see, thanks. (The additional file doesn't actually matter, apart from the other dynamic_arch codes not having one...)

> The goal here would be to enable the C910 kernels via DYNAMIC_ARCH=1 as well? I must admit I didn't try this. It's not clear to me what the runtime detection code for the C910 kernels should look like.

One can dream... I wish this mess of incompatible rvv specifications would not exist. Probably not worth the hassle to try building a library that has both in it. (Not sure if the C9xx have a distinctive hwprobe signature, the build-time detection simply tries to grab a model name from /proc/cpuinfo. I have a MilkV Pioneer here that is "a bit" faster than the Lichee Pi, can check later what https://github.com/clementleger/hwprobe_dump reports)

PR #4802, opened 2024-07-16
Fix axpby_rvv kernels for cases where inc_y = 0

markdryan:
The following openblas_utest tests fail when the RISCV64_ZVL128B is enabled.

TEST 89/103 axpby:zaxpby_inc_0 [FAIL]
TEST 92/103 axpby:caxpby_inc_0 [FAIL]
TEST 95/103 axpby:daxpby_inc_0 [FAIL]
TEST 98/103 axpby:saxpby_inc_0 [FAIL]

The issue is that the vectorized kernels do not work when inc_y == 0. This patch updates the kernels to fall back to the scalar algorithms when inc_y == 0, fixing the failing tests.


martin-frbg:
Thank you

PR #4803, opened 2024-07-16
A64FX: Add support for SVE to SGEMV/DGEMV kernels.

### LINKED ISSUE #4786: SVE support for SGEMV kernel and DGEMV kernel on A64FX
Issue #4779 may also discuss similar improvements, but I am not sure if the target platform is the same as mine, so I am posting this issue.
The SGEMV kernel and DGEMV kernel on A64FX do not currently use SVE, so they do not seem to have enough performance. This may not be a problem on NeoverseV1, since it can execute four 128-bit NEON instructions per cycle, which is the same number of floating operations as executing two 256-bit SVE instructions.  However, A64FX executes two 512-bit SVE or two 128-bit NEON arithmetic instructions per cycle, so the current SGEMV kernel and DGEMV kernel do not perform well.
So I am thinking of providing new kernels using SVE specifically for the A64FX.

iha-taisei:
This pull request proposes a fix to the issue #4786.
I added the SGEMV/DGEMV kernels that use SVE for A64FX and confirmed performance improvements.

DGEMV(TRANS=N) with 1 thread
DGEMV_N_1-thread

DGEMV(TRANS=T) with 1 thread
DGEMV_T_1-thread

The kernels can work with the SVE of a length other than 512 bits of the A64FX,but Neoverse V1 and Neoverse V2, which can execute four 128-bit NEON instructions per cycle do not necessarily need to adopt this SVE version.

### DIFF
--- a/kernel/arm64/KERNEL.A64FX
+++ b/kernel/arm64/KERNEL.A64FX
@@ -1,1 +1,6 @@
 include $(KERNELDIR)/KERNEL.ARMV8SVE
+
+SGEMVNKERNEL = gemv_n_sve.c
+DGEMVNKERNEL = gemv_n_sve.c
+SGEMVTKERNEL = gemv_t_sve.c
+DGEMVTKERNEL = gemv_t_sve.c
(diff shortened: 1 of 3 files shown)


martin-frbg:
This is looking great, thank you very much

PR #4809, opened 2024-07-19
Change computational order in GEMV and GEMM Power6 kernel

penghongbo:
The current develop branch reports about 200 failures in LAPACK testing. After changing the computational order and remove the reduction in tail loop, we only get 0 failure (AIX7.3 + OpenXL compiler) or 2 failures  (RHEL9.2 + GNU compiler).
  1. Remove the FADD insn from the GEMV Transpose code.
  2. Remove the FADD insn from GEMM and ZGEMM code.
  3. Reorder the compution of the Imaginary part in ZGEMM code.

### DIFF
--- a/kernel/power/gemm_kernel_power6.S
+++ b/kernel/power/gemm_kernel_power6.S
@@ -864,15 +864,15 @@ LL(22):
 	LFD	f22, 10 * SIZE(BO)
 	LFD	f23, 11 * SIZE(BO)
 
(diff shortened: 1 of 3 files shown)


martin-frbg:
Thank you very much for the quick fix.

martin-frbg:
(CI failures are almost certainly unrelated - I cannot even log into the OSUOSL PowerCI Jenkins server right now)

PR #4811, opened 2024-07-23
Add A64FX to the list of CPUs supported by DYNAMIC_ARCH

### LINKED ISSUE #4808: DYNAMIC_ARCH does not support A64FX
Hello

DYNAMIC_ARCH does not currently support A64FX.
https://github.com/OpenMathLib/OpenBLAS/blob/develop/Makefile.system#L683
When I built OpenBLAS with DYNAMIC_ARCH=1 and ran it on an A64FX, ARMV8SVE was selected.
However, as mentioned in Pull Request #4803  , A64FX has different characteristics from other CPUs with SVE.
Therefore, it is better to have separate kernels for A64FX from those for ARMV8SVE.

yamazakimitsufumi:
This pull request addresses issue #4808 , which add A64FX to the list of CPUs supported by DYNAMIC_ARCH.
A64FX and Neoverse CPUs previously shared the same implementation, but I’ve made them separated so that their changes don’t cannibalize each other.


yamazakimitsufumi:
Fixed build errors in gemv_n_sve.c and gemv_t_sve.c on Mac OS.

Mousius (approved):

PR #4816, opened 2024-07-24
Unroll SVE gemv further and use on all SVE cores

Mousius:
Unrolling the new SVE kernels looks to be good enough for all of the SVE cores I tested on.


Mousius:
@yamazakimitsufumi I think the `ARMV8SVE` target in DYNAMIC_ARCH now works for your use case. It would also be good if you could double-check these new kernels.

Mousius:
Closing temporarily, running some more benchmarks 🤔

PR #4817, opened 2024-07-26
Fix SCAL on x86 and RISCV_GENERIC

martin-frbg:
fixes two oversights from #4807

### DIFF
--- a/kernel/riscv64/scal.c
+++ b/kernel/riscv64/scal.c
@@ -43,7 +43,7 @@ int CNAME(BLASLONG n, BLASLONG dummy0, BLASLONG dummy1, FLOAT da, FLOAT *x, BLAS
 	if ( (n <= 0) || (inc_x <= 0))
 		return(0);
 	
-	if (dummy2 == 0) {
+	if (dummy2 == 1) {
 		while(j < n)
 		{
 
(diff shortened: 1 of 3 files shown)


mattip:
That was quick, thanks!

mattip:
Dunno if you want to fix it, there are some errors with Windows_cl [starting here](https://dev.azure.com/martinkroeker/martinkroeker/_build/results?buildId=1181&view=logs&j=2d8ea0da-d599-5234-e06f-743200a8aa8e&t=30994467-7236-5890-7de5-bc1a1b096700&l=9907)
```
utest\test_gemv.c(22,15): error C2057: expected constant expression
```

I think that is coming from declarations like this https://github.com/OpenMathLib/OpenBLAS/blob/fb7c53c5e56469d95bcf13345bc3532381873fb6/utest/test_gemv.c#L21-L23

where N is not a constant expression.

mattip:
Maybe you could `#define N` instead

martin-frbg:
thanks for the pointer - I had seen the failure but did not find the actual error when wading through the Azure log, so was hoping it would be the same C99 thing that broke the manylinux build. indeed these should be defines (but at least that wasn't my fault except for not noticing in review)

PR #4820, opened 2024-07-26
Opened PR in the wrong repo. Sorry!

mtsokol:
Opened PR in the wrong repo. Sorry!


martin-frbg:
all been there, no problem. (btw you probably don't need to go to the trouble of adding meson build files for the netlib blas and cblas subtrees unless you consider a PR at Reference-LAPACK/lapack - they're never built in an OpenBLAS context, they just happen to have been imported from the reference implementation together with LAPACK. more btw, "netlib" used to have meson support - briefly - but threw it out again as they didn't know what to do with it)

PR #4826, opened 2024-07-29
Add fallback compile options for A64FX target

Mousius:
According to https://gcc.gnu.org/gcc-10/changes.html, `mtune=a64fx` should be supported in 10.3.


martin-frbg:
Thanks

PR #4827, opened 2024-07-29
Fix GCC11 check for A64FX target

Mousius:
Realised that I hadn't included GCC 11 as an option instead of GCC 10.3 🙀


martin-frbg:
thought I could clean it up in the corresponding fixes for cmake without anybody noticing...

PR #4829, opened 2024-07-30
Update zarch SCAL kernels to handle INF and NAN arguments 

martin-frbg:


mattip:
Let me know if there is something I can do to help verify this.

PR #4835, opened 2024-08-02
Temporarily revert to the coarse-grained locking in the Windows thread server

martin-frbg:
this effectively undoes #4359 due to thread safety regressions found in numpy (scipy/sunpy) contexts - whether caused directly or only exposed by more efficient multithreading with that PR
https://github.com/numpy/numpy/issues/27036
https://github.com/conda-forge/openblas-feedstock/issues/160


mattip:
Thanks, this does seem like the only thing to do right now

PR #4847, opened 2024-08-06
Assume cross-compilation if EMBEDDED was specified

martin-frbg:
from discussion #4845


martin-frbg:
makes no sense to try doing it there...

PR #4848, opened 2024-08-07
cmake: update minimum_required cmake version

haampie:
It's better to enable https://cmake.org/cmake/help/latest/policy/CMP0042.html#policy:CMP0042 than to explicitly set CMAKE_MACOSX_RPATH, since users may want to set the latter from the command line.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,9 +2,7 @@
 ## Author: Hank Anderson <hank@statease.com>
 ##
 
-cmake_minimum_required(VERSION 2.8.5)
-
-cmake_policy(SET CMP0042 NEW)
+cmake_minimum_required(VERSION 3.16.0)
 
 project(OpenBLAS C ASM)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,9 @@
 ## Author: Hank Anderson <hank@statease.com>
 ##
 
-cmake_minimum_required(VERSION 2.8.5...3.0)
+cmake_minimum_required(VERSION 2.8.5)
+
+cmake_policy(SET CMP0042 NEW)
 
 project(OpenBLAS C ASM)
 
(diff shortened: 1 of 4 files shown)


martin-frbg:
Thanks - I'm always a bit hesitant when it comes to tightening minimum version requirements, as at least "historically" some users were unable to update the systems they worked on. The original CMakeLists.txt under lapack-netlib are not used when building LAPACK as a part of OpenBLAS, so I think this can be deferred to 0.3.29 (?)

haampie:
Ah, I saw some `add_subdirectory` and assumed that's how it worked, sorry for the misunderstanding.

In that case let me use the most conservative option: `cmake_policy(SET CMP0042 NEW)`

conradsnicta:
@martin-frbg @haampie I'd suggest to bump the min version of CMake, rather than messing around with policies.

Using ancient versions of CMake adds to the technical debt and increases the maintenance burden.

CMake 2.8.5 (currently required by OpenBLAS) is truly prehistoric, released circa 2010.  
Cmake 3.0 (released in 2014) is generally known as the starting point for "modern" CMake.

A good indicator of the oldest allowable version of CMake would be across old mainstream long-term linux distros (Red Hat, Ubuntu, Debian) that are still within their normal support period:

* RHEL 8 has CMake 3.26.5
* Debain 11 has CMake 3.18.4
* Ubuntu 20.04 LTS has 3.16.3

This suggests CMake 3.16 as the oldest baseline version.  Anything older than that has a rapidly diminishing number of users.

(Similar comment applies to ancient versions of GCC and Clang.  In my view, the required workarounds for out-of-support compilers are simply not worth it.)

martin-frbg:
turns out that updating the minimum required cmake version (or adding even more cmp... policies) is also necessary for cross-compilations with the current Android NDK (r27), so I've finally hacked conradsnicta's suggestion into your PR branch

haampie:
great

PR #4849, opened 2024-08-07
Fix invocation of gemm3m tests in gmake builds

martin-frbg:


martin-frbg:
not sure what happened with this branch being so far in the past...redoing

Issue #4856, opened 2024-08-09
stest_rfp fails on Power10 


penghongbo:
When run with the develop branch of OpenBLAS on POWER10, AIX7.3, we found this test fails which may be a minor failure as discussed in previous issue https://github.com/OpenMathLib/OpenBLAS/issues/3648. The reference LAPACK still set threshold to 30.0 but OpenBLAS set it to 42.0. Is that OK to set it larger such as 45.0?

```
 All tests for SPF drivers  passed the threshold (   2304 tests run)
 All tests for SLANSF auxiliary routine passed the threshold (   384 tests run)
 All tests for the RFP conversion routines passed (    72 tests run)

  *** Error(s) or Failure(s) while testing STFSM               ***
      Failure in STFSM, CFORM='T', SIDE='L', UPLO='U', TRANS='T', DIAG='N', M= 50, N = 50, test=  42.893
  STFSM auxiliary routine:     1 out of  7776 tests failed to pass the threshold
 All tests for SSFRK auxiliary routine passed the threshold (  2592 tests run)
```

martin-frbg:
Well, _I_ think it should be OK as long as this seems to be more of a cosmetic problem. I tried raising this issue with the Reference-LAPACK team back then but they seem to consider it "not their problem" (and probably rightly so, as the topic of "minor failures" is mentioned in their FAQ and the testsuite still does what it was designed for - verifying non-optimized reference LAPACK builds with a non-optimized reference BLAS)

PR #4866, opened 2024-08-12
set larger threshold in stest_rfp.in for POWER10 

penghongbo:
This PR increases the threshold in stest_rfp.in to 45.0. This helps issue #4856.


penghongbo:
Sorry, there is already a PR #4865 for this. So I will close this one. Thanks for creating #4865.

PR #4870, opened 2024-08-14
Fix recursive variable expansion in Makefiles for LOONGSON3A

chenx97:
replace `=` with `:=` to avoid `*** Recursive variable 'FFLAGS' references itself (eventually).`


martin-frbg:
Thank you. (Seems older versions of gmake tolerated this, or I should have noticed ?)

PR #4880, opened 2024-08-15
[POWER] Vectorize SGEMV transpose reduce stage

ChipKerchner:
Vectorize SGEMV transpose (vector x matrix) reduce stage.

Decreases number of instructions in reduce by 5-6X.


martin-frbg:
Unfortunately I am (belatedly) seeing test failures for SGEMV and SSYMV on POWER8 under CentOS Linux, ppc64le (host cfarm112 in the Gcc Compile Farm, which is said to be POWER8 8247-22L provided by OSUOSL) that can be fixed by reverting the changes to sgemv_t.c

ChipKerchner:
I'm not 100% sure how to reproduce this failure.

martin-frbg:
It was a surprise find from just running `make` on the aforementioned machine - however I notice now that it only has gcc8 installed. I'll try to reproduce this with a newer compiler if possible

ChipKerchner:
Yes, I would recommend gcc11+ or later.

ChipKerchner:
I could update this PR to only vectorize this data for P9.

ChipKerchner:
Could update this PR (or a new one) to only vectorize this data for P9. Mainly it would be

#ifndef _ARCH_PWR9
old way
#else
new way
#endif

martin-frbg:
I'm willing to leave it as it is now, if the error only appears with the ancient compiler. But the only other (gccfarm) little-endian power8 I have access to appears to be in Moscow of all places and its connection is flaky. So I'll have to see if building gcc14 on the old CentOS is faster than trying to git pull OpenBLAS _a la russe_

martin-frbg:
Sorry, dropped the ball on this until being reminded by #5122 today - the failures on POWER8 turn out to be reproducible with gcc14 (also when building for TARGET=POWER8 on POWER10

Issue #4805, opened 2024-07-17
GCC14 issues on loongarch64 and amd64


azuresky01:
Hello,

I just tested current develop version of OpenBLAS on Loongson-3A6000 (loongarch64 architecture) and found two issues. The first one also appears on AMD Zen4 7940HS with Debian 12.

 *** The first issue (loongarch64 & amd64):  

many tests in ctest fail for cmake build with GCC14 and setting  INTERFACE64=1:
```
OS:  Loong Arch Linux  or Debian GNU/Linux trixie/sid loongarch64                      
Host: Loongson-3A6000-HV-7A2000-1w-V0.1-EVB                               
Kernel: 6.8.6-2 or 6.9.7-loongarch-00002-gfbf7212b7ab3
gcc version 14.1.0 compiled from the source
cmake version: 3.28.2-1 or 3.30.0
```

compiling and test commands:

```
CC=gcc-14.1.0 FC=gfortran-14.1.0 cmake -B build64 -DBUILD_TESTING=OFF -DINTERFACE64=1
cmake --build build64
cd build64
ctest
```

ctest result:

```
Test project /home/loongson/test/cbuild64/build64
      Start  1: openblas_utest
 1/26 Test  #1: openblas_utest ...................***Exception: SegFault  0.01 sec
      Start  2: openblas_utest_ext
 2/26 Test  #2: openblas_utest_ext ...............***Exception: SegFault  0.01 sec
      Start  3: sblas1
 3/26 Test  #3: sblas1 ...........................Bus error***Exception:   0.03 sec
      Start  4: sblas2
 4/26 Test  #4: sblas2 ...........................   Passed    0.03 sec
      Start  5: sblas3
 5/26 Test  #5: sblas3 ...........................***Failed    0.02 sec
      Start  6: dblas1
 6/26 Test  #6: dblas1 ...........................Bus error***Exception:   0.03 sec
      Start  7: dblas2
 7/26 Test  #7: dblas2 ...........................   Passed    0.03 sec
      Start  8: dblas3
 8/26 Test  #8: dblas3 ...........................***Failed    0.02 sec
      Start  9: cblas1
 9/26 Test  #9: cblas1 ...........................Bus error***Exception:   0.03 sec
      Start 10: cblas2
10/26 Test #10: cblas2 ...........................   Passed    0.03 sec
      Start 11: cblas3
11/26 Test #11: cblas3 ...........................   Passed    0.02 sec
      Start 12: zblas1
[...]

PR #4881, opened 2024-08-15
Use fld.d/fst.d in PROLOGUE/EPILOGUE in LOONGSON3R5 GEMM

martin-frbg:
same underlying issue of potential data loss as detailed in #4872 - this is most likely the proper fix for the spurious utest errors seen in #4805

### DIFF
--- a/kernel/loongarch64/cgemm_kernel_16x4_lasx.S
+++ b/kernel/loongarch64/cgemm_kernel_16x4_lasx.S
@@ -196,17 +196,17 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
     SDARG      $r25,   $sp,   16
     SDARG      $r26,   $sp,   24
     SDARG      $r27,   $sp,   32
-    ST         $f23,   $sp,   40
-    ST         $f24,   $sp,   48
-    ST         $f25,   $sp,   56
-    ST         $f26,   $sp,   64
-    ST         $f27,   $sp,   72
-    ST         $f28,   $sp,   80
-    ST         $f29,   $sp,   88
-    ST         $f30,   $sp,   96
-    ST         $f31,   $sp,   104
-    ST         ALPHA_R,$sp,   112
-    ST         ALPHA_I,$sp,   120
+    fst.d         $f23,   $sp,   40
+    fst.d         $f24,   $sp,   48
+    fst.d         $f25,   $sp,   56
+    fst.d         $f26,   $sp,   64
+    fst.d         $f27,   $sp,   72
+    fst.d         $f28,   $sp,   80
+    fst.d         $f29,   $sp,   88
+    fst.d         $f30,   $sp,   96
+    fst.d         $f31,   $sp,   104
+    fst.d         ALPHA_R,$sp,   112
+    fst.d         ALPHA_I,$sp,   120
 
     xvldrepl.w  VALPHAR, $sp, 112
     xvldrepl.w  VALPHAI, $sp, 120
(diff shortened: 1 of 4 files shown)


azuresky01:
This new fix does not work yet. On Debian sid loong64 with gcc 14.2.0:

Building command: `CC=gcc FC=gfortran make`
The results:
```
...
TEST 103/107 potrf:smoketest_trivial [FAIL]
  ERR: test_potrs.c:535  U s(0,0) difference: 1.19209e-07
TEST 104/107 potrf:bug_695 [OK]
TEST 105/107 kernel_regress:skx_avx [OK]
TEST 106/107 fork:safety [OK]
TEST 107/107 fork:safety_after_fork_in_parent [OK]
RESULTS: 107 tests (106 ok, 1 failed, 0 skipped) ran in 201 ms
make[1]: *** [Makefile:86: run_test] Error 1
...
```
The problem remains.

martin-frbg:
Ah, sad. Guess it was a bit premature to remove the pragma, though it seemed to work well without it on Loongson's 3C5000 in the gcc compile farm (using both gcc14.1 and a git snapshot of 15)

azuresky01:
My CPU is 3A6000. By the way, the problem appears only when building with gmake. cmake builds seem always fine before/after patches.

martin-frbg:
cmake build probably uses -O0 when no CMAKE_BUILD_TYPE is specified, and -O3 when building with setting "Release", while gmake build defaults to -O2
As far as I know 3A6000 should use same kernels as 3C5000 (basically the build system only distinguishes between cpu models that support LASX, LSX or neither), but I can only test on the 3C5000. I'll just re-enable the pragma in test_potrs.c in a minute

martin-frbg:
curiously I do not get the error on 3C5000 even with gcc 14.2.0 (built from source) and the changes from this PR. (without them, the fault appears on each run of utest, with them I can do 100 runs of utest without a single failure). Probably the difference is in the version of binutils used, but can you please check what value of CORE gets put into your Makefile.conf (should be LOONGSON3R5) just to be sure ?

azuresky01:
Currently the version of binutils on Debian sid loong64 is 2.43-2. And yes the value of CORE in my Makefile.conf is LOONGSON3R5, see below:
```
OSNAME=Linux
ARCH=loongarch64
C_COMPILER=GCC
BINARY32=
BINARY64=1
CEXTRALIB= -L/usr/lib/gcc/loongarch64-linux-gnu/14 -L/usr/lib/gcc/loongarch64-linux-gnu/14/../../../loongarch64-linux-gnu -L/usr/lib/gcc/loongarch64-linux-gnu/14/../../.. -L/lib/loongarch64-linux-gnu -L/usr/lib/loongarch64-linux-gnu  -lc 
F_COMPILER=GFORTRAN
FC=gfortran
BU=_
FEXTRALIB= -L/usr/lib/gcc/loongarch64-linux-gnu/14 -L/usr/lib/gcc/loongarch64-linux-gnu/14/../../../loongarch64-linux-gnu -L/usr/lib/gcc/loongarch64-linux-gnu/14/../../.. -L/lib/loongarch64-linux-gnu -L/usr/lib/loongarch64-linux-gnu  -lgfortran -lm -lm -lc 
CORE=LOONGSON3R5
LIBCORE=loongson3r5
NUM_CORES=8
MAKEFLAGS += -j 8
...
```

azuresky01:
And I report that on Loong Arch Linux with your new fix the build can complete without errors. But when I run `./utest/openblas_utest` the problem still exists. On Loong Arch linux the version of binutils is 2.42-1 and the version of gcc is 14.0.1 20240421 (experimental).

martin-frbg:
strange - host "cfarm400.cfarm.net" where I am running my tests claims to be "Debian trixie/sid",  kernel version 6.1.0rc7+ and has binutils 2.43 installed

azuresky01:
My Debian unstable system is similar but it has a newer kernel:
```
OS: Debian GNU/Linux trixie/sid loongarch64
Host: Loongson-3A6000-HV-7A2000-1w-V0.1-EVB
Kernel: 6.9.7-loongarch-00002-gfbf7212b7ab3
...
```
And my apt source is:
`deb [arch=loong64] http://ftp.ports.debian.org/debian-ports unstable main contrib non-free non-free-firmware`

azuresky01:
@martin-frbg

update: I recently tried to build OpenBLAS 0.3.30/develop with different gcc compilers (13.4/14.2/14.3/15.1/15.2), and found that the pragma in `utest/test_potrs.c` can be safely removed. The tested system/host/Kernel are:

```
OS: Debian GNU/Linux forky/sid loongarch64
Host: Loongson-3A6000-HV-7A2000-1w-V0.1-EVB
Kernel: Linux 6.12.38+deb13-loong64
```

However, when building OpenBLAS version 0.3.29 or earlier with gcc>=14, the pragma in `utest/test_potrs.c` is still needed.

martin-frbg:
Interesting, thanks. The only related change in 0.3.30 would seem to be #5248, which addressed precision issues in several level 1 BLAS functions

PR #4883, opened 2024-08-16
Fix SBGEMM unit test to handle zero elements.

ChipKerchner:
Fix SBGEMM unit test to handle zero elements - since malloc zero bytes fails on some systems.

### DIFF
--- a/test/compare_sgemm_sbgemm.c
+++ b/test/compare_sgemm_sbgemm.c
@@ -96,7 +96,7 @@ main (int argc, char *argv[])
   char transA = 'N', transB = 'N';
   float alpha = 1.0, beta = 0.0;
 
-  for (x = 0; x <= loop; x++)
+  for (x = 1; x <= loop; x++)
   {
     if ((x > 100) && (x != SBGEMM_LARGEST)) continue;
     m = k = n = x;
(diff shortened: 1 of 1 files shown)


ChipKerchner:
Fixes malloc(0) failing for AIX.

martin-frbg:
Hmm, that's annoying. M,N,K equal 0 might be a valid corner case to test, though not that important.

ChipKerchner:
I agreed.  I say that we leave that corner case for another PR?

ChipKerchner:
Fixed to handle zero elements.

ChipKerchner:
GEMV returns an error for zero elements.

```
 ** On entry to SGEMV  parameter number  6 had an illegal value
 ** On entry to SBGEMV  parameter number  6 had an illegal value
```

ChipKerchner:
I believe this PR is ready for merging.

ChipKerchner:
Actually zero length GEMV will not work because of this (and other checks)

```
if (lda < MAX(1, m))   info = 6;
```

If it is included, I get this

```
 ** On entry to SGEMV  parameter number  6 had an illegal value
 ** On entry to SBGEMV  parameter number  6 had an illegal value
```

ChipKerchner:
So this PR is ready.  Sorry about all the changes.

martin-frbg:
Thank you (sorry for causing extra work with what was just a spontaneous thought before going on a long drive)

PR #4918, opened 2024-10-05
TST,BUG: Explicitly allow running tests multiple times

HaoZeke:
These are extracted from the `meson` port effort [1] but are not directly tied to a build system. The main fixes involve:
- Calling `abort` over `stop`
  + Equally well supported (in terms of language standards / compilation), but `stop` reports no errors (i.e. return code was always 0), and doesn't include a backtrace.
- Changing the test write status from `new` to `replace`
  + Allows tests to run multiple times without a clean step in between 

[1] https://github.com/HaoZeke/OpenBLAS/tree/mesonBasic


HaoZeke:
M1 failure seems unrelated..

martin-frbg:
@HaoZeke any objections to replacing your `call abort` with a standard-conforming `error stop` for portability ?

HaoZeke:
> @HaoZeke any objections to replacing your `call abort` with a standard-conforming `error stop` for portability ?

I use `error stop [err code]` in newer code, and it would generally be a step up. 

For OpenBLAS though, I'm not sure since it is a Fortran 2008 standard feature, and might not be as widely supported. `call abort` is an extension ([GNU](https://gcc.gnu.org/onlinedocs/gfortran/ABORT.html), [Intel](https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2025-0/abort.html)) but it has been around a lot longer AFAIK.

EDIT: My bad, I hadn't had a copy of NAG lying around (licensing fees) so I had forgotten it wasn't supported. The same comment stands though, as noted in https://github.com/OpenMathLib/OpenBLAS/issues/5002 a backfill function I'd be happier with, unless we're sure OpenBLAS is always compiled with Fortran 2008 compliant compilers.

EDIT-2: Actually I'd have just recommended `f90_unix_proc` in https://github.com/OpenMathLib/OpenBLAS/pull/5003#issuecomment-2518769477

PR #4924, opened 2024-10-09
LoongArch64: Update README.md

XiWeiGu:
@azuresky01 @martin-frbg @yinshiyou , Could you help me review this PR? Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,26 @@ e.g.:
     HOSTCC=gcc HOSTFC=gfortran -j
   ```
 
+#### LOONGARCH64
+
+- **LA64_GENERIC**: Optimized Level-3, Level-2 and Level-1 BLAS with scalar instruction
+  ```sh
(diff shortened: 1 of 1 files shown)


martin-frbg:
Looks good to me (though of course I am not that familiar with the Loongson hardware range - I assume a user would know whether their hardware is LSX or LASX ?) Incidentally, is there a known performance or correctness problem that (still) makes it necessary to use the "simple" level3 threading code on LoongArch64 ?

XiWeiGu:
Using the "simple" level 3 threading code is mainly for performance reasons, and performance data will be provided later.

XiWeiGu:
Test data on the 3A5000 2.5GHz with a peak double-precision floating-point performance of 160 GFlops is as follows:
图片2
When using `USE_SIMPLE_THREADED_LEVEL3=1` on the LoongArch64 platform, there are fewer L3 cache conflicts, resulting in better performance. However, I have also compared this with the X86_64 platform and found the opposite phenomenon, where using `USE_SIMPLE_THREADED_LEVEL3=1` often leads to the lowest performance. This may be related to cache design (just a guess).
So I still recommend enabling `USE_SIMPLE_THREADED_LEVEL3` on the LoongArch64 platform for now.

martin-frbg:
Thank you very much for the detailed explanation and data.

PR #4648, opened 2024-04-15
Add implementation of WhereAmI() to support NO_AFFINITY=0 on ARM64

martin-frbg:
fixes #4623

PR #4927, opened 2024-10-09
Exclude the WhereAmI function on Arm64 when compiling with MSVC

martin-frbg:
fixes compilation problem for WoA flagged in comment on my original PR #4648


martin-frbg:
closing as basically equivalent to #4926 (at least as far as WoA is concerned)

PR #4928, opened 2024-10-09
CGEMM & ZGEMM using C code, Power only, P10 only.

austinpagan:
We originally uploaded this nearly a year ago, but discovered a bug in the XL compiler.
We waited for the compiler to be fixed for AIX, and it is now fixed in version XL 17.1.2.10.
We are now ready to upload this code again!


martin-frbg:
Great, thanks.

austinpagan:
That one failure looks pretty far away from what I'm changing...

martin-frbg:
sure, I just haven't gotten round to helping osx/homebrew find its android-ndk package again

PR #4933, opened 2024-10-11
Change multi-threading logic for SBGEMV to be the same as SGEMV.

ChipKerchner:
The logic for threading SBGEMV was only using the size of the vector when determining whether or not to thread it.

This makes the logic the same as SGEMV (F32).

With 32 threads, I'm seeing a 4.9X improvement for SBGEMV.


ChipKerchner:
The failures don't seem to be related to my change since they are not BF16.

martin-frbg:
> The failures don't seem to be related to my change since they are not BF16.

right. that is fallout from yesrerday's pr merge that made the ctests raise sigabrt on failure. this uncovered a latent failure that makes little sense except as a compiler error (given that it affects the pure C mips64_generic target, and only the cblas version of the test)

PR #4940, opened 2024-10-15
SVE Implementation for Level-1 BLAS Routines 

garadeaniket:
We have optimized Level-1 BLAS routines (scal, swap, and rot) utilizing ARM SVE, resulting in significant performance enhancements in OpenBLAS on two variants of the A64FX—FUJITSU PRIMEHPC FX700 and the FUGAKU supercomputer. These optimizations achieved performance improvements ranging from 1.80x to 4x through effective code vectorization. This research has been accepted as a full paper and presented at the 28th Annual IEEE High Performance Extreme Computing (HPEC) Conference in September 2024, under the title "Optimization Strategies to Accelerate BLAS Operations with ARM SVE."


kernel/arm64/dgemv_kernel_sve.c
```diff
@@ -0,0 +1,41 @@
+#include "common.h"
```
Mousius, line 1: I don't think this file is used and is already implemented in `kernel/arm64/gemv_n_sve.c` / `kernel/arm64/gemv_t_sve.c`

kernel/arm64/rot.c
```diff
@@ -0,0 +1,19 @@
+#include "common.h"
+
+#include "rot_kernel_sve.c"
```
Mousius, line 3: This should be guarded so as not to include SVE when it's not supported, see:
https://github.com/OpenMathLib/OpenBLAS/blob/develop/kernel/arm64/dot.c#L32-L40

kernel/arm64/rot.c
```diff
@@ -0,0 +1,19 @@
...
+#include "rot_kernel_c.c"
+
+int CNAME(BLASLONG n, FLOAT *x, BLASLONG inc_x, FLOAT *y, BLASLONG inc_y, FLOAT c, FLOAT s)
+{
+  
+       
+        if ( n <= 0 )  return(0);
+        if ( inc_x == 1 && inc_y==1)
+	   rot_kernel_sve( n, x, y, c, s);
```
Mousius, line 12: As above, it requires a guard:
https://github.com/OpenMathLib/OpenBLAS/blob/develop/kernel/arm64/dot.c#L59-L63

Mousius (changes requested):
Great to see more routines making use of SVE 😸 .

I have a few questions:
* Compared to the existing `rot`/`scal` for none `inc_x==1 && inc_y==1` cases, how does this perform?
* Can we make the generic `rot.c`/`scal.c`/`swap.c` compile without SVE enabled, preferably with similar performance?
* Could you run a code formatter and clean up the whitespace?
* Could we add some license headers? 😸 
* Did you try unrolling the loops further?

garadeaniket:
> Great to see more routines making use of SVE 😸 .
> 
> I have a few questions:
> 
> * Compared to the existing `rot`/`scal` for none `inc_x==1 && inc_y==1` cases, how does this perform?
> * Can we make the generic `rot.c`/`scal.c`/`swap.c` compile without SVE enabled, preferably with similar performance?
> * Could you run a code formatter and clean up the whitespace?
> * Could we add some license headers? 😸
> * Did you try unrolling the loops further?

We will review the recommended points and implement the necessary changes. Once completed, we will submit a pull request through the organization's GitHub account.

kernel/arm64/rot.c
garadeaniket: I tried adding the sve  guards, but after adding these  guards our sve kernels are not being called. 
Please help to rectify it.

martin-frbg: What did you add ? Did you copy the entire block that @Mousius marked, plus the actual `#ifdef HAVE_SVE...` include guard for your sve kernel ?
This should look something like
```
#include "common.h"
// Some compilers will report feature support for SVE without the appropriate
// header available
#ifdef HAVE_SVE
#if defined __has_include 
#if __has_include(<arm_sve.h>) && __ARM_FEATURE_SVE
#define USE_SVE
#endif 
#endif
#endif

#ifdef USE_SVE
#include "rot_kernel_sve.c"
#endif
```

garadeaniket: yes , I did the same.  I tried debugging the already sve implemented dot product also.   Even in dot product the "dot_kernel_sve.c" is not being called instead "dot_kernel_asimd.c" is called. Do I need to target  ARMV8SVE while running make command?

martin-frbg: Curious. Do you see HAVE_SVE in the compiler command lines, or in the generated config.h ?
Setting target ARMV8SVE should work, maybe autodetection did not pick the correct cpu (what kind of cpu are you building this on?)

martin-frbg: Nevermind, there is some broken logic in the autodetection code - right now HAVE_SVE will only be set when you specify a target. I'll produce a fix.

garadeaniket: It's in the generated config.h. I am using A64FX. When I build using TARGET=ARMV8SVE , test of ssyrk routine is failing.
 ******* SSYRK  FAILED ON CALL NUMBER:
   1732: SSYRK ('U','N', 31,  2, 1.0, A, 32, 0.0, C, 32)           .

 SSYR2K PASSED THE TESTS OF ERROR-EXITS

       1     -0.160742         -0.626985E-01
      THESE ARE THE RESULTS FOR COLUMN   1
 ******* SSYR2K FAILED ON CALL NUMBER:
   1732: SSYR2K('U','N', 31,  2, 1.0, A, 32, B, 32, 0.0, C, 32)

martin-frbg: does it work when you set TARGET=A64FX ?

martin-frbg: at least it works for me - you could try adding `#error using sve` after the `#ifdef USE_SVE` in either dot.c or your rot.c to see if compilation stops there. Obviously the other explanations would be that your toolchain lacks the arm_sve header or does not define the __ARM_FEATURE_SVE

garadeaniket: Yes. it works with TARGET=A64FX.

PR #4944, opened 2024-10-17
[POWER] Vectorize BF16 GEMV

ChipKerchner:
Vectorize BF16 GEMV - Up to 26X faster.  Versions for VSX and MMA.

Add GEMM_GEMV_FORWARD_BF16 compilation flag to allow BF16 GEMV forwarding (like F32).

Updated BF16 to F32 unit test to allow for non inc_x/inc_y equal to 1 for GEMV.


ChipKerchner:
All these failures either are not related to the PR or it took too long to compile.

martin-frbg:
Right, some of the jobs on Azure tend to time out when they get scheduled on ancient hardware. And I'm still trying to figure out locally why the mingw32 job is failing.

giordano:
I believe this broke compilation of OpenBLAS with GCC 6-7: #5068

PR #4953, opened 2024-10-24
fix: return types of *trtrs routines

NickelWenzel:
There are still mismatches of function signatures for the f2c'ed lapack sources.

This PR fixes those mismatches for the trtrs routine family.


martin-frbg:
Thanks - looks like I even got confused at some point in the past and edited the TRTRS prototypes _away_ from their canonical definition to match the type of their reimplementations in OpenBLAS

NickelWenzel:
I'm not sure what to do about the apple android CI build, looks like some toolchain file is missing on the build agent.
https://github.com/OpenMathLib/OpenBLAS/pull/4953/checks?check_run_id=31998029185

PR #4954, opened 2024-10-25
gh-actions: Test LoongArch64 with gcc14 from Ubuntu 24.04

XiWeiGu:


martin-frbg:
Thanks, also good to see that we no longer need to spend time building our own copy of qemu here

PR #4955, opened 2024-10-25
Optimize gemv forwarding on ARM64 systems

### LINKED ISSUE #4951: Performance regression in version 0.3.28 on aarch64 because of GEMM to GEMV transformation
We have found a DGEMM performance regression in OpenBLAS-0.3.28 on aarch64 platforms. It happens because of the GEMM to GEMV forwarding that was introduced and enabled on aarch64 in this version of OpenBLAS. Here is the difference in performance in GFLOP/s for a single-threaded dgemm('N','T',1,91,90,1,a,50,b,91,0,c,50) with and without GEMM to GEMV forwarding:

| Processor     | With forwarding (default) | Without forwarding |
|---------------|----------------:| ------------------:|
| Ampere Altra  | 0.77            | 4.05               |
| AWS Graviton3 | 0.72            | 7.64               |
| NVIDIA Grace  | 0.92            | 12.21              |

The results show that the forwarding can cause an order of magnitude performance loss. A perf profile shows that the time is spent in the scalar code at https://github.com/OpenMathLib/OpenBLAS/blob/v0.3.28/kernel/arm64/gemv_n.S#L295.

The GEMM to GEMV forwarding performs better after copying KERNEL.A64FX to KERNEL.NEOVERSEV1 and KERNEL.NEOVERSEV2, but it is still below the performance without GEMM to GEMV forwarding:

| Processor     | KERNEL.A64FX |
|---------------|-------------:|
| AWS Graviton3 | 3.11         |
| NVIDIA Grace  | 4.72         |

A perf profile shows that the time is spent in the scalar code at: https://github.com/OpenMathLib/OpenBLAS/blob/v0.3.28/kernel/arm64/gemv_n_sve.c#L81.

It seems to be beneficial to either disable the GEMM to GEMV forwarding or restrict its usage as follows:

```
diff -ruN OpenBLAS-0.3.28-original/interface/gemm.c OpenBLAS-0.3.28/interface/gemm.c
--- OpenBLAS-0.3.28-original/interface/gemm.c   2024-08-08 13:41:46.000000000 -0700
+++ OpenBLAS-0.3.28/interface/gemm.c    2024-10-22 07:59:31.538919000 -0700
@@ -521,7 +521,7 @@
       GEMV(&NT, &m, &n, args.alpha, args.a, &lda, args.b, &inc_x, args.beta, args.c, &inc_y);
       return;
     }
-    if (args.m == 1) {
+    if (args.m == 1 && args.ldc == 1) {
       blasint inc_x = args.lda;
       blasint inc_y = args.ldc;
       // These were passed in as blasint, but the struct translates them to blaslong
```
https://github.com/OpenMathLib/OpenBLAS/blob/v0.3.28/interface/gemm.c#L524.

This appears to be different to issue #4939, which has DGEMM calls with m != 1.

cdaley:
Add a modified version of the optimization mentioned in issue #4951. The code now avoids goto statements. The code now also only restricts the gemv forwarding on ARM64 systems since this is where we observed poor gemv performance.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -226,3 +226,6 @@ In chronological order:
 
 * Dirreke <https://github.com/mseminatore>
   * [2024-01-16] Add basic support for the CSKY architecture
+
+* Christopher Daley <https://github.com/cdaley>
+  * [2024-01-24] Optimize GEMV forwarding on ARM64 systems
--- a/interface/gemm.c
+++ b/interface/gemm.c
@@ -39,6 +39,7 @@
 
 #include <stdio.h>
 #include <stdlib.h>
+#include <stdbool.h>
 #include "common.h"
 #ifdef FUNCTION_PROFILE
 #include "functable.h"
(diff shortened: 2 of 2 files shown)


Mousius (approved):
Looks good to me, thanks for removing the `goto`s 😸

martin-frbg:
Thank you (both), merging this.

PR #4978, opened 2024-11-17
remove any optimization flags when building in DEBUG mode on POWER

### LINKED ISSUE #4891: Makefiles need to turn off optimizations for DEBUG=1
It makes it challenging to debug issues when optimizations are turned on and using gdb.  All optimizations should be set to -O0 for DEBUG=1.

martin-frbg:
fixes #4891

### DIFF
--- a/Makefile.power
+++ b/Makefile.power
@@ -93,6 +93,11 @@ FCOMMON_OPT += -O2 -Mrecursive
 endif
 endif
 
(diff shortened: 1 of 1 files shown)


martin-frbg:
actually does not work due to our complex makefile structure - in effect the -DCNAME=<function> from Makefile.system gets dropped, leading to confused and confusing compiler errors

PR #4979, opened 2024-11-17
Remove any optimization flags from DEBUG builds on POWER architecture

### LINKED ISSUE #4891: Makefiles need to turn off optimizations for DEBUG=1

martin-frbg:
fixes #4891


ChipKerchner:
Thanks!

PR #4982, opened 2024-11-20
Restore libsuffix support in pkg-config file

svillemot:
It had been mistakenly removed in 9ef10ffa496b919c25aedbb4aa2fdb930901475a.

### DIFF
--- a/openblas.pc.in
+++ b/openblas.pc.in
@@ -2,6 +2,6 @@ Name: openblas
 Description: OpenBLAS is an optimized BLAS library based on GotoBLAS2 1.13 BSD version
 Version: ${version}
 URL: https://github.com/xianyi/OpenBLAS
-Libs: -L${libdir} -l${libprefix}openblas${libnamesuffix}
+Libs: -L${libdir} -l${libprefix}openblas${libnamesuffix}${libsuffix}
 Libs.private: ${extralib}
 Cflags: -I${includedir} ${omp_opt}


martin-frbg:
Ugh, not again. But I don't think your fix is entirely correct either, the libsuffix is the same as the symbolsuffix so would come before the libnamesuffix (?)

svillemot:
`libsuffix` is empty for 32-bit indexing and equal to `64` for 64-bit indexing. It’s used for the Debian packages.

svillemot:
Also note that I did the same as in `cmake/openblas.pc.in`.

martin-frbg:
then there is probably more chaos in the naming. cmake builds differ from gmake "historically" by automatically setting SYMBOLSUFFIX=64 (which carries over into LIBSUFFIX=64)  for 64bit indexing, but in principle that SYMBOLSUFFIX can be overriden by the user in either case. And any other suffix on the library name (such as _numpy, _scipy or whatever else has been used in the past) would need to come after the usually-numerical-when-present SYMBOLSUFFIX _I think_

svillemot:
Ok. I don’t know the purpose of `libnamesuffix` and I don’t use it, so I don’t have a strong opinion. Please implement whatever solutions looks correct to you. From the perspective of Debian, it’s ok as long as the `64` suffix appears in the generated pkg-config file.

martin-frbg:
will follow up with another pr, there are some unfortunate cases now where we can get duplication of suffixes in both the .pc and .cmake file

PR #4984, opened 2024-11-21
doc: update README to link to the html docs and fix links

rgommers:
Also some minor formatting improvements and linking the home page.


rgommers:
@martin-frbg I think the canonical link for the homepage is now http://www.openmathlib.org/OpenBLAS/, so it would also be useful to edit the link in the right sidebar on the front page for this repo.

martin-frbg:
I kind of like the shorter url, so I've added a redirect for now

rgommers:
Thanks, that works too. Certainly easier to remember the shorter one:)

PR #4992, opened 2024-11-26
CI (MinGW): Remove CLANG32 environment from build matrix.

mmuetzel:
The CLANG32 environment is in the process of being removed from MSYS2 currently:
https://www.msys2.org/news/#2024-09-23-starting-to-drop-the-clang32-environment

Remove it from the build matrix ahead of its complete removal from MSYS2.

### DIFF
--- a/.github/workflows/dynamic_arch.yml
+++ b/.github/workflows/dynamic_arch.yml
@@ -158,7 +158,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        msystem: [UCRT64, MINGW32, CLANG64, CLANG32]
+        msystem: [UCRT64, MINGW32, CLANG64]
         idx: [int32, int64]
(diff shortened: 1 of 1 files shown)


mmuetzel:
Rebased on a current head to resolve a merge conflict.

martin-frbg:
I wonder if it would be worth looking into the llvm-mingw that the linked news file suggests as a replacement, but 32bit clang on Windows is probably not that important to have in CI anymore

mmuetzel:
> 32bit clang on Windows is probably not that important to have in CI anymore

Yeah. 32-bit Clang on Windows was never that big in use anyway as far as I can tell.

Imho, the only thing that might be worth keeping covered in CI is building C_LAPACK (if that isn't covered by any of the other CI jobs anyway).

PR #5003, opened 2024-12-04
Fixes for NAG Compiler

### LINKED ISSUE #5002: NAG CMake support is incomplete/broken
All,

I recently tried to build OpenBLAS with NAG Fortran using CMake and it was...not happy. 

In my testing I've found incomplete NAG support in CMake itself. This was fixable but in doing so, I found there were Fortran Standards issues in:

* lapack-netlib
* local ctest code

So now here's the crux. I have a set of fixes for the CMake and the Fortran Standards issues and can issue a PR at anytime, but I'm not sure how to proceed. I'll detail my reasoning below.

---

### CMake

These should be safe to take since, well, without them, NAG just has issues. They are partly bringing stuff from Makefile to CMake systems and partly from changes I found from @ACSimon33 from https://github.com/Reference-LAPACK/lapack/pull/686 (also seen below)

### lapack-netlib

The lapack-netlib fixes are all thanks to @ACSimon33 who fixed the same issues in a couple Reference-LAPACK PRs:

* https://github.com/Reference-LAPACK/lapack/pull/686
* https://github.com/Reference-LAPACK/lapack/pull/951

But, my question is: How often are Reference-LAPACK changes brought into OpenBLAS? If a sync with the Reference is coming, well, those changes in *my* PR might be moot as they are fixed "upstream". But if the sync might not happen for a bit, I can put them in my PR.

### local ctest code

These changes are caused by #4918 which changed all the `STOP` calls to `CALL ABORT`. `ABORT` is *not* part of the Fortran Standard though many compilers support it via extensions. But, NAG is very strict and will not support extensions that have equivalents in the Fortran Standard. That equivalent is [`ERROR STOP`](https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2025-0/stop-and-error-stop.html). 

The main reason I'm hesitant here is I don't know how old of compilers you need to support. `ERROR STOP` is from Fortran 2008 and while every compiler I use is happy with it, I don't have to build things with GCC 4 or whatever. (The oldest compiler I can find is GCC 7 and it is happy with `ERROR STOP`.)

So, if you'd prefer to keep `CALL ABORT` (and thus not touch "working" code), I believe it should be fairly simple to make a routine that is only compiled when NAG Fortran is used that can provide an `ABORT` routine (that underneath calls `ERROR STOP`).

mathomp4:
Closes #5002 

As detailed in the above issue, this PR collects changes for the NAG compiler using CMake.

The changes are:

1. Adding a NAG bit to `cmake/fc.cmake`. The flags here are mainly from both `Makefile.system` and from https://github.com/Reference-LAPACK/lapack/pull/686 via @ACSimon33
2. Adding the NAG PIC flag to `cmake/system.cmake`
3. Converting the `ctest` `CALL ABORT` calls to Fortran Standard `ERROR STOP`
4. Various Fortran Standard issues in `lapack-netlib` (via https://github.com/Reference-LAPACK/lapack/pull/686 and https://github.com/Reference-LAPACK/lapack/pull/951, again thanks to @ACSimon33)
5. A weird CMake oddity in `cmake/lapack.cmake`

I'll go a bit more into that last one. Apparently, the line:
```cmake
    set_source_files_properties(${LA_SOURCES} PROPERTIES COMPILE_FLAGS "${LAPACK_FFLAGS}")
```
was duplicating Fortran flags when `make` was actually run. Now, for most compilers this probably isn't a bad thing, but NAG does not allow some flags to be duplicated:
```
[  0%] Building Fortran object CMakeFiles/LAPACK_OVERRIDES.dir/lapack-netlib/SRC/la_constants.f90.o
Option error: -ieee= option specified twice
Option error: -thread_safe option specified twice
```
To be extra safe, I only turned that off when building with NAG as I don't want to change behaviors for any other compiler.

I did run `ctest` with NAG and got 100% (see [runctest.NAG.log](https://github.com/user-attachments/files/18011951/runctest.NAG.log)) as well as GCC 14 with my changes. Hopefully the CI will catch more.

If you'd like me to run any other tests with NAG (in case `ctest` doesn't cover everything), let me know.

Also, as mentioned in #5002 , the use of `ERROR STOP` might break some very old compilers. If so, I believe I could revert back to `CALL ABORT` and then write an `ABORT` external call for NAG.


martin-frbg:
As far as I can tell, the redundant addition of LAPACK_FFLAGS dates back to the introduction of CMake support in version 0.2.15, a bit over nine years ago (though I cannot seem to find hpanderson's PR from back then)

ACSimon33:
Changes looking good to me (I'm not familiar with the OpenBLAS CMake system, though). Just a quick note on the support of `ABORT` in nagfor:

There are some non-standard intrinsic NAG modules that contain POSIX interfaces and, therefore, also support some C system functions (`abort`, `atexit`, ...). The module in question would be [`f90_unix_proc`](https://support.nag.com/nagware/np/r72_doc/manual/compiler_7_8.html#AUTOTOC_7_8). 

But changing `CALL ABORT` to `ERROR STOP` makes way more sense.

mathomp4:
> Changes looking good to me (I'm not familiar with the OpenBLAS CMake system, though). Just a quick note on the support of `ABORT` in nagfor:
> 
> There are some non-standard intrinsic NAG modules that contain POSIX interfaces and, therefore, also support some C system functions (`abort`, `atexit`, ...). The module in question would be [`f90_unix_proc`](https://support.nag.com/nagware/np/r72_doc/manual/compiler_7_8.html#AUTOTOC_7_8).
> 
> But changing `CALL ABORT` to `ERROR STOP` makes way more sense.

Huh. I have *never* heard of that `f90_unix_proc`. Thanks for letting me know! In most cases, I've just moved to Fortran Standard intrinsics if there was an equivalent, but I'm sure in the past I've had some where the behavior was different and I just...assumed that was the price!

mathomp4:
Hmm. Not sure about the failing checks. The qemu one seems to be Github hiccup, but the Windows one I'm not sure about...

martin-frbg:
All good, the Windows one is also unrelated - the free Azure CI uses a variable bunch of older hardware, so long-running build jobs tend to time out when they land on a particularly slow machine

HaoZeke:
Sorry I just saw the original ping, so I'm copying the crux of my comment there (https://github.com/OpenMathLib/OpenBLAS/pull/4918#issuecomment-2519436545), basically, I'm not sure the Fortran 2008 compliance requirement is OK for OpenBLAS, so I'd prefer documentation for `f90_unix_proc` instead.

martin-frbg:
Well, gfortran appears to support this language feature since about version 4.6, so I would expect this to turn into a fight of ancient vendor compilers if anything. My question was more along the lines of if you know a case where it is definitely unsupported. (I assume that the "proper error return" is important enough for the Meson environment that it cannot be solved with the previous kludge of grepping the test logs for a failure phrase)

mathomp4:
> Well, gfortran appears to support this language feature since about version 4.6, so I would expect this to turn into a fight of ancient vendor compilers if anything. My question was more along the lines of if you know a case where it is definitely unsupported. (I assume that the "proper error return" is important enough for the Meson environment that it cannot be solved with the previous kludge of grepping the test logs for a failure phrase)

If there is a compiler it isn't supported with, the best bet would probably then be to write our own "abort" routine (openblas_abort?) that calls `error stop` in most cases, but can call `abort` in the other(s)?

martin-frbg:
guess so, I understand this was your plan b in the first post ? the only drawback from my viewpoint would be that it makes the code deviate more from the Reference-LAPACK original - but as the actual tests are unaffected, that would not make it harder to import any future changes

mathomp4:
> guess so, I understand this was your plan b in the first post ? the only drawback from my viewpoint would be that it makes the code deviate more from the Reference-LAPACK original - but as the actual tests are unaffected, that would not make it harder to import any future changes

Well, I suppose we could work with Reference-LAPACK to update their tests with `error stop` or something else that raises a non-zero status?

martin-frbg:
certainly - it is just that there might be less motivation for it, if the change is mostly useful for Meson build support that they don't have

PR #5007, opened 2024-12-05
Revert the NRM2 kernels for NeoverseN2 and ARMV8SVE targets to the generic NEON version 

### LINKED ISSUE #5006: Regression in dgeqrf_ QR factorization since 0.3.21 - Output contains NaN values
This was discovered from a PyTorch unit test failure on aarch64 ( link will be attached )

Code to reproduce the problem is attached

[reproducer.zip](https://github.com/user-attachments/files/18023988/reproducer.zip)
   
See attached code for more details
   
```
   dgeqrf_(&m, &n, input_buffer, &lda, tau_buffer, workspace, &lwork, &info);

    for(int i = 0; i < TAU_STRIDE; i++){
        if(std::isnan(tau_buffer[i]))
            throw std::invalid_argument( "TAU contains NaN" );
    }
```

Under libopenblas/openblas 0.3.20 there is no NaN contained in the tau output, however in 0.3.21+ the output contains NaN values. It appears this is when the implementation was changed from Fortrain to C.

martin-frbg:
fixes #5006

### DIFF
--- a/kernel/arm64/KERNEL.ARMV8SVE
+++ b/kernel/arm64/KERNEL.ARMV8SVE
@@ -104,10 +104,10 @@ IDAMAXKERNEL   = iamax_thunderx2t99.c
 ICAMAXKERNEL   = izamax_thunderx2t99.c
 IZAMAXKERNEL   = izamax_thunderx2t99.c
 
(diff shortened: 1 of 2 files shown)


Mousius (approved):
Thanks for the quick fix @martin-frbg 😸

Issue #4029, opened 2023-04-30
Building Error of OpenBLAS C910V Target 


luyanaa:
When building OpenBLAS with T-head 2.6.1 toolchain, the following multiples errors are found. 

```
error: incompatible types when assigning to type 'vfloat32m4_t' from type 'int'
   35 | #define VLSEV_FLOAT vlse_v_f32m4
      |                     ^~~~~~~~~~~~
note: in expansion of macro 'VLSEV_FLOAT'
  130 |                         vy = VLSEV_FLOAT(&y[iy], stride_y, gvl);
      |                              ^~~~~~~~~~~

error: incompatible types when assigning to type 'vfloat32m4_t' from type 'int'
   34 | #define VLEV_FLOAT vle_v_f32m4
      |                    ^~~~~~~~~~~
note: in expansion of macro 'VLEV_FLOAT'
  145 |                         vx = VLEV_FLOAT(&x[j], gvl);
```

Refer to https://github.com/xianyi/OpenBLAS/pull/4024 for the build CI to test the problem.

martin-frbg:
can you please try the riscv branch instead of develop ? unfortunately the rvv extension and intrinsics are still a bit unstable, the develop branch still uses the older rvv-0.7.1 syntax as far as I know, which may no longer be supported by newer toolchains

RevySR:
in risc-v branch

../kernel/riscv64/max_vector.c:62:31: error: incompatible types when assigning to type 'vfloat64m1_t' from type 'int'
   62 | #define VFREDMAXVS_FLOAT JOIN(__riscv_vfredmax_vs_f,  ELEN,   LMUL,   _f, JOIN2( ELEN,   m1))

../kernel/riscv64/max_vector.c:61:30: error: incompatible types when assigning to type 'vfloat64m8_t' from type 'int'
   61 | #define VLSEV_FLOAT     JOIN(__riscv_vlse,      ELEN,   _v_f,   ELEN,   LMUL)

../kernel/riscv64/max_vector.c:65:30: error: incompatible types when assigning to type 'vfloat64m8_t' from type 'int'
   65 | #define VFMVVF_FLOAT    JOIN(__riscv_vfmv,      _v_f_f, ELEN,   LMUL,   _)

martin-frbg:
Too bad - there are currently no open PRs for RISCV issues, and I cannot suggest with any certainty which toolchain is currently expected to work but please see #3808

martin-frbg:
[...]

PR #5010, opened 2024-12-06
[WIP]Include riscv v0p10 header when building C910V with T-Head toolchain

martin-frbg:
from a comment by zejianzhang1982 on closed issue #4029


martin-frbg:
Apparently the presence of this header depends on the version of the T-head toolchain, or possibly even the Linux kernel version on the build host. Trying to clarify this...

PR #5022, opened 2024-12-18
Replace the __riscv_vid_v_i32m2 and __riscv_vid_v_i64m2 with __riscv…_vid_v_u32m2 and __riscv_vid_v_u64m2 for riscv64-unknown-linux-gnu-gcc compiling.

tingboliao:
Since the functions __riscv_vid_v_i32m2 and __riscv_vid_v_i64m2, which are not standard RVV-Intrinsic API, fail to be compiled by the riscv64-unknown-linux-gnu-gcc compiler, replacing them with __riscv_vid_v_u32m2 and __riscv_vid_v_u64m2 turns out to be an effective solution to fix the bug.


martin-frbg:
Thanks - I guess these were a remnant from one of the earlier drafts or vendor extensions. It will probably make sense to change the associated data types to unsigned int as well, but I can take care of this separately

PR #5031, opened 2024-12-24
Fix invalid link to FAQ

david-cortes:
Current link it offers is this:
```
https://github.com/OpenMathLib/OpenBLAS/blob/develop/www.openmathlib.org/OpenBLAS/docs/faq
```

Needs the http part to work.


martin-frbg:
Oops, thanks.

PR #5054, opened 2025-01-07
Update Makefile.riscv64, remove --fast-math from default build options on x280

fleclairTT:
The default flags for various targets should assume compliance with IEEE-754. fast math and co should be opt in only

### DIFF
--- a/Makefile.riscv64
+++ b/Makefile.riscv64
@@ -3,7 +3,7 @@ CCOMMON_OPT += -march=rv64imafdcv0p7_zfh_xtheadc -mabi=lp64d -mtune=c920
 FCOMMON_OPT += -march=rv64imafdcv0p7_zfh_xtheadc -mabi=lp64d -mtune=c920 -static
 endif
 ifeq ($(CORE), x280)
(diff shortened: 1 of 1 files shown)


martin-frbg:
I had wondered about this one too - IIRC the -ffast-math was in the initial code drop on the riscv branch, but whether it was there to paper over some compiler issue or just out of habit of the author I do not know @HellerZheng

martin-frbg (commented):
wait, that won't work without all the other infrastructure changes for a new target - KERNEL file, param.h entries for GEMM block sizes etc. Better make that a separate PR

fleclairTT:
Yeah, the second PR wasn't supposed to be tracked here, apologies

HellerZheng:
> I had wondered about this one too - IIRC the -ffast-math was in the initial code drop on the riscv branch, but whether it was there to paper over some compiler issue or just out of habit of the author I do not know @HellerZheng
Sure, please go ahead.

fleclairTT:
resolved by https://github.com/OpenMathLib/OpenBLAS/pull/5055

PR #5055, opened 2025-01-07
remove fastmath from x280 target

fleclairTT:
Fast math typically disallows IEEE 754 compliance and is considered an opt in only feature. removing it from default stops it from being opt out to being opt-in


martin-frbg:
(from additional information received off-site, this change actually fixes incorrect results)

PR #5069, opened 2025-01-13
Further rearranged the rotm kernel for the different architectures.

tingboliao:


martin-frbg:
I think it should be sufficient to have the declarations of SROTMKERNEL and DROTMKERNEL in the plain "KERNEL" files under each architecture, no need to repeat them in the specific files (except possibly the KERNEL.generic) as long as they are the same. Also, could you please add a line for the "quad precision" version in common_level.1 as
` int    qrotm_k (BLASLONG, xdouble, BLASLONG, xdouble, BLASLONG, xdouble);` and similarly `int (*qrotm_k...` in common_param.h,
`#define ROTM_K                  QROTM_K` in common_macro.h, 
```
ifndef QROTMKERNEL
QROTKERNEL = rotm.S
endif
```
in kernel/Makefile.L1 and 
```
ifndef QROTMKERNEL
QROTMKERNEL  = ../generic/rotm.c
endif
```
in kernel/x86_64/KERNEL ? The "quad precision" (EXPRECISION=1) kernels (exclusively for x86_64 are probably not fully functional currently (and may never have been), but that would fix compilation problems arising from your PR.
Also please add srotm_kTS, drotm_kTS or qrotm_kTS as appropriate everywhere you see the corresponding srot_kTS etc in kernel/setparam-ref.c to make the DYNAMIC_ARCH builds work.

tingboliao:
> I think it should be sufficient to have the declarations of SROTMKERNEL and DROTMKERNEL in the plain "KERNEL" files under each architecture, no need to repeat them in the specific files (except possibly the KERNEL.generic) as long as they are the same. Also, could you please add a line for the "quad precision" version in common_level.1 as ` int qrotm_k (BLASLONG, xdouble, BLASLONG, xdouble, BLASLONG, xdouble);` and similarly `int (*qrotm_k...` in common_param.h, `#define ROTM_K QROTM_K` in common_macro.h,
> 
> ```
> ifndef QROTMKERNEL
> QROTKERNEL = rotm.S
> endif
> ```
> 
> in kernel/Makefile.L1 and
> 
> ```
> ifndef QROTMKERNEL
> QROTMKERNEL  = ../generic/rotm.c
> endif
> ```
> 
> in kernel/x86_64/KERNEL ? The "quad precision" (EXPRECISION=1) kernels (exclusively for x86_64 are probably not fully functional currently (and may never have been), but that would fix compilation problems arising from your PR. Also please add srotm_kTS, drotm_kTS or qrotm_kTS as appropriate everywhere you see the corresponding srot_kTS etc in kernel/setparam-ref.c to make the DYNAMIC_ARCH builds work.

Hi, @martin-frbg 
      I've made the corrections according to the suggestions you provided. However, it seems that there are still some failures for some architectures.

PR #5070, opened 2025-01-13
LoongArch64: Fix dsymv and ssymv LASX version

xry111:
"fmov.d $f2, $f4" leaves all the bits higher than the 63-th bit unpredictable but it's obvious that the following code uses the value of those high bits.  We actually want to replicate the lower 64 bits here, so we should use xvreplve0.d instead.

LA464 (Loongson 3[A-Z]-5000) happens to replicate them for us due to some uarch internal details so the issue was not detected, but for LA664 (Loongson 3[A-Z]-6000) and future uarch we need to do things correctly or we end up getting a lot of test failures.

Closes: https://bbs.aosc.io/t/topic/302

### DIFF
--- a/kernel/loongarch64/dsymv_L_lasx.S
+++ b/kernel/loongarch64/dsymv_L_lasx.S
@@ -288,7 +288,7 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
     //Acc U2
     GACC   xvf, d, U4, U2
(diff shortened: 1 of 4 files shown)


XiWeiGu:
Yes, I have introduced a bug. We should broadcast the accumulated result instead of simply moving it.

azuresky01:
I confirm that your changes do fix the problem. The compilation error disappears.

martin-frbg:
Thank you for the quick fix.

PR #5071, opened 2025-01-13
Add thread throttling profile for SGEMM on NEOVERSEV1

annop-w:
This PR introduces a new thread throttling profile for SGEMM on NEOVERSEV1 CPU core.
Benchmarking for M=N=K=[2, 1024] shows a geometric mean speedup of ~1.1x.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -232,3 +232,6 @@ In chronological order:
 
 * Aniket P. Garade <https://github.com/garadeaniket>   Sushil Pratap Singh <https://github.com/SushilPratap04>  Juliya James <https://github.com/Juliya32> 
   *  [2024-12-13] Optimized swap and rot  Level-1 BLAS routines with ARM SVE  
+
+* Annop Wongwathanarat <annop.wongwathanarat@arm.com>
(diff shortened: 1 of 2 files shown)


annop-w:
@martin-frbg Could I please get a review for this patch ?

martin-frbg:
As soon as I can concentrate on it. Would you happen to have any benchmark numbers available ? The idea of using varying thread counts has been floated a few times in the past, but the added logic seemed to outweigh gains on x86_64 at least

annop-w:
@martin-frbg Thank you for your comment. As mentioned in the description, the geo. mean speedup is ~1.1 for M=N=K up to 1024. Here is a figure that shows the data on NEOVERSEV1 with 64 cores.
sgemm_v1

martin-frbg:
Hmm, I don't really like the added complexity with yet another "hidden" table of parameters (I assume the V2 would need different thresholds), but what's not to like about the added performance...

annop-w:
Yes, V2 would need another set of parameters. I think the complexity is inevitable if we want optimal performance. We could try to somehow organize these tuning parameters, e.g. into files/folders. I am open to suggestions.

martin-frbg:
Merging to give it more exposure, but I think ideally the table part should be in (or closer to) param.h

PR #5086, opened 2025-01-22
Add comments to the functions in the BFLOAT16 and INT8 extensions

moluopro:
Add comments to the `sbgemv`, `sbgemm`, and `sbgemm_batch` functions in the BFLOAT16 and INT8 extensions.


martin-frbg:
Not sure if these are really necessary, also what INT8 extensions do you refer to here ?

moluopro:
This is a practice PR for a newcomer, so I have only modified the doc or comments. If this is unnecessary, please close the PR directly. As for `INT8 extensions`, it is just the classification of these functions: `BFLOAT16 and INT8 extensions`.  

image

martin-frbg:
Ah I see, thanks, no need for practice PRs here (if we had such a rule, maybe I would still be doing practice PRs myself...).
_If_ you like to write documentation, it occurs to me that something quite similar to your PR might be useful in the documentation under docs/ - a page listing the BLAS functions with a one-line description of each. So far we have always been telling users to "look in the documentation of the Reference BLAS, or of MKL and similar", but now that we have moved the documentation out of the github wiki, maybe we could build something for local/offline use.

PR #5089, opened 2025-01-22
Simplify gemv_t_sve_v1x3 kernel

annop-w:
This patch slightly simplifies gemv_t_sve_v1x3 kernel by removing calculation of predicates in the inner loop which can be precomputed instead. This improves performance for small M and N. 
gemv_t_sve_v1x3


martin-frbg:
Jenkins CI failure (IBM hosts) appears to be caused by a hardware or configuration problem at OSUOSL

PR #5097, opened 2025-01-24
Fix Windows on ARM build instructions

matthew-brett:
The command as merged used the compiler target as the compiler path.

I have run and tested a build with this command.

@Mugundanmcw - is this correct?

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -480,13 +480,13 @@ the LLVM toolchain enables native compilation of the Fortran sources of LAPACK a
 
 4. Navigate to the OpenBLAS source code directory and start building OpenBLAS
    by invoking Ninja:
-   
+
        ```cmd
        cd OpenBLAS
        mkdir build
        cd build
-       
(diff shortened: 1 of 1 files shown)


MugundanMCW:
> The command as merged used the compiler target as the compiler path.
> 
> I have run and tested a build with this command.
> 
> @Mugundanmcw - is this correct?

Yes, it's true that the command as merged, uses the compiler target as the compiler path. Adding "_TARGET" to specify the compiler target would be the correct fix else would leads to error.

Thanks

martin-frbg:
Can either of you confirm that the cmake_c_compiler_target option is actually required in this context ? In my experience, it seems to compile just fine without, which seems to agree with a comment in the cmake documentation that suggests that clang should be able to figure it out on its own.

matthew-brett:
I can confirm the command runs without error without the `_TARGET` defines, and the tests pass, on a native WoA system. 

```
dumpbin /headers lib\Release\openblas.lib | findstr /i machine
```

confirms these are AA64 binaries.

MugundanMCW:
> I can confirm the command run

Yes, it's true that the CMake setup can automatically detect clang-cl on a native WoA system, even without the target flag.

martin-frbg:
Thanks for confirming this. So, do we really want to present redundant options that are likely to be misspelled and that will make the compilation look less straightforward than it actually is on WoA ?

MugundanMCW:
@martin-frbg yes, the following command could be sufficient to compile OpenBLAS for WoA without any errors:
`cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Release -DTARGET=ARMV8 -DBINARY=64 -DCMAKE_C_COMPILER=clang-cl -DCMAKE_Fortran_COMPILER=flang-new`

Thanks!

Issue #5106, opened 2025-02-02
signedness mismatch in pointer to integer type passed to blas_lock()


jhgit:
blas_lock() takes a BLASULONG pointer, but level3_thread.c passed a BLASLONG pointer.  clang warns about this:

    ./level3_thread.c:565:17: warning: passing 'volatile BLASLONG *' (aka 'volatile long *') to parameter of type 'volatile BLASULONG *' (aka 'vola\
    tile unsigned long *') converts between pointers to integer types with different sign [-Wpointer-sign]
      565 |     blas_unlock(&init_lock);
          |                 ^~~~~~~~~~
    ../../common.h:737:54: note: passing argument to parameter 'address' here
      737 | static __inline void blas_unlock(volatile BLASULONG *address){
          |                                                      ^

martin-frbg:
closing as fixed by your #5107

Issue #5104, opened 2025-01-31
[regression] 0.3.29 test failure on ARMV6


cdluminate:
openblas `0.3.29` fails to build on Debian's armhf machine: https://buildd.debian.org/status/fetch.php?pkg=openblas&arch=armhf&ver=0.3.29%2Bds-1&stamp=1738175608&raw=0

It seems to be a regression in the `cblas_stbmv` function:
```
 ******* FATAL ERROR - COMPUTED RESULT IS LESS THAN HALF ACCURATE *******
           EXPECTED RESULT   COMPUTED RESULT
       1      0.157343          0.127373    
       2       0.00000           0.00000    
       3      0.447053          0.447053    
       4      0.277223          0.277223    
       5      0.806693          0.806693    
 ******* cblas_stbmv  FAILED ON CALL NUMBER:
    580: cblas_stbmv (    CblasUpper,  CblasNoTrans,     CblasUnit,
            5,  0, A,  2, X,-1) .

 ******* FATAL ERROR - TESTS ABANDONED *******
ERROR STOP 
```

Debian uses a modified source tree. But I can reproduce this problem with the upstream git repo:
```
TARGET=ARMV6 make
```

The error message reads:
```
 ******* FATAL ERROR - COMPUTED RESULT IS LESS THAN HALF ACCURATE *******                                                                                                                                            
           EXPECTED RESULT   COMPUTED RESULT                                                                                                                                                                         
       1      0.245164          0.245164                                                                                                                                                                             
       2       0.00000           0.00000                                                                                                                                                                             
       3      0.539071          0.539071                                                                                                                                                                             
       4       1.22516           1.22516                                                                                                                                                                             
       5      0.601097E-01      0.374625E-01                                                                                                                                                                         
 ******* cblas_stbmv  FAILED ON CALL NUMBER:                                                                                                                                                                         
    582: cblas_stbmv (    CblasUpper,  CblasNoTrans,  CblasNonUnit,                                                                                                                                                  
            5,  0, A,  2, X, 1) .                                                                                                                                                                                    
 ******* cblas_stbmv  FAILED ON CALL NUMBER:                                                                                                                                                                         
      2: cblas_stbmv (    CblasUpper,  CblasNoTrans,     CblasUnit,                                                                                                                                                  
            1,  0, A,  2, X, 1) .

 ******* FATAL ERROR - TESTS ABANDONED *******
ERROR STOP 
```

[...]

PR #5107, opened 2025-02-02
fix signedness of pointer to integer type passed to blas_lock()

jhgit:
fixes issue #5106


martin-frbg:
Thanks - I had noticed this already, but I do not think it would cause anything worse than the annoying log spam seen e.g. in #5104

jhgit:
> Thanks - I had noticed this already, but I do not think it would cause anything worse than the annoying log spam seen e.g. in #5104

I was pretty sure that was true as well, although I didn't dig into the various blas_lock implementations (assembly code) to confirm.  Thank you for merging in the change.

PR #5125, opened 2025-02-12
Fix SGEMV on POWER8 by reverting to the non-vectorized earlier code

### LINKED ISSUE #5122: test_sbgemm fails with FATAL ERROR SBGEMV - Return code: 16862 on powerpc64le
OpenBLAS 0.3.29, FreeBSD 14.2-RELEASE, powerpc64le

```
OPENBLAS_NUM_THREADS=1 OMP_NUM_THREADS=1 ./test_sbgemm > SBBLAT3.SUMM
FATAL ERROR SBGEMV - Return code: 16862
```

Everything worked fine on OpenBLAS 0.3.28.

martin-frbg:
reverts part of PR #4880 as already discussed there earlier, fixes #5122

### DIFF
--- a/kernel/power/sgemv_t.c
+++ b/kernel/power/sgemv_t.c
@@ -78,7 +78,17 @@ static void sgemv_kernel_4x8(BLASLONG n, BLASLONG lda, FLOAT *ap, FLOAT *x, FLOA
             temp7 += v_x[i] * va7[i]; 
         }
     
-  
+ #if defined(POWER8)
+    y[0] += alpha * (temp0[0] + temp0[1]+temp0[2] + temp0[3]);
+    y[1] += alpha * (temp1[0] + temp1[1]+temp1[2] + temp1[3]);
+    y[2] += alpha * (temp2[0] + temp2[1]+temp2[2] + temp2[3]);
+    y[3] += alpha * (temp3[0] + temp3[1]+temp3[2] + temp3[3]);
+
+    y[4] += alpha * (temp4[0] + temp4[1]+temp4[2] + temp4[3]);
+    y[5] += alpha * (temp5[0] + temp5[1]+temp5[2] + temp5[3]);
+    y[6] += alpha * (temp6[0] + temp6[1]+temp6[2] + temp6[3]);
+    y[7] += alpha * (temp7[0] + temp7[1]+temp7[2] + temp7[3]);
+ #else
     register __vector float t0, t1, t2, t3;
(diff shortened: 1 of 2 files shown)


ChipKerchner:
This was probably an endianness issue but I'm OK with the fix.

martin-frbg:
We can always improve on this later, but it would seem pointless to leave broken code in place for the next release. Curious how the exact same code works on POWER9&10 ppc64le but not on POWER8 ppc64le - does the POWER8 implement the vecmerge(l/h) differently ?
(As an aside, the 8x8 versions of the sgemv kernel are currently unused - looks like Abdel wrote them "for future use" when only POWER8 was available)

ChipKerchner:
It requires some investigation of what exactly is the cause.  vec_mergeh/l should be implemented the same on P8, P9 or P10.

martin-frbg:
Compiler options are identical except for the `-mcpu` and `-mtune` part. I can even compile for either of the three targets on P10 (OSUOSL's cfarm120 in the GCC Compile Farm) and see only the TARGET=POWER8 build fail (without this PR)

ChipKerchner:
Looks like the failure is related to inc_x and/or inc_y not being equal to one.  It seems P8 isn't handling that correctly for SBGEMV without your latest fix.  I know I updated the SBGEMV unit test to test for inc_x/y = 2 last year.  I made sure it worked for P9 & P10 when I improved the SBGEMV for those architecture.

ChipKerchner:
I think your change will fix that failure for P8 but I see it will probably fail for P7 and earlier.  Instead of

```
#if defined(POWER8)
```

it should be

```
#ifndef _ARCH_PWR9
```

martin-frbg:
P7 and earlier do not use these kernels, they fall back to the older gemv_(n/t).S
can't check right now but I think SBGEMV was only affected through the SGEMV comparison breaking - there were actual BLAS2 test failure (the "LESS THAN HALF ACCURATE") in SGEMV and related on P8 without this PR

ChipKerchner:
Ok, after digging into it a bit, it looks like an epsilon difference for P8.  Still trying to figure out why but as I said before, your PR is probably the best approach.

martin-frbg:
Maybe gcc doing something wild when `-mcpu=power8` ? But I do not have the energy right now to compare the generated assembly

xry111:
FTR https://gcc.gnu.org/PR119234 (**not** a valid GCC bug report) has root caused the issue.

PR #5128, opened 2025-02-12
Add -O2 to flang flags when building on WoA in Release mode

### LINKED ISSUE #5120: Enable Optimizations Flags for Fortran Targets on WoA
I noticed that the syev function in OpenBLAS on Windows on ARM (WoA) was not performing optimally because the flang-new compiler does not enable optimization flags by default. To improve performance, I added the -O2 optimization flag during the build process: 

 
`-DCMAKE_Fortran_FLAGS_RELEASE="-O2" `
 

After applying this change, the syev function showed a huge performance improvement. Can we enable this flag by default for Fortran builds on WoA to ensure better optimization.

Below is the benchmark data before and after applying the <strong>-O2</strong> optimization flag (<strong>time in milliseconds</strong>):
Function Name | Before (ms) | After (ms) | Improvement (%)
-- | -- | -- | --
test_syev[200-d] | 55.6886 | 6.4854 | 88%
test_syev[200-s] | 45.8937 | 4.4355 | 90%
test_syev[50-d] | 1.0646 | 0.1641 | 85%
test_syev[50-s] | 0.8456 | 0.1253 | 85%

martin-frbg:
fixes #5120

### DIFF
--- a/cmake/system.cmake
+++ b/cmake/system.cmake
@@ -628,6 +628,18 @@ set(CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS} ${CCOMMON_OPT}")
 endif()
 # TODO: not sure what PFLAGS is -hpa
 set(PFLAGS "${PFLAGS} ${CCOMMON_OPT} -I${TOPDIR} -DPROFILE ${COMMON_PROF}")
+if ("${CMAKE_BUILD_TYPE}" STREQUAL "Release")
+
+if ("${F_COMPILER}" STREQUAL "FLANG")
+if (${CMAKE_Fortran_COMPILER_VERSION} VERSION_LESS_EQUAL 3)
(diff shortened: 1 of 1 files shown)


rgommers (approved):
LGTM. I can't test locally, but should work. I considered whether `-O2` should work also with clang-cl (rather than `/O2` style flag), and https://flang.llvm.org/docs/FlangCommandLineReference.html#optimization-level says it's always `-O2`.

martin-frbg:
> LGTM. I can't test locally, but should work. I considered whether `-O2` should work also with clang-cl (rather than `/O2` style flag), and https://flang.llvm.org/docs/FlangCommandLineReference.html#optimization-level says it's always `-O2`.

Thanks - I do have the hardware to test, and in fact it was one of the Qualcomm/WoA-associated engineers who raised the issue (mcw=Multicoreware Inc.)

PR #5108, opened 2025-02-05
Add SBGEMM for arm neoversev1

taoye9:
This PR is to add an optimised bf16 gemm kernel for arm neoversev1 machine (sve-256 bit).

PR #5129, opened 2025-02-13
Fix Numeric Error in SBGEMM Kernel for `NEOVERSEV2` when DYNAMIC_ARCH=1

taoye9:
In our previous PR ([#5108](https://github.com/OpenMathLib/OpenBLAS/pull/5108)), we introduced an SBGEMM kernel optimized for NEOVERSEV1 (SVE-256 bit). However, when running on NEOVERSEV2 (SVE-128 bit) under the dynamic architecture setting, the SBGEMM kernel incorrectly falls back to the NEOVERSEV1 implementation due to [this line of code](https://github.com/OpenMathLib/OpenBLAS/blob/develop/driver/others/dynamic_arm64.c#L153). This fallback behavior leads to numeric errors.

This PR addresses the issue by explicitly adding NEOVERSEV2 to DYNAMIC_ARCH, ensuring that the correct SBGEMM kernel is selected for NEOVERSEV2 instead of defaulting to NEOVERSEV1.


martin-frbg:
Ah, did not think of that in time - we'd need to make the V2 a full-fledged TARGET  for this to work (parameter assignment and all) which we had initially wanted to avoid in order to keep the total number of DYNAMIC_ARCH configurations (and library size) small

Issue #4921, opened 2024-10-06
Very poor `gemmt` performance compared to `gemm` and `syrk`


david-cortes:
I'm running some timings on operation `t(X)*X` on row-major matrices having many more rows than columns.

I'm finding that for these types of inputs, function `gemmt` is much slower than the equivalent from `syrk` or `gemm`, with a very wide margin.

Timings in milliseconds for input size 1,000,000 x 32, intel i12700H, average of 3 runs:
* `gemmt`: 216.178
* `syrk`: 41.0468
* `gemm`: 39.55553

Version: OpenBLAS 0.3.28, built with OpenMP, compiled from source (gcc with cmake system). Same issue happen with pthreads, and same timing difference is observed when running single-threaded.

For reference, timings for other libraries:
* MKL `gemmt`: 25.66533
* MKL `syrk`: 12.57197
* MKL `gemm`: 15.69447
* `tabmat`'s "sandwich" op: 29.3

Code to reproduce:

<details>

```cpp
#include <iostream>
#include <chrono>
#include <random>
#include <memory>


#include <cblas.h>
extern "C" void cblas_dgemmt(const CBLAS_LAYOUT Layout, const CBLAS_UPLO uplo, const CBLAS_TRANSPOSE transa, const CBLAS_TRANSPOSE transb, const int n, const int k, const double alpha, const double *a, const int lda, const double *b, const int ldb, const double beta, double *c, const int ldc);

using std::chrono::high_resolution_clock;
using std::chrono::duration_cast;
using std::chrono::duration;
using std::chrono::milliseconds;

int main()
{
    const size_t nrows = 1'000'000;
    const size_t ncols = 32;
    const size_t tot = nrows * ncols;

    std::mt19937 rng{123};
    std::normal_distribution norm_distr{0.0, 1.0};

    std::unique_ptr<double[]> X(new double[tot]);
    std::unique_ptr<double[]> out(new double[ncols*ncols]());
    for (size_t ix = 0; ix < tot; ix++) X[ix] = norm_distr(rng);

    auto t1 = high_resolution_clock::now();
    cblas_dgemmt(
        CblasRowMajor, CblasUpper, CblasTrans, CblasNoTrans,
        ncols, nrows,
        1., X.get(), ncols,
        X.get(), ncols,
        0., out.get(), ncols
    );
    auto t2 = high_resolution_clock::now();
    duration<double, std::milli> ms_double = t2 - t1;

    double sum_res = 0.;
[...]

PR #5143, opened 2025-02-20
Fix GEMMT transforming the input array B in some complex cases

### LINKED ISSUE #5111: C/ZGEMMT changes the inputs in the UNC case
The Z/CGEMMT routine changes the second input when calling 
```
cgemmt_("U", "N", "C", &n, &k, &alpha, A, &lda, B, &ldb, &beta, C, &ldc);
```
or 
```
zgemmt_("U", "N", "C", &n, &k, &alpha, A, &lda, B, &ldb, &beta, C, &ldc);
```

After the call B is its conjugate. 

In the `ZGEMMT` case this leads:
```
Matrix A:
(1.000000, 2.000000) (2.000000, 3.000000) (3.000000, 4.000000) 
(4.000000, 5.000000) (5.000000, 6.000000) (6.000000, 7.000000) 
(7.000000, 8.000000) (8.000000, 9.000000) (9.000000, 10.000000) 
Matrix B:
(1.000000, 1.000000) (2.000000, 2.000000) (3.000000, 3.000000) 
(4.000000, 4.000000) (5.000000, 5.000000) (6.000000, 6.000000) 
(7.000000, 7.000000) (8.000000, 8.000000) (9.000000, 9.000000) 
Matrix C:
(144.000000, 12.000000) (0.000000, 0.000000) (0.000000, 0.000000) 
(171.000000, 15.000000) (201.000000, 15.000000) (0.000000, 0.000000) 
(198.000000, 18.000000) (234.000000, 18.000000) (270.000000, 18.000000) 
After cgemmtMatrix A:
(1.000000, 2.000000) (2.000000, 3.000000) (3.000000, 4.000000) 
(4.000000, 5.000000) (5.000000, 6.000000) (6.000000, 7.000000) 
(7.000000, 8.000000) (8.000000, 9.000000) (9.000000, 10.000000) 
Matrix B:
(1.000000, -1.000000) (2.000000, -2.000000) (3.000000, -3.000000) 
(4.000000, -4.000000) (5.000000, -5.000000) (6.000000, -6.000000) 
(7.000000, -7.000000) (8.000000, -8.000000) (9.000000, -9.000000) 
```

[cgemmt_example.c.txt](https://github.com/user-attachments/files/18691046/cgemmt_example.c.txt)

[zgemmt_example.c.txt](https://github.com/user-attachments/files/18691063/zgemmt_example.c.txt)

martin-frbg:
fixes #5111

### DIFF
--- a/interface/gemmt.c
+++ b/interface/gemmt.c
@@ -688,5 +688,19 @@ void CNAME(enum CBLAS_ORDER order, enum CBLAS_UPLO Uplo,
 
 	IDEBUG_END;
 
+/* transform B back if necessary */
+#if defined(COMPLEX)
+	if (transb > 1){
+#ifndef CBLAS
+		IMATCOPY_K_CNC(nrowb, ncolb, (FLOAT)(1.0), (FLOAT)(0.0), b, ldb);
+#else
+		if (order == CblasColMajor)
+			IMATCOPY_K_CNC(nrowb, ncolb, (FLOAT)(1.0), (FLOAT)(0.0), b, ldb);
+		if (order == CblasRowMajor)
+			IMATCOPY_K_RNC(nrowb, ncolb, (FLOAT)(1.0), (FLOAT)(0.0), b, ldb);
+#endif
+	}
+#endif
+
 	return;
 }


interface/gemmt.c
```diff
@@ -688,5 +688,19 @@ void CNAME(enum CBLAS_ORDER order, enum CBLAS_UPLO Uplo,
 
 	IDEBUG_END;
 
+/* transform B back if necessary */
```
conradsnicta, line 691: @martin-frbg Not sure if I'm interpreting the overall code correctly, but shouldn't B not be written to at all?

While GEMMT is not part of standard BLAS and only exists as an "extension", it does have a definition within MKL, where B is marked as "const".  In other words, writing to B should never happen within GEMMT (ie. no in-place operations on B).

https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2025-0/cblas-gemmt.html

martin-frbg: Technically the promise is only that the inputs will be unchanged on exit, IMHO what happens behind closed doors is implementation-defined. And the current implementation is inefficient and needs to be redone eventually. It was only a quick, GEMV-based hack to provide _any_ support for GEMMT at all when the Reference implementation did not yet have it.

conradsnicta: Agree that it's better to have something rather than nothing.  Looking purely at the implementation side, writing to B would still be ill-advised as it's promised as "const" by MKL.  One possible case is that B is shared with another thread (say in a non-OpenBLAS function), which expects it unchanged.  There may also be a possible performance degradation due to cache pollution stemming from unnecessary writes to B.

martin-frbg: As I say the entire implementation is an ugly hack and needs to be replaced - but that is another can of worms, namely #4921, and I can only do so much in a day (and sometimes not even that)

conradsnicta: https://xkcd.com/2347/

PR #5144, opened 2025-02-21
dispatch NEOVERSEV2 to NEOVERSEN2 under dynamic setting

taoye9:
In our previous PR (https://github.com/OpenMathLib/OpenBLAS/pull/5108), we introduced an SBGEMM kernel optimized for NEOVERSEV1 (SVE-256 bit). However, when running on NEOVERSEV2 (SVE-128 bit) under the dynamic architecture setting, the SBGEMM kernel incorrectly falls back to the NEOVERSEV1 implementation due to [this line of code](https://github.com/OpenMathLib/OpenBLAS/blob/develop/driver/others/dynamic_arm64.c#L153). This fallback behavior leads to numeric errors.

This PR is to force SBGEMM kernel on NEOVERSEV2 falls to NEOVERSEN2 SBGEMM instead of NEOVERSEV1 SBGEMM.

### DIFF
--- a/driver/others/dynamic_arm64.c
+++ b/driver/others/dynamic_arm64.c
@@ -150,7 +150,7 @@ extern gotoblas_t  gotoblas_A64FX;
 #endif
 extern gotoblas_t  gotoblas_THUNDERX3T110;
 #endif
-#define gotoblas_NEOVERSEV2 gotoblas_NEOVERSEV1
+#define gotoblas_NEOVERSEV2 gotoblas_NEOVERSEN2
 
 extern void openblas_warning(int verbose, const char * msg);
 #define FALLBACK_VERBOSE 1


taoye9:
hi, @martin-frbg. seems make NEOVERSEV2 a separate requires many changes. we are planning to dispatch it to kernels of  NEOVERSEN2. 

https://github.com/OpenMathLib/OpenBLAS/pull/5129

martin-frbg:
Right, we can certainly merge this one first to resolve the SBGEMM error, and maybe still rework the V2 support after that - either in time for 0.3.30 or shortly after that. Your colleague @Mousius had originally floated the idea of a runtime rescaling of ARMV8SVE parameters instead of adding NEOVERSEV2 and potentially more cpu targets, but that did not turn out that simple either and I pretty much dropped the ball at some point. Conceptually, adding the target should not be that complicated (and we could restructure it later on in favor of Mousius' scheme if we can make that work), less sure if I can manage it this week though.

taoye9:
Hi, martin, thanks help with this patch. and one of my colleagues is working on scaling my previous sbgemm kernel to NEOVERSEV2.

PR #5145, opened 2025-02-21
Run CI on Github-hosted Arm instances too

Rohanjames1997:
Extend the CI to run on Github-hosted Arm instances that are now [available](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/).

Note that running on `ubuntu-24.04-arm` currently fails with `E: Unable to locate package libtinfo5`.

### DIFF
--- a/.github/workflows/arm64_graviton.yml
+++ b/.github/workflows/arm64_graviton.yml
@@ -20,13 +20,15 @@ permissions:
 jobs:
   build:
     if: "github.repository == 'OpenMathLib/OpenBLAS'"
-    runs-on: "cirun-aws-runner-graviton--${{ github.run_id }}"
 
     strategy:
       fail-fast: false
       matrix:
         fortran: [gfortran]
         build: [cmake, make]
+        os: ["cirun-aws-runner-graviton--${{ github.run_id }}", ubuntu-22.04-arm]
+    
+    runs-on: ${{ matrix.os }}
 
     steps:
       - name: Checkout repository


rgommers:
Thanks @Rohanjames1997. Could you please elaborate on what this tests that wasn't already tested? I suspect that this overlaps with the Cirrus CI NeoverseN1 jobs.

Rohanjames1997:
Thanks @rgommers. I actually didn't know of the Cirrus CI jobs, I just saw them 😄
I thought all the CI scripts were on Github Actions.
 
And you're right, while the `ubuntu-22.04-arm` runner is a NeoverseN2 CPU, this would still overlap with the Cirrus NeoverseN1 job or the one in `dynamic_arch.yml`

I guess I can close this if it doesn't add any additional testing.

martin-frbg:
There is already an Arm-on-gh runner (added to the bottom of the dynamic_arch job by #5080 right when these runners became available), but it is currently executing on Neoverse N1 (exactly like the pre-existing Cirrus jobs) rather than N2 as gh had to pull back the runners from the new Cobalt-100 hardware to generic dpdsv5 instances due to software issues - see https://github.com/orgs/community/discussions/148648#discussioncomment-12205917

martin-frbg:
Long- term it would probably make sense to move the comparable Arm64 jobs from Cirrus to gh as we tend to run out of free minutes there towards the end of the month. But this needs to wait until the end of the current "public beta" when it becomes clear what they plan to provide to opensource projects on a permanent basis

Rohanjames1997:
Thanks @rgommers and @martin-frbg!
Do you know of popular repositories that currently lack support for Arm64 CI, where community help would be beneficial?

PR #5157, opened 2025-02-28
Optimize gemv_n_sve kernel

manaalmj:
Loop-unrolling of sgemv_n kernel is implemented with svmla_lane, along with a main loop and a tail loop.
The graph below shows performance improvement for OMP_NUM_THREADS=1 on Graviton-3:

speedup

### DIFF
--- a/kernel/arm64/KERNEL.ARMV8SVE
+++ b/kernel/arm64/KERNEL.ARMV8SVE
@@ -74,7 +74,7 @@ DSCALKERNEL  = scal.S
 CSCALKERNEL  = zscal.S
 ZSCALKERNEL  = zscal.S
 
-SGEMVNKERNEL = gemv_n.S
+SGEMVNKERNEL = gemv_n_sve.c
 DGEMVNKERNEL = gemv_n.S
 CGEMVNKERNEL = zgemv_n.S
 ZGEMVNKERNEL = zgemv_n.S
--- a/kernel/arm64/gemv_n_sve.c
+++ b/kernel/arm64/gemv_n_sve.c
@@ -1,5 +1,5 @@
 /***************************************************************************
-Copyright (c) 2024, The OpenBLAS Project
+Copyright (c) 2024-2025, The OpenBLAS Project
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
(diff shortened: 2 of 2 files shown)


kernel/arm64/gemv_n_sve.c
```diff
@@ -57,25 +57,89 @@ int CNAME(BLASLONG m, BLASLONG n, BLASLONG dummy1, FLOAT alpha, FLOAT *a, BLASLO
...
+        SV_TYPE a02_vec = svld1(pg_true, a2_ptr + i);
+
+        SV_TYPE y_vec = svld1(pg_true, y + i);
+        y_vec = svmla_lane(y_vec, a00_vec, x0_vec, 0);
+        y_vec = svmla_lane(y_vec, a01_vec, x1_vec, 0);
+        y_vec = svmla_lane(y_vec, a02_vec, x2_vec, 0);
+
+        svst1(pg_true, y + i, y_vec);
+
+        i += sve_size * 1;
```
fadara01: what's the point of multiplying by 1 here?
Same comment for the while loop condition

fadara01:
Can we also update the plot s.t:
- it's discontinuous - don't connect the points to each other
- showing clearly what the current state of affairs is vs this PR - maybe a red horizontal line at `speedup=1`
- It might also make sense to use log scale for the Y-axis. 
- X-axis label is `MxKxN`, right?
- add the number of threads/machine to the title

kernel/arm64/gemv_n_sve.c
```diff
@@ -57,25 +57,89 @@ int CNAME(BLASLONG m, BLASLONG n, BLASLONG dummy1, FLOAT alpha, FLOAT *a, BLASLO
...
+    FLOAT *a0_ptr = a + lda * width * 0;
+    FLOAT *a1_ptr = a + lda * width * 1;
+    FLOAT *a2_ptr = a + lda * width * 2;
+
+    for (j = 0; j < width; j++) {
       i = 0;
-      svbool_t pg = SV_WHILE(i, m);
-      while (svptest_any(SV_TRUE(), pg)) {
-        SV_TYPE a_vec = svld1(pg, a_ptr + i);
+      while ((i + sve_size * 1 - 1) < m) {
```
fadara01, line 72: A for-loop seems more suitable here and easier to read and follow

fadara01: nit: A for-loop might be more suitable here (easier to read and follow) ... same comment for the loop below

martin-frbg:
yes, an explanation of the benchmark graph would be nice - I'm not sure I understand why your PR would result in two very narrow regions where there is more than tenfold speedup ?

manaalmj:
> yes, an explanation of the benchmark graph would be nice - I'm not sure I understand why your PR would result in two very narrow regions where there is more than tenfold speedup ?

Please refer to the updated graph. For the regions where we see spikes, The baseline performance of these problem sizes are poor and that's why we see spike in speedup.

manaalmj:
> Can we also update the plot s.t:
> 
> * it's discontinuous - don't connect the points to each other
> * showing clearly what the current state of affairs is vs this PR - maybe a red horizontal line at `speedup=1`
> * It might also make sense to use log scale for the Y-axis.
> * X-axis label is `MxKxN`, right?
> * add the number of threads/machine to the title

Please see the updated graph.

fadara01:
@martin-frbg could you please re-review?

kernel/arm64/KERNEL.ARMV8SVE
```diff
@@ -1,4 +1,5 @@
 
+
```
fadara01: please avoid white-space changes

kernel/arm64/gemv_n_sve.c
```diff
@@ -57,25 +57,85 @@ int CNAME(BLASLONG m, BLASLONG n, BLASLONG dummy1, FLOAT alpha, FLOAT *a, BLASLO
 
   ix = 0;
   a_ptr = a;
-
```
fadara01, line 60: same here

kernel/arm64/gemv_n_sve.c
```diff
@@ -57,25 +57,85 @@ int CNAME(BLASLONG m, BLASLONG n, BLASLONG dummy1, FLOAT alpha, FLOAT *a, BLASLO
...
+        SV_TYPE x_vec = SV_DUP(alpha * x[(ix)]);
+        SV_TYPE a_vec = svld1(pg, a_ptr + i);
+        y_vec = svmla_m(pg, y_vec, a_vec, x_vec);
+        svst1(pg, y + i, y_vec);
+      }
+
       a_ptr += lda;
       ix += inc_x;
     }
-    return(0);
```
fadara01, line 78: same here

kernel/arm64/KERNEL.ARMV8SVE
manaalmj: Done.

kernel/arm64/gemv_n_sve.c
manaalmj: Done

kernel/arm64/gemv_n_sve.c
manaalmj: Done

martin-frbg:
Looks good to me, thank you very much.

PR #5160, opened 2025-02-28
Add SBGEMVN Kernel for ARM64

taoye9:
should be applied along with sbgemv_t kernel in https://github.com/OpenMathLib/OpenBLAS/pull/5159.


annop-w:
@taoye9 Could you please add this to N2 and V2 as well ? since this is NEON code.

PR #5183, opened 2025-03-13
Fix bug in ARM64 sbgemv_t

annop-w:


annop-w:
@martin-frbg Sorry I didn't catch this in the previous patch.

PR #5190, opened 2025-03-24
Fix missing commas in gensymbol.pl and DLL symbol pre/postfixing in CMake builds

Harishmcw:
Fixes: https://github.com/OpenMathLib/OpenBLAS/issues/5154

This PR fixes missing commas in gensymbol.pl, particularly in list declarations like @cblasobjsc. The missing commas could cause syntax errors or unintended behavior in Perl. The corrections ensure proper parsing and execution of the script.

PR #5188, opened 2025-03-19
Fix DLL symbol name pre/postfixing in CMake builds on Windows

martin-frbg:
fixes #5154

PR #5191, opened 2025-03-26
Fix DLL symbol name pre/postfixing in CMake builds on Windows

Harishmcw:
This PR is related to [#5190](https://github.com/OpenMathLib/OpenBLAS/pull/5190), which supersedes [#5188](https://github.com/OpenMathLib/OpenBLAS/pull/5188). However, the CMake changes from 5188 were not included in 5190. This PR ensures those missing changes are applied.

It modifies the CMake script to correctly handle symbol pre/post fixing for CMake builds.

Additionally, I have commented out #if (${REMOVE_STATIC_LIB}) as it was causing build errors. The REMOVE_STATIC_LIB variable is intended to control the removal of the static library, but it is not defined anywhere in the file, leading to build failures.


martin-frbg:
I can only assume that your force-push on the other PR triggered some unexpected behavior in git, as the CMakeLists change was still visible in the web interface when I triggered the merge. I'll add proper initialization of the REMOVE_STATIC_LIB variable later (indeed my draft PR only had it defined inside the conditional)

MehdiChinoune:
Could you stop assuming that `Windows == MSVC`. Also there is a specific variable for Windows `if(WIN32)` that works for both native and cross-compiling.

PR #5194, opened 2025-03-26
Fix lapack-netlib compile

zhiqiang-hhhh:
Some c src file should be added to avoid undefined symbols.


martin-frbg:
There is probably something else going wrong with your build - these are LAPACK functions that are overridden by reimplementations from the lapack (no netlib suffix) folder

Issue #5192, opened 2025-03-26
Cannot use some openBLAS routines on my m1 Mac ( example : cgeqp3rk(...))


Ouasselhabti:
I am encountering a linking error when trying to compile a Fortran project that uses OpenBLAS on my M1 MacBook Pro. The error message indicates that the symbol _cgepq3rk_ cannot be found, which I believe is related to OpenBLAS or LAPACK functions.

### Steps to Reproduce:
Install OpenBLAS using Homebrew:
```
brew install openblas
Set up a basic Fortran program that uses OpenBLAS (or any program that links OpenBLAS).
Attempt to compile the program using gfortran with OpenMP and OpenBLAS flags:
gfortran -fopenmp -I/opt/homebrew/opt/openblas/include -L/opt/homebrew/opt/openblas/lib -lopenblas -g -O0 main.f90 -o fmain
```
### Expected Behavior:

> The program should compile and link successfully without errors.

### Actual Behavior:
The linker returns the following error:

```
Undefined symbols for architecture arm64:
  "_cgepq3rk_", referenced from:
      ___blockcompress_MOD_compresstruncated in cciveV9U.o
ld: symbol(s) not found for architecture arm64
collect2: error: ld returned 1 exit status
make: *** [f90] Error 1
```
### Additional Information:
I am using macOS on an Apple M1 (ARM64) architecture.
OpenBLAS is installed via Homebrew and is in the /opt/homebrew/opt/openblas directory.
The error suggests that the required OpenBLAS symbol _cgepq3rk_ is missing, which may indicate a mismatch between the OpenBLAS library and the code being compiled.
OpenBLAS is keg-only on my system, and I have set the environment variables as per the Homebrew output:
```
export LDFLAGS="-L/opt/homebrew/opt/openblas/lib"
export CPPFLAGS="-I/opt/homebrew/opt/openblas/include"
export PKG_CONFIG_PATH="/opt/homebrew/opt/openblas/lib/pkgconfig"
```
### System Details:
macOS Version: [e.g., macOS 13.2.1]
Homebrew Version: [e.g., 3.5.0]
gfortran Version: [e.g., 12.2]
OpenBLAS Version: 0.3.29
Architecture: ARM64 (Apple M1)
### Troubleshooting Steps Taken:
Checked OpenBLAS Installation: OpenBLAS is installed in /opt/homebrew/opt/openblas.
[...]

PR #5195, opened 2025-03-26
Re-synchronize gensymbol.pl with the posix shell version

martin-frbg:
related to #5192


Ouasselhabti:
worked , just recompiled blas from source another time and I can now find the symbols in the dylib.

martin-frbg:
thanks for the feedback

PR #5203, opened 2025-04-03
Add vector registers to clobber list to prevent compiler optimization.

vaiskv:
SME based SGEMMDIRECT kernel uses the vector registers (z) and adding clobber list informs compiler not to optimize these registers.

### DIFF
--- a/kernel/arm64/sgemm_direct_arm64_sme1.c
+++ b/kernel/arm64/sgemm_direct_arm64_sme1.c
@@ -7,7 +7,6 @@
 #include <stdlib.h>
 #include <inttypes.h>
 #include <math.h>
-
 #if defined(HAVE_SME)
 
 /* Function prototypes */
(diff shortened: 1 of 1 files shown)


vaiskv:
Hi @martin-frbg,
This patch fixes an issue of #5084 where in O3 optimization compiler was directly loading values from z registers instead of reading from memory, which was causing undesirable output. This patch adds the vector registers to clobber list informing compiler not to optimize these registers.

vaiskv:
I came across an issue in the way the library is compiled and used in an application.

When the library is compiled with `TARGET=ARMV8 DYNAMIC_ARCH=1` and run on a target with SME feature, the library is hitting an exception with an error "**OpenBLAS: Architecture Initialization failed. No initialization function found**" 

But if the library is compiled with `TARGET=ARMV9SME DYNAMIC_ARCH=1`, the application works and call to `cblas_sgemm` gets directed to SME based sgemm implementation.

Upon debugging,  `gotoblas_dynamic_init` is populating gotoblas with gotoblas_ARMV9SME, but in the case where the libary is failing, the init function is NULL. But when compiled with ARMV9SME, init function is not NULL and is working fine.

Our requirement is use a common library for all the targets and if the target supports SME, `cblas_sgemm` should direct to SME implementation. 

Is there something missing in the integration? Please let me know if any changes are needed.

martin-frbg:
huh, looks like cpu type autodetection for ARMV9SME went missing in dynamic_arm64.c - at least I'm fairly sure we had it already - basically like it's done for ARMV8SVE but using the support_sme1() function

vaiskv:
I added the auto detection for sme by referring to this [commit](https://github.com/OpenMathLib/OpenBLAS/commit/f971ef55f2ce09d60b08137ca0608a6475a7611f#diff-d5d8fc115656969af2961dbe1b6bf7f8a722df4998be94e0936d8be0dee4f733). Still the issue persists  :(

martin-frbg:
Are you testing with an Apple cpu, or with something else ?

vaiskv:
I am testing on QEMU with SME enabled

martin-frbg:
I have now uploaded my current version of dynamic_arm64.c as PR #5222 - I think this should fix the ARMV9SME issue though I have not specifically tested it with QEMU. In any case I think that part of your PR is unrelated to the original task of fixing the over optimization issue in the sgemmdirect kernel ?

vaiskv:
Hi @martin-frbg,

Yes, the primary intention of this PR is to fix the compiler optimization. This fix can go as a separate commit. 

The fix for dynamic detection of the ARMV9SME can continue as part of PR #5222 .

Thanks!

vaiskv:
Hi @martin-frbg , can we please get this commit for fixing compiler optimization issue merged?

martin-frbg:
If you could please drop the other, unrelated change from this PR again as suggested earlier ?

vaiskv:
Hi @martin-frbg , I have removed the unrelated commit from this PR now. Thanks!

PR #5206, opened 2025-04-05
Update zsum.c -- fixed spelling error to successfully compile

ColumbusAI:
spelling error where zsum_kernel is used and it should be zasum_kernel. Will not compile without fix.


martin-frbg:
Thank you

Issue #4239, opened 2023-09-25
Building OpenBLAS 0.3.24 fails on MacPorts running macOS 13.5.2


kickingvegas:
# Description

Updating to OpenBLAS 0.3.24 on MacPorts seg faults due to invalid memory reference. Issue [reported to MacPorts](https://trac.macports.org/ticket/68225#comment:3) with response from MacPorts maintainers to escalate upstream to OpenBLAS.

Error message below:

```text
OPENBLAS_NUM_THREADS=1 OMP_NUM_THREADS=1 ./cblat2 < ./cblat2.dat

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:

Program received signal SIGTRAP: Trace/breakpoint trap.

Backtrace for this error:
OPENBLAS_NUM_THREADS=1 OMP_NUM_THREADS=1 ./zblat2 < ./zblat2.dat
```

# Environment

Apple M1 Pro, macOS 13.5.2 (22G91)

# Steps to Reproduce

Upgrade OpenBLAS from 0.3.23\_0 to 0.3.24\_0

# Expected Result

OpenBLAS 0.3.24 is built and installed.

# Actual Result

Build hangs, no backtrace is emitted in build log.

[openblas-upgrade.log](https://github.com/OpenMathLib/OpenBLAS/files/12718568/openblas-upgrade.log)

martin-frbg:
Is that Apple's clang you used, or something from macports as well ? I see the Fortran part seems to have been compiled with gfortran from GCC12, but I can't seem to find the clang version in your log. (No problems are/were seen in CI with clang14.0.6 and gfortran 12.2 (both from homebrew) under macOS 12)

michaellass:
I am facing the same issue and was able to reproduce the issue just running `make` in a checked-out git master. It seems to be related to Apple Clang. There was very recently an XCode update (15.0) which could play a role here. By default, on my system the following compilers are used:

```
% cc --version
Apple clang version 15.0.0 (clang-1500.0.40.1)
Target: arm64-apple-darwin22.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

% gfortran --version
GNU Fortran (MacPorts gcc12 12.3.0_0+stdlib_flag) 12.3.0
Copyright (C) 2022 Free Software Foundation, Inc.
[...]

PR #5217, opened 2025-04-10
test_potrs.c: do not use GCC pragma on darwin-aarch64

haampie:
Using GCC 14.2.0 on darwin aarch64, the pragma ultimately causes a linker
error "ld: invalid r_symbolnum=".  As a minimal example:

```c
#pragma GCC optimize("no-gcse")
float a = 1.0;
void b();
void c() { b(a); }
```

```console
$ gcc-14 -O1 -c example.c
$ ld example.o
ld: invalid r_symbolnum=1 in 'example.o'
```

The current workaround is to use the old linker, but (a) it's deprecated and
(b) it can produce libraries that are subsequently not linkable with the newer
linker in dependents: the new ld64 pedantically errors when linking libraries
with duplicate rpaths created by the classic linker.

In my opinion that's sufficient reason for build systems not to force `-ld_classic`

### DIFF
--- a/Makefile.system
+++ b/Makefile.system
@@ -436,10 +436,6 @@ CCOMMON_OPT += -Wl,-ld_classic
 FCOMMON_OPT += -Wl,-ld_classic
 endif
 ifeq (x$(XCVER), x 16)
-ifeq ($(C_COMPILER), GCC)
-CCOMMON_OPT += -Wl,-ld_classic
-FCOMMON_OPT += -Wl,-ld_classic
-endif
 ifeq ($(F_COMPILER), GFORTRAN)
 override CEXTRALIB := $(filter-out(-lto_library, $(CEXTRALIB))) 
 endif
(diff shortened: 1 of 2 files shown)


adamjstewart:
Can we get rid of ld_classic for Xcode 15 too?

martin-frbg:
Thanks - intriguing that this pragma would lead to the linker error. But can you please make the line conditional on ARCH being LOONGARCH64, instead of deleting it ? The no-gcse directive was put there to work around a spurious test failure specific to that hardware's gcc14 port

haampie:
Thanks! `ARCH_LOONGARCH64` is the right define?

martin-frbg:
Yes, ARCH_LOONGARCH64 it is, thanks for updating the PR. 
Regarding xcode15, the `ld_classic` was added there to solve a different problem, a segfault in the BLAS tests (issue #4239) so I expect it will have to stay. (The imminent deprecation/removal of ld_classic should not be a problem in that context, as the user would need to upgrade xcode to encounter it)

PR #5220, opened 2025-04-11
Further performance improvements to non-transposed [SD]GEMV kernels for A64FX and Neoverse V1.

### LINKED ISSUE #5210: Further performance improvements of [SD]GEMV on A64FX and Neoverse V1.
[Pull Request #5157](https://github.com/OpenMathLib/OpenBLAS/pull/5157) did excellent work and the performance of non-transposed [SD]GEMV improved.
I think Neoverse V1 has room for further performance improvement and the A64FX has a better optimal loop unrolling number. So, I would like to propose a patch for the A64FX and Neoverse V1.

iha-taisei:
close #5210 

This pull request proposes a patch for issue #5210.
I have implemented a loop unrolling in the kernel of the non-transposed [SD]GEMV for A64FX and Neoverse V1.
This PullRequest improves performance by 1.7x for A64FX and 2x for Neoverse V1 compared to v0.3.29.
image
image

### DIFF
--- a/kernel/arm64/KERNEL.A64FX
+++ b/kernel/arm64/KERNEL.A64FX
@@ -1,6 +1,6 @@
 include $(KERNELDIR)/KERNEL.ARMV8SVE
 
-SGEMVNKERNEL = gemv_n_sve.c
-DGEMVNKERNEL = gemv_n_sve.c
+SGEMVNKERNEL = gemv_n_sve_v4x3.c
+DGEMVNKERNEL = gemv_n_sve_v4x3.c
 SGEMVTKERNEL = gemv_t_sve_v4x3.c
 DGEMVTKERNEL = gemv_t_sve_v4x3.c
--- a/kernel/arm64/KERNEL.NEOVERSEV1
+++ b/kernel/arm64/KERNEL.NEOVERSEV1
@@ -1,5 +1,7 @@
 include $(KERNELDIR)/KERNEL.ARMV8SVE
 
+SGEMVNKERNEL = gemv_n_sve_v1x3.c
+DGEMVNKERNEL = gemv_n_sve_v1x3.c
 SGEMVTKERNEL = gemv_t_sve_v1x3.c
 DGEMVTKERNEL = gemv_t_sve_v1x3.c
 ifeq ($(BUILD_BFLOAT16), 1)
--- a/kernel/arm64/gemv_n_sve_v1x3.c
+++ b/kernel/arm64/gemv_n_sve_v1x3.c
@@ -1,0 +1,138 @@
+/***************************************************************************
+Copyright (c) 2025, The OpenBLAS Project
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+
+   1. Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+
+   2. Redistributions in binary form must reproduce the above copyright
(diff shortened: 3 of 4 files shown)


annop-w:
@iha-taisei Thank you very much for your contribution. Could we please then remove `gemv_n_sve.c` as well ? since it will no longer be used. I think we should try keeping the number of kernels low.

annop-w:
@iha-taisei I have just benchmarked the `gemv_n_sve_v1x3.c` against the NEON kernel in #5225 on NEOVERSEV2 and the SVE version wins slightly.

c8g_sgemvn

So, would you mind adding `gemv_n_sve_v1x3.c` in `kernel/arm64/KERNEL.NEOVERSEN2` as well please ? I will close #5225 in favor of this one. Thank you.

martin-frbg:
> @iha-taisei Thank you very much for your contribution. Could we please then remove `gemv_n_sve.c` as well ? since it will no longer be used. I think we should try keeping the number of kernels low.

I'm not convinced that we need to remove kernel files simply because they are (currently) not in use by any hardware

abhishek-iitmadras:
Hi @annop-w 

Can we use **gemv_n_sve_v1x3.c** for **KERNEL.ARMV8SVE**, like we have already for [S/D]GEMVTKERNEL with patch https://github.com/OpenMathLib/OpenBLAS/pull/5215? 

cc @iha-taisei

iha-taisei:
Hi @annop-w

> So, would you mind adding gemv_n_sve_v1x3.c in kernel/arm64/KERNEL.NEOVERSEN2 as well please ? I will close https://github.com/OpenMathLib/OpenBLAS/pull/5225 in favor of this one. Thank you.

Yes, you can add the kernel to KERNEL.NEOVERSEN2, although I haven't evaluated it on such CPUs.

annop-w:
@iha-taisei 

> Yes, you can add the kernel to KERNEL.NEOVERSEN2, although I haven't evaluated it on such CPUs.

I have results for NEOVERSEV2, which currently uses the same settings as NEOVERSEN2 for DYNAMIC_ARCH, in my above comment. I have not benchmarked on N2 but I believe the result will hold as well and we will see speedup.

annop-w:
@martin-frbg 

> I'm not convinced that we need to remove kernel files simply because they are (currently) not in use by any hardware

From a quick look, the kernel `gemv_n_sve.c` in #5157 is very similar to the one in this PR (the same no. of unrolling). That's why I suggested removing the old one. In fact, it is interesting to try to understand why this one performs significantly better.

annop-w:
@abhishek-iitmadras 

> Can we use gemv_n_sve_v1x3.c for KERNEL.ARMV8SVE, like we have already for [S/D]GEMVTKERNEL with patch https://github.com/OpenMathLib/OpenBLAS/pull/5215?

Yes, but I have not tried benchmarking on those CORTEX-As and -Xs. But, seeing how this new SVE kernel outperforms the assembly one on V1 and V2, I expect the same on those cores perhaps.

martin-frbg:
> @abhishek-iitmadras
> 
> > Can we use gemv_n_sve_v1x3.c for KERNEL.ARMV8SVE, like we have already for [S/D]GEMVTKERNEL with patch #5215?
> 
> Yes, but I have not tried benchmarking on those CORTEX-As and -Xs. But, seeing how this new SVE kernel outperforms the assembly one on V1 and V2, I expect the same on those cores perhaps.

I can benchmark on a Pixel8, if we can agree that it is an underrated supercomputer (and if I can find the time and energy for non-trivial work again)

Issue #2166, opened 2019-06-19
OpenBLAS build fails due to segmentation fault on Power8 platform


kavanabhat:
OpenBLAS build fails for Power8 target in 64-bit mode(Both Redhat Big endian and AIX ). Details are as below:

#make DEBUG=1 BINARY=64 TARGET=POWER8
Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x3FFF95A692C3
#1  0x3FFF95A69E23
#2  0x3FFF95CB0477
#3  0x10078920 in dtrmm_ounucopy at trmm_uncopy_4.c:93
OPENBLAS_NUM_THREADS=1 OMP_NUM_THREADS=1 ./cblat2 < ./cblat2.dat
OPENBLAS_NUM_THREADS=1 OMP_NUM_THREADS=1 ./zblat2 < ./zblat2.dat
/bin/sh: line 1: 54413 Segmentation fault      (core dumped) OPENBLAS_NUM_THREADS=1 OMP_NUM_THREADS=1 ./dblat3 < ./dblat3.dat
make[1]: *** [level3] Error 139
make[1]: *** Waiting for unfinished jobs....
rm -f ?BLAT2.SUMM
OMP_NUM_THREADS=2 ./sblat2 < ./sblat2.dat
OMP_NUM_THREADS=2 ./dblat2 < ./dblat2.dat
OMP_NUM_THREADS=2 ./cblat2 < ./cblat2.dat
OMP_NUM_THREADS=2 ./zblat2 < ./zblat2.dat
make[1]: Leaving directory `/home/kavana/OpenBLAS/test'
make: *** [tests] Error 2
[root@pokndd8 OpenBLAS]# cd test
[root@pokndd8 test]# gdb ./dblat3
GNU gdb (GDB) Red Hat Enterprise Linux 7.6.1-100.el7
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "ppc64-redhat-linux-gnu".
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>...
Reading symbols from /home/kavana/OpenBLAS/test/dblat3...done.
[...]

PR #5235, opened 2025-04-20
Explicit unaligned vector load/stores in PPC64LE GEMV kernels

quickwritereader:
- Explicit unaligned vector load/stores in PPC64LE GEMV kernels
- Formatted code


quickwritereader:
do not we have ppc64le GitHub build and tests?

quickwritereader:
#2166

martin-frbg:
The "Jenkins" build runs on both POWER and ZARCH, however the underlying PPC hardware is POWER9 now I think.

martin-frbg:
@xry111 reminds me that we have been there two months ago with #5122 (and my subsequent #5125 to revert parts of 
#4880 which is where the `if defined(POWER8)` comes from), so perhaps the code can be further streamlined again. I'll need to do build tests in the GCC Compile Farm when I'm able

martin-frbg:
Thank you very much for the quick fix - it does indeed look like reverting my #5125 will be possible with this fix, restoring optimizations from #4880 that increase accuracy.

Issue #4073, opened 2023-06-05
Why I can't use builed lib and include after building with Visual Studio 2019?


Kunlei-Hong:
sorry,something shoule be wrong in my project.
step1:make
step2:make install
and i got bin include and libe folders and adding them to my project environment. when generating in vs 2019,it annonunced 128 errors.
could anyone tell me why and how to solve it. thanks all the way.
![image](https://github.com/xianyi/OpenBLAS/assets/64198368/e92ca5ee-43ce-48e1-808d-68559902116e)

martin-frbg:
probably related to #3661

martin-frbg:
which version of OpenBLAS did you build ?

Kunlei-Hong:
> which version of OpenBLAS did you build ?

hey,the version was the lastest so far. I doubt I used the wrong compiler in VS 2019. I am trying to change deault compiler and to be continued.

martin-frbg:
Nothing wrong with the compiler (except that it is VS :) ), I was just wondering if it could be fixed similarly to what seemed to be working for #3661 - does your project build when you add something like
```
#if defined(_MSC_VER)
#if !defined(LAPACK_COMPLEX_CUSTOM)
    #include <complex>
    #define LAPACK_COMPLEX_CUSTOM
    #define lapack_complex_float _Fcomplex
    #define lapack_complex_double _Dcomplex
#endif
#endif

before you include lapack.h ?
```

nanthony21:
The #include should be <complex.h>, not <complex>. Also it looks like newer versions of MSVC need to #define _CRT_USE_C_COMPLEX_H before including <complex.h> .  It would be nice if the documentation was updated to reflect this. https://github.com/microsoft/STL/issues/3280

martin-frbg:
Thank you - added to lapack.h in #4368 , which should be more effective than adding to documentation (I think)

PR #5244, opened 2025-04-24
allow the use of LAPACK_COMPLEX_CPP when using MSVC compiler

chitao1234:
Currently, openblas force the use `complex.h` when building lapack for msvc, this commit allow it to also use the C++ `complex` using `LAPACK_COMPLEX_CPP` macro as it is in the non msvc case.


martin-frbg:
MSVC is clearly not my forte, but I added this in response to #4073 - would we need an MSVC version check here perhaps ?

chitao1234:
I guess we do not need a version check for MSVC, I checked that std::complex is present all the way back in visual studio .net 2003.

PR #5256, opened 2025-05-10
Bump xuantie toolchains

RevySR:


martin-frbg:
This change seems to lead to several intrinsics becoming undeclared - would this require additional changes to the (perhaps now outdated) xuantie-specific part of common_riscv64.h, or is it a deeper issue ?

pz9115:
> This change seems to lead to several intrinsics becoming undeclared - would this require additional changes to the (perhaps now outdated) xuantie-specific part of common_riscv64.h, or is it a deeper issue ?

Agree, I think the xtheadvector intrinsic APIs add the prefix 'th_' to follow the vendor intrinsic naming rule, see https://github.com/riscv-non-isa/riscv-toolchain-conventions/blob/main/src/toolchain-conventions.adoc#list-of-vendor-prefixes

so may update the intrinsic calls with new prefix may fix this problem.

PR #5265, opened 2025-05-15
 kernel/riscv64:Added support for omatcopy on RISCV64_ZVL256B

guoyuanplct:
I modified the omatcopy implementation for riscv64_zvl256b, achieving a certain performance improvement.
image

### DIFF
--- a/.github/workflows/c910v.yml
+++ b/.github/workflows/c910v.yml
@@ -83,9 +83,39 @@ jobs:
 
       - name: test
         run: |
-          export PATH=$GITHUB_WORKSPACE/qemu-install/bin/:$PATH
-          qemu-riscv64 ./utest/openblas_utest
-          qemu-riscv64 ./utest/openblas_utest_ext
+          run_with_retry() {
+              local cmd="$1"
+              local time_out=10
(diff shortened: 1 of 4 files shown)


guoyuanplct:
It might be worth considering moving this test — which appears to occasionally hang — to the end, or perhaps adding a timeout (e.g., 3 minutes) to allow for a safe restart if necessary. 
image

guoyuanplct:
I tried a function that restarts the deadlock test after it gets hanged. This function restarts the program if it runs beyond a certain time limit, with a maximum of 10 retries. In my forked branch, I did some basic testing. On average, the number of retries was around 5 to 8 times. Or we can continue to increase the number of restarts, since each run is very short and won't have a significant impact on the overall time.

martin-frbg:
Okay, thanks, let's go with your version then - certainly more elegant than my quick fix of letting the entire job time out faster. Eventually this can hopefully be fixed by a newer qemu, or changes to the virtual hardware environment of the job.
(The hanging utest is most likely the fork-and-run-dgemm one, which I haven't seen hang on actual hardware)

PR #5276, opened 2025-05-21
Improvement of 2D thread-partitioned GEMM for M << N case

### LINKED ISSUE #5270: Improvement of 2D thread-partitioned GEMM for M << N case
The 2D thread partitioning in GEMM ([PR#4655](https://github.com/OpenMathLib/OpenBLAS/pull/4655)) requires nthreads_m % 2 == 0. This can prevent optimal nthreads_m and nthreads_n combinations on architectures like A64FX (48 cores) or Grace (144 cores) when M<<N, due to core counts having divisors other than 2. A fix is planned.

nakagawa-fj:
Closes #5270
  The 2D thread partitioning in GEMM (PR#4655) requires nthreads_m % 2 == 0. This can prevent optimal nthreads_m and nthreads_n combinations on architectures like A64FX (48 cores) or Grace (144 cores) when M<<N, due to core counts having divisors other than 2.
  Specifically, when matrix size N is significantly larger than M, the number of threads for N direction should be increased.
However, if nthreads_m includes divisors other than 2, such as 3, the increase of nthreads_n is prevented by ' nthreads_m % 2 == 0 '.
  This modification removes the nthreads_m % 2 == 0 restriction and selects the combination that minimizes the following objective function 'n * nthreads_m + m * nthreads_n'.
  This change improves the performance of multi-threaded GEMM for M << N cases.
  
image
  
image

### DIFF
--- a/driver/level3/level3_thread.c
+++ b/driver/level3/level3_thread.c
@@ -851,9 +851,19 @@ int CNAME(blas_arg_t *args, BLASLONG *range_m, BLASLONG *range_n, IFLOAT *sa, IF
     /* Objective function come from sum of partitions in m and n.             */
     /* (n / nthreads_n) + (m / nthreads_m)                                    */
     /* = (n * nthreads_m + m * nthreads_n) / (nthreads_n * nthreads_m)        */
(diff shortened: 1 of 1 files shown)


martin-frbg:
Thank you

PR #5278, opened 2025-05-21
Fix compilation with pre-C99 compilers

martin-frbg:
small fix to make #5276 work with the old manylinux setup in CI


nakagawa-fj:
Martin-san, thank you for the follow-up.

PR #5287, opened 2025-05-28
Bgemm for arm64

taoye9:


common.h
```diff
@@ -306,6 +307,13 @@ typedef int blasint;
 #define SIZE	8
 #define  BASE_SHIFT 3
 #define ZBASE_SHIFT 4
+#elif defined(BFLOAT16_ONLY)
```
annop-w, line 310: I do not think we need to introduce BFLOAT16_ONLY build flag.
The type of FLOAT is the only difference. Can we simplyl use XFLOAT instead of FLOAT for C matrix, for example, in kernel/arm64/bgemm_beta.c ?

martin-frbg:
Thanks. Failures on CirrusCI are from running out of compute credits (as we're close to the end of the month), I'll rectify that in a minute. Not sure how DGEMM got hit by your PR on at least RISCV and LOONGARCH64 platforms, haven't looked closely yet but possibly some unintentional shifting of GEMM_UNROLL parameters ? (Also not sure why this touches SBGEMM settings ?)

driver/level3/level3.c
```diff
@@ -169,6 +170,22 @@
 #define STOP_RPCC(COUNTER)
 #endif
 
+#if defined(HALF)
```
annop-w, line 175: If we remove BUILD_BFLOAT16_ONLY flag, this is not needed I suppose.

driver/level3/level3_thread.c
```diff
@@ -216,6 +216,22 @@ typedef struct {
 #define STOP_RPCC(COUNTER)
 #endif
 
+#if defined(HALF)
```
annop-w, line 219: Ditto. See comment in level3.c.

interface/gemm.c
```diff
@@ -56,6 +56,9 @@
 #elif defined(BFLOAT16)
 #define ERROR_NAME "SBGEMM "
 #define GEMV BLASFUNC(sbgemv)
+#elif defined(BFLOAT16_ONLY)
```
annop-w, line 59: Maybe a different naming ? Perhaps change to SBFLOAT16 for SBGEMM and BFLOAT16 for BGEMM. What do you think ?

kernel/Makefile.L3
```diff
@@ -660,15 +687,15 @@ $(KDIR)$(SBGEMMONCOPYOBJ) : $(KERNELDIR)/$(SBGEMMONCOPY)
 $(KDIR)$(SBGEMMOTCOPYOBJ) : $(KERNELDIR)/$(SBGEMMOTCOPY)
 	$(CC) $(CFLAGS) -c -DBFLOAT16 -UDOUBLE -UCOMPLEX $< -o $@
 
-ifneq ($(SBGEMM_UNROLL_M), $(SBGEMM_UNROLL_N))
+#ifneq ($(SBGEMM_UNROLL_M), $(SBGEMM_UNROLL_N))
```
annop-w, line 738: Why is this commented out ?

kernel/arm64/bgemm_beta.c
```diff
@@ -0,0 +1,38 @@
...
+ * POSSIBILITY OF SUCH DAMAGE.
+ * *****************************************************************************/
+
+#include "common.h"
+
+#include <arm_neon.h>
+
+int CNAME(BLASLONG m, BLASLONG n, BLASLONG dummy1, FLOAT beta, IFLOAT *dummy2,
+          BLASLONG dummy3, IFLOAT *dummy4, BLASLONG dummy5, FLOAT *c,
+          BLASLONG ldc) {
```
annop-w: Change FLOAT to IFLOAT or XFLOAT ?

interface/gemm.c
Mousius: Is there a need to split it? `BUILD_BFLOAT16` could cover both cases and enable all bfloat16 ops?

martin-frbg: @Mousius there's a slight difference between the global BUILD_BFLOAT16 and this BFLOAT16 (that IIRC is local to the interface and gets added by the (C)Makefile ). Maybe we could have a "BGEMM" define that gets set by the Makefile specifically when the function is bgemm and that is processed inside the "elif defined(BFLOAT16)" ? I haven't looked into the assumed requirement for a BUILD_BFLOAT16_ONLY in the level3 driver yet, but my gut feeling is that a local variable in interface/Makefile (and eventually CMakeLists.txt) to modify gemm.c build behaviour for bgemm.o should be sufficient.

kernel/Makefile.L3
taoye9: My changes introduced a bug that raises an error when the kernels are reused. this line need to be uncomment when the fix is fixed.

martin-frbg:
Closing as superseded by #5357 ; thank you very much to all involved !

Issue #285, opened 2013-08-25
Add an axpby Kernel


FreddieWitherden:
An extremely useful BLAS extension, found in ATLAS, MKL and others is: axpby defined as: `y = alpha*x + beta*y` and is a generalization of axpy.  Incorporating this into OpenBLAS would greatly simplify the use of OpenBLAS in existing projects which assume the availability of such a kernel.

PR #5288, opened 2025-05-29
kernel/riscv64:Optimized the implementation of axpby on TARGET=RISCV64_ZVL256B.

guoyuanplct:
The specific improvements are shown in the figure below.
image
image

### DIFF
--- a/kernel/riscv64/KERNEL.RISCV64_ZVL256B
+++ b/kernel/riscv64/KERNEL.RISCV64_ZVL256B
@@ -169,10 +169,14 @@ SSYMV_U_KERNEL =  symv_U_vector.c
 SSYMV_L_KERNEL =  symv_L_vector.c
 DSYMV_U_KERNEL =  symv_U_vector.c
 DSYMV_L_KERNEL =  symv_L_vector.c
+
+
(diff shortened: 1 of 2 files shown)


abhishek-iitmadras:
Just out of curiosity and for my learning, i have below question :

What are the key practical scenarios or algorithms where a **dedicated AXPBY** kernel from openBLAS provides a significant performance advantage given that if we already have a highly optimized **AXPY** ?

Thanks

@martin-frbg @guoyuanplct

martin-frbg:
We may not have a highly optimized AXPY on all architectures, and the current default for AXPBY is a naive C loop instead of combining calls to SCAL and AXPY in the interface. (The git log suggests that axpby was added a decade ago for compatibility with MKL - #285 - and nobody looked at it - or its performance - ever since)

martin-frbg:
(small correction - the Loongson crew did add optimized kernels for their hardware in late 2023, so this is not entirely without precedent. There are no callers in Reference-LAPACK, and the only user in OpenBLAS itself seems to be the generic GEADD, so this may have gone mostly unnoticed)

Issue #5279, opened 2025-05-22
Lack of FP16/BF16 Precision Support for GEMM Kernels on RISCV


Srangrang:
Hello,

We are currently working on porting LLMs to RISCV for inference, and due to hardware limitations, we need to maintain FP16 or BF16 precision. In this process, we are using the widely-used and highly-regarded OpenBLAS as the underlying computational support. However, the computational speed didn't meet our expectations.

It seems that this issue is due to the lack of pure FP16 or mixed-precision GEMM kernels in OpenBLAS for the RISCV architecture. Although we have seen sbgemm kernels on other architectures, they are not available for RISCV.

We noticed that the RISCV instruction set manual released on May 8, 2025, already supports RVV 1.0 for mixed-precision FP16 and BF16 [Instruction Set Manual](https://drive.google.com/file/d/1uviu1nH-tScFfgrovvFCrj7Omv8tFtkp/view?usp=sharing), and the RVV intrinsics have also been updated for FP16 mixed-precision [RVV Intrinsic Document](https://github.com/riscv-non-isa/rvv-intrinsic-doc).
Based on the vectorized sgemm kernel, we have implemented a mixed-precision GEMM kernel for RISCV, which we believe should be shgemm, similar to the approach in #2767.

We would like to know if there is any plan to add support for FP16/BF16 precision GEMM kernels on the RISCV architecture in OpenBLAS. Specifically, we are interested in whether there are plans to support these kernels on the RISC-V 64-bit architecture with vlen of 128 bits (RISCV64_ZVL128B) and 256 bits (RISCV64_ZVL256B). If not, we are willing to contribute our implementation and provide the code.

martin-frbg:
There is currently very little organizational structure or developtment planning behind OpenBLAS, but I am not aware of anybody currently working on BF16 kernels for RISCV. For FP16, I _think_ it would be trivial to copy all the "infrastructure" bits (additions to various headers and makefiles, etc) from the early BF16 pull requests, and there was someone expressing interest in writing x86_64 kernels a few years ago but he switched jobs and that work stalled. So your contribution would be very welcome.

LaNasilDark:
hello,
I am very interested in the "mixed-precision GEMM kernel for RISCV" you mentioned in this issue,can you  introduce more infomation？ Thanks a lot.

gkdddd:
@LaNasilDark 
hello,
[...]

PR #5290, opened 2025-06-04
Add support for FP16 to openBLAS and shgemm on RISCV

Srangrang:
-add HFLOAT16 and BUILD_HFLOAT16 macro define to distinguish BFLOAT16 and BUILD_BFLOAT16
-add shgemm for RISCV_ZVL128B and RISCV_ZVL256B
-using fp16 on RISCV requires zfh and zvfh instruction sets
-enable fp16 support in Makefile.rule

Related to issue #5279 
Co-authored-by Ao Dong


martin-frbg:
I _think_ we'll need some adjustments in interface/gemm.c to disable the "small matrix" pathway, adjust the function name used in error messages, and also keep hfloat16 out of the optimum thread number computations, now that it is separate from bfloat16. Please see attached (and add this or similar to the PR if you agree)
[gemm.c.txt](https://github.com/user-attachments/files/20716611/gemm.c.txt)

Also it probably makes sense to seperate the two types to such an extent that one can be built without the other - I have further modified your version of gensymbol(.pl) to take another parameter for the BUILD_HFLOAT16, and adjusted exports/Makefile accordingly:
[Makefile.txt](https://github.com/user-attachments/files/20716654/Makefile.txt)
[gensymbol.pl.txt](https://github.com/user-attachments/files/20716656/gensymbol.pl.txt)
[gensymbol.txt](https://github.com/user-attachments/files/20716661/gensymbol.txt)

martin-frbg:
In similar vein, the affected files in the benchmark folder probably need adjusting as well, please check
[Makefile.txt](https://github.com/user-attachments/files/20716702/Makefile.txt)

[gemm.c.txt](https://github.com/user-attachments/files/20716704/gemm.c.txt)

martin-frbg:
Lastly, I found that I cannot build the "generic" replacements on a non-RISCV host unless I remove the `ifneq ($(SHGEMM_UNROLL_M), $(SHGEMM_UNROLL_N))` conditional around the compiler commands for the SHGEMMINCOPYOBJ and SHGEMMITCOPYOBJ in kernel/Makefile.L3 (and I wonder if it would make sense to provide hfloat16 conversion functions in kernel/generic/gemm_kernel2x2.c similar to what is present there for bfloat16) ?

Unfortunately, I have not had the time to try a CMake build yet.

martin-frbg:
cmake/utils.cmake
[utils.cmake.txt](https://github.com/user-attachments/files/20741995/utils.cmake.txt)
cmake/kernel.cmake
[kernel.cmake.txt](https://github.com/user-attachments/files/20741997/kernel.cmake.txt)
kernel/CMakeLists 
[CMakeLists.txt](https://github.com/user-attachments/files/20742181/CMakeLists.txt)
interface/CMakeLists
[CMakeLists.txt](https://github.com/user-attachments/files/20742183/CMakeLists.txt)

toplevel CMakeLists (added status messages, updated gensymbol calls, updated benchmark build to include B/HFLOAT16)
[CMakeLists.txt](https://github.com/user-attachments/files/20742002/CMakeLists.txt)

Srangrang:
> I _think_ we'll need some adjustments in interface/gemm.c to disable the "small matrix" pathway, adjust the function name used in error messages, and also keep hfloat16 out of the optimum thread number computations, now that it is separate from bfloat16. Please see attached (and add this or similar to the PR if you agree) [gemm.c.txt](https://github.com/user-attachments/files/20716611/gemm.c.txt)
> 
> Also it probably makes sense to seperate the two types to such an extent that one can be built without the other - I have further modified your version of gensymbol(.pl) to take another parameter for the BUILD_HFLOAT16, and adjusted exports/Makefile accordingly: [Makefile.txt](https://github.com/user-attachments/files/20716654/Makefile.txt) [gensymbol.pl.txt](https://github.com/user-attachments/files/20716656/gensymbol.pl.txt) [gensymbol.txt](https://github.com/user-attachments/files/20716661/gensymbol.txt)

Thank you for your comment, @martin-frbg . I agree with your modification and will incorporate your changes into this PR.

Srangrang:
> In similar vein, the affected files in the benchmark folder probably need adjusting as well, please check [Makefile.txt](https://github.com/user-attachments/files/20716702/Makefile.txt)
> 
> [gemm.c.txt](https://github.com/user-attachments/files/20716704/gemm.c.txt)

I have reviewed the contents of the benchmark folder and noticed that references to BF16 appear to be defined as `HALF`. However, the modifications provided in your Makefile.txt suggest changing this to `BFLOAT16`. I believe this change could reduce ambiguity and more effectively differentiate between FP16 and BF16. Could we consider standardizing the use of `HFLOAT16` for FP16 and `BFLOAT16` for BF16 within OpenBLAS (by modifying the definition of `GOTO_HF16_TARGETS` to `GOTO_HFLOAT16_TARGETS` in this file)?

Additionally, I have identified other instances where `HALF` is used to define BF16, such as in the driver/level3/Makefile. Should we also adjust these to use the `BFLOAT16` definition?

martin-frbg:
You are right - any occurences of HALF must be leftovers from the initial BFLOAT16 code contribution that was using "half precision" and the "h" prefix throughout.

ChipKerchner:
I believe the vector instructions within the if ( M & 4 ) in 128b can be m1 versus what they are currently m2.

kernel/riscv64/shgemm_kernel_8x8_zvl128b.c
```diff
@@ -0,0 +1,767 @@
...
+            _Float16 B4 = B[bi+4];
+            _Float16 B5 = B[bi+5];
+            _Float16 B6 = B[bi+6];
+            _Float16 B7 = B[bi+7];
+            bi += 8;
+
+            vfloat16m1_t A0 = __riscv_vle16_v_f16m1(&A[ai + 0 * gvl], gvl);
+            ai += 4;
+
+            vfloat32m2_t result0 = __riscv_vfwmul_vf_f32m2( A0, B0, gvl);
```
ChipKerchner, line 121: I believe the vector instructions within the if ( M & 4 ) in 128b can be m1 versus what they are currently m2.

kernel/riscv64/shgemm_kernel_16x8_zvl256b.c
```diff
@@ -0,0 +1,969 @@
...
+            float result11 = 0;
+            float result12 = 0;
+            float result13 = 0;
+            float result14 = 0;
+            float result15 = 0;
+            BLASLONG ai = m_top * K;
+            BLASLONG bi = n_top * K;
+            
+            for(BLASLONG k=0; k<K; k++) {
+                result0+=(float)(A[ai+0]*B[bi+0]);
```
ChipKerchner, line 298: I believe this is wrong. A * B will produce a _Float16 result before it is cast. _Float16 is a storage type (at least before C++23) - unsigned short. So A * B is unsigned short * unsigned short and then cast to float.

It should be

result0+=(float)(A[ai+0])*(float)(B[bi+0]);

Srangrang: Thanks for pointing this out!

kernel/riscv64/shgemm_kernel_8x8_zvl128b.c
Srangrang: Thanks again for the detailed review — and especially for resolving this issue in #5451!

kernel/riscv64/KERNEL.RISCV64_ZVL128B
```diff
@@ -245,3 +245,14 @@ endif
 ifndef ZGEMM_BETA
 ZGEMM_BETA = zgemm_beta_rvv.c
 endif
+
+ifeq ($(BUILD_BFLOAT16), 1)
```
ChipKerchner, line 249: Bug here - should be BUILD_HFLOAT16.  Fixed in MR 5454.

kernel/riscv64/KERNEL.RISCV64_ZVL256B
```diff
@@ -209,5 +209,23 @@ COMATCOPY_CN = zomatcopy_cn_vector.c
 DOMATCOPY_CN = omatcopy_cn_vector.c
 SOMATCOPY_CN = omatcopy_cn_vector.c
 
+
+ifeq ($(BUILD_BFLOAT16), 1)
```
ChipKerchner, line 213: Bug here - should be BUILD_HFLOAT16.  Fixed in MR 5454.

Srangrang: Indeed, thanks.

Issue #5286, opened 2025-05-28
The Performance Issues Caused by TARGET=RISCV64_ZVL256B


guoyuanplct:
When compiling OpenBLAS using TARGET=RISCV64_ZVL256B, there can be significant performance issues, as shown in the two tables below. The tables show the performance when using RISCV64_GENERIC and RISCV64_ZVL256B, as well as the corresponding speedup ratios. HBMV is a Level 2 function. I am trying to offload its computation to the kernel to make specific modifications for RISC-V. However, I don't know exactly how to do it. Do you have any suggestions?

<img width="307" alt="Image" src="https://github.com/user-attachments/assets/7623843d-9de8-4350-97c3-5433dfa98ac4" />

<img width="303" alt="Image" src="https://github.com/user-attachments/assets/8e845c1a-689f-47e4-93a1-a64fa59b1143" />

martin-frbg:
Strange that the optimized kernels would be so much slower than their generic C counterpart - internally HBMV appears to be implemented as a series of AXPY calls, I wonder if the ZVL256B AXPY is slow too ? The only other specialty I see is that this
function appears to be implemented in single-threaded form only, but that should affect both codes equally.

guoyuanplct:
I also find it very strange because the performance of RISCV64_ZVL256B is really good in the AXPY part.

![Image](https://github.com/user-attachments/assets/413e5ba4-33b1-4565-b243-f939947dbabf)

martin-frbg:
Then either it is some weird corner case that is holding everything up, or the CDOT/ZDOT kernels are slower, but they'd have to be dramatically slower than the simple C implementation ??

guoyuanplct:
I’ve tested the CDOT/ZDOT kernels and they actually perform very well. I’m still looking into it to figure out what’s causing the problem.

martin-frbg:
Unfortunately I cannot get perf to produce any useful output on my system (Banana Pi F3, seems to be a hardware issue. On x86_64, the benchmark seems to spend a bit more time in DOT than in AXPY, but the only "hot" instruction recorded is packing the result back into a complex variable. So nothing that would provide a clue. I wish I had more experience with RISCV, I'm currently stuck on a bug with NAN handling in ZSCAL and can't even get it to build a working mask with vmfne.vv

martin-frbg:
[...]

PR #5291, opened 2025-06-05
kernel/riscv64:fixed the performance problem in RISCV64_ZVL256 when OPENBLAS_K is small

guoyuanplct:
I made these two code modifications to address the HBMV issue. When the computation scale is too small, the performance of RVV is very poor. Therefore, I call the unvectorized code when the scale is small.
The numbers 8 and 16 in the code are the balance points I found. Around these values, the performance of the RVV version and the unvectorized version is close.

### DIFF
--- a/kernel/riscv64/zaxpy_vector.c
+++ b/kernel/riscv64/zaxpy_vector.c
@@ -43,8 +43,55 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define VFNMSACVF_FLOAT RISCV_RVV(vfnmsac_vf_f64m4)
 #endif
 
+#if !defined(DOUBLE)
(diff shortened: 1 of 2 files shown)


guoyuanplct:
This PR is directly related to issue   #5286

martin-frbg:
Thanks - that's pretty much the same as what I came up with in my initial experimentation. I do wonder if the _rvv kernels (used by the ZVL128B and x280 targets) perform better - one oddity I noticed is that the _vector kernels always request maximum vector length (VSETVL_MAX) while their _rvv counterparts seem to try to match the vector length to the actual amount of data. (I'm still rather new to RISCV though, so may be misreading the code...)

guoyuanplct:
I'm also relatively new to RVV. My understanding is that both kernels will try to match the vector length to the actual amount of data. VSETVL_MAX is generally only used outside the computation loop to initialize some variables (or perform other types of operations), which need to be long enough to ensure they can be used for subsequent vector computations. Once inside the loop, both kernels will use VSETVL to match the appropriate length.

martin-frbg:
Ah, you're right of course, I missed the later vsetvl() in zdot_vector.c

PR #5292, opened 2025-06-09
Optimize gemv_n_sve_v1x3 kernel

isharif168:
- Calculate predicate outside the loop
- Divide matrix in blocks of 3

Comparison

x-axis -> M = N
y-axis -> GFLOPS (timing)


kernel/arm64/gemv_n_sve_v1x3.c
```diff
@@ -73,57 +72,79 @@ int CNAME(BLASLONG m, BLASLONG n, BLASLONG dummy1, FLOAT alpha, FLOAT *a,
...
+        i += sve_size;
       }
 
       if (i < m) {
-        svbool_t pg0 = SV_WHILE(i + sve_size * 0, m);
-
-        pg00 = svand_z(SV_TRUE(), pg0, pg00);
-        pg01 = svand_z(SV_TRUE(), pg0, pg01);
-        pg02 = svand_z(SV_TRUE(), pg0, pg02);
+        svbool_t pg_tail = SV_WHILE(i, m);
```
annop-w: Is it better to pre-calculate this predicate outside of the loop ? 
This is re-used again below.

isharif168: I think since we are calculating the predicate for the tail elements , it depends on i value , so if we remove outside of the loop then we have to calculate for (0 , m % sve_size) but that can go wrong sometime , since we want from (i, m) and not from 0 , whats your thought on this?

annop-w: I don't see why (0 , m % sve_size) wouldn't work since we increment i by sve_size in the main loop. Please also soo https://github.com/OpenMathLib/OpenBLAS/pull/5089/files#diff-d0b63f332b08eef9b57a1eec785ff43afc468108c60f237b0c4e9401df08b510R68

isharif168: Yes it will work, but its just that it will give the predicate for (0 , m % sve_size) index rather than the correct index as (i , m) , sure will make this change , Thanks.

annop-w:
LGTM. @martin-frbg any further comment ?

Issue #112, opened 2012-05-26
Workaround for >64 processors


gborzi:
I tried to compile openblas-git on a 4-Xeon E7-4850 Linux system which has 80 threads (4 procs \* 10 core \* 2 threads) using the following commands

export OMP_NUM_THREADS=64
make USE_OPENMP=1 NO_LAPACK=1

but it fails in the testing phase with the error message:

OpenBLAS Warining : The number of CPU/Cores(80) is beyond the limit(64). Terminated.

(note: it should be Warning not Warining)
Looking into init.c under driver/others I realised that OMP_NUM_THREADS is used to set the numprocs variable in gotoblas_affinity_init(), but that the check which gives the above error message is made on common->num_procs. This is set with a call to get_nprocs() at line 647. I changed this line to

common -> num_procs = get_nprocs() > 64? 64: get_nprocs();

and was finally able to pass the test.

Regards.

PR #5295, opened 2025-06-10
Fix some hyperthreading errors.

Pengzhou0810:
1.When there are multiple NUMA nodes and hyper-threading causes adjacent logical cores to share a physical core (e.g., common -> avail[i] = 0x5555555555555555UL), the numa_mapping function should not use a bitmask for filtering, as this would lead to redundant masking with the subsequent local_cpu_map function. 
2.In the scenario described above, the final_num_procs parameter cannot accurately represent the actual number of valid CPU cores. The num_procs parameter can be used as a replacement, so the final_num_procs parameter has been removed.

### DIFF
--- a/driver/others/init.c
+++ b/driver/others/init.c
@@ -391,7 +391,15 @@ static void numa_mapping(void) {
     core = 0;
     for (cpu = 0; cpu < common -> num_procs; cpu ++) {
       bitmask_idx = CPUELT(cpu);
+/*
+ * When common->avail[i] = 0x5555555555555555UL (indicating that adjacent logical cores share a physical core),
+ * using it as a mask may overlap with the local_cpu_map function's role, leading to only half of the real physical cores being detected.
+ */
+#ifdef ARCH_LOONGARCH64
+      if (common -> node_info[node][bitmask_idx]) {
+#else
       if (common -> node_info[node][bitmask_idx] & common -> avail[bitmask_idx] & CPUMASK(cpu)) {
+#endif
 	common -> cpu_info[count] = WRITE_CORE(core) | WRITE_NODE(node) | WRITE_CPU(cpu);
 	count ++;
 	core ++;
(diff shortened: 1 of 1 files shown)


Pengzhou0810:
Hi @XiWeiGu, @martin-frbg,

On LoongArch systems with hyper-threading (SMT), I noticed that OpenBLAS fails to correctly bind threads to physical CPU cores, which can lead to suboptimal performance due to logical core contention. I’ve implemented fixes to address some of these hyper-threading core-binding issues—could you kindly review the changes when you have time?

martin-frbg:
This is mostly original GotoBLAS code, last modified by Xianyi over ten years ago in response to #112. I must admit that I'm a bit wary to change it globally, I wonder if this is something more likely to occur on Loongarch hardware, or simply not reported during all these years.

Pengzhou0810:
> Collaborator

Pengzhou0810:
> This is mostly original GotoBLAS code, last modified by Xianyi over ten years ago in response to #112. I must admit that I'm a bit wary to change it globally, I wonder if this is something more likely to occur on Loongarch hardware, or simply not reported during all these years.

From the code analysis, this modification indeed makes the code more compatible with a wider range of platforms. I’ve conducted functionality tests on some x86 platforms without any issues, but there aren't enough test samples to cover all hardware platforms. If you're concerned about potential issues on other platforms, I can restrict this modification to the Loongarch platform.

XiWeiGu:
I recommend taking a more conservative approach: keep the logic unchanged for other platforms, and implement separate handling only for the LoongArch platform.

driver/others/init.c
```diff
@@ -415,7 +402,7 @@ static void numa_mapping(void) {
     }
   }
   for (i = 0; i < MAX_BITMASK_LEN; i++)
-    cpu_count += popcount(common -> node_info[current_node][i] & common -> avail[i]);
+    cpu_count += popcount(common -> node_info[current_node][i]);
```
XiWeiGu: The revised cpu_count now represents the number of logical cores per NUMA node, whereas before the revision it reflected the number of physical cores. This change may cause errors in condition checks like if (numprocs <= cpu_count), where tasks that originally required execution across two nodes might now be incorrectly judged as runnable within a single node

Pengzhou0810:
Hi @XiWeiGu, @martin-frbg, I've updated the code, limiting the changes to the LoongArch platform. Could you review it when possible?

XiWeiGu:
LGTM, Thanks

Pengzhou0810:
Hi @martin-frbg, would you mind taking a look at this commit? I've restricted the modifications to only affect the LoongArch architecture.

XiWeiGu (approved):

martin-frbg:
could you rebase this to 0.3.30 (e.g. use the "sync fork" in the github web interface of your branch) please ?

Pengzhou0810:
> could you rebase this to 0.3.30 (e.g. use the "sync fork" in the github web interface of your branch) please ?


 @martin-frbg Thanks for the update! The changes are now aligned with the latest community modifications. Let me know if there's anything else needed from my side.

PR #5310, opened 2025-06-16
Bug Fix: Problem with identifying some ARM64 processors

nakagawa-fj:
getarch command have a problem that it can not identify some ARM64 processors.
I have fixed this problem with implementing to get the correct values of cpu_implementer and cpu_part.


martin-frbg:
Thank you. Guess I must have broken this with one of the final changes in 0.3.29 - and of course it would primarily hit your processors, sorry.

PR #5312, opened 2025-06-17
Work around X86 POTRS/CDOT bug on old systems and add CI job for 32bit manylinux

martin-frbg:
fixes utest errör seen in MacPython wheel builds with the modified C/ZSCAL - https://github.com/MacPython/openblas-libs/issues/205 - apparently the limited NAN propagation capabilities of the old CSCAL (as called from CPOTF2) masked the erroneous NAN returned by cdot


mattip:
Thanks!

PR #5326, opened 2025-06-21
CMake: Make sure to find OpenMP dependency before usage.

MehdiChinoune:
cmake will throw an error if OpenMP::OpenMP_C or OpenMP::OpenMP_Fortran were not found first. see: https://github.com/msys2/MINGW-packages/issues/24616

### DIFF
--- a/.github/workflows/apple_m.yml
+++ b/.github/workflows/apple_m.yml
@@ -104,7 +104,7 @@ jobs:
                     -DUSE_OPENMP=${{matrix.openmp}} \
                     -DOpenMP_Fortran_LIB_NAMES=omp \
                     -DINTERFACE64=${{matrix.ilp64}} \
-                    -DNOFORTRAN=0 \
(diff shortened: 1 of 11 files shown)


MehdiChinoune:
There are two issues:
- NOFORTRAN is not a Boolean variable, need to be changed so it could be replaced by ON or OFF.
I see that it is set to 2 in one place but  never tested with that value, all conditionals are 'if(NOFORTRAN)' or 'if(NOT NOFORTRAN)'
- CMake couldn't find OpenMP Fortran component without a Fortran compiler, so the user has to install one which is a little annoying giving that no Fortran code is going to be compiled.

martin-frbg:
NOFORTRAN used to be tri-state (something like "the user wanted it but the compiler doesn't work" IIRC) I'll take a look as soon as I can. CMake macro features are elegant but they  bring in so many new logical dependencies compared to a naive gmake build ...

PR #5338, opened 2025-06-25
Update install.md

mennodeij:
Extra instructions about package versions in conda-forge and use full paths to compilers to prevent using other LLVM compilers installed on the system.


rgommers (changes requested):
I'm sorry but this isn't right. There is no reason to make all users jump through hoops for some very specific issue on your machine. You shouldn't be getting a compiler that old, and adding full paths to compilers is never necessary and extremely cumbersome.

PR #5353, opened 2025-06-30
Multi-thread Performance Improvement of GEMM with DIVIDE_RATE=1 for A64FX

### LINKED ISSUE #5347: Inefficiency of thread control with DIVIDE_RATE in GEMM
Currently, GEMM thread control uses default DIVIDE_RATE=2, which always splits N dimension of matrix into two parts for computation.     However, this division occurs even when N is small (e.g., N=2), leading to a decrease in computational efficiency.
 I've observed some performance issues related to this splitting on the A64FX architecture and would like to explore modifications to address them.

nakagawa-fj:
Closes #5347
 This PR improves the multi-thread performance of GEMM on A64FX by setting `DIVIDE_RATE` to 1.
The thread control in GEMM currently uses default value of DIVIDE_RATE=2, which always splits N dimension of matrix into two parts for computation. However, this splitting occurs even when N is small (e.g., N=2), leading to a decrease in computational efficiency.
For GEMM on A64FX, I tried DIVIDE_RATE=1 and confirmed performance improvements as shown in the graphs below.
While improvements were expected for narrow matrices with small N dimensions, performance gains were also observed for square matrices.

gemm_divide_rate_1
gemm_divide_rate_2
gemm_divide_rate_3

### DIFF
--- a/driver/level3/gemm.c
+++ b/driver/level3/gemm.c
@@ -59,6 +59,10 @@
 #define GEMM_Q 128
 #endif
 
+#ifdef GEMM_DIVIDE_RATE
+#define DIVIDE_RATE GEMM_DIVIDE_RATE
+#endif
+
 #ifdef THREADED_LEVEL3
 #include "level3_thread.c"
 #else
--- a/param.h
+++ b/param.h
@@ -3701,6 +3701,8 @@ is a big desktop or server with abundant cache rather than a phone or embedded d
 
 #elif defined(A64FX) // 512-bit SVE
 
+#define GEMM_DIVIDE_RATE  1
+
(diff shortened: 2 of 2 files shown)


martin-frbg:
That's a very insteresting result - somewhat counterintuitive, but I guess if it helps to make optimum use of the vector length... Certainly a reminder that some core design decisions for OpenBLAS (GotoBLAS really) were made on and for the cpu architectures of some twenty years ago.
I notice that DIVIDE_RATE is also used in the multithreaded SYRK code (and GETRF/POTRF, though they may be speed-limited by other constraints). And I guess it will be interesting to see if a non-default value would also be beneficial for other (SVE/SME) arm64 or RISCV-vector platforms

PR #5354, opened 2025-07-01
Add Support for POWER11

pratiklp00:
This is the Initial patch to support build infrastructure for POWER11 architecture.


pratiklp00:
This failure dosent seems to be caused  by this PR, could you please help me with review.

driver/others/dynamic_power.c
```diff
@@ -70,6 +70,9 @@ static int cpuid(void)
 #endif
 #ifdef POWER_10
     else if (arch >= POWER_10) return CPU_POWER10;
```
ChipKerchner, line 72: @pratiklp00 Shouldn't this be

else if arch == POWER_10 return CPU_POWER10;

I only made it >= as a temp change till POWER11 came out.

pratiklp00: @ChipKerchner Yes Chip.

PR #5357, opened 2025-07-03
Add infrastructure for BGEMM

Mousius:
Setting up all the infrastructure for BGEMM support in OpenBLAS, hopefully I found all the right places.

Derived mostly from the previous work done in https://github.com/OpenMathLib/OpenBLAS/pull/5287


Mousius:
@martin-frbg looks like the tests are timing out due to not optimized variant, probably worth disabling them before we burn through a large amount of credits?

https://github.com/Mousius/OpenBLAS/commit/8cd4be8d47a797f7912ffbe6d377732c859a6501

martin-frbg:
Hmm, seems to be something wrong with test_bgemm in the DYNAMIC_ARCH case, sadly I missed that CI timeout... probably something wrong with the change to setparam-ref.c

martin-frbg:
reproducible on Neoverse N1 (and probably any other armv8 target) as well, so no need to burn credits... test seems to be looping with weird data even at the first step of the test_bgemm loop, while non-dynamic arch runs without issues.

PR #5360, opened 2025-07-06
Fix cpuid.S on arm

sertonix:
The ARM assembly syntax differs a bit

Fixes 61b9339d3a1f getarch/cpuid.S: Fix warning about executable stack

Ref https://github.com/OpenMathLib/OpenBLAS/discussions/5313#discussioncomment-13674817


martin-frbg:
Thank you

PR #5362, opened 2025-07-07
Fix SBGEMM BFLOAT16 build

Mousius:
Whilst developing https://github.com/OpenMathLib/OpenBLAS/pull/5357 I realised SBGEMM was failing tests, and I traced it back to https://github.com/xianyi/OpenBLAS/commit/ca31c32693bbb70cd8eeee5f2be09a7e9d1b363c - just a few places missed.

Hopefully this fixes them and also enables `BUILD_BFLOAT16` in cirun to catch this in future.


martin-frbg:
Eek. So that big rename is still the gift that keeps on giving... the BLAS_HALF in potrf_parallel.c is a lie as well, I'll fix it in a minute...

PR #5367, opened 2025-07-09
Temporarily disable test_bgemm

Mousius:


Mousius:
@martin-frbg I'll look at a proper fix next week.

PR #5371, opened 2025-07-10
Complete the infrastructure changes for adding BGEMM

### LINKED ISSUE #5370: build broken in compare_sgemm_bgemm.c for ppc64le after a recent change
Seems that commit f95e7b0 breaks build on ppc64le with at least the POWER8 kernel as it hardcodes an `#include <arm_neon.h>`
````
...
compare_sgemm_bgemm.c:31:10: fatal error: arm_neon.h: No such file or directory
   31 | #include <arm_neon.h>
      |          ^~~~~~~~~~~~
compilation terminated.
...
````

martin-frbg:
Fixes improper operation in DYNAMIC_ARCH builds and fixes #5370 while also restoring the test disabled in #5367

### DIFF
--- a/driver/others/parameter.c
+++ b/driver/others/parameter.c
@@ -72,6 +72,11 @@ BLASLONG shgemm_p = DEFAULT_GEMM_P;
 #else
 BLASLONG shgemm_p = SHGEMM_P;
 #endif
+#if BGEMM_P == bgemm_p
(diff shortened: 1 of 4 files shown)


Mousius:
There are a lot of places to update for a new operator - thanks for the save @martin-frbg!

PR #5380, opened 2025-07-15
SME1 based direct kernel (with alpha and beta) for cblas_sgemm level 3

matcraje:
This PR contains support for sgemm_direct kernel (with support for alpha and beta) based on SME1 architecture.


martin-frbg:
Thanks - this looks like an interesting addition (if not competitor) to the present sgemm_direct kernel. The CI error suggests that perhaps the entire kernel needs to be guarded with the __ARM_FEATURE_SME define (or the ...2VLx2VL function should have an empty alternative for when that feature macro is undefined) ?
Also, could you please review the attribution at the top of the kernel file - you are certainly welcome to put your copyright statement there, but I expect having a "confidential and proprietary" warning in the middle of a BSD-licensed code base would be confusing

matcraje:
Hi Martin. Thanks for your quick review. It’s very helpful.

I have updated the PR to address the two issues. (Add empty alternative for when feature macro is undefined. Modify the copyright statement)

martin-frbg:
seems now we have AppleClang acting up over things in its own arm_sme header file

matcraje:
> seems now we have AppleClang acting up over things in its own arm_sme header file

From the error log, I understand that error coming from clang-15.0.0 not supporting 'arm_streaming_' attributes. I tried to reproduce locally but found sme isn't supported in clang-15.


For mitigation, I am adding extra guard along with __ARM_FEATURE_SME.

I have pushed below update.
#if defined(__ARM_FEATURE_SME) && defined(__ARM_FEATURE_LOCALLY_STREAMING)

If the above doesn't work, I can think of explicitly checking clang version.
#if defined(__ARM_FEATURE_SME) && defined(__clang__) && __clang_major__ >= 16

matcraje:
@martin-frbg 
Two checks are still failing. Can you please let me know how to resolve these failures.

martin-frbg:
Both are unrelated - the loongarch job ran out of time and the IBM-Z build on Jenkins failed to access github (and still does today)

matcraje:
Thanks!

Issue #5388, opened 2025-07-21
Having troubles cross compiling with RISCV64 and cmake


ChipKerchner:
I'm trying to setup OpenBLAS to cross compiling from x86_64 to RISCV_ZVL256B.  I'm getting errors with cmake.  Can someone help with this?

export FC=riscv64-unknown-linux-gnu-gfortran
export CC=riscv64-unknown-linux-gnu-gcc
export CXX=riscv64-unknown-linux-gnu-g++
export AR=riscv64-unknown-linux-gnu-ar
export AS=riscv64-unknown-linux-gnu-gcc
export LD=riscv64-unknown-linux-gnu-ld
export HOSTCC=gcc
export HOSTFC=gfortran
export CXXFLAGS="-DTARGET=RISCV64_ZVL256B"
export CFLAGS="-DTARGET=RISCV64_ZVL256B"
export LD_LIBRARY_PATH=/home/ckerchner/tools/tt-riscv-toolchain-ae8a01f3/sysroot/lib

cmake -DTARGET=RISCV64_ZVL256B ..

  /home/ckerchner/tools/tt-riscv-toolchain-ae8a01f3/bin/riscv64-unknown-linux-gnu-gcc
  -DTARGET=RISCV64_ZVL256B -DFORCE_RISCV64_ZVL256B -march=native
  -DGEMM_MULTITHREAD_THRESHOLD=4 -DNO_AVX512 -DNO_PARALLEL_MAKE=0
  -I"/proj_sw/user_dev/ckerchner/OpenBLAS/build/getarch_build"
  -I"/proj_sw/user_dev/ckerchner/OpenBLAS"
  -I"/proj_sw/user_dev/ckerchner/OpenBLAS/build" -o
  CMakeFiles/cmTC_a895f.dir/getarch.c.o -c
  /proj_sw/user_dev/ckerchner/OpenBLAS/getarch.c

  riscv64-unknown-linux-gnu-gcc: error: '-march=native': ISA string must

This seems to be the offending area in system.cmake

# On x86_64 build getarch with march=native. This is required to detect AVX512 support in getarch.
if (X86_64 AND NOT (${CMAKE_C_COMPILER_ID} STREQUAL "PGI" OR ${CMAKE_C_COMPILER_ID} STREQUAL "NVHPC"))
  set(GETARCH_FLAGS "${GETARCH_FLAGS} -march=native")
endif ()

[...]

PR #5389, opened 2025-07-21
Add cross-compilation parameters for RISCV64 targets in CMake

martin-frbg:
for #5388

### DIFF
--- a/cmake/prebuild.cmake
+++ b/cmake/prebuild.cmake
@@ -1407,6 +1407,78 @@ endif ()
     set(ZGEMM_UNROLL_M 8)
     set(ZGEMM_UNROLL_N 2)
     set(SYMV_P 8)
+  elseif ("${TCORE}" STREQUAL "C910V")
+    file(APPEND ${TARGET_CONF_TEMP}
+      "#define L1_DATA_SIZE 32768\n"
+      "#define L1_DATA_LINESIZE 32\n"
+      "#define L2_SIZE 1048576\n"
+      "#define L2_LINESIZE 32 \n"
+      "#define DTB_DEFAULT_ENTRIES 128\n"
+      "#define DTB_SIZE 4096\n"
+      "#define L2_ASSOCIATIVE 4\n")
+    set(SGEMM_UNROLL_M 16)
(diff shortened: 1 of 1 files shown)


cmake/prebuild.cmake
```diff
@@ -1407,6 +1407,78 @@ endif ()
...
+    set(CGEMM_UNROLL_M 8)
+    set(CGEMM_UNROLL_N 4)
+    set(ZGEMM_UNROLL_M 4)
+    set(ZGEMM_UNROLL_N 4)
+    set(SYMV_P 16)
+  elseif ("${TCORE}" STREQUAL "RISCV64_ZVL256B")
+    file(APPEND ${TARGET_CONF_TEMP}
+      "#define L1_DATA_SIZE 65536\n"
+      "#define L1_DATA_LINESIZE 32\n"
+      "#define L2_SIZE 262144\n"
```
ChipKerchner, line 1468: @martin-frbg Is this right?  Seems a bit small for L2

cmake/prebuild.cmake
```diff
@@ -1407,6 +1407,78 @@ endif ()
...
+    set(CGEMM_UNROLL_M 2)
+    set(CGEMM_UNROLL_N 2)
+    set(ZGEMM_UNROLL_M 2)
+    set(ZGEMM_UNROLL_N 2)
+    set(SYMV_P 16)
+  elseif ("${TCORE}" STREQUAL "x280")
+    file(APPEND ${TARGET_CONF_TEMP}
+      "#define L1_DATA_SIZE 65536\n"
+      "#define L1_LINESIZE 32 \n"
+      "#define L2_SIZE 262144\n"
```
ChipKerchner, line 1432: Same here

cmake/prebuild.cmake
ChipKerchner: Maybe 2097152?

martin-frbg:
You would probably know - not sure offhand where I got that number from, likely I did not find an official datasheet or other reliable source and just put a "safe" smallish number that should fit all kinds of hardware configurations

PR #5393, opened 2025-07-23
Bump xuantie toolchains

xctan:


martin-frbg:
hmm. there was something fishy about the test_fork being surprisingly prone to hanging before, but at least it didn't segfault. does the newer toolchain work for you on actual hardware ?

xctan:
I can confirm this segfault occurs on physical hardware. After debugging with GDB, I suspect the issue stems from a subtle linking problem that is producing an incorrectly built test binary.

martin-frbg:
Curiously I do not get the segfault on C910V hardware (MILKV Pioneer) with binaries from that new version of the cross-compiler (built on Debian11).

PR #5396, opened 2025-07-23
ARM64: Enable bfloat16 kernels by default

abhishek-iitmadras:
This patch enables bfloat16 kernels by default for ARM64 processors.


martin-frbg:
Can get rid of the "duplicate -lgfortran" error by using `$(filter-out -lgfortran,$(EXTRALIB))` instead of $(EXTRALIB) in the build commands but it's not immediately clear to me why this comes up now

abhishek-iitmadras:
> Can get rid of the "duplicate -lgfortran" error by using `$(filter-out -lgfortran,$(EXTRALIB))` instead of $(EXTRALIB) in the build commands but it's not immediately clear to me why this comes up now

- the **macOS + make** jobs link the BLAS regression tests (test_bgemv, test_sbgemv, …) with Apple clang without telling the linker where Home-brew’s libgfortran.dylib lives.
The Home-brew path /opt/homebrew/Cellar/gcc/<ver>/lib/gcc/current is not searched by default, so ld64 aborts with `ld: library 'gfortran' not found.`

- The `duplicate -lgfortran` notice is only a warning; the link fails because the library is missing, not because it appears twice.

martin-frbg:
Right, thanks. Filtering out the redundant `-lgfortran` did appear to work as well - but I haven't checked where the "other" instance of `-lgfortran` is coming from (must be from `f_check`'s querying of Fortran dependencies, if it happens to have the correct `-L` as well). So not sure what's going to be the final solution (seems that the homebrew script does not add the Cellar path to the LIBRARY_PATH by default, so this could be a gotcha for others too). That's unrelated to your PR though, so merging - thank you.

PR #5399, opened 2025-07-24
Add optimized BGEMM for NEOVERSEN2 target

Mousius:
This re-uses the existing NEOVERSEN2 8x4 `sbgemm` kernel to implement `bgemm`.

### DIFF
--- a/kernel/arm64/KERNEL.NEOVERSEN2
+++ b/kernel/arm64/KERNEL.NEOVERSEN2
@@ -188,6 +188,20 @@ ZGEMMOTCOPY    =  ../generic/zgemm_tcopy_$(ZGEMM_UNROLL_N).c
 ZGEMMONCOPYOBJ =  zgemm_oncopy$(TSUFFIX).$(SUFFIX)
 ZGEMMOTCOPYOBJ =  zgemm_otcopy$(TSUFFIX).$(SUFFIX)
 
+ifeq ($(BUILD_BFLOAT16), 1)
+BGEMM_BETA    =  sbgemm_beta_neoversen2.c
+BGEMMKERNEL    = sbgemm_kernel_$(BGEMM_UNROLL_M)x$(BGEMM_UNROLL_N)_neoversen2.c
+BGEMMINCOPY    = sbgemm_ncopy_$(BGEMM_UNROLL_M)_neoversen2.c
+BGEMMITCOPY    = sbgemm_tcopy_$(BGEMM_UNROLL_M)_neoversen2.c
+BGEMMONCOPY    = sbgemm_ncopy_$(BGEMM_UNROLL_N)_neoversen2.c
+BGEMMOTCOPY    = sbgemm_tcopy_$(BGEMM_UNROLL_N)_neoversen2.c
+BGEMMINCOPYOBJ =  bgemm_incopy$(TSUFFIX).$(SUFFIX)
+BGEMMITCOPYOBJ =  bgemm_itcopy$(TSUFFIX).$(SUFFIX)
+BGEMMONCOPYOBJ =  bgemm_oncopy$(TSUFFIX).$(SUFFIX)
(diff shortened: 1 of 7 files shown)


mattip:
I think there is a missing include here: it does not build in the weekly openblas-libs tests because `vcvtah_f32_bf16` is not declared. [It seems](https://scalable.uni-jena.de/opt/hpc/chapters/assignment_bfloat16.html) including both `#include <arm_bf16.h>` and `#include <arm_neon.h>` is needed. The failed run is [here](https://github.com/MacPython/openblas-libs/actions/runs/16700000662/job/47269507749). Is there something different about the openblas-libs CI that is missing those includes?

martin-frbg:
maybe it is a toolchain question, or you are using additional code checking options ? I have only limited options for testing the most recent Neoverse cpus - our Cirun job uses an Ubuntu Jammy image that appears to be stuck at gcc11, and the most modern hardware in the GCC Compile Farm is a N1. The code in question still appears to compile on my Pixel8 phone with gcc-15 though

Mousius:
@mattip potentially a bigger issue is that this is building `BGEMM`/`SBGEMM` and `GEMV` variants by default thanks to https://github.com/OpenMathLib/OpenBLAS/pull/5396 - which likely increases the binary size for no gain to numpy.

mattip:
> maybe it is a toolchain question, or you are using additional code checking options?

This fails compilation on CI for macos-arm64. When I run it locally on a macbook M1, I do not see compilation of the `sbgemm_kernel_8x4_neoversen2.c` kernel.
```
CFLAGS=' -ftrapping-math -mmacos-version-min=11.0 -fvisibility=protected -Wno-uninitialized'
make BUFFERSIZE=20 DYNAMIC_ARCH=1 QUIET_MAKE=1 USE_OPENMP=0 NUM_THREADS=64 BINARY=64 INTERFACE64=1 SYMBOLSUFFIX=64_ LIBNAMESUFFIX=64_ OBJCONV=/Users/runner/work/openblas-libs/openblas-libs/objconv/objconv SYMBOLPREFIX=scipy_ LIBNAMEPREFIX=scipy_ FIXED_LIBNAME=1 TARGET=VORTEX
```

and the similar command on linux-arm64 fails tests, since it does not actually have bfloat
```
CFLAGS=' -fvisibility=protected -Wno-uninitialized'
2025-08-03T02:05:19.0182311Z + make BUFFERSIZE=20 DYNAMIC_ARCH=1 QUIET_MAKE=1 USE_OPENMP=0 NUM_THREADS=64 BINARY=64 OBJCONV=/io/objconv/objconv SYMBOLPREFIX=scipy_ LIBNAMEPREFIX=scipy_ FIXED_LIBNAME=1 TARGET=ARMV8
```

> building BGEMM/SBGEMM and GEMV variants by default

Right, we should probably use a `DYNAMIC_LIST` to not use the bfloat kernels on linux-arm64

mattip:
Ahh, the difference is that the CI run specifies `MACOSX_DEPLOYMENT_TARGET="11.0"`, which then allows SVE and SME here https://github.com/OpenMathLib/OpenBLAS/blob/d23680b81d5179ce6ae1ca5546303b81646ecac1/Makefile.system#L425-L431

So for me a minimal reproducer for the build failure is this. Maybe the CI here does not care about the undefined function warning.
```
export MACOSX_DEPLOYMENT_TARGET="11.0"
export CFLAG=-Werror
make TARGET=NEOVERSEN2
```

martin-frbg:
probably the xcode 15.4 toolchain ?
and dynamic_list won't change anything about bfloat16 support, you need to build with BUILD_BFLOAT16=0

martin-frbg:
Looks like arm_neon.h should be included, while arm_bf16.h would be included automatically from either this or arm_sve.h if needed - #5396 had already fixed this in the N1/V1 kernels, but not here

PR #5408, opened 2025-07-30
Fix bad vector zero initializer and other compiler warnings for RISC-V.

ChipKerchner:
Fix bad vector zero initializer and other compiler warnings for RISC-V.

Previous vector zero initializer didn't handle NaN and Inf correctly.


ChipKerchner:
[5211](https://github.com/OpenMathLib/OpenBLAS/pull/5211)

martin-frbg:
Thank you

PR #5411, opened 2025-07-30
CI: Update WoA job to use LLVM 20.1.8 and avoid stray preinstalled LLVM19

martin-frbg:


Greenie0701:
@martin-frbg any reason why Windows ARM64 CI are failing?

martin-frbg:
somehow the current 20.1.6 based CI job appears to have lost its (intrinsic) iso_fortran_env module - I have not seen this problem locally. should be fixed when I merge this update

Greenie0701:
> somehow the current 20.1.6 based CI job appears to have lost its (intrinsic) iso_fortran_env module - I have not seen this problem locally. should be fixed when I merge this update

Thanks for the update

martin-frbg:
actually, for some weird reason, the pre-installed MS compiler seems to have grown an outdated installation of LLVM19 in its Tools directory lately, which took precedence over the 20.1.x due to the vcvars script's overwriting of our preset PATH

PR #5423, opened 2025-08-18
Split VORTEXM4 from VORTEX target and fix SGEMM_DIRECT support for SME-capable targets

### LINKED ISSUE #5414: test_extensions/test_sgemmt.c fails with SME on Apple M4
Building commit d23680b81d5179ce6ae1ca5546303b81646ecac1 with `make -j DYNAMIC_ARCH=1 TARGET=VORTEX` results in test failures on Apple M4:

```
TEST 1135/1522 sgemmt:c_api_rowmajor_upper_M_50_K_50_a_notrans_b_notrans [FAIL]
  ERR: test_extensions/test_sgemmt.c:797  expected 0.000e+00, got 2.741e-01 (diff -2.741e-01, tol 1.000e-04)
TEST 1139/1522 sgemmt:c_api_rowmajor_upper_alpha_zero [FAIL]
  ERR: test_extensions/test_sgemmt.c:888  expected 0.000e+00, got 8.187e-03 (diff -8.187e-03, tol 1.000e-04)
TEST 1140/1522 sgemmt:c_api_rowmajor_upper_beta_one [FAIL]
  ERR: test_extensions/test_sgemmt.c:910  expected 0.000e+00, got 3.662e-01 (diff -3.662e-01, tol 1.000e-04)
TEST 1141/1522 sgemmt:c_api_rowmajor_lower_M_50_K_50_a_notrans_b_notrans [FAIL]
  ERR: test_extensions/test_sgemmt.c:933  expected 0.000e+00, got 2.784e-01 (diff -2.784e-01, tol 1.000e-04)
TEST 1145/1522 sgemmt:c_api_rowmajor_lower_alpha_zero [FAIL]
  ERR: test_extensions/test_sgemmt.c:1024  expected 0.000e+00, got 8.250e-03 (diff -8.250e-03, tol 1.000e-04)
TEST 1146/1522 sgemmt:c_api_rowmajor_lower_beta_one [FAIL]
  ERR: test_extensions/test_sgemmt.c:1046  expected 0.000e+00, got 3.539e-01 (diff -3.539e-01, tol 1.000e-04)
```

These failures only occur with the SME SGEMM direct kernel. I no longer see the test failures if I disable SME support with the following patch:

```
diff --git a/interface/gemm.c b/interface/gemm.c
index 54e5604fd..cde3038e6 100644
--- a/interface/gemm.c
+++ b/interface/gemm.c
@@ -429,7 +429,7 @@ void CNAME(enum CBLAS_ORDER order, enum CBLAS_TRANSPOSE TransA, enum CBLAS_TRANS
 #endif
 #if defined(ARCH_ARM64) && (defined(USE_SGEMM_KERNEL_DIRECT)||defined(DYNAMIC_ARCH))
 #if defined(DYNAMIC_ARCH)
- if (support_sme1())
+ if (false)
 #endif
   if (beta == 0 && alpha == 1.0 && order == CblasRowMajor && TransA == CblasNoTrans && TransB == CblasNoTrans) {
 	SGEMM_DIRECT(m, n, k, a, lda, b, ldb, c, ldc);
```

This was initially reported on the [numpy bug tracker](https://github.com/numpy/numpy/issues/29393).

### LINKED ISSUE #5528: cblas_sgemm has wrong result on apple M4
0.3.30 version
It seems that cblas_sgemm() always output the zero result.
```
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <cblas.h>
#include <math.h>

void generate_random_matrix(float *matrix, int rows, int cols) {
    for (int i = 0; i < rows; i++) {
        for (int j = 0; j < cols; j++) {
            matrix[i * cols + j] = (float)rand() / RAND_MAX * 2.0f - 1.0f; // [-1, 1]
        }
    }
}

void print_matrix(const char *name, float *matrix, int rows, int cols) {
    printf("%s (%dx%d):\n", name, rows, cols);
    for (int i = 0; i < rows; i++) {
        for (int j = 0; j < cols; j++) {
            printf("%8.4f ", matrix[i * cols + j]);
        }
        printf("\n");
    }
    printf("\n");
}

int main() {
    const int M = 3;  // out_channels - 
    const int N = 4;  // tiles - 
    const int K = 2;  // channels - 
    
    printf("Matrix dimensions: M=%d, N=%d, K=%d\n", M, N, K);
    printf("Computing: C(MxN) = A(MxK) * B(KxN)\n\n");
    
    float *A = (float*)malloc(M * K * sizeof(float));
    float *B = (float*)malloc(K * N * sizeof(float));
    float *C_cblas = (float*)malloc(M * N * sizeof(float));
    float *C_ref = (float*)malloc(M * N * sizeof(float));
    
    srand(42); // 
    generate_random_matrix(A, M, K);
    generate_random_matrix(B, K, N);
    
    memset(C_cblas, 0, M * N * sizeof(float));
    
    print_matrix("Matrix A", A, M, K);
    print_matrix("Matrix B", B, K, N);
    

    const int lda = K;  
    const int ldb = N;  
    const int ldc = N;  
    
    printf("Leading dimensions: lda=%d, ldb=%d, ldc=%d\n", lda, ldb, ldc);
    
    printf("Calling cblas_sgemm...\n");
    cblas_sgemm(CblasRowMajor, CblasNoTrans, CblasNoTrans, 
                M, N, K,                    // M, N, K
                1.0,                        // alpha
                A, lda,                     // A, lda
                B, ldb,                     // B, ldb  
                0.0,                        // beta
                C_cblas, ldc);              // C, ldc
    
    print_matrix("CBLAS Result", C_cblas, M, N);
    
    free(A);
    free(B);
    free(C_cblas);
    
    return 0;
}
```
and output
```
Matrix dimensions: M=3, N=4, K=2
Computing: C(MxN) = A(MxK) * B(KxN)

Matrix A (3x2):
 -0.9993   0.0492 
  0.4708  -0.4734 
 -0.2476  -0.6074 

Matrix B (2x4):
  0.9517   0.0246   0.0609  -0.4858 
 -0.7858   0.6310   0.8011  -0.0959 

Leading dimensions: lda=2, ldb=4, ldc=4
Calling cblas_sgemm...
CBLAS Result (3x4):
  0.0000   0.0000   0.0000   0.0000 
  0.0000   0.0000   0.0000   0.0000 
  0.0000   0.0000   0.0000   0.0000 

```

### LINKED ISSUE #5585: new SME kernels are breaking clang on Windows
Trying to build the new SME kernels fails with this error in llvm-21:
```
error: Incorrect size for ssyrk_direct_sme1_2VLx2VL prologue: 108 bytes of instructions in range, but .seh directives corresponding to 96 bytes
```

This seems to be fixed in llvm-22 (The newer version moves the rdsvl after .seh_endprologue, avoiding the SEH metadata issue.)

But it runs into this error in llvm-22:
```
  LLVM ERROR: SME hazard padding is not supported on Windows
```

Further analysis of the PR suggests the combination of alloca with SME is not implemented on Windows:

  | LLVM Version    | Behavior                                                            |
  |-----------------|---------------------------------------------------------------------|
  | 21.1.7 (old)    | Generates invalid .seh_save_reg x9/x0 directives                    |
  | 22.0.0git (new) | Explicitly errors: "SME hazard padding is not supported on Windows" |

  The newer LLVM "fixed" the assembler issues by rejecting the combination entirely rather than fixing the SEH generation. The check is in AArch64PrologueEpilogue.cpp: when hasSVECalleeSavesAboveFrameRecord and hasStackHazardSlotIndex are both true.

fatal usage error added in https://github.com/llvm/llvm-project/pull/138609/files#diff-abc2806d934cd854992bdf139a4ab9405859556f01b2fc4ab17aa6b8a50e72c4R1902-R1918

https://github.com/llvm/llvm-project/pull/156467/files#diff-3da805da491ea568bb4b1b76dcbf5066b69923612f30e6cf97737f7eb30194cfR99

```
$ /home/jameson/llvm-mingw/build/bin/aarch64-w64-mingw32-clang -g -DSMALL_MATRIX_OPT -DGEMM_GEMV_FORWARD -DSBGEMM_GEMV_FORWARD -DBGEMM_GEMV_FORWARD -DMS_ABI -DMAX_STACK_ALLOC=2048 -Wall -DF_INTERFACE_GFORT -DDYNAMIC_ARCH -DSMP_SERVER -DNO_WARMUP -DMAX_CPU_NUMBER=512 -DMAX_PARALLEL_NUMBER=1 -DBUILD_BFLOAT16 -DBUILD_SINGLE=1 -DBUILD_DOUBLE=1 -DBUILD_COMPLEX=1 -DBUILD_COMPLEX16=1 -DVERSION=\"0.3.30.dev\" -march=armv9-a+sve2+sme -UASMNAME -UASMFNAME -UNAME -UCNAME -UCHAR_NAME -UCHAR_CNAME -DASMNAME=ssyrk_direct_alpha_betaLT_ARMV9SME -DASMFNAME=ssyrk_direct_alpha_betaLT_ARMV9SME_ -DNAME=ssyrk_direct_alpha_betaLT_ARMV9SME_ -DCNAME=ssyrk_direct_alpha_betaLT_ARMV9SME -DCHAR_NAME=\"ssyrk_direct_alpha_betaLT_ARMV9SME_\" -DCHAR_CNAME=\"ssyrk_direct_alpha_betaLT_ARMV9SME\" -DNO_AFFINITY -DTS=_ARMV9SME -I.. -DBUILD_KERNEL -DTABLE_NAME=gotoblas_ARMV9SME -DHAVE_SME -march=armv9-a+sve2+sme -DBUILD_KERNEL -DTABLE_NAME=gotoblas_ARMV9SME -UDOUBLE  -UCOMPLEX -UDOUBLE -UCOMPLEX -UUPPER -DTRANSA ../kernel/arm64/ssyrk_direct_alpha_beta_arm64_sme1.c -Wclang -o /dev/null -c
warning: unknown warning option '-Wclang'; did you mean '-Wvla'? [-Wunknown-warning-option]
In file included from ../kernel/arm64/ssyrk_direct_alpha_beta_arm64_sme1.c:6:
In file included from ../common.h:61:
../config_kernel.h:22:9: warning: 'HAVE_SME' macro redefined [-Wmacro-redefined]
   22 | #define HAVE_SME
      |         ^
<command line>:35:9: note: previous definition is here
   35 | #define HAVE_SME 1
      |         ^
In file included from ../kernel/arm64/ssyrk_direct_alpha_beta_arm64_sme1.c:6:
../common.h:378:9: warning: 'YIELDING' macro redefined [-Wmacro-redefined]
  378 | #define YIELDING        __asm__ __volatile__ ("nop;nop;nop;nop;nop;nop;nop;nop; \n");
      |         ^
../common.h:373:9: note: previous definition is here
  373 | #define YIELDING        SwitchToThread()
      |         ^
../kernel/arm64/ssyrk_direct_alpha_beta_arm64_sme1.c:191:35: warning: passing 'const float *' to parameter of type 'float *' discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
  191 |       kernel_2x2(&a_ptr[row_idx], &b_ptr[col_idx],
      |                                   ^~~~~~~~~~~~~~~
../kernel/arm64/ssyrk_direct_alpha_beta_arm64_sme1.c:35:35: note: passing argument to parameter 'B' here
   35 | kernel_2x2(const float *A, float *B, float *C, size_t shared_dim,
      |                                   ^
../kernel/arm64/ssyrk_direct_alpha_beta_arm64_sme1.c:21:17: warning: unused function 'sve_cntw' [-Wunused-function]
   21 | static uint64_t sve_cntw() {
      |                 ^~~~~~~~
error: Incorrect size for ssyrk_direct_sme1_2VLx2VL prologue: 108 bytes of instructions in range, but .seh directives corresponding to 96 bytes

5 warnings and 1 error generated.
```

It appears the code may be considered to have alloca because of the use of `vscale` functions

This is clearly an upstream bug, but it might be necessary to disable those new AArch64 kernels on Windows with Clang?

martin-frbg:
fixes #5414
fixes #5528
fixes #5585


kernel/arm64/sgemm_direct_sme1_2VLx2VL.S
```diff
@@ -35,16 +35,17 @@
...
 #define M_cntr       x16 //M loop counter
 #define C1           x17 //Constant1: N*(SVLs+1);SVLs-No. of 32-bit elements
-#define C2           x18 //Constant2: N + SVLs
-#define C3           x19 //Constant3: K*SVLs + SVLs
-#define C4           x20 //Constant4: SVLs-2
-#define C5           x21 //Constant5: K*SVLs
-#define C6           x22 //Constant6: N*SVLs
+#define C2           x19 //Constant2: N + SVLs
+#define C3           x20 //Constant3: K*SVLs + SVLs
+#define C4           x21 //Constant4: SVLs-2
```
matcraje, line 40: Modifying x20 to x21 will require below dependent changes.
At line 65: sub w21, w21, #2
At line 202: cmp w13, w21

martin-frbg: Oops, sorry, I had already corrected this locally but pushed the wrong version. Unfortunately this correction has no effect on the wrong xscblat3 test results seen for M odd (and contrary to my expectations this PR also does not fix the divergence between SGEMM and SGEMMT seen in test_sgemmt of utest/openblas_utest_ext that was flagged in #5414)

interface/gemm.c
```diff
@@ -564,6 +541,36 @@ void CNAME(enum CBLAS_ORDER order, enum CBLAS_TRANSPOSE TransA, enum CBLAS_TRANS
...
+        SGEMM_DIRECT(m, n, k, a, lda, b, ldb, c, ldc);
+        return;
+  }
+#endif
+#if defined(ARCH_ARM64) && (defined(USE_SGEMM_KERNEL_DIRECT)||defined(DYNAMIC_ARCH))
+#if defined(DYNAMIC_ARCH)
+if (strcmp(gotoblas_corename(), "armv9sme") == 0 || strcmp(gotoblas_corename(), "vortexm4") == 0)
+// if (support_sme1())
+#endif
+  if (order == CblasRowMajor && m==lda && n ==ldb && k==ldc && beta == 0 && alpha == 1.0 && TransA == CblasNoTrans && TransB == CblasNoTrans&& SGEMM_DIRECT_PERFORMANT(m,n,k)) {
```
matcraje, line 565: For RowMajor, shouldn't the leading dimension check be (lda==k && ldb==n && ldc==n) ?

kernel/arm64/sgemm_direct_performant.c
```diff
@@ -0,0 +1,31 @@
+#include "common.h"
+/* helper for the direct sgemm code written by Arjan van der Ven */
+
+
+
+
+int CNAME(BLASLONG M, BLASLONG N, BLASLONG K)
```
matcraje, line 4: So, this helper function checks for when the SME implementation wouldn't be performant?
Are these checks applicable explicitly for Apple M4 only?

martin-frbg: In principle I'd expect them to be relevant for future SME hardware as well - I think it is unlikely that the direct path will outperform Goto's block algorithm at **any** matrix size and shape (and we should have an SME GEMM kernel compatible with that at some point - there already is a draft PR that only lacks the TRMM part).
This is just a quick copy of the x86_64 implementation for now, so numbers will need to be tuned once we're certain that the codes are correct

interface/gemm.c
martin-frbg: normally yes but arguments have already been reshuffled at this point (I think - I'll recheck when I get back to this later this week)

PR #5425, opened 2025-08-21
Increase L2 defaults for RISCV X280 / ZVL256B and ARM SVE targets in CMake cross-compilation

martin-frbg:
addresses comments on #5389 - these targets are unlikely to be used in underpowered designs, so don't need overly "conservative" defaults that can hurt performance


ChipKerchner:
Great.  Those were the values I was going to suggest.

PR #5431, opened 2025-08-28
disable fp16 flags on RISC-V unless BUILD_HFLOAT16=1

### LINKED ISSUE #5428: RISC-V RISCV64_ZVL128B  and RISCV64_ZVL256B targets are broken
There are a number of issues here.

Commit [Added shgemm_kernel_8x8 for RISCV64_ZVL128B and shgemm_kernel_16x8 for RISCV64_ZVL256B](https://github.com/OpenMathLib/OpenBLAS/commit/670ec6f7576ecc74fff96be7c00ec8fffed8647b) changes the --march option used to build the RISCV64_ZVL128B and RISCV64_ZVL256B targets from

```
-march=rv64imafdcv
```
to
```
-march=rv64imafdcv_zvfh_zfh
```

The justification for this change in the commit message was

> The instruction sets used are ZVFH and ZFH, which need to be supported by RVV1.0

However, zvfh and zvh are not part of the V: Vector Extension for Application Processors.  There's no guarantee that a device that supports V will also support zvfh and zfh.  In fact it's likely that devices will not as zfh and zvfh are not mandatory extensions in any of the RVA profiles.  For example, in RVA23U64 they are both listed as [expansion options](https://github.com/riscv/riscv-profiles/blob/main/src/rva23-profile.adoc#expansion-options). 

~This means that the RISCV64_ZVL128B and RISCV64_ZVL256B targets built from the current develop will be unlikely to work on any real hardware for the foreseeable future.~ (This turned out not to be the case.  The Banana Pi BPI F3 does seem to have Zvfh and Zfh, but there's no guarantee that other devices support V will have these extensions).

Not only will they not work, but if you do a dynamic build, they'll probably SIGILL as the dynamic detection code [only checks for v](https://github.com/OpenMathLib/OpenBLAS/blob/develop/driver/others/dynamic_riscv64.c#L168) and not zvfh and zvh.

To make matters worse, the use of -march=rv64imafdcv_zvfh_zfh means that the RVV kernels can no longer be compiled with gcc-14 and so cannot be built in the current riscv64 manylinux images ( quay.io/pypa/manylinux_2_39_riscv64  ), meaning that it will not be possible to enable a riscv64 build of openblas-libs that includes the vectorized kernels.

In summary,

```
make TARGET="RISCV64_ZVL128B"
make TARGET="RISCV64_ZVL256B"
make TARGET="RISCV64_GENERIC" DYNAMIC_ARCH=1
```

Should all build with gcc-14 and the resulting binaries should work without issue on all devices fully supporting only rv64imafdcv (with the matching vlen).  Extra detection (and possibly makefile magic) is required to enable zvfh and zfh specific kernels.

Would it be possible to revert https://github.com/OpenMathLib/OpenBLAS/commit/670ec6f7576ecc74fff96be7c00ec8fffed8647b and re-introduce the functionality in a better way?

markdryan:
The compiler options that enable 16 bit floating point instructions
should not be enabled by default when building the RISCV64_ZVL128B
and RISCV64_ZVL256B targets.  The zfh and zvfh extensions are not part
of the 'V' extension and are not required by any of the RVA profiles.
There's no guarantee that kernels built with zfh and zvfh will work
correctly on fully compliant RVA23U64 devices.

To fix the issue we only build the RISCV64_ZVL128B and RISCV64_ZVL256B
kernels with the half float flags if BUILD_HFLOAT16=1.  We also update
the RISC-V dynamic detection code to disable the RISCV64_ZVL128B and
RISCV64_ZVL256B kernels at runtime if we've built with DYNAMIC_ARCH=1
and BUILD_HFLOAT16=1 and are running on a device that does not support
both Zfh and Zvfh.

Fixes: https://github.com/OpenMathLib/OpenBLAS/issues/5428

### DIFF
--- a/Makefile.prebuild
+++ b/Makefile.prebuild
@@ -64,11 +64,11 @@ TARGET_FLAGS = -march=rv64imafdcv_zba_zbb_zfh -mabi=lp64d
 endif
 
 ifeq ($(TARGET), RISCV64_ZVL256B)
-TARGET_FLAGS = -march=rv64imafdcv_zvfh_zfh -mabi=lp64d
+TARGET_FLAGS = -march=rv64imafdcv -mabi=lp64d
 endif
 
 ifeq ($(TARGET), RISCV64_ZVL128B)
-TARGET_FLAGS = -march=rv64imafdcv_zvfh_zfh -mabi=lp64d
+TARGET_FLAGS = -march=rv64imafdcv -mabi=lp64d
 endif
 
 ifeq ($(TARGET), RISCV64_GENERIC)
--- a/driver/others/dynamic_riscv64.c
+++ b/driver/others/dynamic_riscv64.c
@@ -97,6 +97,8 @@ struct riscv_hwprobe {
 
 #define RISCV_HWPROBE_KEY_IMA_EXT_0	4
 #define		RISCV_HWPROBE_IMA_V		(1 << 2)
+#define		RISCV_HWPROBE_EXT_ZFH		(1 << 27)
+#define		RISCV_HWPROBE_EXT_ZVFH		(1 << 30)
 
 #ifndef NR_riscv_hwprobe
 #ifndef NR_arch_specific_syscall
@@ -147,6 +149,7 @@ char* gotoblas_corename(void) {
 }
 
 static gotoblas_t* get_coretype(void) {
+	uint64_t vector_mask;
 	unsigned vlenb = 0;
 
 #if !defined(OS_LINUX)
@@ -165,14 +168,23 @@ static gotoblas_t* get_coretype(void) {
 	};
 	int ret = syscall(NR_riscv_hwprobe, pairs, 1, 0, NULL, 0);
 	if (ret == 0) {
-		if (!(pairs[0].value & RISCV_HWPROBE_IMA_V))
+#if defined(BUILD_HFLOAT16)
+		vector_mask = (RISCV_HWPROBE_IMA_V | RISCV_HWPROBE_EXT_ZFH | RISCV_HWPROBE_EXT_ZVFH);
+#else
+		vector_mask = RISCV_HWPROBE_IMA_V;
+#endif
+		if ((pairs[0].value & vector_mask) != vector_mask)
 			return NULL;
 	} else {
+#if defined(BUILD_HFLOAT16)
+		return NULL;
+#else
 		if (!(getauxval(AT_HWCAP) & DETECT_RISCV64_HWCAP_ISA_V))
 			return NULL;
 
 		if (!detect_riscv64_rvv100())
 			return NULL;
+#endif
 	}
 
 	/*
--- a/Makefile.riscv64
+++ b/Makefile.riscv64
@@ -7,12 +7,22 @@ CCOMMON_OPT += -march=rv64imafdcv_zba_zbb_zfh_zvl512b -mabi=lp64d
 FCOMMON_OPT += -march=rv64imafdcv_zba_zbb_zfh -mabi=lp64d -static
 endif
 ifeq ($(CORE), RISCV64_ZVL256B)
+ifeq ($(BUILD_HFLOAT16), 1)
 CCOMMON_OPT += -march=rv64imafdcv_zvl256b_zvfh_zfh -mabi=lp64d
 FCOMMON_OPT += -march=rv64imafdcv_zvfh_zfh -mabi=lp64d
+else
+CCOMMON_OPT += -march=rv64imafdcv_zvl256b -mabi=lp64d
+FCOMMON_OPT += -march=rv64imafdcv -mabi=lp64d
+endif
 endif
 ifeq ($(CORE), RISCV64_ZVL128B)
-CCOMMON_OPT += -march=rv64imafdcv_zvfh_zfh -mabi=lp64d 
+ifeq ($(BUILD_HFLOAT16), 1)
+CCOMMON_OPT += -march=rv64imafdcv_zvfh_zfh -mabi=lp64d
 FCOMMON_OPT += -march=rv64imafdcv_zvfh_zfh -mabi=lp64d
+else
+CCOMMON_OPT += -march=rv64imafdcv -mabi=lp64d
+FCOMMON_OPT += -march=rv64imafdcv -mabi=lp64d
+endif
 endif
 ifeq ($(CORE), RISCV64_GENERIC)
 CCOMMON_OPT += -march=rv64imafdc -mabi=lp64d
--- a/driver/others/Makefile
+++ b/driver/others/Makefile
@@ -218,7 +218,7 @@ mulx.$(SUFFIX) : $(ARCH)/mulx.c
 	$(CC) $(CFLAGS) -c -DXDOUBLE -UCOMPLEX $< -o $(@F)
 
 detect_riscv64.$(SUFFIX): detect_riscv64.c
-	$(CC) $(CFLAGS) -c -march=rv64imafdcv_zvfh_zfh $< -o $(@F)
+	$(CC) $(CFLAGS) -c -march=rv64imafdcv $< -o $(@F)
 
 xerbla.$(PSUFFIX) : xerbla.c
 	$(CC) $(PFLAGS) -c $< -o $(@F)


Makefile.prebuild
```diff
@@ -63,18 +63,6 @@ ifeq ($(TARGET), x280)
 TARGET_FLAGS = -march=rv64imafdcv_zba_zbb_zfh -mabi=lp64d
 endif
 
-ifeq ($(TARGET), RISCV64_ZVL256B)
```
markdryan, line 66: Does anyone see any issue with removing this?

driver/others/dynamic_riscv64.c
```diff
@@ -165,9 +168,17 @@ static gotoblas_t* get_coretype(void) {
...
+		vector_mask = (RISCV_HWPROBE_IMA_V | RISCV_HWPROBE_EXT_ZFH | RISCV_HWPROBE_EXT_ZVFH);
+#else
+		vector_mask = RISCV_HWPROBE_IMA_V;
+#endif
+		if ((pairs[0].value & vector_mask) != vector_mask)
 			return NULL;
 	} else {
+#if defined(BUILD_HFLOAT16)
+		return NULL;
+#endif
```
markdryan: I think I should probably put a #else here.

sergei-lewis: It's fine as is, no? If BUILD_HFLOAT16 isn't defined, we just get the existing pre-your-changes behaviour

Makefile.prebuild
martin-frbg: I don't think you want to remove these entirely, just revert them to their state before the addition of the HFLOAT16 PR (i.e. rv64imafdcv), unless we can be absolutely certain that the getarch utility and everything downstream of c_check still works on any affected RISCV64 platform (I'm especially worried about the `-mabi=lp64d` part) ?

markdryan: I was removing these to fix 

https://github.com/OpenMathLib/OpenBLAS/issues/5424

but perhaps this is not the way to do this.  Perhaps it's best to remove

https://github.com/OpenMathLib/OpenBLAS/pull/5431/commits/05c8654d0a3b6ef1339989678bfc1b15673bb27f

from this PR and fix the issue separately.

markdryan: I've dropped the commit that removes these lines entirely.  The remaining commit just removes the half float flags.

driver/others/dynamic_riscv64.c
martin-frbg: Hmm. Aren't we dropping out of the entire "vector extension supported" case with this change, when "only" the fp16 shgemm kernel is not supported by the detected hardware ? (Though I don't have an easy solution for how to keep using the others regardless, and switch just shgemm to the generic C implementation. Perhaps "rvv+fp16" would need to be a separate TARGET even)

markdryan: For the case where DYNAMIC_ARCH=1, yes.  The approach this patch takes is if we've built with DYNAMIC_ARCH=1 and BUILD_HFLOAT16=1 the vector kernels will only be used if we can detect Zfh, Zvfh and V at runtime, which seems reasonable as this is how they've been compiled.  Not ideal perhaps, but better than the current code on develop where a DYNAMIC_ARCH=1 build will happily execute the vector kernels on machines without half float support and probably crash.

martin-frbg: agree for shgemm but not sure why you'd expect the vector kernels in general to crash if Z(v)fh isn't available ?

markdryan: > agree for shgemm but not sure why you'd expect the vector kernels in general to crash if Z(v)fh isn't available ?

Sorry, poor choice of words.  By "vector kernels" I meant the RISCV64_ZVL128B and RISCV64_ZVL256B targets which IIUC will contain shgemm code when  BUILD_HFLOAT16=1.

martin-frbg: but these would crash only when actually executing SHGEMM - while your change makes them inexplicably slow for any other BLAS operation if only the build host happened to have these extensions. This might create a problem with third-party packagers (like Linux distributions) depending on how likely they are to active the BUILD_HFLOAT16 option

markdryan: Currently all the kernels in RISCV64_ZVL128B and RISCV64_ZVL256B are compiled with Zfh, Zvfh and V but at runtime we only check for V before executing them.  In general I don't think it's a good idea to execute code compiled with optional extensions like Zfh and Zvfh without first checking to see whether the extensions are supported at runtime.  Even if the other full float kernels aren't explicitly using half float instructions, by compiling them with Zfh and Zvfh, we give the compiler permission to use the half float instructions in those kernels, should it find some clever reason to do so.

> This might create a problem with third-party packagers (like Linux distributions) depending on how likely they are to active the BUILD_HFLOAT16 option

I think it's very unlikely that Linux distributions will be able to activate BUILD_HFLOAT16=1 in its current form for riscv64 any time soon as Zvfh and Zfh are not mandatory in any of the existing RVA profiles, even RVA23U64.  They could only safely build with BUILD_HFLOAT16=1 if runtime detection of zfh and zvfh were performed. 

I'm happy to explore another solution, e.g., introducing new targets as you suggested above.

Incidentally, what is done on X86?  How are bf16 and fp16 handled?

martin-frbg: On x86_64 (and power), bf16 capability is tied to TARGET, and fp16 is as yet unimplemented - it made its debut in the riscv64 PR.
Perhaps what could be done here is issue a warning about lack of Zfh (and consequentially SHGEMM) support before activating the RISCV64_ZVL128B or RISCV64_ZVL256B kernel regardless ?

martin-frbg: Actually I think I'll just add a FALLBACK_VERBOSE type message (as a followup commit) so that the user has a chance to find out why performance is lower than expected

markdryan: Yes, that sounds like a good idea.

Apologies for not replying to your previous comment.  I got a little distracted.

ChipKerchner:
Does this also need to be done for BUILD_BFLOAT16 since it now has vector support for SBGEMM?

markdryan:
> Does this also need to be done for BUILD_BFLOAT16 since it now has vector support for SBGEMM?

By this do you mean update the [dynamic_riscv64.c](https://github.com/OpenMathLib/OpenBLAS/blob/develop/driver/others/dynamic_riscv64.c#L171) code to check for the bfloat16 extensions?  I'd say we do need to do that, particularly as the BF16 extensions are not mandatory in RVA23.

martin-frbg:
The runtime check for extension support in the DYNAMIC_ARCH case ? I guess so, if we cannot be sure that these are going to be present in any relevant -zvl128b or -zvl256b cpu. 
That is unless we accept the risk that SBGEMM calls will just SIGILL on underequipped hardware - I still think any other BLAS calls should be fine, and I'm still not sure if I like the tradeoff of falling back to generic kernels for __everything__ just because the optimized ones were built with either flavor of half-precision support - which the user _may_ know to avoid calling. There is the risk of the compiler using half-precision tricks in optimizing regular precision code if we tell it they're available, but I assume it is largely theoretical ?

driver/others/dynamic_riscv64.c
```diff
@@ -165,14 +168,23 @@ static gotoblas_t* get_coretype(void) {
 	};
 	int ret = syscall(NR_riscv_hwprobe, pairs, 1, 0, NULL, 0);
 	if (ret == 0) {
-		if (!(pairs[0].value & RISCV_HWPROBE_IMA_V))
+#if defined(BUILD_HFLOAT16)
+		vector_mask = (RISCV_HWPROBE_IMA_V | RISCV_HWPROBE_EXT_ZFH | RISCV_HWPROBE_EXT_ZVFH);
```
ChipKerchner, line 172: Don't you need to check for these also?

#define RISCV_HWPROBE_EXT_ZFHMIN (1 << 28)
#define RISCV_HWPROBE_EXT_ZVFHMIN (1 << 31)

markdryan: Are these extensions being used?

They don't appear to be enabled in the [compiler flags](https://github.com/OpenMathLib/OpenBLAS/blob/develop/Makefile.riscv64#L11)

ChipKerchner: `__riscv_vle16_v_f16m1` (and other variations) are listed under the Zvfhmin extension.  I would think the scalar form is also being used in situations like this `_Float16 B0 = B[bi+0];`

The ones that you currently have are for the mult and madd instructions.

ChipKerchner: Ok, I looked it up.  Minimum extension is a subset of the regular extension.  So what we and you have seem correct.

PR #5439, opened 2025-09-08
clean up duplicate assignment of cpus newer than POWER10

martin-frbg:
trivial cosmetic fix for #5354 as suggested in review


pratiklp00:
@martin-frbg Thanks for the Changes.

PR #5444, opened 2025-09-12
riscv：Remove  the manual unrolling in gemv_t_vector.c  to performance better

yuanjia111:
**1. The conclusion and detailed discussion can be found in the following link：**
[https://github.com/OpenMathLib/OpenBLAS/issues/5430](url)
[https://github.com/OpenMathLib/OpenBLAS/pull/5427](url)

2. After modifying the code, the function is normal and the performance is indeed improved. The performance verified on K1 [C908, vlen = 256].
image
image
image

### DIFF
--- a/kernel/riscv64/gemv_t_vector.c
+++ b/kernel/riscv64/gemv_t_vector.c
@@ -72,123 +72,34 @@ int CNAME(BLASLONG m, BLASLONG n, BLASLONG dummy1, FLOAT alpha, FLOAT *a, BLASLO
     FLOAT_V_T_M1 v_res;
     size_t vlmax = VSETVL_MAX_M1();
 
-#ifndef RISCV_0p10_INTRINSICS
-    FLOAT_V_T va0, va1, va2, va3, vr0, vr1, vr2, vr3;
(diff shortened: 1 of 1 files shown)


kernel/riscv64/gemv_t_vector.c
```diff
@@ -72,123 +72,34 @@ int CNAME(BLASLONG m, BLASLONG n, BLASLONG dummy1, FLOAT alpha, FLOAT *a, BLASLO
     FLOAT_V_T_M1 v_res;
     size_t vlmax = VSETVL_MAX_M1();
```
ChipKerchner: Remove unused variable - as well as VSETVL_MAX_M1

ChipKerchner:
Looks good - seeing a 4x+ speedup from previous version.

kernel/riscv64/gemv_t_vector.c
yuanjia111: Done! Thank you for your thorough review!

martin-frbg:
Thank you for the quick update

Issue #5372, opened 2025-07-10
test_dgemmt (actually DGEMM) failure on ppc64le with GCC 15


sharkcz:
Looks like the `test_extensions/test_dgemmt` test has some failure when OpenBLAS is built with GCC 15 on ppc64le (Fedora 42). Building with GCC 14 (Fedora 41) there is no such issue.
````
...
TEST 1197/1522 dgemmt:c_api_rowmajor_lower_beta_one [OK]
TEST 1198/1522 dgemmt:c_api_rowmajor_lower_alpha_zero [OK]
TEST 1199/1522 dgemmt:c_api_rowmajor_lower_M_50_K_50_a_trans_b_trans [OK]
TEST 1200/1522 dgemmt:c_api_rowmajor_lower_M_50_K_100_a_notrans_b_trans [OK]
TEST 1201/1522 dgemmt:c_api_rowmajor_lower_M_100_K_50_a_trans_b_notrans [OK]
TEST 1202/1522 dgemmt:c_api_rowmajor_lower_M_50_K_50_a_notrans_b_notrans [OK]
TEST 1203/1522 dgemmt:c_api_rowmajor_upper_beta_one [OK]
TEST 1204/1522 dgemmt:c_api_rowmajor_upper_alpha_zero [OK]
TEST 1205/1522 dgemmt:c_api_rowmajor_upper_M_50_K_50_a_trans_b_trans [OK]
TEST 1206/1522 dgemmt:c_api_rowmajor_upper_M_50_K_100_a_notrans_b_trans [OK]
TEST 1207/1522 dgemmt:c_api_rowmajor_upper_M_100_K_50_a_trans_b_notrans [FAIL]
  ERR: test_extensions/test_dgemmt.c:820  expected 0.000e+00, got 6.352e-02 (diff -6.352e-02, tol 1.000e-13)
TEST 1208/1522 dgemmt:c_api_rowmajor_upper_M_50_K_50_a_notrans_b_notrans [OK]
TEST 1209/1522 dgemmt:c_api_colmajor_lower_beta_one [OK]
TEST 1210/1522 dgemmt:c_api_colmajor_lower_alpha_zero [OK]
TEST 1211/1522 dgemmt:c_api_colmajor_lower_M_50_K_50_a_trans_b_trans [OK]
TEST 1212/1522 dgemmt:c_api_colmajor_lower_M_50_K_100_a_notrans_b_trans [OK]
[...]

PR #5448, opened 2025-09-17
Fix clobber list entries for arrays in POWER kernels that use inline asm

martin-frbg:
proper fix for #5372 as pointed out by GCC developers there


sharkcz:
There are few little glitches in the power8 routines, going thru the failures right now ...

sharkcz:
fixed in PR#5463

pratiklp00:
Hi @martin-frbg we have observed the few typos with Power10 kernel as well. I have fixed them in PR #5472. Please have a look.

sharkcz:
Thanks for the P10 fixes, I didn't have a P10 system handy when fixing the P8 issues.

martin-frbg:
I have (in the GCC Compile Farm) but I'm stretched a bit thin right now, didn't manage to put in the time over the weekend as planned

sharkcz:
no problem and thanks for your work, these PRs show there are other people who care about ppc64le/Power support :-)

PR #5452, opened 2025-09-21
Rescale Infinity input in CSRSCL/ZDRSCL early to avoid infinite loop later (Reference-LAPACK PR 1158)

martin-frbg:


martin-frbg:
Interesting, looks like OpenBLAS missed a (largely cosmetic) update to these two files in Reference-LAPACK ages ago.

martin-frbg:
Actually I applied the changes to the wrong files, oops. This is not supposed to happen

PR #5463, opened 2025-09-26
fix typos in Power8 routines

sharkcz:


martin-frbg:
Thank you (did not expect this merge to get picked up so quickly...)

PR #5465, opened 2025-09-26
Add vectorized packing for FP16 and BF16 for RISC-V.  Reactivate vector packing for FP64 transposed

ChipKerchner:
Add vectorized packing for FP16 and BF16 - up to a 3X improvement.

Reactivate vector packing for FP64 transposed - turned out slowdown in previous MR was from use of vector load/store segment (which is slow on some platforms for FP64).


ChipKerchner:
[https://github.com/OpenMathLib/OpenBLAS/issues/5457](https://github.com/OpenMathLib/OpenBLAS/issues/5457)

PR #5471, opened 2025-09-29
Support for SME1 based ssymm_direct kernel for cblas_ssymm level 3 API

matcraje:
This PR introduces a specialized kernel that utilizes ARM SME1 (Scalable Matrix Extension) capabilities to optimize the cblas_ssymm function.


matcraje:
Looks like cmake builds are failing. I will check and modify the changes.

matcraje:
The checks using 'make' are successful.
The checks using 'cmake' fail due to undefined symbols (Eg. undefined symbol '_ssymm_direct_alpha_betaLL_A64FX' etc). 

@martin-frbg Do you know if I missed something?

martin-frbg:
kernel/CMakeLists.txt has a second block of definitions starting around line 1100 that specifically handles DYNAMIC_ARCH builds - you need to add the equivalent lines for your ssymm_direct_alpha_betaLL etc. with an added ${TSUFFIX} there

martin-frbg:
WoA build still fails with the 
```
2025-09-30T05:58:18.3418812Z lld-link: error: duplicate symbol: __arm_tpidr2_save
2025-09-30T05:58:18.3419188Z >>> defined at kernel\CMakeFiles\kernel.dir\CMakeFiles\ssymm_direct_alpha_betaLU.c.obj
```
that we've come to understand happens when arm_sme.h is included unconditionally - please fix

Also please move your additions in interface/symm.c after the error handling (the line where xerbla is called if "info" is not zero) - this will probably require another `#if defined CBLAS` but protects your kernel from getting called with invalid arguments

matcraje:
Thanks @martin-frbg for the comments. Earlier errors are now resolved.

yuyichao:
Is it intended that this only support clang and will error if gcc is used?

martin-frbg:
Right now the only practical use case is Apple M4 as far as I know, and gcc requires SVE capability in order to use SME. (Same with the earlier SGEMM_DIRECT PRs). I'm planning to spin off an actual VORTEXM4 target in #5423, which should also fix a couple of recent issues related to inadvertent use of the sgemm_direct code path on non-SME targets

PR #5474, opened 2025-09-29
fix `-lto_library` filtering for apple-clang + gfortran on sdk26

### LINKED ISSUE #5473: Link error with apple-clang and gfortran: ld: -lto_library library filename must be 'libLTO.dylib'
This is the same problem as in https://github.com/OpenMathLib/OpenBLAS/issues/5202. However, the fix is not robust with new xcode sdk 26.

On Sequoia 15.7 with[ xcode 26](https://developer.apple.com/documentation/xcode-release-notes/xcode-26-release-notes) where XCode version jumped from 16 -> 26.

If Xcode is installed, the check [here](https://github.com/OpenMathLib/OpenBLAS/blob/develop/Makefile.system#L436) will return `XCVER=26`:
```
package-id: com.apple.pkg.Xcode
version: 26.0.0.0.1.1757562052
volume: /
location: /
install-time: 1758115875
```
causing the lto fix to not be applied. 
```
ifeq (x$(XCVER), x 16)
override CEXTRALIB := $(filter-out(-lto_library, $(CEXTRALIB))) 
endif
```
However, XCode doesn't have to be installed to get the command line tools. In this case, there is no return value
```
pkgutil --pkg-info=com.apple.pkg.Xcode
No receipt for 'com.apple.pkg.Xcode' found at '/'.
```
causing the second check to be used. Now, weirdly, this can seemingly report either 16 or 26.

I get (aarch64)
```
pkgutil --pkg-info=com.apple.pkg.CLTools_Executables
package-id: com.apple.pkg.CLTools_Executables
version: 26.0.0.0.1.1757719676
volume: /
location: /
install-time: 1758734670
```

whereas a colleague on x86_64 Sequoia 15.7 gets 
```
package-id: com.apple.pkg.CLTools_Executables
version: 16.4.0.0.1.1747106510
volume: /
location: /
install-time: 1753176345
```.

In addition, I don't see `INTERPROCEDURAL_OPTIMIZATION` in any of the CMake files, so this problem doesn't exist under CMake builds.

Chrismarsh:
This fixes https://github.com/OpenMathLib/OpenBLAS/issues/5473 where the filtering of `-lto_library` needs to be applied for xcode >= 16


martin-frbg:
Uh, sorry for mishandling the initial merge conflict here...

Chrismarsh:
I think I made the pr off of tag 0.3.30 too 😬

All correct now?

martin-frbg:
I had removed the `if gfortran` conditional in #5436 already, maybe that's what made github flag a merge conflict where there was none (and unfortunately I compared with a local checkout from before that September 2 commit by mistake).
I'll do a squash-and-merge and hope it will end up looking as originally intended...

PR #5211, opened 2025-04-08
Optimizing the Implementation of GEMV on the RISC-V V Extension

guoyuanplct:
performed loop unrolling, and reduced the number of multiplications. On my development board, there is approximately an 180% performance improvement.

PR #5476, opened 2025-09-30
Tranverse matrix data in a cache friendly manner for GEMV_N (RISCV)

ChipKerchner:
Tranverse matrix data in a cache friendly manner for GEMV_N (RISCV) - up to 10x faster.


martin-frbg:
Thanks - that does look better than just reverting #5211

PR #5481, opened 2025-10-03
Add SBGEMV and SHGEMV routines to RISC-V

ChipKerchner:
Add SBGEMV and SHGEMV routines to RISC-V - up to 32X faster.


ChipKerchner:
Can't tie in SHGEMV since there is no support in the system as of yet.

[https://github.com/OpenMathLib/OpenBLAS/issues/5480](https://github.com/OpenMathLib/OpenBLAS/issues/5480)

martin-frbg:
can probably tie that into interface/sbgemv.c and copy driver/level2/sbgemv_thread.c, then needs adding to common_s.h , common_level2.h , cblas.h , common_interface.h , common_param.h (two times there - once for single target and once for dynamic_arch) , kernel/setparam-ref.c plus exports/gensymbol and the (c)makefile additions.
I'll see if I can prepare a companion PR sometime tomorrow

martin-frbg:
almost done but stuck on a prototype mismatch that doesn't immediately make sense. I'll try to get back to it tomorrow night

martin-frbg:
prototype mismatch is a direct consequence of the kernel function signature being wrong - no support for parameter "beta", and an extraneous "buffer" argument ? (Please compare the CNAME line in generic/gemv_n.c or the x86_64 sbgemv_n.c )

ChipKerchner:
I think that RISC-V used the same GEMV kernel function signature as ARM did - without beta and with extra buffer parameter.

Mousius:
I think the older variants use the buffer and do beta separately, newer ones seem to be doing beta updates as part of the kernel:

* `sgemv` - https://github.com/OpenMathLib/OpenBLAS/blob/develop/kernel/x86_64/sgemv_n.c#L109
* `sbgemv` - https://github.com/OpenMathLib/OpenBLAS/blob/develop/kernel/x86_64/sbgemv_n.c#L103

ChipKerchner:
Is the kernel function signature different for GEMV vs SBGEMV?

The original GEMV (FP32/FP64) for RISC-V seems to use the non-beta format.

Mousius:
> Is the kernel function signature different for GEMV vs SBGEMV?
> 
> The original GEMV (FP32/FP64) for RISC-V seems to use the non-beta format.

Yip, those are the two kernel signatures I linked to 😸 BGEMV also uses the SBGEMV signature.

martin-frbg:
Yes, for some "historic" reason the kernel interface for SBGEMV is different from SGEMV. Not entirely sure whose "fault" it was, probably the Intel folks were first - arguably the "beta and no buffer" version has a clear majority now, so it is probably easier to consolidate on that.
(@Mousius guess it would make sense to rename "your" generic/gemv(_n/_t).c to bgemv or sbgemv too, as a generic kernel for S/DGEMV already lives in kernel/arm like half of the generics do - another old oddity)

ChipKerchner:
@martin-frbg Ok, I changed the signature of SBGEMV for RISCV to the appropriate one.

martin-frbg:
Thank you - I was on it but got interrupted by the chimney sweep :)

ChipKerchner:
[https://github.com/OpenMathLib/OpenBLAS/issues/5279](https://github.com/OpenMathLib/OpenBLAS/issues/5279)

PR #5482, opened 2025-10-05
Fix compilation on ARM

yuyichao:
Define a dummy function if SME is not supported, following what sgemm does.

Also, what's the reason for the condition on `__clang__` only? Is there known problem that is not expected to be fixable on all other compiler vendors? Why is the feature test not enough?


yuyichao:
Updated to fix new breakage from https://github.com/OpenMathLib/OpenBLAS/pull/5450

PR #5492, opened 2025-10-08
Tie in SHGEMV for RISC-V.

ChipKerchner:
Tie in SHGEMV for RISC-V.


ChipKerchner:
@martin-frbg Do we need to add the BUILD_BFLOAT16 and BUILD_HFLOAT16 to the RISC-V CI?

martin-frbg:
Guess we could (assuming that qemu-riscv64 supports it), but looks like the riscv-vector workflow has lost its qemu somehow and isn't actually running its tests lately.

PR #5494, opened 2025-10-08
RISCV64-CI: Update toolchain to unbreak qemu-riscv; add B/HFLOAT16 to VL256 target

martin-frbg:


ChipKerchner:
Thanks Martin!

martin-frbg:
weird error where it apparently fails to include a valid time.h from common.h , but only in the compare_ tests

test/compare_sgemv_shgemv.c
```diff
@@ -26,6 +26,7 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 *****************************************************************************/
 #include <stdio.h>
 #include <stdint.h>
+#include <time.h>
```
ChipKerchner, line 29: Why would only SHGEMV need this?  Shouldn't the others also need it?

martin-frbg:
I'm just trying if it fixes compilation for that one file... time.h should get included by common.h, so maybe that nightly build of the toolchain is incomplete/broken ? (The added `-std=c11` has been there a few months)

ChipKerchner:
BTW, it is recommended that GCC be at least 15.0 for vector RISC-V. Probably 19+ on LLVM

martin-frbg:
Idk how authoritative this riscv-collab project is actually... besides being endorsed by RISC-V International

martin-frbg:
Seems the problem is that (at least) that toolchain does not define ` __USE_POSIX199309` although it appears to define POSIX_MONOTONIC_CLOCK, so the clock_gettime() function is undefined although time.h was included.

martin-frbg:
Further to this, seems to be a glibc oddity where one has to define POSIX_C_SOURCE to 199309 to enable certain POSIX functions (or alternatively, use std=gnu11 instead of c11)

ChipKerchner:
Maybe you are using an older version of glibc?

PR #5485, opened 2025-10-07
Add Infrastructure for SHGEMV

Mousius:
This adds all the relevant bits and pieces to add a `shgemv` path as well as a future `hgemm`/`hgemv` path in a similar model to `sb` and `b` interfaces.

I've also fixed a few bits and pieces around `shgemm` which didn't build in a few situations.

PR #5499, opened 2025-10-10
Add test for SHGEMM

### LINKED ISSUE #5497: No support for SHGEMM (FP16) in the tests
There is support for SBGEMM (BF16) but none for SHGEMM (FP16) in the tests

martin-frbg:
fixes #5497


test/compare_sgemm_shgemm.c
```diff
@@ -0,0 +1,147 @@
...
+#include <stdio.h>
+#include <stdint.h>
+#include "../common.h"
+
+#include "test_helpers.h"
+
+#define SGEMM   BLASFUNC(sgemm)
+#define SHGEMM   BLASFUNC(shgemm)
+#define SGEMV   BLASFUNC(sgemv)
+#define SHGEMV   BLASFUNC(shgemv)
```
Mousius: These shouldn't be necessary as `shgemv` is tested separately?

test/compare_sgemm_shgemm.c
```diff
@@ -0,0 +1,147 @@
...
+int
+main (int argc, char *argv[])
+{
+  blasint m, n, k;
+  int i, j, l;
+  blasint x, y;
+  int ret = 0;
+  int loop = SHGEMM_LARGEST;
+  char transA = 'N', transB = 'N';
+  float alpha = 1.0, beta = 0.0;
```
Mousius, line 47: Given #5485, is it worth setting `beta` to `1.0` here, or testing both variations? Similarly with `alpha`, as `1.0` and `0.0` can have faster paths but in-between values usually don't.

I haven't got a good answer to this in the C tests, but we could try adding some Python-based tests with `hypothesis` or Rust tests with `proptest`.

test/compare_sgemm_shgemm.c
```diff
@@ -0,0 +1,147 @@
...
+
+  for (x = 0; x <= loop; x++)
+  {
+    if ((x > 100) && (x != SHGEMM_LARGEST)) continue;
+    m = k = n = x;
+    float *A = (float *)malloc_safe(m * k * sizeof(FLOAT));
+    float *B = (float *)malloc_safe(k * n * sizeof(FLOAT));
+    float *C = (float *)malloc_safe(m * n * sizeof(FLOAT));
+    hfloat16 *AA = (hfloat16 *)malloc_safe(m * k * sizeof(bfloat16));
+    hfloat16 *BB = (hfloat16 *)malloc_safe(k * n * sizeof(bfloat16));
```
Mousius:
```suggestion
    hfloat16 *AA = (hfloat16 *)malloc_safe(m * k * sizeof(hfloat16));
    hfloat16 *BB = (hfloat16 *)malloc_safe(k * n * sizeof(hfloat16));
```

test/compare_sgemm_shgemm.c
Mousius: Worth noting we also don't test anything but square, and it'd be nice to sample in the various dimensions - though we could maybe just randomly sample to start with?

martin-frbg: Agree on the variations of alpha and beta, but we need to keep in mind that these test are run on each and every build, so time will become a problem sooner or later, and I think non-trivial dependencies like python or rust are completely out of the question for a default build.
Perhaps it would make sense to put those in a completely separate project similar to BLAS-Tester

Mousius: > we need to keep in mind that these test are run on each and every build

Yeah, that's the reason I moved to a more randomly sampled approach, as it'll likely hit the relevant cases and shrink to the specific problem case quite quickly in such a small space. Otherwise, yes, the amount of permutations becomes a bit tricky to manage 😿 

> I think non-trivial dependencies like python or rust are completely out of the question for a default build.

This is where I got to, I haven't found something that fits just yet.

> Perhaps it would make sense to put those in a completely separate project similar to BLAS-Tester

Hmm, does that cause an issue that they're not run on the default build? And does it discourage people from adding tests for the default build? 

It would be fun though, we could also potentially combine it with the benchmarking repo?

martin-frbg: Yes, something like that that runs as CI but does not (necessarily) run as part of a normal build process. I'm not even totally averse to having something like a pytest subdirectory as long as it doesn't become mandatory for user builds and doesn't take up too much space.

martin-frbg: But short-term I'd like to get the code stable enough  to do another release, and perhaps focus on my other work for a while so that I do not run myself out of business.

PR #5505, opened 2025-10-12
Rewrite lapacke headers with pre/postfixes if necessary

### LINKED ISSUE #5493: lapacke.h has incorrect names when using SYMBOLPREFIX
For example on latest `develop`:
```
$ make SYMBOLPREFIX=scipy_ USE_OPENMP=0 DYNAMIC_ARCH=0 USE_THREAD=0 TARGET=PRESCOTT
$ make SYMBOLPREFIX=scipy_ PREFIX=$PWD/install install
$ grep "dlange" install/include/lapacke.h 
double LAPACKE_dlange( int matrix_layout, char norm, lapack_int m,
double LAPACKE_dlange_work( int matrix_layout, char norm, lapack_int m,
$ nm -D install/lib/libscipy_openblas_prescott-r0.3.30.dev.so | grep dlange
00000000005c0cc0 T scipy_dlange_
0000000000b97ee0 T scipy_LAPACKE_dlange
```
But `cblas.h` is fine:
```
$ grep "dnrm2" install/include/cblas.h 
double scipy_cblas_dnrm2 (OPENBLAS_CONST blasint N, OPENBLAS_CONST double *X, OPENBLAS_CONST blasint incX);
$ nm -D install/lib/libscipy_openblas_prescott-r0.3.30.dev.so | grep dnrm2
00000000002ab200 T dnrm2_k
00000000000acae0 T scipy_cblas_dnrm2
00000000000ab580 T scipy_dnrm2_
```
So whatever logic is used for `cblas.h` for example should probably be adopted for `lapacke.h` rather than just copying (as it says during install currently: "Copying LAPACKE header files"). I have verified that a manual `sed` on the header works in another repo where I'm using `scipy_openblas32` wheels with this prefix.

cc @isuruf since you worked on https://github.com/OpenMathLib/OpenBLAS/issues/4771 which seems like the most recent related issue/PR based on a very quick search!

If there are no other volunteers I can attempt a PR to fix this but my `Makefile` skills are not great so it might be a bumpy ride!

martin-frbg:
fixes #5493


larsoner:
Thanks @martin-frbg !

PR #5513, opened 2025-10-20
More arm compilation fix

yuyichao:
Make dummy function have the same linkage as the real one


takeshige-fj:
I am currently encountering the same issue and would appreciate it if a fix could be implemented.
CPU : A64FX
OS : Rocky Linux release 8.6
Compiler : gcc 14.2.1
$ make USE_OPENMP=1 DYNAMIC_ARCH=1
The code compiles without any issues when using GCC version 13.3.1.

PR #5522, opened 2025-11-01
Fix #5521: add @SUFFIX64@ in OpenBLASConfig.cmake.in

### LINKED ISSUE #5521: target_link_libraries not suport SUFFIX64
Thank you for your work on this project.

I’ve encountered an issue where the CMake integration is inconsistent when OpenBLAS is compiled with the SUFFIX64=64 option. The generated config file does not apply the suffix to CMake targets and variables, which can cause conflicts.

Problem:
When building with SUFFIX64=64, the OpenBLASConfig.cmake file still exports a target named OpenBLAS::OpenBLAS and variables like OpenBLAS_LIBRARIES. This is inconsistent with the library filename and can be problematic if a project needs to link against both 32-bit and 64-bit interfaces simultaneously.

Thank you for your time and consideration.

Smilyf:
Fixes #5521

This PR updates OpenBLASConfig.cmake.in to correctly use the @SUFFIX64@ placeholder for CMake targets and variables.

When OpenBLAS is built with SUFFIX64=64, this ensures the exported CMake target is OpenBLAS64::OpenBLAS (instead of OpenBLAS::OpenBLAS).

To support ：
```
    target_link_libraries(${LIBRARY_NAME}Shared PUBLIC
        OpenBLAS64::OpenBLAS
    )
```

### DIFF
--- a/cmake/OpenBLASConfig.cmake.in
+++ b/cmake/OpenBLASConfig.cmake.in
@@ -50,30 +50,30 @@ set(PN OpenBLAS)
 
 # need to check that the @USE_*@ evaluate to something cmake can perform boolean logic upon
 if(@USE_OPENMP@)
-    set(${PN}_openmp_FOUND 1)
+    set(${PN}@SUFFIX64@_openmp_FOUND 1)
 elseif(@USE_THREAD@)
-    set(${PN}_pthread_FOUND 1)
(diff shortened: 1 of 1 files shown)


martin-frbg:
thank you

PR #5525, opened 2025-11-01
CMake: Fix missing support for HFLOAT16 in Windows dll generation

### LINKED ISSUE #5524: CMake build fails on Windows when prefix flag is used
When building OpenBLAS with CMake and using a custom prefix (e.g., scipy_), the build fails with undefined symbol errors:

```
lld-link: error: <root>: undefined symbol: shgemm_
lld-link: error: <root>: undefined symbol: shgemv_
lld-link: error: <root>: undefined symbol: cblas_shgemm
```

**Observed behavior:**

These symbols (shgemm_, shgemv_, cblas_shgemm) are conditionally compiled only when BUILD_HFLOAT16=ON. However, even when this flag is OFF, the generated renamesys.def file still contains entries like:

```
scipy_shgemm=shgemm_ @XX
scipy_shgemv=shgemv_ @YY
scipy_cblas_shgemm=cblas_shgemm @ZZ
```

Because the corresponding object files (shgemm.obj, etc.) are not built, the linker fails to find these symbols, resulting in the errors above.
When BUILD_HFLOAT16=ON, the objects are built and the build completes successfully.

**Expected behavior:**

When BUILD_HFLOAT16=OFF, the .def file should not include the shgemm, shgemv, or cblas_shgemm symbols.

martin-frbg:
fixes #5524

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -498,6 +498,11 @@ if (BUILD_SHARED_LIBS OR DELETE_STATIC_LIBS AND NOT ${SYMBOLPREFIX}${SYMBOLSUFFI
   else ()
 	  set (BBF16 0)
   endif()
+  if (${BUILD_HFLOAT16})
+	  set (BHF16 1)
+  else ()
+	  set (BHF16 0)
+  endif()
   if (${BUILD_SINGLE})
 	  set (BS 1)
   else ()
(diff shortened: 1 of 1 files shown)


mattip:
FWIW, I turned the changeset into a patch in MacPython/openblas-libs#227 and it fixes the failure. Thanks!

martin-frbg:
Thanks for the confirmation

PR #5527, opened 2025-11-04
Prevent possible conversion from bfloat16 to __bf16 - RISC-V

ChipKerchner:
Prevent possible conversion from bfloat16 to __bf16


martin-frbg:
Great, this appears to fix the test errors I saw with the cross build and qemu-riscv64

ChipKerchner:
Great!  Should we reactivate the missing comparison tests?  SBGEMM/V?

PR #5534, opened 2025-11-15
Fix f_check detection of LLVM 21 flang

bartoldeman:
The check for GCC is confused by the GNU-stack in

```
	.file	"FIRModule"
	.text
	.globl	zhoge_
	.p2align	4
	.type	zhoge_,@function
zhoge_:
	xorps	%xmm0, %xmm0
	xorps	%xmm1, %xmm1
	retq
.Lfunc_end0:
	.size	zhoge_, .Lfunc_end0-zhoge_

	.ident	"flang version 21.1.5"
	.section	".note.GNU-stack","",@progbits
```

And displays:
```
./f_check: line 102: [: 	: integer expression expected
```

Since it expects a string with GCC anyway, better to only match GCC and not GNU.

### DIFF
--- a/f_check
+++ b/f_check
@@ -92,7 +92,7 @@ else
 		vendor=FLANG
 		openmp='-fopenmp'
 		;;	
-            *GNU*|*GCC*)
+            *GCC*)
 
                 v="${data#*GCC: *\) }"
                 v="${v%%\"*}"


martin-frbg:
ISTR there is/was a reason for matching GNU in addition to GCC - probably earlier versions of gfortran not using GCC before the CC became associated with "compiler collection"

bartoldeman:
If I look in the history it must have been lost in translation from perl to bash, as the perl code did:

```
} elsif ($data =~ /GNU/ || $data =~ /GCC/ ) {
            $data =~ s/\(+.*?\)+//g;
	    $data =~ /(\d+)\.(\d+).(\d+)/;
	    $major = $1;
	    $minor = $2;
	    if ($major >= 4) {
		$vendor = GFORTRAN;
		$openmp = "-fopenmp";
```
so there was no `GCC` in the $data setting regexps here unlike in the shell script.

bartoldeman:
the oldest gfortran I have on hand has this:
```
	.file	"ftest.f"
	.text
	.p2align 4,,15
	.globl	zhoge_
	.type	zhoge_, @function
zhoge_:
.LFB0:
	.cfi_startproc
	xorpd	%xmm1, %xmm1
	movapd	%xmm1, %xmm0
	ret
	.cfi_endproc
.LFE0:
	.size	zhoge_, .-zhoge_
	.ident	"GCC: (GNU) 4.8.5"
	.section	.note.GNU-stack,"",@progbits
```
That's a 10 year old compiler, not sure if OpenBLAS still supports that?

PR #5538, opened 2025-11-18
Optimize ZROT_RVV for the unit-stride case (inc_x = inc_y = 1)

CheryDan:
Optimized zrot for inc_x = inc_y = 1 by reducing instruction count and avoiding register spills to memory.


martin-frbg:
Thank you

PR #5539, opened 2025-11-18
Refactoring: ARM64 dot Kernel: don't call num_cpu_avail twice

FRosner:
This is a tiny refactoring. In `get_dot_optimal_nthreads`, we'd call `num_cpu_avail(1)` twice. This PR simply uses the value from the first invocation when returning it.


abhishek-iitmadras:
Thanks @FRosner for this PR

martin-frbg:
Thank you

Issue #5545, opened 2025-11-22
build failures using clang to build powerpc64le


mayeut:
git bisect shows the issue was introduced by https://github.com/OpenMathLib/OpenBLAS/pull/4944

reproducer using an ubuntu:24.04 container (doing cross build but the issue also happens in native builds):
```
apt install lsb-release wget software-properties-common gnupg curl wget
wget https://apt.llvm.org/llvm.sh
chmod +x llvm.sh
./llvm.sh 21
apt install make gcc gcc-powerpc64le-linux-gnu gfortran-powerpc64le-linux-gnu
ln -s clang-21 /usr/bin/powerpc64le-linux-gnu-clang
git bisect start
git bisect bad v0.3.29
git bisect good v0.3.28
git bisect run bash -c 'make clean && make -j$(nproc) HOSTCC="clang-21" CC="powerpc64le-linux-gnu-clang" FC="powerpc64le-linux-gnu-gfortran" ARCH=power TARGET=POWER8 DYNAMIC_ARCH=1 USE_OPENMP=0'
....
76227e2948b1f847abef003de5f6d49ea0dd3171 is the first bad commit
commit 76227e2948b1f847abef003de5f6d49ea0dd3171
Author: Chip Kerchner <chip.kerchner@ibm.com>
Date:   Fri Sep 6 14:03:31 2024 -0500

    Initial commit for vectorized BF16 GEMV. Added GEMM_GEMV_FORWARD_BF16 to enable using BF16 GEMV for one dimension matrices.  Updated unit test to support inc_x != 1 or inc_y for GEMV.
...
```

The reported error looks like:
```
In file included from ../kernel/power/sbgemv_n_vsx.c:31:
../kernel/power/sbgemv_common.c:In file included from 61:18: error: type specifier missing, defaults to 'int'; ISO C99 and later do not support implicit int [-Wimplicit-int]
../kernel/power/sbgemv_t_vsx.c   61 | typedef :_31_:
../kernel/power/sbgemv_common.c:61:18: error: type specifier missing, defaults to 'int'; ISO C99 and later vdoe cnott osupportr implicit  IintF L[-Wimplicit-int]OA
T        vec_bf16;
      | ~~~~~~~~~~~~~~~~ ^
      | int
   61 | typedef __vector IFLOAT        ../common.hv:ec_bf13106:;16:
       | note: ~~~~~~~~~~~~~~~~ ^
expanded       from|  macroint 
'IFLOAT'
[...]

PR #5548, opened 2025-11-22
ci: add build with clang on ppc64le

mayeut:
Add a CI run for clang on ppc64le in order to prevent build/tests regressions.

related to #5545


mayeut:
might be worth looking at Jenkins logs but all seem good from the status check point of view.

PR #5551, opened 2025-11-24
Updated SGEMV ramps.

almayne:
Significant performance improvements are gained by the proposed changes on both c7g (NEOVERSEV1) and c8g (NEOVERSEV2) instances. To reproduce these values you need to run with OMP_ADAPTIVE=1. The plots below show the average time taken for 10000 iterations of sgemv operations on increasing square matrix/vector sizes, from 2x2 through to 1024x1024. The x axis reaches 2046 as we first run sgemv without transposition, then with. I've also include plots of the ratio, relative to the original stats (lower is better). These generate the following stats:
Geometric mean for c7g_sgemv.txt: 0.890437968914142
Geometric mean for c8g_sgemv.txt: 0.7884951978206536

image

image

image

image

### DIFF
--- a/interface/gemv.c
+++ b/interface/gemv.c
@@ -1,4 +1,5 @@
 /*********************************************************************/
+/* Copyright 2025 The OpenBLAS Project                               */
 /* Copyright 2009, 2010 The University of Texas at Austin.           */
 /* All rights reserved.                                              */
 /*                                                                   */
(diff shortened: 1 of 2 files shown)


aditew01:
cc: @martin-frbg @annop-w @Mousius 
can you please help with a review?

interface/gemv.c
```diff
@@ -81,9 +82,12 @@ static inline int get_gemv_optimal_nthreads_neoversev1(BLASLONG MN, int ncpu) {
           : (MN < 1050625L)   ? MIN(ncpu, 40)
           : ncpu;
   #else
-      return (MN < 25600L)     ? 1
```
aditew01, line 84: do we want to guard it for NEOVERSEV2 / NEOVERSEN2?

almayne: Discussed offline, happy that this is done by the calling function. No changes needed.

aditew01:
bumping this up?
cc: @martin-frbg @annop-w @Mousius
can you please help with a review?

PR #5564, opened 2025-12-11
Add $(EXE) suffix to executables in Makefiles

vtjnash:
Define EXE variable in Makefile.system (set to .exe on Windows,
empty otherwise) and update all test executables to use it for
Windows compatibility.

- Makefile.system: Add EXE variable definition
- test/Makefile: Update all executables with $(EXE) suffix
- ctest/Makefile: Update all executables with $(EXE) suffix
- utest/Makefile: Update UTESTBIN and UTESTEXTBIN with $(EXE) suffix
- Makefile.install: Update install_tests target with $(EXE) suffix
- .gitignore: Add *.exe pattern


martin-frbg:
The `make` binary included in the Ubuntu/x86_64 version of the mingw-llvm cross-compiler does this automatically (as does regular mingw - or at least the MXE version I use for cross-compiling to Windows/x86_64. Is it the native aarch64/WoA version of this toolchain that gets it "wrong" by default, or which version (and context) of `make` are you using that needs to be told the appropriate EXE extension ?

vtjnash:
The make binary doesn't (couldn't?) do this anywhere ever, since it doesn't know that the rule will produce an executable file. The linker usually does automatically add the suffix, but it means the rules in the make binary don't end up producing the files that was specified in the Makefile to be made by that rule, which can cause `make` to get confused later. It is compiler hack to provide basic support for unix-only makefiles, but not intended to be relied upon:

https://sourceware.org/binutils/docs-2.20/ld/Options.html

> --force-exe-suffix
> Make sure that an output file has a .exe suffix.
> If a successfully built fully linked output file does not have a .exe or .dll suffix, this option forces the linker to copy the output file to one of the same name with a .exe suffix. This option is useful when using unmodified Unix makefiles on a Microsoft Windows host, since some versions of Windows won't run an image unless it ends in a .exe suffix.

vtjnash:
I guess the customary name for this is [EXEEXT](https://www.gnu.org/software/automake/manual/html_node/EXEEXT.html) if you prefer autotools conventions

vtjnash:
Bump?

martin-frbg:
I'm not sure I understand which combination of compiler and operating system actually needs this ?

vtjnash:
Any use targeting Windows requires this for correctness. The compiler has a hack that makes it partly work even when your makefile has this bug, but it is awkward to be specifying that it builds the wrong file name

vtjnash:
bump? You do not get this "for free" with the make binary, only with automake, and only for the build rules, not all of the other uses, but you aren't using automake anyways: https://www.gnu.org/software/automake/manual/html_node/EXEEXT.html

vtjnash:
Without this, the output being literally specified never actually gets built, since the compiler always forces an added extension of `.exe`. That ends up meaning that make is forced to re-running linking every time, since the build never generates the specified output. In essence, it mistakenly has a PHONY target here instead, for Win32-only. As it turns out, making all targets PHONY does mostly work, since `make` is intentionally not very strict about that, but it isn't terribly reliable for others who want to build and test. So this PR aligns the name of the output targets in make with the names of the file actually produced by the linker.

vtjnash:
bump?

vtjnash:
bump?

PR #5565, opened 2025-12-11
build: fix rule for building dynamic files

vtjnash:
Previously the architecture-specific dynamic files were relying on the built-in rules alone.

### DIFF
--- a/driver/others/Makefile
+++ b/driver/others/Makefile
@@ -127,10 +127,10 @@ endif
 xerbla.$(SUFFIX) : xerbla.c
 	$(CC) $(CFLAGS) -c $< -o $(@F)
 
-dynamic.$(SUFFIX) : dynamic.c
+dynamic%.$(SUFFIX) : dynamic%.c
 	$(CC) $(CFLAGS) -c $< -o $(@F)
 
-dynamic.$(PSUFFIX) : dynamic.c
+dynamic%.$(PSUFFIX) : dynamic%.c
 	$(CC) $(PFLAGS) -c $< -o $(@F)
 
 parameter.$(SUFFIX) : parameter.c ../../param.h


martin-frbg:
Sorry, I do not see what this would be fixing, when the built-in rule works fine (and has so for years)

vtjnash:
I appreciate the quick review, but please reopen this. This was accidentally relying on an old-fashioned technique (not my name for it, literally that is what this feature is called https://www.gnu.org/software/make/manual/html_node/Suffix-Rules.html) that does not work reliably for many uses. In particular, the rule is only implemented when $(SUFFIX) is defined to be `.o`, but that means this one file breaks the build if any part of the Makefile uses a different SUFFIX value. This is a really simple fix for that.

vtjnash:
Bump. Please reopen and merge this bugfix.

martin-frbg:
Thanks for the explanation. Can you please let me know what environment/toolchain you are working with ? I _assume_ it would have to be a Windows on Arm target environment (to require a specific dynamic_ARCH.c) and probably a cross-compiler that uses .obj extensions ?

vtjnash:
That is right. I'm using https://github.com/mstorsjo/llvm-mingw.git modified to add flang support (I have a PR there) to test various configurations of builds

martin-frbg:
I see... I wasn't aware of that project, thanks for the pointer. (I guess you are aware that the native LLVM package for WoA does include flang already, unlike their Windows/x64 package ?)

vtjnash:
Yes, I'm working on integration with Julia, which is still mostly Makefile driven, and thus very annoying to deal with on Windows natively. That project lets me use mingw as the libraries to be able to use it from linux (WSL)

vtjnash:
Merge?

vtjnash:
Merge?

vtjnash:
Merge?????

vtjnash:
bump

martin-frbg:
This actually **breaks** the build with mingw - reverting

vtjnash (commented):
mingw32-make[1]: *** No rule to make target 'dynamic.obj', needed by 'libs'.  Stop.

Curious. I thought mingw32-make was the really buggy version deprecated well over a decade ago. I wonder why it can't parse this rule correctly though.

PR #5566, opened 2025-12-11
arm64: fix clang ICE on Windows for zdot_thunderx2t99.c

vtjnash:
Guard .align directive to avoid internal compiler error on
AArch64 Windows with clang.

See: https://github.com/llvm/llvm-project/issues/149547
See: #5076, which applied this same fix to the other files

### DIFF
--- a/kernel/arm64/dznrm2_thunderx2t99_fast.c
+++ b/kernel/arm64/dznrm2_thunderx2t99_fast.c
@@ -155,7 +155,10 @@ static double nrm2_compute(BLASLONG n, FLOAT *x, BLASLONG inc_x)
 	"	cmp	"J", xzr			\n"
 	"	beq	.Lnrm2_kernel_F1		\n"
 
+/* https://github.com/llvm/llvm-project/issues/149547 */
+#if !(defined(__clang__) && defined(OS_WINDOWS))
 	"	.align 5				\n"
+#endif
 	".Lnrm2_kernel_F:				\n"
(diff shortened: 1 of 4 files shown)


martin-frbg:
I'm pretty sure that alignment wasn't the only issue preventing successful compilation of this file at the time, else I wouldn't have treated it differently, just a day after fixing the other thunderx kernels in the same PR. Did you check that it actually works now on real WoA hardware ?

martin-frbg:
This is not wrong, but it serves no purpose as long as there are other issues precluding use of this file on WoA

vtjnash:
I do have real WoA hardware, so I'm happy to do any test for you (it isn't a THUNDERX2T99, but I can ask my colleague who has that hardware vendor if he can test also, if you need that). I don't know why you missed it, but don't see how that is relevant to the present existence of the bug here. The point of this is to be able to even test anything, as otherwise the flang compiler crashes and I don't get to even attempt to test the results. With this fix, (and with my other two bugfixes to the makefile) then everything compiles successfully. The fix I assumed was obviously correct on its own, since it is already required and implemented in many of the other files, so it shouldn't need any more justification.

vtjnash:
I can force push my rebase of this to fix the remaining bugs here, but you need to reopen first, or github will prevent it from being reopened and force me to make a new PR.

vtjnash:
Bump? Would you rather I open a new PR or would you rather re-open this one for the rebased fixes?

PR #5573, opened 2025-12-17
Add interleaving to sgemm and dgemm. Disentangle trmm/symm from gemm.

almayne:
This change adds interleaving to sgemm and dgemm copies and kernels for ARMV8SVE.
This required a degree of disentangling symm and trmm kernels from gemm. It should now be much easier to apply further optimisations to gemm.

The addition of interleaving provides a ~1.4% speedup on c7g (V1), with negligible changes on c8g (V2).

Taken over square matrix operations with size 2->2014, stepsize = 1:
Geometric mean for interleave/c7g_dgemm.txt: 0.9859023206257058
Geometric mean for interleave/c7g_sgemm.txt: 0.9887890902680289
Geometric mean for interleave/c8g_dgemm.txt: 0.9970050554316875
Geometric mean for interleave/c8g_sgemm.txt: 0.9948135816755502

We see an increase in the sgemm speedup (~2.4%) on c7g for larger matrix sizes. 

Taken over square matrix operations with size 2,000->10,000, stepsize = 1,000:
Geometric mean for 64thread_interleave/c7g_dgemm.txt: 0.9865252964543917
Geometric mean for 64thread_interleave/c7g_sgemm.txt: 0.9762227312411808
Geometric mean for 64thread_interleave/c8g_dgemm.txt: 0.9997186302044462
Geometric mean for 64thread_interleave/c8g_sgemm.txt: 0.9996022927667269


aditew01:
@martin-frbg  @Mousius can you please have a look?

martin-frbg:
Thanks for looking into this - though I'm not immediately convinced that the speedup warrants the added complexity @Mousius ? At least I'd like to put this off until after the 0.3.31 release.

Mousius:
> Thanks for looking into this - though I'm not immediately convinced that the speedup warrants the added complexity @Mousius ? At least I'd like to put this off until after the 0.3.31 release.

The speedup is one thing, the other is enabling GEMM kernels without the need to also implement the other kernels. This would enable us to land existing SME kernels, as previously proposed here: https://github.com/OpenMathLib/OpenBLAS/pull/5011#issuecomment-2537071236

Putting off until after 0.3.31 makes perfect sense to me as it's a relatively high risk change.

almayne:
> Thanks for looking into this - though I'm not immediately convinced that the speedup warrants the added complexity @Mousius ? At least I'd like to put this off until after the 0.3.31 release.

Hi Martin. Thanks for taking a look. I'm happy for this to go in after the release. Do you have a rough estimate for when that might be, so I can share internally?

almayne:
@martin-frbg congrats on getting the release out. Would it be possible to get this merged?

martin-frbg:
I'm probably missing something obvious, but I'm still a bit confused why this needs the addition of "COMM" kernels instead of declaring dedicated ?TRMM and ?SYMM COPY kernels within the existing framework, like the riscv64 folks do in e.g. KERNEL.RISCV64_ZVL128B ?

almayne:
> I'm probably missing something obvious, but I'm still a bit confused why this needs the addition of "COMM" kernels instead of declaring dedicated ?TRMM and ?SYMM COPY kernels within the existing framework, like the riscv64 folks do in e.g. KERNEL.RISCV64_ZVL128B ?

The intent is to capture what is still common between trmm and symm, and reduce duplicate code. I can refactor if preferred? @Mousius let me know if you have opinions either way. No preference on my part.

Mousius:
I agree with @martin-frbg, it'd be better to name them aligned with the algorithms they're influencing. 

@martin-frbg we're targeting these types of blocks:
https://github.com/OpenMathLib/OpenBLAS/blob/1a9cf8e2912b0597e1a6150180cd6aeca123bbe9/driver/level3/trmm_L.c#L151-L161

I think that'd be adding `SYMM_ONCOPY`/`TRMM_ONCOPY` to replace `GEMM_ONCOPY` if we override it, else falling back to the `GEMM_ONCOPY`?

martin-frbg:
Yes, at least that's my current understanding - that mapping ?SYMM_ONCOPY et al in the KERNEL file should do. I haven't tried it though, as my lowly M4 doesn't handle the SVE bits - guess I'd need to look into hijacking our Numpy-sponsored AWS CI job to run on M9g ?

Mousius:
I don't think we currently have `?SYMM_ONCOPY` or `?TRMM_ONCOPY`, the `RISCV` kernels replace the triangle copies along `M`:
https://github.com/OpenMathLib/OpenBLAS/blob/1a9cf8e2912b0597e1a6150180cd6aeca123bbe9/kernel/riscv64/KERNEL.RISCV64_ZVL128B#L169-L172

Which was originally done for SVE:
https://github.com/OpenMathLib/OpenBLAS/blob/1a9cf8e2912b0597e1a6150180cd6aeca123bbe9/kernel/arm64/KERNEL.ARMV8SVE#L149-L152

So we're aware of these ones being overridable, but the `GEMM_ONCOPY` doesn't have an override as yet.

For a sanity check, I couldn't find any reference to `SYMM_ONCOPY` / `TRMM_ITCOPY`:
* [SYMM_ONCOPY](https://github.com/search?q=repo%3AOpenMathLib%2FOpenBLAS%20SYMM_ONCOPY&type=code) / [SYMMONCOPY](https://github.com/search?q=repo%3AOpenMathLib%2FOpenBLAS%20SYMMONCOPY&type=code) / [SSYMM_ONCOPY](https://github.com/search?q=repo%3AOpenMathLib%2FOpenBLAS%20SSYMM_ONCOPY&type=code) / [SSYMMONCOPY](https://github.com/search?q=repo%3AOpenMathLib%2FOpenBLAS%20SSYMMONCOPY&type=code) 
* [TRMM_ITCOPY](https://github.com/search?q=repo%3AOpenMathLib%2FOpenBLAS%20TRMM_ITCOPY&type=code) / [TRMMITCOPY](https://github.com/search?q=repo%3AOpenMathLib%2FOpenBLAS%20TRMMITCOPY&type=code) / [STRMM_ITCOPY](https://github.com/search?q=repo%3AOpenMathLib%2FOpenBLAS%20STRMM_ITCOPY&type=code) / [STRMMITCOPY](https://github.com/search?q=repo%3AOpenMathLib%2FOpenBLAS%20STRMMITCOPY&type=code) 
* [TRMM_TCOPY](https://github.com/search?q=repo%3AOpenMathLib%2FOpenBLAS%20TRMM_TCOPY&type=code) / [TRMMTCOPY](https://github.com/search?q=repo%3AOpenMathLib%2FOpenBLAS%20TRMMTCOPY&type=code) / [STRMM_TCOPY](https://github.com/search?q=repo%3AOpenMathLib%2FOpenBLAS%20STRMM_TCOPY&type=code) / [STRMMTCOPY](https://github.com/search?q=repo%3AOpenMathLib%2FOpenBLAS%20STRMMTCOPY&type=code) 

Though it's worth noting this would also include `OTCOPY` / `INCOPY` / `ITCOPY` which are used in different kernels 🙀

Mousius:
> guess I'd need to look into hijacking our Numpy-sponsored AWS CI job to run on M9g ?

Just circling back here, @almayne can run this on relevant hardware 😸

martin-frbg:
> I don't think we currently have `?SYMM_ONCOPY` or `?TRMM_ONCOPY`, the `RISCV` kernels replace the triangle copies along `M`:

Oh, right. But I think it should be possible to do this within the existing framework, much like it was done for the SVE kernels you mentioned ? Basically, kernel/Makefile.L3 around line 2000 derives the "unmentioned" COPY kernels unconditionally from the corresponding generic C kernels no matter the architecture, and the salient change in the context of binebrank's SVE work was to add the `ifdef  ?TRMM_UNCOPY build kernel/$TRMM_UNCOPY else build kernel/generic/trmm_uncopy_$UNROLL.c like before`

Mousius:
> > I don't think we currently have `?SYMM_ONCOPY` or `?TRMM_ONCOPY`, the `RISCV` kernels replace the triangle copies along `M`:
> 
> Oh, right. But I think it should be possible to do this within the existing framework, much like it was done for the SVE kernels you mentioned ? Basically, kernel/Makefile.L3 around line 2000 derives the "unmentioned" COPY kernels unconditionally from the corresponding generic C kernels no matter the architecture, and the salient change in the context of binebrank's SVE work was to add the `ifdef ?TRMM_UNCOPY build kernel/$TRMM_UNCOPY else build kernel/generic/trmm_uncopy_$UNROLL.c like before`

I think we're agreeing that those extra `ifdef`s need adding in this patch? 😸

martin-frbg:
> I think we're agreeing that those extra `ifdef`s need adding in this patch? 😸

At least I think so - adding the ifdefs in the kernel makefile instead of redirecting driver calls to extra "COMM" kernels. But I haven't actually tried that rewrite, I hope I'm not missing something obvious ...

almayne:
Thanks for the feedback @martin-frbg and @Mousius. I'll give your recommendations a go as soon as I get a chance to pick this up again.

aditew01:
I think this will not be pushed in time for 0.3.33 release. cc: @martin-frbg  
@almayne please correct me if my assumption is wrong.

almayne:
The failing build looks unrelated. @Mousius, @martin-frbg, apologies for how long it's taken me to carve out some time for this. Please could you take a look when you get a chance.

PR #5575, opened 2025-12-19
Make the thunderx2 zdot kernel compatible with LLVM21 in Windows on Arm

martin-frbg:


kernel/arm64/zdot_thunderx2t99.c
```diff
@@ -1,359 +1,381 @@
...
+	"	bne	5f //dot_kernel_S_BEGIN		\n"
+
+	"1: //dot_kernel_F_BEGIN:			\n"
+	"	lsl	"INC_X", "INC_X", "INC_SHIFT"	\n"
+	"	lsl	"INC_Y", "INC_Y", "INC_SHIFT"	\n"
+	"	asr	"J", "N", #"N_DIV_SHIFT"	\n"
+	"	cmp	"J", xzr			\n"
+	"	beq	3f //dot_kernel_F1		\n"
+
+	"//	.align 5				\n"
```
vtjnash: Merge my PR (https://github.com/OpenMathLib/OpenBLAS/pull/5566) to avoid regressing quality on non-windows platforms? With `#if !(defined(__clang__) && defined(OS_WINDOWS))`

martin-frbg: certainly - this is just the second part required for making the change do anything useful for WoA when compiled on device, where LLVM (and CMake) is currently the only option.
the huge diff due to CRLF line endings is also spurious and temporary - but I really had to prioritize unrelated work when you brought it up

vtjnash: Not actually the only option: I am using clang from WSL, so that I can use all the familiar make tools and linux environment on-device, with native support for `.exe` files running natively (via binfmt magic). Only oddness is needing to set CROSS=1 to compile, then CROSS=0 to run the tests.

I can share the details of the setup if you like.

kernel/arm64/zdot_thunderx2t99.c
```diff
@@ -235,8 +243,9 @@ static void zdot_compute(BLASLONG n, FLOAT *x, BLASLONG inc_x, FLOAT *y, BLASLON
 	"	asr	"J", "N", #"N_DIV_SHIFT"	\n"
 	"	cmp	"J", xzr			\n"
 	"	beq	3f //dot_kernel_F1		\n"
-
+#ifndef _MSC_VER
 	"	.align 5				\n"
+#endif
```
vtjnash, line 248: The MSC compiler versions (aka cl) don't even support asm, so it cannot even parse this file. It is a bit confusing to have this macro being used to mean `defined __clang__ && defined _WIN32` here

PR #5577, opened 2025-12-22
add visibility=default to exported functions, visibility=hidden to most assembler

### LINKED ISSUE #5560: Ensure only public API has default symbol visibility, hide private symbols
This was attempted once in gh-3658 but abandoned. I think the right way of doing it is as pointed out in https://github.com/OpenMathLib/OpenBLAS/pull/3658#issuecomment-1176201131, something like this in a header and then marking public functions with `OPENBLAS_API`:

```C
#if defined (_WIN32) || defined (__CYGWIN__)
#  if defined (__GNUC__)
     /* GCC */
#    define OPENBLAS_EXPORT __attribute__ ((dllexport))
#    define OPENBLAS_IMPORT __attribute__ ((dllimport))
#  else
     /* MSVC */
#    define OPENBLAS_EXPORT __declspec(dllexport)
#    define OPENBLAS_IMPORT __declspec(dllimport)
#  endif
#else
   /* All other platforms. */
#  define OPENBLAS_EXPORT __attribute__ ((visibility ("default")))
#  define OPENBLAS_IMPORT
#endif

#if defined (OPENBLAS_SHLIB)
#  define OPENBLAS_API OPENBLAS_EXPORT 
#else
#  define OPENBLAS_API OPENBLAS_IMPORT
#endif
```

This resurfaced again in https://github.com/numpy/numpy/pull/30299, specifically on Alpine (a Musl-based distro) because Musl libc has different behavior from glibc and doesn't support dlopen'ing a shared library with `RTLD_LOCAL`. This is what NumPy does for its vendored `libscipy_openblas.so` which hides the (non-suffixed) private symbols well enough on glibc-based distros.

The method above is the standard way of exporting a public API. I think we should try to implement that; it's a fair bit of plumbing but it should be doable and will prevent obscure issues (e.g., see https://github.com/MacPython/openblas-libs/issues/79 for a previous one).

### LINKED ISSUE #5934: Ilp32 and ilp64 can't coexist in same process
Hi, I am trying to use an ilp64 build of OpenBlas alongside an ilp32. The ilp64 build is being built with SYMBOLSUFFIX=64_ to keep the names separate.

Turns out not all the names are modified with a suffix which leads to the ilp32 library calling an internal function in the ilp64 library. This can cause wrong memory accesses due to the 32/64 bit difference. Perhaps internal names (e.g., dsdot_k) should not be public in the build .so?

Here is a Dockerfile that shows the crash happening when linking against both ilp32+ilp64-with-suffixes.

```
FROM debian:bookworm

ARG OPENBLAS_REF=v0.3.33
ARG MAKE_JOBS=4

ENV DEBIAN_FRONTEND=noninteractive

RUN apt-get update && \
    apt-get install -y --no-install-recommends \
      ca-certificates curl file g++ gfortran make binutils perl xz-utils && \
    rm -rf /var/lib/apt/lists/*

WORKDIR /src

RUN curl -fsSL "https://github.com/OpenMathLib/OpenBLAS/archive/refs/tags/${OPENBLAS_REF}.tar.gz" \
      -o openblas.tar.gz && \
    mkdir openblas-ilp32 openblas-ilp64 && \
    tar -xzf openblas.tar.gz -C openblas-ilp32 --strip-components=1 && \
    tar -xzf openblas.tar.gz -C openblas-ilp64 --strip-components=1 && \
    rm openblas.tar.gz

RUN make -C /src/openblas-ilp32 -j"${MAKE_JOBS}" \
      USE_THREAD=0 NO_LAPACKE=1 && \
    make -C /src/openblas-ilp32 install \
      PREFIX=/opt/openblas-ilp32 USE_THREAD=0 NO_LAPACKE=1 && \
    make -C /src/openblas-ilp64 -j"${MAKE_JOBS}" \
      USE_THREAD=0 NO_LAPACKE=1 INTERFACE64=1 SYMBOLSUFFIX=64_ LIBNAMESUFFIX=cpp && \
    make -C /src/openblas-ilp64 install \
      PREFIX=/opt/openblas-ilp64 USE_THREAD=0 NO_LAPACKE=1 \
      INTERFACE64=1 SYMBOLSUFFIX=64_ LIBNAMESUFFIX=cpp && \
    rm -rf /src/openblas-ilp32 /src/openblas-ilp64

COPY link_time_reproducer.cpp /src/link_time_reproducer.cpp
COPY docker_run_reproducer.sh /usr/local/bin/docker_run_reproducer

RUN g++ -O0 -g -Wall -Wextra -std=c++17 \
      -o /usr/local/bin/link_time_reproducer /src/link_time_reproducer.cpp \
      -Wl,-rpath,/opt/openblas-ilp64/lib \
      -Wl,-rpath,/opt/openblas-ilp32/lib \
      -Wl,--no-as-needed \
      /opt/openblas-ilp64/lib/libopenblas64_cpp.so \
      /opt/openblas-ilp32/lib/libopenblas.so \
      -ldl && \
    chmod +x /usr/local/bin/docker_run_reproducer

CMD ["/usr/local/bin/docker_run_reproducer"]
```

[docker_run_reproducer.sh](https://github.com/user-attachments/files/30193431/docker_run_reproducer.sh)

[link_time_reproducer.cpp](https://github.com/user-attachments/files/30193435/link_time_reproducer.cpp)

mattip:
Towards #5560
- Add a `OPENBLAS_EXPORT` macro that maps to `__attribute__ ((visibility ("default")))`
- Add the macro to all the exported functions: everything in `interfaces` and some in `kernels/generic`
- Add a directive to the `PROLOGUE` assembler macro to hide functions by default, and use a `PROLOGUE_EXPORT` macro where functions should be exported

Where would be the best place to add a non-MSVC default `-fvisibility=hidden`? Somewhere in `Makefile.system` I guess, but how/where?

Note most of this was done as part of MacPython/openblas-libs#241, and I tested it there on the most popular builds.


lapack-netlib/LAPACKE/src/Makefile
```diff
@@ -39,7 +39,7 @@ include $(TOPSRCDIR)/make.inc
 
 .SUFFIXES: .c .o
 .c.o:
-	$(CC) $(CFLAGS) -I../include -c -o $@ $<
+	$(CC) $(filter-out -fvisibility=hidden, $(CFLAGS)) -I../include -c -o $@ $<
```
mattip, line 42: Rather than add `OPENBLAS_EXPORT` to all the functions here and in `../utils`, I removed the flag from the Makefile. I should do the same for the CMake. How does one tweak the flags per-directory in CMake?

mattip:
The clang builds are complaining about the assembler directive `.hidden`
```
  ../kernel/x86_64/qconjg.S:42:35: error: unknown directive
 .text;.align 5; .globl _qconjg_; .hidden _qconjg_; _qconjg_:
                                  ^
```

But the passing gcc builds are happy enough with it. I wonder if there is another way to tell clang to hide the assembler function without resorting to `objcopy`

mattip:
I got the documentation for `.hidden` [here](https://sourceware.org/binutils/docs/as/Hidden.html) which mentions `ELF`. So I will put it behind `#ifdef __ELF__`

mattip:
The Windows_cl job is failing since on windows I need the full `dllexport`, `dllimport` dance in `cblas.h`. It is used both internally during compilation (hence `dllimport`) and externally shipped (hence `dllexport`).

The other failures are due to timeouts or network issues, and I don't think they are connected to this PR.

I am still thinking about the best way to add `-fvisibility=hidden` to `CFLAGS`, any hints are welcome.

mattip:
Ahh, the shipped `cblas.h` is actually heavily modified in the build process (in both CMakeLists.txt and Makefile.install, look for "Generating cblas.h"), so I could add the OPENBLAS_EXPORT for the build and remove it in the generated file.

lapack-netlib/LAPACKE/src/Makefile
martin-frbg: Sorry for being a bit behind the curve on this - I think you could do a `string(REPLACE "-fvisibility=hidden" "" LAPACK_CFLAGS ${LAPACK_CFLAGS})` in cmake/system.cmake (or at the end of cmake/lapacke.cmake where there is the set_source_files_properties() to apply these LAPACK_CFLAGS)

mattip:
I think I have now removed the `hidden` flag if it is used, but so far I have not turned it on by default.

Where would be the best place to add a non-MSVC default -fvisibility=hidden? Somewhere in `Makefile.system` or `cmake/system.cmake` I guess, but how/where?

mattip:
Rebased off HEAD and added `-fvisbility=hidden` in fc243a3

martin-frbg:
Jenkins error is (on zarch):
```
[  6%] Building C object CMakeFiles/LAPACK_OVERRIDES.dir/lapack-netlib/SRC/slaruv.c.o
/home/alfred/jenkins/workspace/OpenBLAS-Z_PR-5577/lapack-netlib/SRC/slaruv.c:7:10: fatal error: exports.h: No such file or directory
    7 | #include <exports.h>
      |          ^~~~~~~~~~~
compilation terminated.
```
same as the HarmonyOS one

mattip:
Ahh, right, I adjusted the Makefiles but not the CMake files.

mattip:
I am not sure whether adjusting the include paths or copyng the `exports.h` file into LAPACKE/include is the correct solution to adding `OPENBLAS_EXPORTS` to the lapack functions. In the last commit I copied the file, it seems simpler.

martin-frbg:
Hmm, the CMake REPLACE doesn't work as the LAPACK_CFLAGS declared as a source property in lapack.cmake are only added to the general CMAKE_C_FLAGS, so in effect the LAPACKE symbols are all hidden. I haven't come up with s better idea than adding visibility=default to LAPACK_CFLAGS do that the settings cancel out. As far as I'm aware, there is nothing in LAPACKE that we'd want hidden anyway, so I guess having exports.h in LAPACKE/include is moot

martin-frbg:
Not sure what's up with the stubborn x86 Mac here - I see no fundamental reason why a DYNAMIC_ARCH build wouldn't find LAPACKE symbols when a targeted build does. Unfortunately I do not have easy access to such a system locally

mattip:
Can I replicate it with rosetta or does it require an actual macos-x86_64 machine?

martin-frbg:
nvm - I've reproduced it with Linux now too, it's the question of how to handle visibility in LAPACKE that I thought we had solved back in February, not entirely sure why this didn't show up in the other CI jobs too

PR #5579, opened 2025-12-22
fix define for YIELDING

vtjnash:
The intent seem to be to define this in one way, but previously it was
then immediately overriding it for various architectures, causing a
compiler warning.

### DIFF
--- a/common.h
+++ b/common.h
@@ -390,18 +390,19 @@ typedef int blasint;
 #define YIELDING
 #endif
 
+#if defined(_MSC_VER) && !defined(__clang__)
+#undef YIELDING // MSVC doesn't support assembly code
+#define YIELDING    	YieldProcessor()
+#endif
+
 #ifndef YIELDING
 #if defined(OS_SUNOS)
 #define YIELDING	thr_yield()
 
 #elif defined(OS_WINDOWS)
-# if defined(_MSC_VER) && !defined(__clang__)
-# define YIELDING    	YieldProcessor()
-# else
-# define YIELDING	SwitchToThread()
-# endif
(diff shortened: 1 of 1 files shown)


vtjnash:
Alternatively if the intent wasn't to override this, then I can swap the order of the declarations to define the architecture first.

martin-frbg:
intention is to override, but I haven't checked which of the two is actually preferable on Windows - though I suspect it will be the asm no-op that currently "wins" amid the noisy warning.

vtjnash:
Alright, can change that. I see past issues now that indicate why it uses the spin lock with nops (https://github.com/OpenMathLib/OpenBLAS/issues/1881#issuecomment-450359064)

martin-frbg:
seems to be missing an `#endif` now somewhere, although the cause is not immediately obvious to me from the diff

vtjnash:
ah, yep, it was missing at the obvious place

common.h
```diff
@@ -390,18 +390,19 @@ typedef int blasint;
...
 #elif defined(OS_WINDOWS)
-# if defined(_MSC_VER) && !defined(__clang__)
-# define YIELDING    	YieldProcessor()
-# else
-# define YIELDING	SwitchToThread()
-# endif
+#define YIELDING	SwitchToThread()
 
-#else
+#else // assume linux
```
martin-frbg: not only linux but *bsd as well - and probably other unixoid systems as well - this is just "assume sched_yield may work better with an unknown scheduler than doing nothing at all", and was historically what GotoBLAS/OpenBLAS would use on any host. perhaps simply leave out the comment ?

PR #5581, opened 2025-12-23
cmake: document options and set defaults for influential variables

vtjnash:
This makes `ccmake` list all of these values, by letting it know that they can be influential, and let the user know what they do. Adds documentation strings and defaults for CMake cache variables including TARGET, BINARY, threading options, vector extensions, memory tuning, and precision types. Treat empty string values as "not set" to allow auto-detection while keeping options visible in ccmake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,6 +79,61 @@ set(SYMBOLPREFIX "" CACHE STRING  "Add a prefix to all exported symbol names in 
 
 set(SYMBOLSUFFIX "" CACHE STRING  "Add a suffix to all exported symbol names in the shared library, e.g. _64 for INTERFACE64 builds" )
 
+# Target architecture - auto-detected if not specified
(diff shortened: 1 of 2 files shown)


martin-frbg:
Looks like your (or  Claude's ?) change to system.cmake broke Cooperlake option handling, please check.

vtjnash:
A little bit of both of our fault. Claude wanted to re-write all the files to use the explicit `"${VAR}" = "string"` form instead of the automatic `VAR = string`, and I wanted to reduce the diff so I undid that, but didn't realize that leaving the added quotes here would break it.

vtjnash:
Bump? Seems like everything passed (unless it got killed when CI ran out of time) now that I removed the unnecessary part of this change.

vtjnash:
bump?

vtjnash:
rebased (no changes)

PR #5584, opened 2025-12-23
build: remove unused REFNAME macro from SANITY_CHECK

vtjnash:
REFNAME was defined as a preprocessor macro (-DREFNAME=$(*F)f$(BU)) but
appears to have never been consumed by any source code. The SANITY_CHECK
feature uses common_reference.h which has its own independent mechanism
via BLASFUNC_REF macros with hardcoded REF_BU=f.

For context, REFNAME was imported with GotoBLAS2 in 2011 but became dead
code when the unit test framework added common_reference.h with its own
naming scheme. The CMake version also had a broken $(*F) reference that
could never work, but this was also moot since the macro was unused.


martin-frbg:
pretty sure the entirevsanity_check is dead code from gotoblas

vtjnash:
It enables building the code in `reference/` and adding those to the common.h file, so it would be a functional change to delete it, but does seem unlikely to be used now

martin-frbg:
It may or may not build it ("it" being a <=2010 version of the Reference BLAS), but the code to run a calculation with both implementations and compare the results was never implemented. Kazushige Goto put several such empty hooks and dead ends into the initial GotoBLAS and never got around to implementing them - but it isn't worthwile to throw them out.
The separate BLAS-Tester project - adapted from the ATLAS testsuite - serves mostly the same purpose - just in a systematic way rather than tracing a specific user-specified sequence of calls.

vtjnash:
Reducing/removing broken options that can result in broken configs seems meaningful to me. If you want to merge or close this, I can make another PR that deletes all of the dead code.

vtjnash:
bump

PR #5591, opened 2026-01-05
Support for SME1 based ssyr2k_direct kernel for cblas_ssyr2k level 3 API

zhiqxie:
Support for SME1 based ssyr2k_direct kernel for cblas_ssyr2k level 3 API


martin-frbg:
thanks - I think you have a missing endif() in your addition to kernel/CMakeLists.txt around line 255, could you please have a look ?

martin-frbg:
Failed checks on Cirrus were due to a network error, but the other ones are caused by missing changes to common_param.h and setparam-ref.c (both used in DYNAMIC_ARCH builds, compare e.g. #5471)

kernel/CMakeLists.txt
```diff
@@ -311,6 +315,16 @@ function (build_core TARGET_CORE KDIR TSUFFIX KERNEL_DEFINITIONS)
       endif ()
     endif()
 
+    if (USE_DIRECT_SSYR2K)
+      if (ARM64)
+	    set (SSYR2KDIRECTKERNEL_ALPHA_BETA ssyr2k_direct_alpha_beta_arm64_sme1.c)
+      GenerateNamedObjects("${KERNELDIR}/${SSYR2KDIRECTKERNEL_ALPHA_BETA}" "" "ssyr2k_direct_alpha_betaUN" false "" "" false SINGLE)
+      GenerateNamedObjects("${KERNELDIR}/${SSYR2KDIRECTKERNEL_ALPHA_BETA}" "" "ssyr2k_direct_alpha_betaUT" false "" "" false SINGLE)
```
martin-frbg: has to be "syr2k_direct_alpha_betaUN" etc here - the "SINGLE" adds the leading "s"

zhiqxie: Fixed, thanks, to avoid this build error, is there a guidle for locally build?

martin-frbg: nothing too specific - if you have an arm64 machine with clang and cmake, you'd just need to create a build directory, change into it and run "cmake -DTARGET=ARMV8 DYNAMIC_ARCH=1 .." followed by "cmake --build ." and "ctest" to make sure the tests pass. (I guess it would make sense to describe the cmake build layout in the docs - somehow everybody always goes from "I understand too little to document it" directly to "it's so obvious now" - or they run away screaming)

PR #5594, opened 2026-01-05
Fix bug where openblas_set_threads_callback_function does not support modyfing symbol prefix and suffix in shared library

### LINKED ISSUE #5593: openblas_set_threads_callback_function missing prefix/suffix in shared library when SYMBOLPREFIX/SYMBOLSUFFIX are set in Makefile
# Description
An **undefined reference** linking error was detected during testing. This error arises because the symbol prefix/suffix for the function `openblas_set_threads_callback_function` in the dynamic library is missing after `SYMBOLPREFIX` or `SYMBOLSUFFIX` is defined (this feature was introduced by PR #4577). When `SYMBOLPREFIX` or `SYMBOLSUFFIX` is configured in the Makefile, the function signature of `SYMBOLPREFIX_openblas_set_threads_callback_function_SYMBOLSUFFIX` in the header file `cblas.h` is modified (with the configured prefix/suffix added), but the function name in the dynamic link library remains unmodified, resulting in an undefined reference.

# Steps to Reproduce
1. Set `SYMBOLPREFIX` or `SYMBOLSUFFIX` in the Makefile (e.g., `SYMBOLSUFFIX=_smp`).
2. Compile the OpenBLAS dynamic library.
3. Compile and link the following test code against the compiled library:
    ```c
    #include <omp.h>
    #include "cblas.h"

    void myfunction_(int sync, openblas_dojob_callback_smp dojob, int numjobs, size_t jobdata_elsize, void *jobdata, int dojob_data)
    {
        #pragma omp parallel for
        for(int i = 0; i < numjobs; i++)
        {
            void *element_addr = (void *) (((char *)jobdata) + ((unsigned) i) * jobdata_elsize);
            dojob(i, element_addr, dojob_data);
        }
        return;
    }

    int main() {
        openblas_set_num_threads_smp(4);
        omp_set_num_threads(4);
        openblas_set_threads_callback_function_smp(myfunction_);
        return 0;
    }
    ```
4. Trigger the linking error: **undefined reference to `openblas_set_threads_callback_function_smp'**

# Root Cause
`openblas_set_threads_callback_function` is not included in the `exports/gensymbol` and `exports/gensymbol.py` files. As a result, the symbol generation script does not apply the `SYMBOLPREFIX`/`SYMBOLSUFFIX` configured in the Makefile to the `openblas_set_threads_callback_function` when processing the dynamic library.

# Related PR
The original PR that introduced the `openblas_set_threads_callback_function` feature: #4577

lujiaweics:


martin-frbg:
thank you

al3xtjames:
Should these symbols be limited to multithreaded builds? I'm seeing this error while trying to build a single-threaded build on macOS:

```
Undefined symbols for architecture x86_64:
  "_openblas_set_threads_callback_function", referenced from:
     -exported_symbol[s_list] command line option
ld: symbol(s) not found for architecture x86_64
make[1]: *** [Makefile:166: libopenblas-r0.3.31.dylib] Error 1
make[1]: Leaving directory '/nix/var/nix/b/20zrcf6cgm5ah1rq0f4byy2dcm/source/exports'
make: *** [Makefile:158: shared] Error 2
```

It looks like this is because blas_server_callback.c (which defines `openblas_set_threads_callback_function`) is only built if multithreading is enabled: https://github.com/OpenMathLib/OpenBLAS/blob/45e9820118b7747f2e34f1de54f5491562f3371b/driver/others/Makefile#L8-L13

PR #5595, opened 2026-01-06
Fixing warning messages in dgemm and dgemv kernels

amritahs-ibm:
Fixing warning messages in dgemm and dgemv kernels


martin-frbg:
thank you

PR #5596, opened 2026-01-09
docs: fix iOS build script & use xcrun SDK path

moluopro:
The original script had an invalid CFLAGS assignment (CFLAGS= -O2 ...), which caused zsh to treat -O2 as a command and fail with:
```
zsh: command not found: -O2
```

It also hardcoded the iPhoneOS SDK and Xcode toolchain paths, which may not exist on all machines and can prevent the build from running successfully.

This update:

1. Fix the shell syntax by assigning CFLAGS correctly

2. Use `xcrun --sdk iphoneos --show-sdk-path` to resolve the active iOS SDK dynamically

3. Build only the libraries (make libs) to avoid running iOS test binaries on macOS

I have tested this on a real macOS machine and the build completes successfully.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -712,11 +712,12 @@ fully working OpenBLAS for this platform.
 
 Go to the directory where you unpacked OpenBLAS,and enter the following commands:
 ```bash
-CC=/Applications/Xcode_12.4.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang
+CC="/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang"
 
-CFLAGS= -O2 -Wno-macro-redefined -isysroot /Applications/Xcode_12.4.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS14.4.sdk -arch arm64 -miphoneos-version-min=10.0
+SDKROOT="$(xcrun --sdk iphoneos --show-sdk-path)"
+CFLAGS="-O2 -Wno-macro-redefined -isysroot $SDKROOT -arch arm64 -miphoneos-version-min=10.0"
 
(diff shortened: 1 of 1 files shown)


docs/install.md
```diff
@@ -712,11 +712,12 @@ fully working OpenBLAS for this platform.
...
 ```bash
-CC=/Applications/Xcode_12.4.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang
+CC="/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang"
 
-CFLAGS= -O2 -Wno-macro-redefined -isysroot /Applications/Xcode_12.4.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS14.4.sdk -arch arm64 -miphoneos-version-min=10.0
+SDKROOT="$(xcrun --sdk iphoneos --show-sdk-path)"
+CFLAGS="-O2 -Wno-macro-redefined -isysroot $SDKROOT -arch arm64 -miphoneos-version-min=10.0"
 
-make TARGET=ARMV8 DYNAMIC_ARCH=1 NUM_THREADS=32 HOSTCC=clang NOFORTRAN=1
+make libs TARGET=ARMV8 DYNAMIC_ARCH=1 NUM_THREADS=32 HOSTCC=clang NOFORTRAN=1
```
martin-frbg: I do not think this change is correct - why would you want to make "libs" only instead of a full build ?

moluopro: Build only the libraries (make libs) to avoid running iOS test binaries on macOS

moluopro: If you think `make libs` is wrong, I can roll it back. Are there any other problems?

martin-frbg: Cross-build condition should already prevent running the tests

moluopro: Done.

PR #5598, opened 2026-01-11
build: skip tests when building for iOS

moluopro:
I tried building for the iOS following the doc and encountered the following error:
```
...
TEST 1520/1522 csbmv:xerbla_inc_c_zero [OK]
TEST 1521/1522 csbmv:xerbla_k_invalid [OK]
TEST 1522/1522 csbmv:xerbla_lda_invalid [OK]
RESULTS: 1522 tests (1516 ok, 6 failed, 0 skipped) ran in 184 ms
make[1]: *** [run_test] Error 6
make: *** [tests] Error 2
```

Initially, I thought this was caused by the tests being built, so I tried modifying the doc's `make` to `make libs`. However, I was informed that ["Cross-build condition should already prevent running the tests"](https://github.com/OpenMathLib/OpenBLAS/pull/5596#discussion_r2678618947).

The original build process didn’t distinguish the `iOS` platform and treated it the same as `Darwin`. I updated the `Darwin` build process as follows:

1. Add in `ctest.c`:
```c
#if defined(TARGET_OS_IPHONE)
OS_IOS
#endif
```

2. Add an `is_ios` variable in `c_check`, setting `cross=1` when it is true.

I tested this on a physical device, and the `iOS` build now works correctly.

### DIFF
--- a/c_check
+++ b/c_check
@@ -23,6 +23,7 @@ config="$2"
 compiler_name="$3"
 shift 3
 flags="$*"
+is_ios=false
 
 # First, we need to know the target OS and compiler name
 {
@@ -78,6 +79,7 @@ case "$data" in *OS_CYGWIN_NT*) os=CYGWIN_NT ;; esac
 case "$data" in *OS_INTERIX*) os=Interix ;; esac
 case "$data" in *OS_ANDROID*) os=Android ;; esac
 case "$data" in *OS_HAIKU*) os=Haiku ;; esac
+case "$data" in *OS_IOS*) is_ios=true ;; esac
 
 case "$data" in
     *ARCH_X86_64*) architecture=x86_64 ;;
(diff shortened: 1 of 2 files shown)


martin-frbg:
Curious as the CI job on Cirrus does not run into such problems

moluopro:
> Curious as the CI job on Cirrus does not run into such problems

From the logic of the build script, this issue does exist. I ran some tests locally, and the results support my thinking. Is there anything else I need to adjust?

moluopro:
I found the reason. In `.cirrus.yml`, when building the iOS library, `CROSS=1` was set manually. I think `iOS` and `Android` builds should be consistent (i.e., it should work properly without manually specifying `CROSS=1`). Of course, this is also how it is described in the documentation.
```yml
macos_instance:
  image: ghcr.io/cirruslabs/macos-sonoma-xcode:latest
task:
  name: AppleM1/LLVM armv8-ios xbuild
  ......
  - make TARGET=ARMV8 NUM_THREADS=32 HOSTCC=clang NOFORTRAN=1 CROSS=1
```

martin-frbg:
Ah, I see - good catch, thank you.

PR #5618, opened 2026-01-19
arm64: fix clang ICE on Windows for thunderx2t kernels

vtjnash:
Guard .align directive to avoid internal compiler error on AArch64 Windows with clang.

See: https://github.com/llvm/llvm-project/issues/149547
Update of https://github.com/OpenMathLib/OpenBLAS/pull/5566 to fix remaining issues
See: #5076

### DIFF
--- a/kernel/arm64/zdot_thunderx2t99.c
+++ b/kernel/arm64/zdot_thunderx2t99.c
@@ -243,7 +243,8 @@ static void zdot_compute(BLASLONG n, FLOAT *x, BLASLONG inc_x, FLOAT *y, BLASLON
 	"	asr	"J", "N", #"N_DIV_SHIFT"	\n"
 	"	cmp	"J", xzr			\n"
 	"	beq	3f //dot_kernel_F1		\n"
-#ifndef _MSC_VER
+/* https://github.com/llvm/llvm-project/issues/149547 */
+#if !(defined(__clang__) && defined(OS_WINDOWS))
 	"	.align 5				\n"
 #endif
 	"2: //dot_kernel_F:				\n"
--- a/kernel/arm64/dznrm2_thunderx2t99_fast.c
+++ b/kernel/arm64/dznrm2_thunderx2t99_fast.c
@@ -155,7 +155,10 @@ static double nrm2_compute(BLASLONG n, FLOAT *x, BLASLONG inc_x)
 	"	cmp	"J", xzr			\n"
 	"	beq	.Lnrm2_kernel_F1		\n"
 
+/* https://github.com/llvm/llvm-project/issues/149547 */
+#if !(defined(__clang__) && defined(OS_WINDOWS))
 	"	.align 5				\n"
+#endif
 	".Lnrm2_kernel_F:				\n"
 	"	"KERNEL_F"				\n"
 	"	subs	"J", "J", #1			\n"
--- a/kernel/arm64/scnrm2_thunderx2t99.c
+++ b/kernel/arm64/scnrm2_thunderx2t99.c
@@ -238,7 +238,10 @@ static double nrm2_compute(BLASLONG n, FLOAT *x, BLASLONG inc_x)
 	"	cmp	"J", xzr			\n"
 	"	beq	5f //nrm2_kernel_S_BEGIN	\n"
 
+/* https://github.com/llvm/llvm-project/issues/149547 */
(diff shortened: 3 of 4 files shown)


kernel/arm64/zdot_thunderx2t99.c
```diff
@@ -243,7 +243,8 @@ static void zdot_compute(BLASLONG n, FLOAT *x, BLASLONG inc_x, FLOAT *y, BLASLON
 	"	asr	"J", "N", #"N_DIV_SHIFT"	\n"
 	"	cmp	"J", xzr			\n"
 	"	beq	3f //dot_kernel_F1		\n"
-#ifndef _MSC_VER
+/* https://github.com/llvm/llvm-project/issues/149547 */
+#if !(defined(__clang__) && defined(OS_WINDOWS))
```
martin-frbg, line 247: I believe the _MSC_VER to be equivalent here (it will never match plain MSVC as that can't handle the inline assembly anyway, but it matches any LLVM derivative that uses the clang-cl shim) ?

martin-frbg:
Thanks - did you happen to check if these changes are sufficient for using these files with WoA ? I think it is likely that they might need changes to the storing of the complex results (similar to the earlier zdot changes) but I won't get around to checking this myself today

kernel/arm64/zdot_thunderx2t99.c
vtjnash: It might be similar? Hard to check exactly if all libc specify that clang+_WIN32 is exactly the same as MSC when this needs to do is to specify clang+_WIN32 and exclude MSVC. It is just very confusing to intentionally use a macro to work around a bug in one specific compiler by specifying that this should be disabled for a completely different compiler

vtjnash:
> Thanks - did you happen to check if these changes are sufficient for using these files with WoA ?

I'm not precisely sure what you want checked. It compiles and seems to pass tests. clang has always supported c99 just fine (unlike MSVC, which still is buggy): https://github.com/OpenMathLib/OpenBLAS/blob/5ffbf38b41a1fe494d038efcb09cf22f4d527c22/common.h#L579-L583

kernel/arm64/zdot_thunderx2t99.c
martin-frbg: Oh well, I'm not opposed to switching to your version if it feels less confusing. My idea was that it would catch any LLVM-derived compiler that goes through the MSVC backend as well, not immediately sure if those would also define `__clang__`

martin-frbg:
CI failures are spurious BTW

> > Thanks - did you happen to check if these changes are sufficient for using these files with WoA ?
> 
> I'm not precisely sure what you want checked. It compiles and seems to pass tests. clang has always supported c99 just fine (unlike MSVC, which still is buggy):
> 
> https://github.com/OpenMathLib/OpenBLAS/blob/5ffbf38b41a1fe494d038efcb09cf22f4d527c22/common.h#L579-L583

Back in december I had to work around the OPENBLAS_COMPLEX_FLOAT macro in zdot_thunderx2t99.c not compiling (with LLVM21 on a WoA laptop I had on loan from Qualcomm), but could be the conditional in common.h is borked.
BTW the two NRM2 kernels had been "mostly" retired in 2024 due to low accuracy and inability to handle some corner cases - this will probably need a revisit from someone who really knows ARM assembly before they can be put to good use again.

kernel/arm64/zdot_thunderx2t99.c
vtjnash: My thought is that eventually it gets fixed in llvm, and then these checks can be changed to an exact `__clang_major__` check

There is clang/c2, which is clang with msvc backend (https://devblogs.microsoft.com/cppblog/clang-c2-we-need-your-advice/), but that wouldn't have this bug. I don't think there is an implementation for using msvc frontend with llvm backend

vtjnash:
It might be because I'm using mingw-w64 headers, which seem to have full c99 support, while it sounded like you might have been using MSVC headers?

martin-frbg:
> It might be because I'm using mingw-w64 headers, which seem to have full c99 support, while it sounded like you might have been using MSVC headers?

I guess that would explain it - I'm doing my best to keep environments separate so that both MSYS2 and "native" LLVM package builds will work on WoA.

PR #5619, opened 2026-01-20
Serialize accesses to parallelized syrk functions from multiple calle…

lujiaweics:
…rs, like it was already done for GEMM in level3_thread.c and GEMM3M in level3_gemm3m_thread.c


martin-frbg:
Thanks, not sure why that got missed. I worry that something similar might be needed for getrf/potrf as well

lujiaweics:
Thanks for pointing this out! I'll prioritize checking getrf and potrf soon and follow up with any necessary changes.

lujiaweics:
@martin-frbg Hi Martin, I've checked the code and agree that getrf/potrf are missing the concurrency-limiting code as well. Could you confirm if I should create a new PR to fix this bug? Thanks！

PR #5626, opened 2026-01-23
Fix CMake/LLVM compilation issues seen under Windows-on-Arm

martin-frbg:


martin-frbg:
need to redo this without the overzealous attemp to remove the sme_abi header

PR #5634, opened 2026-02-04
Fix: Remove invalid parentheses after endif

yuanjia111:
Solve the compilation problem shown in the image below：
image


martin-frbg:
Thank you

PR #5640, opened 2026-02-10
Added ability to accumulate in FP16.  Convert BF16 to FP32.  For FP16 and BF16 GEMM in RISC-V (BF16 now works for pre-RVA23)

ChipKerchner:
Added ability to accumulate in FP16 for GEMM.  Widens once at the end of loops.

```
Testing LLVM FP16 LMUL1 VLEN256 GEMM 1 0 0  512  512  512   1  2.0  1.0  1

Total time =         24948910

Testing LLVM FP16_N LMUL1 VLEN256 GEMM 1 0 0  512  512  512   1  2.0  1.0  1

Total time =         18968190
```

Accumulation differences are about 4 epsilons - compared to the widening (previous) version.  But the performance it up to 2.7X faster - Note: BananaPi shows only 1.85X faster.

BF16 is 1.5X faster and now works for RVA22 systems like BananaPi


ChipKerchner:
Unfortunately BF16 only has widening MADD instructions.  So the same changes can not be made for BF16.

ChipKerchner:
These are for VLEN = 256 only currently

ChipKerchner:
It now works for VLEN = 128.

ChipKerchner:
Main loop now uses LMUL = 2

ChipKerchner:
Even faster!!!

```
Testing LLVM FP16_N LMUL1 VLEN256 GEMM 1 0 0  512  512  512   1  2.0  1.0  1

Total time =         13400067
```

ChipKerchner:
Convert inputs from BF16 to FP32 and use FP32 vector madds.  18% faster.

ChipKerchner:
Convert BF16 values once (and vectorized) - 3-4% faster.

ChipKerchner:
Latest BF16 version is 48% faster than the current version.  And is 3% faster than the FP32 .

It uses 1.5X more memory unfortunately.

It would be possible to make this version to work on the BananaPi - which doesn't support BF16 vector instructions.     The conversions could be done with uint32 conversion and shift.  And fixing the remaining BF16 vectors to not uses vector BF16 MADDs.

ChipKerchner:
BF16 GEMM now works for pre-RVA23 systems like BananaPi

ChipKerchner:
This is ready to go unless someone has comments.

PR #5643, opened 2026-02-11
Fix SGEMM returning wrong results in multithreading on NeoverseV2

martin-frbg:
Fixes https://github.com/numpy/numpy/issues/30816

### DIFF
--- a/param.h
+++ b/param.h
@@ -3662,6 +3662,8 @@ is a big desktop or server with abundant cache rather than a phone or embedded d
 
 #elif defined(NEOVERSEN2) || defined(NEOVERSEV2)
 
+#define GEMM_DIVIDE_LIMIT       3
+
 #if defined(XDOUBLE) || defined(DOUBLE)
 #define SWITCH_RATIO            8
 #else


mattip:
Is this ready to go?

martin-frbg:
It definitely fixes the problem on NeoverseV2, but (a) there may be other arm64 cpus similarly affected and (b) I haven't fully understood the underlying issue with that specific parameter that was introduced fairly recently - it may be papering over a missing tail call in the gemm kernel it rode in on, or something else entirely.

martin-frbg:
@Mousius I'm a bit confused now as I notice NeoverseV2 was already using the SVE SGEMM kernel (via ARMV8SVE) until you switched its KERNEL file to be based on NEOVERSEN2 rather than ARMV8SVE in order to reuse N2's sbgemm kernel in the otherwise unrelated #5399 . Was that a conscious decision to return the V2 to the basic Neon kernel due to its shorter vector register size compared to V1&A64FX, or collateral damage that I missed at the time ?

martin-frbg:
Reverting as it turns out to be entirely unrelated to the actual numpy issue, a heisenbug in the asimd dot kernel

PR #5646, opened 2026-02-12
use 100 minute timeout for azure mingw32 job

mattip:
The Windows_mingw_gmake job is consistently timing out at the default 60 minutes. Add an explicit timeout and make it 100 minutes.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -91,6 +91,7 @@ jobs:
       openblas_utest.exe
 
 - job: Windows_mingw_gmake
+  timeoutInMinutes: 100
   pool:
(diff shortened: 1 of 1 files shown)


mattip:
CI is passing, including azure. Looking at the [logs](https://dev.azure.com/martinkroeker/martinkroeker/_build?definitionId=2&_a=summary), this is the first green CI run on azure since Dec 25. 

Edit: the Dec 25 run was 58 minutes, the current run is 1hr4min. 100 minutes seems safe.

martin-frbg:
Thanks - still annoying that it takes so long, but IIRC that's just mingw's way of handling files

PR #5659, opened 2026-03-02
Handle Intel's emerald rapids and do some formatting in the cpuid_x86 file

lindsayad:
The instruction set for emerald rapids is the same as sapphire rapids so I select the same arch's for the emerald rapids branches as for the sapphire rapids branches, e.g.
```c
        case 15: // Emerald Rapids
          if (support_amx_bf16()) return CORE_SAPPHIRERAPIDS;
          if (support_avx512_bf16()) return CORE_COOPERLAKE;
          if (support_avx512()) return CORE_SKYLAKEX;
          if (support_avx2()) return CORE_HASWELL;
          if (support_avx()) return CORE_SANDYBRIDGE;
          else return CORE_NEHALEM;
```
I was also having a very hard time following the code with the different indentation levels so I've clang-formatted the file. I can understand if this is undesirable so I will revert if asked, but in my opinion it significantly improves the file's readability.

### DIFF
--- a/cpuid_x86.c
+++ b/cpuid_x86.c
@@ -1513,6 +1513,11 @@ int get_cpuname(void)
         break;
       case 12: //family 6 exmodel 12
         switch (model) {
+        case 6: // Arrow Lake
+          if (support_avx512()) return CPUTYPE_SAPPHIRERAPIDS;
+          if (support_avx2()) return CPUTYPE_HASWELL;
+          if (support_avx()) return CPUTYPE_SANDYBRIDGE;
+          else return CPUTYPE_NEHALEM;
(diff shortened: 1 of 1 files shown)


martin-frbg:
Thanks - indeed normally I don't like to see major reformatting mixed in with functional changes, but here the functional changes are harmless enough and those scraggly case lists really had it coming...

lindsayad:
Commits 1 and 3 are the functional changes (well really just commit 1; commit 3 just re-orders a couple of the case statements). Commit 2 isolates the clang-formatting

lindsayad:
thanks for the quick turn-around!

PR #5670, opened 2026-03-09
powerpc: Bypass FP16 as BF16 on Power

amritahs-ibm:
typedef FP16 as BF16 on Power as FP16 is
not yet supported

### DIFF
--- a/openblas_config_template.h
+++ b/openblas_config_template.h
@@ -40,7 +40,11 @@ typedef uint16_t bfloat16;
 #endif
 
 #if defined(__GNUC__) && (__GNUC__ > 12)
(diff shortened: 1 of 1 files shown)


martin-frbg:
Thanks - is uint16_t unsupported there too ?

RajalakshmiSR:
> Thanks - is uint16_t unsupported there too ?
 
If the bfloat16 type is unavailable, the implementation defaults to uint16_t.
https://github.com/OpenMathLib/OpenBLAS/blob/55b16e5923d48cd3de42ff82c2c024a4a0c7e043/common.h#L266

PR #5671, opened 2026-03-10
Update x86 cache size decoding table from current sandpile.org data

martin-frbg:


cpuid_x86.c
```diff
@@ -489,10 +489,10 @@ int get_cacheinfo(int type, cache_info_t *cacheinfo)
     info[12] = BITMASK(edx, 8, 0xff);
     info[13] = BITMASK(edx, 16, 0xff);
     info[14] = BITMASK(edx, 24, 0xff);
-
+fprintf(stderr,"numcalls needed to retrieve all cache data %d\n",numcalls);
```
mattip, line 492: Formatting is a bit strange. Did you mean to remove this debug info?

martin-frbg: Meh, that was some last-minute thought to check if repeat calls for cache info are still required on somewhat recent hardware. Thanks for notifying me

PR #5674, opened 2026-03-12
Improve performance on edges of GEMM for RISC-V

ChipKerchner:
Improve performance on edges of GEMM for RISC-V - up to 9X faster.

image

image


ChipKerchner:
Small values of M and/or N are useful for situations like AI inferencing.

ChipKerchner:
This increases the utilization to nearer peak numbers.

ChipKerchner:
@martin-frbg What's going on here?

`TEST 126/128 fork:safety Attempt 10 timed out (retrying...)
All 10 attempts failed, giving up.`

ChipKerchner:
This has nothing to do with this patch but could we limit the number of failures (or differences) to a reasonable number (like 200 or so)?  It makes for some huge log files and makes looking for failures difficult.

`2026-03-12T13:22:24.3299509Z CC 261.500000 DD  261.181854 
2026-03-12T13:22:24.3328792Z CC 268.875000 DD  269.182281 
2026-03-12T13:22:24.3329295Z CC 259.125000 DD  258.630676 
2026-03-12T13:22:24.3330928Z CC 256.062500 DD  256.341339 
2026-03-12T13:22:24.3332162Z CC 261.937500 DD  262.319489 
2026-03-12T13:22:24.3373825Z SHGEMM FAILURES: 527559!!!
2026-03-12T13:23:13.9093248Z BGEMM FAILURES: 224`

BTW, these are probably reasonable differences for SHGEMM - since they are only off in the last bit.

martin-frbg:
> @martin-frbg What's going on here?
> 
> `TEST 126/128 fork:safety Attempt 10 timed out (retrying...) All 10 attempts failed, giving up.`

If it's C910V, that's a (qemu) thread lockup in a forked process that is not reproducible on actual  hardware so far

ChipKerchner:
Ok, let me report the QEMU issue to my team members.

martin-frbg:
> This has nothing to do with this patch but could we limit the number of failures (or differences) to a reasonable number (like 200 or so)? It makes for some huge log files and makes looking for failures difficult.
> 
> `2026-03-12T13:22:24.3299509Z CC 261.500000 DD 261.181854 2026-03-12T13:22:24.3328792Z CC 268.875000 DD 269.182281 2026-03-12T13:22:24.3329295Z CC 259.125000 DD 258.630676 2026-03-12T13:22:24.3330928Z CC 256.062500 DD 256.341339 2026-03-12T13:22:24.3332162Z CC 261.937500 DD 262.319489 2026-03-12T13:22:24.3373825Z SHGEMM FAILURES: 527559!!! 2026-03-12T13:23:13.9093248Z BGEMM FAILURES: 224`
> 
> BTW, these are probably reasonable differences for SHGEMM - since they are only off in the last bit.

That's probably why we had them printed at some point, to see if they're all reasonable (and the test criteria need to be adjusted) ?

ChipKerchner:
FYI - in the graphs above M and N are modulo the block size - in this case 16x8.  But the biggest performance gains are for small M values.

martin-frbg:
> Ok, let me report the QEMU issue to my team members.

Unfortunately it's a bit complicated - stock qemu doesn't support C910V (to my knowledge) so CI is using some Xuantie fork of qemu9 that may or may not be actively maintained.

ChipKerchner:
Since C910V only supports RVV 0.7.1 - maybe this build shouldn't get kicked off on every check-in.  RVV 1.0 has been ratified about 3 years ago.  Unfortunately I guess until there is more silicon, maybe this is needed?

martin-frbg:
> Since C910V only supports RVV 0.7.1 - maybe this build shouldn't get kicked off on every check-in. RVV 1.0 has been ratified about 3 years ago. Unfortunately I guess until there is more silicon, maybe this is needed?

I wonder how many (other) early adopters bought a MilkV Pioneer or similar... by the time everybody has dumped them, we're probably looking at RVV1.5 or the like and the cycle begins anew. I guess I could look into disabling the utest in CI so that we get at least the build and BLAS tests

ChipKerchner:
The way I hear after RVA23 there won't be a major upgrade until RVA30 (?).  There may be a minor one though.  But there are plenty of optional extension to be used and are being ratified.

I think RVV 1.0 is the way to go.  RVV 0.7.1 was mainly a hack.

martin-frbg:
reminds me of Felix LeClerc's FOSDEM talk... we'll see how RISCV64 standards evolve but the earlier generation(s?) of miscreants probably won't go away soon

kernel/riscv64/sgemm_kernel_16x8_zvl256b.c
```diff
@@ -40,33 +40,2200 @@ AUTOGENERATED KERNEL
...
 
+int CNAME(BLASLONG M, BLASLONG N, BLASLONG K, FLOAT alpha, FLOAT* A, FLOAT* B, FLOAT* C, BLASLONG ldc)
 {
     BLASLONG gvl = 0;
     BLASLONG m_top = 0;
     BLASLONG n_top = 0;
-
+    const BLASLONG m_edge = M & 15;
+    const bool S = (M == (ldc & 0xF));
+    if (K <= 0) return 0;
```
ChipKerchner: There was actually a bug in the original for K = 0.  It would add the first A times B onto C when it should not.

martin-frbg: Good catch, very unfortunate that the tests apparently did not catch it

ChipKerchner: Maybe the check length is related to K (?) - which doesn't make sense since it should check the whole M x N input/output.  Either that or something isn't calling GEMM for K = 0

martin-frbg: Ah right, there is an early exit for K=0 (or alpha=0) in the level3 driver (driver/level3/level3.c or driver/level3/level3_thread.c depending on whether multithreading or not)  that must have taken care of it.
(Can't hurt to have it correct in the kernel too, in case the driver code changes for whatever reason)

ChipKerchner:
It is possible that the failures with BGEMM are due to bfloat16 rounding of C.  I mentioned this priorly since I think this is incorrect.  @Mousius 

float res0;
FLOAT *C0;
C0[0] = TO_OUTPUT(TO_F32(C0[0])+res0);

For BGEMM

#define TO_F32(x) (bfloat16tof32(x))
#define TO_OUTPUT(x) (f32tobfloat16(x))

This causes double (2X) rounding.

I'm wondering if we are getting rounding differences between the way the hardware is converting _Float16 or _bf16 and the software is.  We should probably allow the tests to use native conversions if it has them (like RISC-V).  I've seen rounding unit test failures in BF16 in Eigen because of this.

martin-frbg:
Sorry, truncation has nothing to do with it - the flagged mismatches have matrix elements below 10 where they should be around 110. This looks more like your new SGEMM kernel replacing the original value of C(i,j) in _some_ places instead of adding to it (given that the C matrix is initialized to all 100s )

example:
Mismatch at i=7, j=2, k=8: BGEMM bfloat CC=110.000000, SGEMM truncatedfloat C=10.187500 original float C=10.174612
where with the "old" SGEMM kernel:
at i=7, j=2, k=8: BGEMM bfloat CC=110.000000, SGEMM truncatedfloat C=110.000000 original float C=110.174614
NOTE all wrong matrix elements occur at K=8

ChipKerchner:
It seems like you are correct.  I found a typo in my program.  Not sure why my testing didn't show the problem.

martin-frbg:
Great, thanks. I've been trying to find the problem, but being at best half-literate in riscv didn't exactly help. I guess if this passes, you could still un-draft it for 0.3.32 ?

ChipKerchner:
It seems my testing wasn't handling NaNs correctly.

ChipKerchner:
I'm OK with waiting on this one till the next release (0.3.33) - since it only affects FP32 GEMM VLEN = 256.

ChipKerchner:
I did find a typo in the BGEMM testing code.  Since we are only testing square matrices, it doesn't cause any problems but technically it should be fixed.

`    float *C = (float *)malloc_safe(m * n * sizeof(FLOAT));
    bfloat16 *AA = (bfloat16 *)malloc_safe(m * k * sizeof(bfloat16));
    bfloat16 *BB = (bfloat16 *)malloc_safe(k * n * sizeof(bfloat16));
    bfloat16 *CC = (bfloat16 *)malloc_safe(k * n * sizeof(bfloat16));  <- This should be m times n`

ChipKerchner:
Latest graph for small N values.

image

ChipKerchner:
Note:  Even though the graphs show only values of M <= 16 or N <= 8, these changes will have some improvement for all (M % 16) != 0 or (N % 8) != 0 - AKA not a multiple of the block size.

ChipKerchner:
Improved FP64 edges - up to more than 3X faster.

PR #5677, opened 2026-03-16
Optimize ZROT_RVV for the non-unit-stride case

CheryDan:
Optimize the RVV implementation of ZROT when inc_x and inc_y are non-unit strides (inc_x != 1, inc_y != 1).

Reorder several operations to reduce vector register pressure and avoid unnecessary vector register spill to the stack. This helps GCC keep vector values in registers and reduces redundant spill/reload instructions, improving runtime performance.

No functional change.


martin-frbg:
Thank you

PR #5684, opened 2026-03-17
Fix utest issues flagged by Address Sanitizer

martin-frbg:


utest/test_fork.c
```diff
@@ -117,14 +118,17 @@ exit(0);
         if (fork_pid_nested == -1) {
             perror("fork");
             CTEST_ERR("Failed to fork nested process.");
+	    free(a);free(b);free(c);free(d);
```
ChipKerchner, line 121: Bad indentation

Issue #4023, opened 2023-04-27
Improving performance with WebAssembly


rth:
Thanks to the discussions and fixes in https://github.com/xianyi/OpenBLAS/issues/3640 and follow up work by @lesteve and @ogrisel  we now have a build of OpenBLAS with emscripten for WebAssembly in Pyodide. It works quite well when used via scipy.

I recently run some benchmarks for square matrix multiplications (DGEMM) to get some ideas about the performance, which can be found [here](https://github.com/pyodide/pyodide/issues/3763#issuecomment-1526577657). The good news is that the scipy build with OpenBLAS is around 2-3x times faster for DGEMM than with the reference BLAS. The less good news is that it's still around 10x slower than almost the same OpenBLAS version built for a modern x86-64 CPU (single-threaded) natively.

For now, the constraint of that runtime is single-threaded, and without SIMD. (Though we should investigate whether it would be possible optionally built with SIMD and have some browser feature detection.)

I was wondering if is there anything else we could try to improve the performance of OpenBLAS for the WebAssembly platform ? 

It's currently built with Emscripten using the following options,
```bash
make libs shared CC=emcc HOSTCC=gcc TARGET=RISCV64_GENERIC NOFORTRAN=1 NO_LAPACKE=1 \
        USE_THREAD=0 -O2
```

Thank you!

rth:
And a follow-up question, if by chance, there is anything OpenBLAS specific that we could be doing to reduce the size of the produced shared library? Currently, it's 1.75 MB compressed which is non-negligible when loading on a web page.

martin-frbg:
The current level of emscripten support from #3640 was expected to be the bare minimum to get it working at all, if this already provides a 2x speediup over the reference implementation I consider that quite impressive.

As far as I know, there is work underway to have emscriptem understand at least SSE and AVX intrinsics, though not the corresponding assembly instructions themselves, as used in the current NEHALEM and SANDYBRIDGE kernels.

About the only part of BLAS where intrinsics are currently used is in the DOT kernels, so at least GEMM and probably AXPY would need to be (re)written to achieve some speedup. (Not sure if your 10x figure comes from a native build for such a fairly outdated cpu - definitely do not get your hopes too high by comparing to current AVX2/AVX512 cpus).

As for library size, there is nothing obviously unnecessary compiled in by default (like profiling code or a testsuite),
so the most likely means of a size reduction would be by dropping support for individual precisions - if, for instance,
[...]

PR #2867, opened 2020-09-29
Optimize the performance of dot by using universal intrinsics in X86/ARM

Qiyu8:
## Introduction

Compare with develop branch, when operating float dot, the new implementation has about 3x faster in AVX2, about 10%~30% faster in SSE2/SSE3/NEON. 

## Benchmark

  Each test has run for 10 loops and then take the average result, The radio is the MFlops division of Baseline and optimized branch(*The bigger* the radio is, the *better* performance is achieved).

- X86 BaseLine(auto vectorized)

| data size(10^3) | develop(float)               |
| --------------- | ---------------------------- |
| 4000            | 772.81 MFlops   0.010352 sec |
| 8000            | 769.16 MFlops   0.020802 sec |

- X86-SSE2 enabled

  | data size(10^3) | usimd-floatdot(float)        | ratio |
  | --------------- | ---------------------------- | ----- |
  | 4000            | 819.17 MFlops   0.009766 sec | 1.05  |
  | 8000            | 815.09 MFlops   0.019630 sec | 1.05  |
  
- X86-SSE3 enabled

  | data size(10^3) | usimd-floatdot(float)        | ratio |
  | --------------- | ---------------------------- | ----- |
  | 4000            | 853.32 MFlops   0.009375 sec | 1.10  |
  | 8000            | 848.88 MFlops   0.018848 sec | 1.10  |
  
- X86-AVX2 enabled

  | data size(10^3) | usimd-floatdot(float)       | ratio |
  | --------------- | --------------------------- | ----- |
  | 4000            | 2482.31 MFlops 0.003223 sec | 3.21  |
  | 8000            | 2307.47 MFlops 0.006934 sec | 2.99  |
  
- ARM BaseLine(dot.s)

  | data size(10^3) | develop(float)                |
  | --------------- | ----------------------------- |
  | 4000            | 1009.15 MFlops   0.007927 sec |
  | 8000            | 1042.69 MFlops   0.015345 sec |
  
- ARM-Neon enabled

  | data size | usimd-floatdot(float)         | ratio |
  | --------- | ----------------------------- | ----- |
  | 4000      | 1311.65 MFlops   0.006099 sec | 1.30  |
  | 8000      | 1274.01 MFlops   0.012559 sec | 1.22  |

## System Info

|           | Arm                                                          | x86                                      |
| --------- | ------------------------------------------------------------ | ---------------------------------------- |
| Hardware  | KunPeng                                                      |                                          |
[...]

PR #5685, opened 2026-03-18
Add a WebAssembly SIMD backend for reusable intrinsics kernels

teddygood:
Follow-up to #5676 and #4023.

This PR adds a WebAssembly SIMD backend for the shared SIMD intrinsics layer. The goal is to let existing reusable kernels that already use the common intrinsics API execute through WebAssembly SIMD on `ARCH_WASM` instead of falling back to scalar code. As part of that, it adds `kernel/simd/intrin_wasm.h`, wires it up `from kernel/simd/intrin.h` for `__wasm_simd128__,` and fills in the missing reduction helpers needed for backend completeness.

I also enabled the generic vector path for `srot` on `ARCH_WASM`. In local Pyodide/Emscripten testing, contiguous `daxpy` improved by about 1.23x–1.40x over the current baseline, and contiguous `srot` improved by about 1.41x–2.13x. Stride-2 cases were approximately flat. `drot` was evaluated too, but local testing did not show a clear or consistent benefit on WASM SIMD, so I left it out of this PR.

This PR intentionally keeps the scope small. It does not add new WASM-specific kernels, and it does not try to cover every reusable kernel that could potentially use the shared intrinsics layer at once.

If this is not the direction you would prefer upstream, I would be happy to adjust it.

### DIFF
--- a/kernel/simd/intrin.h
+++ b/kernel/simd/intrin.h
@@ -56,6 +56,11 @@ extern "C" {
 #include <arm_neon.h>
 #endif
 
+/** WASM SIMD **/
+#if defined(ARCH_WASM) && defined(__wasm_simd128__)
+#include <wasm_simd128.h>
+#endif
+
 // distribute
 #if defined(HAVE_AVX512VL) || defined(HAVE_AVX512BF16)
 #include "intrin_avx512.h"
@@ -67,6 +72,10 @@ extern "C" {
 
 #ifdef HAVE_NEON
 #include "intrin_neon.h"
+#endif
+
+#if defined(ARCH_WASM) && defined(__wasm_simd128__)
+#include "intrin_wasm.h"
 #endif
 
 #ifndef V_SIMD
--- a/kernel/simd/intrin_wasm.h
+++ b/kernel/simd/intrin_wasm.h
@@ -34,6 +34,23 @@ BLAS_FINLINE v_f64 v_mulsub_f64(v_f64 a, v_f64 b, v_f64 c)
 { return v_sub_f64(v_mul_f64(a, b), c); }
 
 /***************************
+ * reduction
+ ***************************/
+BLAS_FINLINE float v_sum_f32(v_f32 a)
+{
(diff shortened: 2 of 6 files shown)


martin-frbg:
Thanks - having the universal simd header should also allow leveraging the simd-based optimizations from #2867 in the generic dot.c (SDOT/DDOT/DSDOT) - the kernel file copied out of the riscv64_generic support currently uses its own copy of the more trivial generic code from arm/dot.c for some reason.

teddygood:
> Thanks - having the universal simd header should also allow leveraging the simd-based optimizations from #2867 in the generic dot.c (SDOT/DDOT/DSDOT) - the kernel file copied out of the riscv64_generic support currently uses its own copy of the more trivial generic code from arm/dot.c for some reason.

Thanks, that is very helpful. Would you prefer to include that in this PR as well, or keep it as a follow-up once this backend is in place?

martin-frbg:
As you prefer - come to think of it, it should also be possible to trivially copy the "ifndef DOUBLE" part of the SIMD fallback kernel from x86_64 daxpy.c to saxpy.c (which currently has only a trivial C loop for when no assembly microkernel is available). But I guess that is all existing SIMD kernels then. :)

teddygood:
> Thanks - having the universal simd header should also allow leveraging the simd-based optimizations from #2867 in the generic dot.c (SDOT/DDOT/DSDOT) - the kernel file copied out of the riscv64_generic support currently uses its own copy of the more trivial generic code from arm/dot.c for some reason.

Thanks, that is very helpful. In that case I’ll keep this PR to the intrinsics backend plus SAXPY, and leave the dot-related changes for a follow-up PR.

teddygood:
I went ahead and added SAXPY in this PR by switching SAXPYKERNEL to x86_64/saxpy.c and adding the same kind of SIMD fallback used in daxpy.c. In local direct WASM benchmarking, contiguous saxpy improved by about 2.14x, 1.46x, and 1.10x for the sizes I tested, while a stride-2 case was roughly flat at about 1.04x.

PR #5687, opened 2026-03-18
Improve tests for SHGEMM and BGEMM

### LINKED ISSUE #5686: Fix SBGEMM and SHGEMM test tolerances so that unit tests pass
SBGEMM and SHGEMM should have an acceptable tolerances so that unit tests pass.

SBGEMM has 7 bits of fraction and SHGEMM has 10.

Unit tests are failing (I know RVV is) because they are off by a couple of epsilon differences.

Please change so that we don't see these failures for each patch updated.

martin-frbg:
fixes #5686

### DIFF
--- a/test/compare_sgemm_shgemm.c
+++ b/test/compare_sgemm_shgemm.c
@@ -122,12 +122,16 @@ main (int argc, char *argv[])
                 (float)AA[k * j + l] * (float)BB[i + l * n];
             }
           if (!is_close(CC[i * m + j], C[i * m + j], 0.01, 0.001)) {
-		  fprintf(stderr,"CC %f C %f \n",(float)CC[i*m+j],C[i*m+j]);
-            ret++;
+#ifdef DEBUG
+		  	fprintf(stderr,"CC %f C %f \n",(float)CC[i*m+j],C[i*m+j]);
+#endif
+			ret++;
           }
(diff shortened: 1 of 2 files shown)


test/compare_sgemm_shgemm.c
```diff
@@ -122,12 +122,16 @@ main (int argc, char *argv[])
                 (float)AA[k * j + l] * (float)BB[i + l * n];
             }
           if (!is_close(CC[i * m + j], C[i * m + j], 0.01, 0.001)) {
```
ChipKerchner, line 124: Is this really an absolute difference (AKA 0.01)?  I would think the difference would be related to the maximum value in the output?

martin-frbg: 0.01 plus C scaled by 0.001 like I wrote in the ticket - not sure if ARM's "Mousius" came up with this, or if he just copied it from an existing test

ChipKerchner:
The same type change should probably be made to test SBGEMM

test/compare_sgemm_shgemm.c
ChipKerchner: Scaled should probably be at least 8 times bigger than the BF16 version since there are 3 less bits in the fraction.  So I think 0.01 for the scale value might work.  Same for BGEMM

PR #5688, opened 2026-03-19
Make PREFERRED_SIZE, GEMM_DIVIDE_LIMIT and _RATE available to DYNAMIC_ARCH builds

martin-frbg:
Closes a performance gap compared to dedicated builds for cpus that have non-default settings (mainly newer x86_64, Arm Neoverse and Fujitsu A64FX) - noticed while looking into https://github.com/numpy/numpy/issues/30816


mattip:
I think this is still needed to get the parameter through, right?

martin-frbg:
Yes, but it is "only" a performance-related fix (and only relevant to Neoverse V1 - also Fugaku if/when I add the related GEMM_DIVIDE_RATE passthrough)

PR #5689, opened 2026-03-19
Use generic dot kernels for WASM128_GENERIC

teddygood:
Follow-up to #5685, #2867, and #4023.

This switches SDOTKERNEL and DDOTKERNEL for WASM128_GENERIC from the trivial riscv64 fallback to `kernel/generic/dot.c`, and adds a WASM SIMD widening path there for DSDOT.

In local direct WASM benchmarking with Emscripten/Node, contiguous inputs improved over the current baseline by about 8.96x at n=1048576, 2.00x at n=2097152, and 1.72x at n=4194304 for sdot, by about 1.33x at n=1048576, 1.47x at n=2097152, and 1.14x at n=4194304 for ddot, and by about 1.14x at n=1048576, 1.83x at n=2097152, and 1.14x at n=4194304 for dsdot.

I also tried a dedicated WASM-specific DSDOT prototype locally, but it did not show a clear overall benefit over the generic implementation, so this keeps the simpler generic path.

### DIFF
--- a/kernel/generic/dot.c
+++ b/kernel/generic/dot.c
@@ -117,6 +117,50 @@ FLOAT CNAME(BLASLONG n, FLOAT *x, BLASLONG inc_x, FLOAT *y, BLASLONG inc_y)
             i += vstep;
         }
         dot = v_sum_f32(vsum0);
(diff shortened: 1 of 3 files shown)


martin-frbg:
Great, thanks

PR #5712, opened 2026-03-26
Fix incorrect cast from BF16 to FP32 in SBGEMM

murste01:
This change fixes a regression in SBGEMM where C is assumed to be BF16, and so unconditionally attempts to cast elements of C from BF16 to FP32 resulting in incorrect outputs when beta=1.

### DIFF
--- a/kernel/generic/gemmkernel_2x2.c
+++ b/kernel/generic/gemmkernel_2x2.c
@@ -30,6 +30,12 @@
 
 #include "conversion_macros.h"
 
+#ifdef BGEMM
+#define C_TO_F32 TO_F32
+#else
+#define C_TO_F32
+#endif
+
 int CNAME(BLASLONG bm,BLASLONG bn,BLASLONG bk,FLOAT alpha,IFLOAT* ba,IFLOAT* bb,FLOAT* C,BLASLONG ldc
 #ifdef TRMMKERNEL
 		,BLASLONG offset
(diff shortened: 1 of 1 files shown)


aditew01:
@martin-frbg is there a way we can get a patch release for OpenBLAS? 
The current bug in `0.3.32` blocks OpenBLAS upgrade in PyTorch, because it causes a unit test failure.

martin-frbg:
Yes, I'll try to do 0.3.33 this weekend - though the spurious utest failure bothers me a lot less than the weird DDOT bug that my workaround for the weird Neoverse SDOT bug introduced. Coincidentally the Reference-LAPACK team is committing a flurry of fixes for their upcoming 3.13, so it looks like a good chance to import those as well.

aditew01:
that'll be much appreciated! 🙏

ChipKerchner:
I think this is what I've been pointing out for a while now.

aditew01:
> Yes, I'll try to do 0.3.33 this weekend - though the spurious utest failure bothers me a lot less than the weird DDOT bug that my workaround for the weird Neoverse SDOT bug introduced. Coincidentally the Reference-LAPACK team is committing a flurry of fixes for their upcoming 3.13, so it looks like a good chance to import those as well.

@martin-frbg gentle nudge on this. Do you think we've a release candidate for 0.3.33 or are we waiting for more patches / fixes?  
Apologies for the constant nag 🙏

martin-frbg:
I'm still importing PRs from Reference-LAPACK...

aditew01:
Hi @martin-frbg any tentative date for release for 0.3.33 ? TIA

martin-frbg:
Tentatively this weekend or early next week, unless something like the NeoverseV1 DGEMV issue spooks me again

puneetmatharu:
@aditew01 looks like v0.3.33 is out 🎉

PR #5716, opened 2026-03-27
[ARM64] Add optimized fp16 shgemm kernels for Neoverse N2

yuanjia111:
https://github.com/OpenMathLib/OpenBLAS/issues/5681


martin-frbg:
Great, thank you very much

PR #5738, opened 2026-04-02
Add offline documentation building instructions

SuperCowProducts:
Added instructions for building documentation for offline use, including version references for Python, mkdocs, and uv.


README.md
```diff
@@ -24,6 +24,30 @@ For a general introduction to the BLAS routines, please refer to the extensive d
...
+
+```
+uv venv
+source .venv/bin/activate
+uv pip install mkdocs
+uv pip install $(mkdocs-get-deps)
+mkdocs serve
+```
+
+For reference / reproducing in the future, here is a summary of the versions used:
```
martin-frbg, line 39: Thanks - I don't think this part needs to be in the README itself ?

SuperCowProducts: It may has a better place - just wanted to somehow give extra guidance for future users!

PR #5751, opened 2026-04-08
Remove redundant C implemetations from MIPS directories

chenx97:
... and factor out definitions found in both if and else branches.


chenx97:
The key difference between arm/dot.c and mips/dot.c appears to be

```patch
@@ -50,8 +41,11 @@ FLOAT CNAME(BLASLONG n, FLOAT *x, BLASLONG inc_x, FLOAT *y, BLASLONG inc_y)
 
	while(i < n)
	{
-
-		dot += y[iy] * x[ix] ;
+#if defined(DSDOT)
+		dot += (double)y[iy] * (double)x[ix] ;
+#else
+		dot += y[iy] * x[ix];
+#endif
		ix  += inc_x ;
		iy  += inc_y ;
		i++ ;
```

Let me see if this change can be applied to arm/dot.c or not.

martin-frbg:
> Let me see if this change can be applied to arm/dot.c or not.

Yes, there should be no harm from this.

These duplications tended to happen when someone cloned an existing near-generic setup for a new architecture, I think riscv64 still carries a few duplicated "arm" files as well. But apart from carrying a couple of unnecessary files - all small - there was no harm.
 (I guess eventually it may make sense to gather all the pure C kernels that popped up under kernel/arm (or similar) and move them to kernel/generic where they "properly" belong. But it would be a purely cosmetic change with a nonzero chance to break something...)

PR #5753, opened 2026-04-09
Fix DROUNDUP_LWORK: patch was not fully copied

### LINKED ISSUE #5754: Broken memory computation in DSYGVD/DSPGVD routines
## Observed behaviour

With ndim = 17
```
dsygvd(1,'v','u',ndim,h,ndim,p,ndim,e,aux,-1,iwork,liwork,info)
```
returns aux(1) equals zero which is later converts to integer and used as memory requirement. However, since lwork becomes to zero, later calls of `dsygvd` are failing. 

See defs and usage here: https://github.com/grimme-lab/xtb/blob/d1314aa59f84e3668ec6620ee66ca26362529d4d/src/scc_core.f90#L930-L937

## Expected behaviour

`aux(1)` should be 681.

## Solution

See #5753 for fix.

foxtran:
Asm before fix: https://godbolt.org/z/dzoTvaoTq

The most interesting lines are 153-154:
```
         bl      droundup_lwork_
         fcvt    d0, s0
```
Here, `droundup_lwork_` sets d0 register based on integer value in x0 register. Then, `fcvt` sets `d0` fp64 register with value from `s0` fp32 register. However, s0 was set to zero (and, actually, never used), so, after this, d0 is 0.0 that breaks computing of memory requirement. 

Asm after fix: https://godbolt.org/z/MW4rvoETa

Only these two files are affected.

Bug was introduced by commit https://github.com/OpenMathLib/OpenBLAS/commit/99c120916a25b4190ed22fca5cc7928d29fb86af (PR #5702)

Fixes #5754

### DIFF
--- a/lapack-netlib/SRC/dspgvd.f
+++ b/lapack-netlib/SRC/dspgvd.f
@@ -225,7 +225,8 @@
 *     ..
 *     .. External Functions ..
       LOGICAL            LSAME
-      EXTERNAL           LSAME
+      DOUBLE PRECISION   DROUNDUP_LWORK
+      EXTERNAL           LSAME, DROUNDUP_LWORK
 *     ..
 *     .. External Subroutines ..
       EXTERNAL           DPPTRF, DSPEVD, DSPGST, DTPMV, DTPSV, XERBLA
--- a/lapack-netlib/SRC/dsygvd.f
+++ b/lapack-netlib/SRC/dsygvd.f
@@ -245,7 +245,8 @@
 *     ..
 *     .. External Functions ..
       LOGICAL            LSAME
-      EXTERNAL           LSAME
+      DOUBLE PRECISION   DROUNDUP_LWORK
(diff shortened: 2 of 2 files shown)


martin-frbg:
Oops, not sure what happened to #5702 - there appear to be similar differences in the related complex routines. (One of those silly situations where I abused the web interface for PR creating instead of pushing from a local repo due to travelling)
Thanks a lot for spotting and fixing this, I'll follow up with the remaining files in a minute.

foxtran:
Remaining files are OK.

martin-frbg:
I'm seeing differences in chegvd.f (roundup not used on RWORK(1) in line 322 between a fresh checkout of both OpenBLAS and Reference-LAPACK, and also for the other c & z files from 1203/5702

martin-frbg:
Absolutely no idea right now how #5702 ended up with a _partial_ commit of the fixes from 1203, as I had copied the entire files from an up-to-date checkout of Reference , no manual edits...

PR #5759, opened 2026-04-16
Add CPU ID for Neoverse V3

nSircombe:
The PR which added NeoverseV3 detection to openBLAS includes the V3AE
variant, but not V3. This patch includes the missing cpuid and labels
the V3 variants.

This ensures that Neoverse V3 will use existing V2 kernels rather than
falling back to ARMV8SVE.


martin-frbg:
Thank you - I had meant to check if the obvious gap in the sequence was the V3, but seems I forgot.

nSircombe:
No problem, happy to help. Thanks for getting it merged in so quickly, and adding the [build time detection](https://github.com/OpenMathLib/OpenBLAS/pull/5761/commits) - I missed that,

PR #5769, opened 2026-04-22
Fix CMake builds on Apple that use the Ninja generator

### LINKED ISSUE #5768: Build failure on macOS with CMake + Ninja
When building OpenBLAS on macOS with `Ninja` as the generator the build will fail due to the response file workaround in `CMakeLists.txt`. This is because Ninja doesn't generate the same `.rsp` response files that a `Makefile` generated by CMake would. This then has two consequences:

1. A warning is generated during build just before the `libopenblas.a` is generated:
```
cat: .../build/Release/CMakeFiles/openblas_static.dir/objects*.rsp: No such file or directory
ar: creating archive libopenblas.a
```

2. The `libopenblas.a` (and also the `libopenblas.0.3.dylib`) don't contain any code except for `xerbla_`.

3. Then the build fails because the tests that are built can't find the symbols of the actual functions of OpenBLAS:
```
Undefined symbols for architecture arm64:
  "_caxpby_", referenced from:
      ___ctest_axpby_caxpby_inc_0_run in test_axpby.c.o
      ___ctest_axpby_caxpby_inc_1_run in test_axpby.c.o
      ___ctest_axpby_caxpby_inc_2_run in test_axpby.c.o
  "_caxpy_", referenced from:
      ___ctest_axpy_caxpy_inc_0_run in test_axpy.c.o
      ___ctest_axpy_caxpy_incx_0_run in test_axpy.c.o
  "_cgemv_", referenced from:
      ___ctest_cgemv_0_nan_inf_run in test_gemv.c.o
      ___ctest_cgemv_0_nan_inf_incy_2_run in test_gemv.c.o
      ___ctest_cgemv_0_2_nan_1_inf_1_run in test_gemv.c.o
      ___ctest_cgemv_0_2_nan_1_inf_1_incy_2_run in test_gemv.c.o
      ___ctest_cgemv_2_0_nan_1_inf_1_run in test_gemv.c.o
      ___ctest_cgemv_2_0_nan_1_inf_1_incy_2_run in test_gemv.c.o
[...]
```

With Ninja the workaround that is present in the `CMakeLists.txt` is not actually required, because Ninja always uses response files anyway (but uses temporary files it removes again, so there will never be a `.rsp` file in a build folder).

I therefore suggest to replace:

```
if(APPLE)
```
in `CMakeLists.txt` (around line 310) with
```
if(APPLE AND "${CMAKE_GENERATOR}" MATCHES ".*Makefiles")
```

That will still keep the workaround in-place when `make` is used, but allows Ninja to just work out of the box.

martin-frbg:
fixes #5768 as suggested there

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -309,11 +309,9 @@ if (USE_OPENMP)
 endif()
 
 # Fix "Argument list too long" for macOS - mostly seen with older OS versions on POWERPC or Intel CPUs 
-if(APPLE)
+if(APPLE AND "${CMAKE_GENERATOR}" MATCHES ".*Makefiles")
   # Use response files to get around the ARG_MAX limit, unless using the Ninja generator
-  if("${CMAKE_GENERATOR}" MATCHES ".*Makefiles")
-    set(CMAKE_C_USE_RESPONSE_FILE_FOR_OBJECTS 1)
-  endif()
+  set(CMAKE_C_USE_RESPONSE_FILE_FOR_OBJECTS 1)
   # Always build static library first
   if(BUILD_STATIC_LIBS)
     set(STATIC_PATH "${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/lib${OpenBLAS_LIBNAME}.a")
@@ -334,9 +332,7 @@ if(APPLE)
     set(OMP_LIB "")
   endif()
   if(NOT NOFORTRAN)
-    if("${CMAKE_GENERATOR}" MATCHES ".*Makefiles")
-      set(CMAKE_Fortran_USE_RESPONSE_FILE_FOR_OBJECTS 1)
-    endif()
+    set(CMAKE_Fortran_USE_RESPONSE_FILE_FOR_OBJECTS 1)
 	set(CMAKE_Fortran_CREATE_STATIC_LIBRARY ${CREATE_STATIC_LIBRARY_COMMAND})
     if(BUILD_SHARED_LIBS)
       set(CMAKE_Fortran_CREATE_SHARED_LIBRARY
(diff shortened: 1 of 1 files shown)


chris-se:
Note that this doesn't actually fix the issue, because the issue is not the `CMAKE_C_USE_RESPONSE_FILE_FOR_OBJECTS`, which is ignored with `cmake -G Ninja`, but the fact that the lines

```
set(CREATE_STATIC_LIBRARY_COMMAND
    "sh -c 'cat ${CMAKE_BINARY_DIR}/CMakeFiles/${OpenBLAS_LIBNAME}_static.dir/objects*.rsp | xargs -n 1024 ${CMAKE_AR} -ru ${STATIC_PATH} && exit 0' "
    "sh -c '${CMAKE_AR} -rs ${STATIC_PATH} ${CMAKE_BINARY_DIR}/driver/others/CMakeFiles/driver_others.dir/xerbla.c.o && exit 0' ")
```

won't work, because no `objects*.rsp` exist with CMake+Ninja -- they are only written out with CMake + UNIX Makefiles or similar.

But the thing is: because Ninja can handle long command lines directly, the static linking commands don't need to be changed regardless here, so the whole workaround is completely unnecessary. That's why I suggested in #5768 to simply replace the entire ``if (APPLE)`` section.

I just tested your variant of the patch and it fails the same as if that change wasn't there.

martin-frbg:
Ah, I see. Didn't want to exclude the whole section as there appear to be unrelated settings in it, but didn't take a closer look

chris-se:
As far as I can tell the entire section is just for the workaround, and it's just that complicated because basically the whole linking logic inside CMake for up to two libraries (the static one and potentially the shared one) has to be written by hand to make the workaround work properly.

I can't test this on **[edit]** other **[/edit]** Macs, but at least on the M1 I have access to simply replacing the entire if (as I suggested) and having macOS use the same library linking logic as the other platforms "just works" for me.

chris-se:
(But irrespective of this, many thanks for the very quick response to this. :smile:)

Issue #5775, opened 2026-04-23
macOS with CMake + Ninja unfortunately still doesn't work 100%


chris-se:
Yesterday I reported #5768, indicating that building OpenBLAS on macOS with Ninja worked for me in the "out of the box" configuration, so the workaround would only be needed for the Makefile variant.

A colleague of mine unfortunately ran into a problem (with slightly different toolchain, CMake and Ninja versions than I had) and was able to trigger the "argument list too long" issue even with Ninja. It appears to be the case that CMake+Ninja doesn't default to using response files in this case, and because I had an older version of the XCode installed, where OpenBLAS enables less variants, and didn't run into that with Ninja.

But since this is now the second time of me running into this, I decided to investigate this further. I found the following:

 - There is an option to force CMake+Ninja to use response files. This works in this case for shared libraries. This must be enabled **before** the `add_library()` calls are made though, otherwise it will have no effect.
 - For static libraries CMake+Ninja will try to use a response file, but `ar` does not accept them, and then the builds fails.
 - Whereas for CMake+Unix Makefiles the response files will work for shared libraries (if they are enabled), so we don't need the other workaround if **only** the shared library is built.
 - The workaround doesn't work for Ninja as-is though because the generated response file is different.

I've now created a patch (I'll open a pull request after this ticket) that does the following:

 - Before the `add_library()` calls it sets the appropriate options to ensure that response files are used on macOS (regardless of the generator).
 - When building static libraries (regardless of whether shared libraries are also built) the previous workaround is used for both Unix Makefiles and Ninja, but with a different response file list in each case.

After doing some test builds in the configurations make/static, make/shared, make/both, ninja/static, ninja/shared, and ninja/both I've verified that the CMake build system now properly generates the link commands in each case with the response file:

## make, static
```
[ 98%] Linking C static library lib/libopenblas.a
/opt/homebrew/bin/cmake -P CMakeFiles/openblas_static.dir/cmake_clean_target.cmake
/opt/homebrew/bin/cmake -E cmake_link_script CMakeFiles/openblas_static.dir/link.txt --verbose=1
ar: creating archive [...]/OpenBLAS-0.3.32-test/build-make/static/lib/libopenblas.a
[...]

PR #5776, opened 2026-04-23
CMake: properly fix build on macOS with Ninja (cf. to response files)

chris-se:
On macOS several build-related programs (`cc`, `ar`) will not accept the linker line if too many files have been added to it in the DYNAMIC_ARCH case. (This now occurs on all platforms where macOS runs / used to run, as there are enough ARM variants that are build with DYNAMIC_ARCH that the number of files exceeds the intrinsic limit on macOS.)

The workaround for this is to use response files that contain the file list themselves, and pass these to the build system. For dynamic libraries this works fine, but for static libraries it doesn't because `ar` doesn't accept response files on macOS. For this reason there was previously a workaround in CMakeLists.txt to handle this, but the workaround didn't take into consideration that when building only a shared library it is not actually necessary, and when building static libraries the response file CMake generates for Ninja is at a different location than the response files CMake generates for Unix Makefiles.

This commit cleans this all up by setting the necessary options for the response files to be properly generated, and adjusts the additional workaround to only be applicable in the case a static library is built. The workaround is also adjusted to handle the Ninja case.

Fixes GitHub issue #5775

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -243,6 +243,30 @@ if (${DYNAMIC_ARCH})
   endforeach()
 endif ()
 
+# Work around Apple issue with "argument list too long" when linking.
+# Seen for DYNAMIC_ARCH when there are a _LOT_ of variants build, previously
+# mostly with older versions on POWERPC or Intel CPUs, but with newer ARM
+# CPUs and newer compilers this will also occur on ARM.
+#
+# This must happen before the add_library() calls, otherwise this will have
+# no effect.
+#
+# Unfortunately, while this workaround will work for shared libraries (with
+# either Ninja oder Unix Makefiles), but not for static libraries, as ar
+# does not accept response files. So for the case where we (also) build
+# static libraries we need an additional workaround, see below. (The
+# generation of the response files by CMake must still be enabled, so this
+# is required regardless.)
+if(APPLE)
+  set(CMAKE_C_USE_RESPONSE_FILE_FOR_OBJECTS ON)
+  set(CMAKE_C_USE_RESPONSE_FILE_FOR_LIBRARIES ON)
+  if(NOT NOFORTRAN)
+    set(CMAKE_Fortran_USE_RESPONSE_FILE_FOR_OBJECTS ON)
+    set(CMAKE_Fortran_USE_RESPONSE_FILE_FOR_LIBRARIES ON)
+  endif()
+  set(CMAKE_NINJA_FORCE_RESPONSE_FILE ON)
+endif()
+
 # add objects to the openblas lib
 if(NOT NO_LAPACK)
   add_library(LAPACK_OVERRIDES OBJECT ${LA_SOURCES})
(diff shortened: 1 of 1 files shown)


martin-frbg:
Thanks. Perhaps it would (also) make sense to drop a few targets when building for macOS - neither Fugaku nor any of the ThunderX chips seem particularly likely to be present in an iAnything, and even the Neoverse lineup feels questionable in that context.

chris-se:
That would probably also make sense, if you wish I can provide another pull request for this. I suggest doing this separately though. Unfortunately I don't 100% know the nomenclature of OpenBLAS (i.e. how the names OpenBLAS uses translate to different chips). With the most current XCode OpenBLAS wants to generate the following variants:

- A64FX
- ARMV8
- ARMV8SVE
- ARMV9SME
- CORTEXA53
- CORTEXA57
- EMAG8180
- NEOVERSEN1
- NEOVERSEN2
- NEOVERSEV1
- THUNDERX
- THUNDERX2T99
- THUNDERX3T110
- TSV110
- VORTEXM4

If you tell me which ones of these are relevant to macOS/arm64 (which will only run on Apple's own silicon, i.e. A18 Pro, M1, M2, M3, M4, M5, ...) I'd be happy to provide another pull request that restricts this on macOS platforms.

martin-frbg:
I'm not that familiar with Apple's product range, but I suspect the only actually relevant targets would be CortexA57 (and perhaps A53 too ?) for very old models of iPhone (might make sense to look up what SoC the first generation had) - or just the baseline ARMV8 for them. Next perhaps NeoverseN1 as a stand-in for the Vortex target (M1 to M3), and VortexM4 for the current top of the line with SME but no SVE, perhaps ARMV9SME for any future chip that might support non-streaming SVE.

martin-frbg:
Can probably skip the A53/A57 combo as too little seems to be known about the internals of A8..A10, and autodetection would only resolve to ARMV8.
For A11 and beyond, it might actually make sense to use NEOVERSEN1 (just based on instruction set) _if_ anybody has the hardware - again these currently would be autodetected as ARMV8, as their cpu part ids are not known to the autodetection code in dynamic_arm64.c 

So in effect, the DYNAMIC_ARCH list would shrink to ARMV8 NEOVERSEN1 VORTEXM4 ARMV9SME - of course this is
already doable by supplying your own DYNAMIC_LIST. Similarly, the defaults for x86_64 include a host of AMD chips and cpus that came out after Apple's switch to the ARM architecture - both at least officially supported by Apple

chris-se:
I wouldn't necessarily touch the x86_64 list because some people may be running Hackintoshes on these more modern CPUs. But I haven't seen the arm64 version of macOS running on anything other than Apple silicon, so I'll restrict it there to the ones you mentioned here - many thanks for the list.

I'll create another pull request either today or on Monday, after I've tested this.

martin-frbg:
This caused a build failure on old powermacs now

PR #5789, opened 2026-04-30
Fix compilation with DYNAMIC_TARGET=1 and BUILD_BFLOAT16

luhenry:
coremsg isn't defined outside the BUILD_BFLOAT16 and BUILD_HFLOAT16 blocks, leading to the use of an undefined variable

### DIFF
--- a/driver/others/dynamic_riscv64.c
+++ b/driver/others/dynamic_riscv64.c
@@ -152,6 +152,7 @@ char* gotoblas_corename(void) {
 static gotoblas_t* get_coretype(void) {
 	uint64_t vector_mask;
 	unsigned vlenb = 0;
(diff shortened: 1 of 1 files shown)


martin-frbg:
Oops, that looks like a rather embarassing oversight.
Thanks for the fix

luhenry:
@martin-frbg would you have an idea when the next 0.3.34 release is coming? I would love to bump to the new version to fix it in some other projects like [PyTorch](https://github.com/pytorch/pytorch/pull/182278/changes#diff-6a0f6a8ae9d9eb07ff3e6f3440dd32a94de840028ae5131e9da86b17ac85492c). Thank you!

PR #5791, opened 2026-05-01
Fix Apple platform detection and iOS cross-build

moluopro:
This PR fixes Apple platform detection and iOS cross-build instructions.

The main issue is that Apple clang defines `TARGET_OS_IPHONE` on macOS as well,
but with value `0`. The previous detection only checked whether the macro was
defined, so macOS builds could be identified as iOS. This PR checks the macro
value, adds a separate display name for Darwin targets, and updates the iOS CI
and documentation build commands so that iPhoneOS compiler flags are actually
passed to `make`.

### `ctest.c`

- Change iOS detection from `defined(TARGET_OS_IPHONE)` to
  `defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE`.
- This avoids treating normal macOS builds as iOS, because Apple headers define
  `TARGET_OS_IPHONE` as `0` for macOS targets.

### `c_check`

- Replace the boolean `is_ios` state with `darwin_os`, which records the Darwin
  target kind as `macOS` or `iOS`.
- Keep the internal build OS as `Darwin`, while writing `OSNAME_DISPLAY` for
  user-facing output.
- Set `CROSS=1` when the detected Darwin target kind is `iOS`.
- Use `grep -q` when checking the generated assembly symbol prefix, avoiding
  accidental diagnostic output from the detection command.

### `c_check.pl`

- Mirror the shell `c_check` behavior in the Perl fallback script.
- Track `darwin_os` separately from `OSNAME`.
- Emit `OSNAME_DISPLAY` only for Darwin targets.
- Set cross-compilation mode for iOS targets.

### `Makefile`

- Add `OSNAME_DISPLAY ?= $(OSNAME)`.
- Print `OSNAME_DISPLAY` in the final build summary instead of the internal
  `OSNAME`.
- This keeps existing build logic based on `Darwin`, while showing users
  `macOS` or `iOS` in the summary.

### `.github/workflows/apple_m.yml`

- Update the iOS cross-build step to query the iPhoneOS clang and SDK path with
  `xcrun`.
- Pass `CC` and `CFLAGS` directly to `make`, instead of setting shell variables
  that `make` would not necessarily receive.
- Print the resolved `CC` and `SDKROOT` values in CI logs for easier debugging.
- Restore `DYNAMIC_ARCH=1` in the iOS CI build command.

### `docs/install.md`

- Update the iOS build instructions to match the CI command.
- Use `xcrun --sdk iphoneos --find clang` instead of a hard-coded Xcode path.
- Pass `CC` and `CFLAGS` directly to `make`.
- Fix the follow-up note to refer to the actual
  `-miphoneos-version-min` flag instead of `MIN_IOS_VERSION`.

### DIFF
--- a/.github/workflows/apple_m.yml
+++ b/.github/workflows/apple_m.yml
@@ -244,15 +244,12 @@ jobs:
 
       - name: Crossbuild to iOS
         run: |
-          #brew install llvm
-          #export #PATH=/opt/homebrew/opt/llvm/bin:$PATH
-          # export #LDFLAGS="-L/opt/homebrew/opt/llvm/lib"
-          export #CPPFLAGS="-I/opt/homebrew/opt/llvm/include" 
-          export CC=/Applications/Xcode_26.0.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang
-          export CFLAGS="-O2 -unwindlib=none -Wno-macro-redefined -isysroot /Applications/Xcode_26.0.1.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS26.0.sdk -arch arm64 -miphoneos-version-min=10.0"
-          xcrun --sdk iphoneos --show-sdk-path
-          ls -l /Applications
-          make TARGET=ARMV8 NUM_THREADS=32 HOSTCC=clang NOFORTRAN=1 CROSS=1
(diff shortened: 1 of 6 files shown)


moluopro:
I previously fixed the iOS shell script, but didn’t update Perl, CI, or the documentation. I’ve polished those up this time.

martin-frbg:
Thank you very much

PR #5800, opened 2026-05-05
provide a NEON version of arm/sgemm

notaz:
Surprisingly OpenBLAS lacks NEON optimized kernels for armv7, even though it auto-enables NEON during build by default (passes `-mfpu=neon` to the compiler, meaning the compiler will use NEON instructions wherever it can).

The speedup on Cortex-A76 is significant, before:
```
 M= 200, N= 200, K= 200 :     9262.97 MFlops   0.001727 sec
```
after:
```
 M= 200, N= 200, K= 200 :    30223.64 MFlops   0.000529 sec
```


martin-frbg:
Thanks - ARMV7 hasn't received much work in recent years, and I believe in the early days of CORTEXA53/A57 it was assumed/claimed that NEON wasn't significantly faster in V7 code. (Telling the compiler it's allowed to use NEON was required for some code contribution that used universal intrinsics, I think, and it was obviously trivial to do...)

PR #5801, opened 2026-05-06
Makefile.system: when Clang is used, check if gfortran is used

pkubaj:
At least on power, GCCVERSIONGT4 is checked for when gfortran is used, like at Makefile.power:59


martin-frbg:
I wonder if we could simply drop this check - it was somewhat useful back in the day when POWER9 was new and some distributions still defaulted to gcc4 ? (At least now it does no harm except generate a spurious message)

pkubaj:
A spurious message and it downgrades to -mcpu=power8.

PR #5802, opened 2026-05-06
Makefile.power: allow using power10 instructions

pkubaj:
OpenBLAS uses various power9 and power10 instructions and they fail to build with errors like in https://github.com/OpenMathLib/OpenBLAS/issues/4793. Just allow the latest ISA, it will generate the same code, but actually allow the binaries to be assembled.

### DIFF
--- a/Makefile.power
+++ b/Makefile.power
@@ -111,7 +111,7 @@ endif
 endif
 
 ifeq ($(C_COMPILER), CLANG)
-CCOMMON_OPT += -fno-integrated-as
+CCOMMON_OPT += -fno-integrated-as -Wa,-mpower10
 endif
 # workaround for C->FORTRAN ABI violation in LAPACKE
 ifeq ($(F_COMPILER), GFORTRAN)


martin-frbg:
Any minimum LLVM version requirement for this ? ISTR the original problem was specific to BSD

pkubaj:
I'm not sure, it certainly works with LLVM 19 that FreeBSD currently uses, but I think it's more about minimum binutils version, since -Wa is being passed to as.

martin-frbg:
Seems we need at least binutils 2.35 for this option, and I guess it is only needed on POWER10 and newer, with only the equivalent power9 option _possibly_ needed for POWER9.

martin-frbg:
@pkubaj can you check if the `-Wa,power10` or `power9` option is actually required on a POWER9 running BSD ? It doesn't appear to be needed with Linux, and the one BSD system I could access through the GCC Compile Farm is down for the foreseeable future

pkubaj:
Yes, they are required. Without that, I was getting https://github.com/OpenMathLib/OpenBLAS/issues/4793

martin-frbg:
Does POWER9 require `-Wa,power10` though ? (Different binutils version requirements for either) ? I'm inclined to make this entire addition conditional on OS being `*BSD`, but I'm not sure if a POWER9 user would necessarily keep their software current enough to support _future_ hardware. (And POWER8 or earlier probably don't need either option)

pkubaj:
Obviously not, POWER9 requires just `-Wa,power9`. But that flag doesn't cause the compiler to emit POWER10 instructions, it just makes the linker accept those and if you don't set -mcpu=power10, they won't be emitted. Since POWER10's ISA is a superset of POWER9's ISA, it's fine to just accept everything (and POWER11 doesn't introduce a new ISA version. Also, there's no reason why you can't run new software on POWER9 (or even POWER8).

Another point: on powerpc64le DYNAMIC_ARCH builds work, so you might just build with it either on POWER9 or POWER8 and hit the same issue. On FreeBSD, we set that option to be enabled on powerpc64le.

martin-frbg:
Okay, so it is basically a BSD linker issue ? My point is mainly that it has the side effect of requiring specific minimum versions of the binutils - the bunch of machines and docker images maintained by OSUOSL may not be entirely representative, but they have anything from 2.27 (Ubuntu 18LTS), which is "sufficient" for a Linux build on POWER8, onward. 
I think I'll just make this addition conditional on
`ifeq ($(OSNAME), $(filter $(OSNAME),FreeBSD OpenBSD NetBSD DragonFly))`
and hope that any `*BSD` in the wild is sufficiently equipped to match the requirements of its linker. This should also take care of the earlier conflict with AIX where the assembler doesn't like the vowels spelled out.

pkubaj:
I'm not sure if it's a linker issue. `-fno-integrated-as` just forces the use of GAS. Also, DragonflyBSD is only available for amd64. NetBSD obviously supports "everything", but that somehow doesn't cover any 64-bit POWER :) OpenBSD supports powerpc64 (only big-endian, although it actually runs only on POWER8 and newer and I'm not sure about whether OpenBLAS works there).

martin-frbg:
I'm not familiar with the various BSDs obviously.. So Dragonfly and NetBSD are out, probably OpenBSD too... thanks.
  
Curious though if we force the use of the GNU assembler, wouldn't the linker be part of the same binutils package and be supposed to understand its siblings ? (Unless it's some newfangled gold/mold/whatever setup where it's more
reasonable to put the burden on whoever set that thing up)

Another half-baked thought - as it stands right now, that option would also be applied to any truly ancient PPC970 - would it be more likely to cause problems there, or wouldn't they be using LLVM anyway ?

pkubaj:
Yes, the linker is also a part of binutils and it's currently at 2.44. I don't think it will  cause issues on PPC970 -  after all we just allow the existence of new instructions to fix the build, but on BE baseline the compiler will not emit them at all. Dropping that option will not hurt either, but I don't think its existence will matter.

martin-frbg:
Thanks - the whole thing flamed up because an ancient binutils `as` errored out on "unknown option -mpower10", so I try to be on the safe side here

PR #5803, opened 2026-05-10
Makefile.install: fix PKGFILE expression so it works with paths with spaces

Sid-MB:
Previously, running the make file with a PREFIX path which included spaces would write to the wrong path, now it will write to the correct path. 

Also note that the second setting of PKGFILE, at line 194 previously, was useless because that line was treated as a shell command which only assigned the variable for that line's invocation.


martin-frbg:
Thank you

PR #5804, opened 2026-05-11
Use -mpwr10 instead of -mpower10 to support AIX assembler

ayappanec:
#5802 broke OpenBLAS build on AIX with the below error,
```
Assembler:
line 1: 1252-162 Invalid -m flag assembly mode operand: POWER10.
	Valid values are:COM PWR PWR2 PPC 601 603 604 PPC64 620 A35 PWR4 PWR5 PWR5X 970 PPC970 PWR6 PWR6E PWR7 PWR8 PWR9 PWR10 ANY
/tmp/scal-bfa82c.s: line 14: 1252-149 Instruction stdu is not implemented in the current assembly mode PPC64.
/tmp/scal-bfa82c.s: line 15: 1252-149 Instruction std is not implemented in the current assembly mode PPC64.
```
AIX assembler recognise only "-mpwrX" variant instead of "-mpowerX". GNU binutils assembler recognises both. So let's have "-mpwr10" instead of "-mpower10".

### DIFF
--- a/Makefile.power
+++ b/Makefile.power
@@ -111,7 +111,7 @@ endif
 endif
 
 ifeq ($(C_COMPILER), CLANG)
-CCOMMON_OPT += -fno-integrated-as -Wa,-mpower10
+CCOMMON_OPT += -fno-integrated-as -Wa,-mpwr10
 endif
 # workaround for C->FORTRAN ABI violation in LAPACKE
 ifeq ($(F_COMPILER), GFORTRAN)


ayappanec:
Thanks @martin-frbg for the support.

martin-frbg:
Unfortunately this breaks LLVM builds - its integrated assembler doesn't support `-mpwr10`, so we'll need to go with a conditional after all. Sadly this was not visible in CI as the Jenkins job at OSUOSL appears to have been flaky, and its status code is shared with the zarch job that passed.

ayappanec:
> Unfortunately this breaks LLVM builds - its integrated assembler doesn't support `-mpwr10`, so we'll need to go with a conditional after all. Sadly this was not visible in CI as the Jenkins job at OSUOSL appears to have been flaky, and its status code is shared with the zarch job that passed.

Okay. 
Just a doubt though., "-fno-integrated-as" is passed before it. So why the LLVM compiler still used it's integrated assembler ?.

martin-frbg:
Good question - that shouldn't be its integrated as, but then what is it ? The job is running inside a docker container based on OSUOSL's ubuntu-ppc64le image, so the only likely alternative would seem to be the binutils assembler that doesn't complain in the corresponding gcc job.

martin-frbg:
Doesn't help that the OSUOSL images are old - the working gcc job is Ubuntu 18 with gcc-7, the failing clang job is Ubuntu 20 with clang-10 - both probably unaware of a future POWER10. 
But both are `-mtune=power8`, yet the clang one gets the `pwr10` thrown into its assembler options - doing this unconditionally in any clang context is probably the root cause, no matter if we spell it power10 or pwr10

martin-frbg:
Curiously only binutils 2.35 and newer officially support this option, but 2.30 appears to ignore it when passed by the compiler, while 2.27 errors.  Calling as-2.30 with `-mpwr10` on the command line exits with an error too. What a mess.

ayappanec:
> Doesn't help that the OSUOSL images are old - the working gcc job is Ubuntu 18 with gcc-7, the failing clang job is Ubuntu 20 with clang-10 - both probably unaware of a future POWER10. But both are `-mtune=power8`, yet the clang one gets the `pwr10` thrown into its assembler options - doing this unconditionally in any clang context is probably the root cause, no matter if we spell it power10 or pwr10

Okay. 
So it's the original commit #5802 that broke it.
Tagging @amritahs-ibm who works on Linux on Power.

martin-frbg:
Yes, and probably a BSD-specific issue at that. I've asked the original contributor to run some more tests as I cannot access any POWER machines running BSD right now. My suspicion is that the change is certainly unneccessary (and detrimental) for POWER8 and older, and probably unnecessary for POWER9 as well.

PR #5805, opened 2026-05-11
Fix incorrect leading dimension check for SME SGEMM direct kernel path

### LINKED ISSUE #5794: SME kernel not getting hit due to incorrect conditioning on leading dimensions
https://github.com/OpenMathLib/OpenBLAS/blob/8e57c8667da06cb55a777975b6c2435c9b304079/interface/gemm.c#L565

@martin-frbg, 
The correct conditioning is k==lda && n ==ldb && n==ldc.

matcraje:
For row-major matrices, the tight-packing condition should be k==lda (A is m×k), n==ldb (B is k×n), and n==ldc (C is m×n). The old check used m==lda and k==ldc, which prevented the SME/direct kernel from being invoked except when m==k==n (square matrices).

Fixes https://github.com/OpenMathLib/OpenBLAS/issues/5794

### DIFF
--- a/interface/gemm.c
+++ b/interface/gemm.c
@@ -562,12 +562,12 @@ if (strcmp(gotoblas_corename(), "armv9sme") == 0
 )
 // if (support_sme1())
 #endif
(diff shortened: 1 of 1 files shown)


martin-frbg:
Right, sorry for dropping the ball on that earlier.

PR #5807, opened 2026-05-12
riscv64: wire TRSM, complex SYMV, and complex GEMM copy RVV kernels

mengzhuo:
Wire existing RVV-optimized kernels into KERNEL.RISCV64_ZVL128B and KERNEL.RISCV64_ZVL256B that were already implemented but not referenced:

- TRSM (S/D/C/Z, all 4 directions): replace ../generic/trsm_kernel_*.c with trsm_kernel_*_rvv_v1.c
- TRSM copy: add TRSMCOPY*_M and ZTRSMCOPY*_M variables using trsm_*copy_rvv_v1.c and ztrsm_*copy_rvv_v1.c
- Complex SYMV (C/Z): replace ../generic/zsymv_k.c with zsymv_*_rvv.c
- HEMV (C/Z): replace zhemv_*_vector.c with zhemv_*_rvv.c
- Complex GEMM copy (C/Z): replace ../generic/zgemm_ncopy/tcopy with zgemm_ncopy_rvv_v1.c / zgemm_tcopy_rvv_v1.c

Generic fallback counts: ZVL256B 56->28, ZVL128B 63->47

### DIFF
--- a/kernel/riscv64/KERNEL.RISCV64_ZVL128B
+++ b/kernel/riscv64/KERNEL.RISCV64_ZVL128B
@@ -140,27 +140,27 @@ DGEMMITCOPYOBJ =  dgemm_itcopy$(TSUFFIX).$(SUFFIX)
 endif
 
 CGEMMKERNEL    =  cgemm_kernel_$(CGEMM_UNROLL_M)x$(CGEMM_UNROLL_N)_zvl128b.c
-CGEMMONCOPY    =  ../generic/zgemm_ncopy_$(CGEMM_UNROLL_N).c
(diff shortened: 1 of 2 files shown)


martin-frbg:
Sorry, no - you cannot just wire up GEMM copy functions from a different target without regard for the blocking  parameters as defined in param.h that the respective GEMM kernels use. Likewise, TRSM depends in part on the GEMM kernel so needs to be matched to it.
What would _probably_ work from this PR is the SYMV/HEMV parts , and reusing ZGEMMONCOPY/ZGEMMOTCOPY from x280 for ZVL256B - but exactly the zgemm_ncopy_4_rvv/zgemm_tcopy_4_rvv that KERNEL.x280 has. 
But it would still be a fairly pointless exercise without benchmark results showing a clear advantage

PR #5808, opened 2026-05-12
Make OpenBLAS's usage of OpenMP respect `openblas_set_num_threads()`

### LINKED ISSUE #5806: `openblas_set_num_threads()` is silently overridden when built with `USE_OPENMP`
### Summary

When OpenBLAS is built with `USE_OPENMP`, calling `openblas_set_num_threads(1)` before a BLAS operation has no lasting effect.

The thread count is unconditionally overridden back to `omp_get_max_threads()` on every BLAS call, making the API a silent no-op.

When a user uses `openblas_set_num_threads(1)` in the hope to get deterministic BLAS results, they will eventually find that this does not work, and results are still nondeterministic, because of this.

The workaround is to run with the `OMP_NUM_THREADS=1` environment variable set. But it seems very wrong that this cannot be done with code, and that a function that is explicitly named to set the number of threads, does not actually do that.

### Call chain demonstrating the problem

1. User calls [`openblas_set_num_threads(1)`](https://github.com/OpenMathLib/OpenBLAS/blob/8cecf899e21d99f9d8766ed34bfeeb3e2992c844/driver/others/blas_server_omp.c#L124) which calls [`goto_set_num_threads(1)`](https://github.com/OpenMathLib/OpenBLAS/blob/8cecf899e21d99f9d8766ed34bfeeb3e2992c844/driver/others/blas_server_omp.c#L103), setting [`blas_cpu_number = 1`](https://github.com/OpenMathLib/OpenBLAS/blob/8cecf899e21d99f9d8766ed34bfeeb3e2992c844/driver/others/blas_server_omp.c#L113).
2. User then calls `cblas_sgemm()`; code is in [`void CNAME()` in `gemm.c`](https://github.com/OpenMathLib/OpenBLAS/blob/8cecf899e21d99f9d8766ed34bfeeb3e2992c844/interface/gemm.c#L371).
3. Inside, [`args.nthreads = get_gemm_optimal_nthreads(MNK)`](https://github.com/OpenMathLib/OpenBLAS/blob/8cecf899e21d99f9d8766ed34bfeeb3e2992c844/interface/gemm.c#L717) calls [`num_cpu_avail(3)`](https://github.com/OpenMathLib/OpenBLAS/blob/8cecf899e21d99f9d8766ed34bfeeb3e2992c844/interface/gemm.c#L228).
4. [`num_cpu_avail()`](https://github.com/OpenMathLib/OpenBLAS/blob/8cecf899e21d99f9d8766ed34bfeeb3e2992c844/common_thread.h#L142) calls [`omp_get_max_threads()`](https://github.com/OpenMathLib/OpenBLAS/blob/8cecf899e21d99f9d8766ed34bfeeb3e2992c844/common_thread.h#L146), which returns the OpenMP default (all CPUs, since `OMP_NUM_THREADS` is unset).
5. [`if (blas_cpu_number != openmp_nthreads)`](https://github.com/OpenMathLib/OpenBLAS/blob/8cecf899e21d99f9d8766ed34bfeeb3e2992c844/common_thread.h#L164) is true (1 != N).
6. [`goto_set_num_threads(openmp_nthreads)`](https://github.com/OpenMathLib/OpenBLAS/blob/8cecf899e21d99f9d8766ed34bfeeb3e2992c844/common_thread.h#L165) **overrides `blas_cpu_number` back to N**.
7. [`num_cpu_avail()` returns `blas_cpu_number`](https://github.com/OpenMathLib/OpenBLAS/blob/8cecf899e21d99f9d8766ed34bfeeb3e2992c844/common_thread.h#L169) (now N), which flows to [`GEMM_THREAD(..., args.nthreads)`](https://github.com/OpenMathLib/OpenBLAS/blob/8cecf899e21d99f9d8766ed34bfeeb3e2992c844/interface/gemm.c#L748) then to [`exec_blas(num_cpu, queue)`](https://github.com/OpenMathLib/OpenBLAS/blob/8cecf899e21d99f9d8766ed34bfeeb3e2992c844/driver/level3/gemm_thread_n.c#L95) then to [`#pragma omp parallel for`](https://github.com/OpenMathLib/OpenBLAS/blob/8cecf899e21d99f9d8766ed34bfeeb3e2992c844/driver/others/blas_server_omp.c#L458).

### Root cause

[`goto_set_num_threads()`](https://github.com/OpenMathLib/OpenBLAS/blob/8cecf899e21d99f9d8766ed34bfeeb3e2992c844/driver/others/blas_server_omp.c#L103) does NOT call `omp_set_num_threads()` -- confirmed by searching the entire `driver/` directory (zero results). So the OpenMP runtime's idea of the thread count is never updated, and `num_cpu_avail()` always re-syncs `blas_cpu_number` from `omp_get_max_threads()`.

### Expected behavior

`openblas_set_num_threads(1)` should durably limit OpenBLAS to 1 thread, even when built with `USE_OPENMP`. The only current workaround is setting `OMP_NUM_THREADS=1` in the environment, which is a process-global side effect affecting all OpenMP consumers.

### Suggested fix

When `USE_OPENMP` is defined, `goto_set_num_threads()` should also call `omp_set_num_threads(num_threads)` so that subsequent calls to `omp_get_max_threads()` in `num_cpu_avail()` return the value the user requested.

I have not tested this yet, but it seems the most sensical to me that this should work.

### Practical impact

This breaks deterministic computation for any library that links OpenBLAS and tries to force single-threaded BLAS via the documented API. Multi-threaded floating-point reductions in `exec_blas()` produce nondeterministic results due to different summation orders across runs.

I found it when trying to make [llama-cpp](https://github.com/ggml-org/llama.cpp) deterministic with its `--threads 1` option and found that it didn't work.

### Version

Tested on commit `8cecf899e` (v0.3.32).

nh2:
Fixes #5806.

Also

* Documents what `openblas_set_num_threads_local()` does, see https://github.com/OpenMathLib/OpenBLAS/pull/4425#issuecomment-4426791139
* Cleans up the crazy syntax of `num_cpu_avail()` I complained about in https://github.com/OpenMathLib/OpenBLAS/issues/5806#issuecomment-4426885884

Reviewing by commit is recommended.

### DIFF
--- a/common_thread.h
+++ b/common_thread.h
@@ -138,10 +138,19 @@ typedef struct blas_queue {
 extern int blas_server_avail;
 extern int blas_omp_number_max;
 extern int blas_omp_threads_local;
+extern int blas_is_num_threads_set_explicitly;
 
 static __inline int num_cpu_avail(int level) {
 
 #ifdef USE_OPENMP
+  /* If the user explicitly called openblas_set_num_threads(),
+     respect that setting instead of overriding it with
+     `omp_get_max_threads()` below (which is to get a default
+     in case the user hasn't made an explicit choice). */
+  if (blas_is_num_threads_set_explicitly) {
+    return blas_cpu_number;
+  }
+
 int openmp_nthreads;
 	openmp_nthreads=omp_get_max_threads();
 	if (omp_in_parallel()) openmp_nthreads = blas_omp_threads_local;
--- a/driver/others/blas_server_omp.c
+++ b/driver/others/blas_server_omp.c
@@ -70,6 +70,7 @@
 int blas_server_avail = 0;
 int blas_omp_number_max = 0;
 int blas_omp_threads_local = 1;
+int blas_is_num_threads_set_explicitly = 0; // tracks whether the user called openblas_set_num_threads()
 
 extern int openblas_omp_adaptive_env(void);
 
@@ -122,7 +123,7 @@ void goto_set_num_threads(int num_threads) {
 
 }
 void openblas_set_num_threads(int num_threads) {
-
+	blas_is_num_threads_set_explicitly = 1;
 	goto_set_num_threads(num_threads);
 }
 
@@ -145,7 +146,7 @@ extern int openblas_omp_num_threads_env(void);
 
    if(blas_omp_number_max <= 0)
 	   blas_omp_number_max= openblas_omp_num_threads_env();
-   if (blas_omp_number_max <= 0) 
+   if (blas_omp_number_max <= 0)
 	   blas_omp_number_max=MAX_CPU_NUMBER;
 #else
     blas_omp_number_max = omp_get_max_threads();
@@ -365,14 +366,14 @@ static void exec_threads(int thread_num, blas_queue_t *queue, int buf_index){
 #ifdef BUILD_COMPLEX16
 	    sb = (void *)(((BLASLONG)sa + ((ZGEMM_P * ZGEMM_Q * 2 * sizeof(double)
 					    + GEMM_ALIGN) & ~GEMM_ALIGN)) + GEMM_OFFSET_B);
-#else 
+#else
 fprintf(stderr,"UNHANDLED COMPLEX16\n");
 #endif
 	  } else if ((queue -> mode & BLAS_PREC) == BLAS_SINGLE) {
 #ifdef BUILD_COMPLEX
 	    sb = (void *)(((BLASLONG)sa + ((CGEMM_P * CGEMM_Q * 2 * sizeof(float)
 					    + GEMM_ALIGN) & ~GEMM_ALIGN)) + GEMM_OFFSET_B);
-#else 
+#else
 fprintf(stderr,"UNHANDLED COMPLEX\n");
 #endif
 	  } else {


itamarst:
Would be nice to get a fix merged, if someone has time to review.

ngoldbaum:
@frbg - @itamarst is working with me and @kumaraditya303 on improving support for free-threading in Python packages. He's identified the issue this PR in particular fixes as the cause of substantial CPU oversubscription in scikit-learn using the threading joblib backend. We'd appreciate your eyes on this and the original issue.

martin-frbg:
Can you split this please ? openblas_set_num_threads() immediately getting reset to the OMP maximum is clearly unintended (though obviously it should be bounded by the OMP maximum).
 
Regarding the other other things in this PR - openblas_set_num_threads_local is intended to match mkl_set_num_threads_local (but may not yet be complete) - so nothing to do with forcing OMP parallel jobs inside an OMP_PARALLEL region (which IIRC would be invoking undefined behavior at best).
Refactoring num_cpu_avail() _may_ be a good idea at some point, but it bears the risk of introducing new bugs and I have neither the time nor energy to review that now.

nh2:
> Can you split this please ?

Done. I moved the rest to PR #5907.

martin-frbg:
Thank you

PR #5815, opened 2026-05-18
Fix contiguous memory check for SGEMM and DGEMM in RISC-V.

### LINKED ISSUE #5811: RISC-V (riscv64_zvl256b, SpaceMiT K1): DGEMM regression between 0.3.31 and 0.3.33 produces non-PSD matrix from A^T @ A
On SpaceMiT K1 hardware (RVA22 + RVV 1.0 (VLEN=256), runtime DYNAMIC_ARCH selects riscv64_zvl256b), DGEMM produces a matrix with significantly negative eigenvalues when computing `A.T @ A` for a 50x50 random matrix. Mathematically the result must be positive semi-definite. This regressed between OpenBLAS 0.3.31 and 0.3.33.

The downstream symptom is `numpy.linalg.cholesky` raising `LinAlgError: Matrix is not positive definite` on a matrix that should have been PSD by construction.

### Hardware

- BananaPi F3 (SpaceMiT K1 SoC, 8x X60 cores, RVA22 + RVV 1.0, VLEN=256)
- Armbian 25.11.2 trixie, Linux 6.6.99-current-spacemit
- Also confirmed on Muse Pi Pro (also K1) by another user
- NOT reproducible on Pine64 Star64 (SiFive U74, no RVV, dispatches to riscv64_generic), which sharpens the suspicion that the regression is in the riscv64_zvl256b kernels

### Build configuration

OpenBLAS built with `DYNAMIC_ARCH`, `USE64BITINT`, `NO_AFFINITY`. Runtime dispatch picks `riscv64_zvl256b` on the K1 (verified with `OPENBLAS_VERBOSE=2`).

### Bisect

Tested via LD_PRELOAD swap of the OpenBLAS shared library against an unchanged numpy 2.5.0.dev0 wheel. All 0.3.31.x scipy-openblas64 builds produce correct results; 0.3.33.0.0 produces wrong results.

| scipy-openblas64 | OpenBLAS embedded | min eigval of A^T @ A | cholesky |
| --- | --- | --- | --- |
| 0.3.31.22.1 | 0.3.31.dev | +1.405e-02 | OK |
| 0.3.31.126.1 | 0.3.31.dev | +1.405e-02 | OK |
| 0.3.31.126.5 | 0.3.31.dev | +1.405e-02 | OK |
| 0.3.31.159.0 | 0.3.31.159.0 | +1.405e-02 | OK |
| 0.3.31.188.0 | 0.3.31.188.0 | +1.405e-02 | OK |
| 0.3.33.0.0 | 0.3.33.0.0 | -2.779e+00 | LinAlgError |

The Debian 13 trixie system OpenBLAS 0.3.29 (`libopenblas0:riscv64 0.3.29+ds-3`) also produces correct results on the same machine.

scipy-openblas64 has no 0.3.32 release on PyPI, so the regression range at OpenBLAS source level is `v0.3.31..v0.3.33` (~250 commits).

### Minimal reproducer

Requires Python with numpy linked against the OpenBLAS in question:

```python
import numpy as np

np.random.seed(1)
a = np.random.randn(50, 50)
a = (a.T @ a).astype(np.float64)  # mathematically PSD

print("min eigval =", np.linalg.eigvalsh(a).min())
# OpenBLAS 0.3.31: +1.405e-02 (correct)
# OpenBLAS 0.3.33: -2.779e+00 (wrong by ~2 orders of magnitude)

np.linalg.cholesky(a, upper=True)  # raises LinAlgError on 0.3.33
```

The matrix is dense, symmetric (within FP roundoff), 50x50, float64. The error magnitude (~2.8) is far beyond floating-point precision. DGEMM is producing an actively wrong result, not a tiny precision drift.

### Context

Surfaced in numpy/numpy#30995 (riscv64 wheel CI). Confirmed by multiple K1 users. Cross-referencing here so OpenBLAS maintainers can pick up the bisect.

Available to test patches on the F3 if useful.

ChipKerchner:
Fix contiguous memory check for SGEMM and DGEMM in RISC-V.


ChipKerchner:
[5811](https://github.com/OpenMathLib/OpenBLAS/issues/5811)

ChipKerchner:
I also removed some dead code.

ChipKerchner:
[5674](https://github.com/OpenMathLib/OpenBLAS/pull/5674)

martin-frbg:
Thank you for the quick fix

ChipKerchner:
@martin-frbg This shows a need for additional unit tests.  In specific, we need more tests that have situations in which ldc does not equal M.

martin-frbg:
Yeah, there's always one more test we could/should add to utest... I'm not even sure BLAS-Tester (i.e. the ATLAS testsuite modified for OpenBLAS) covers this, but clearly the Reference-BLAS testsuite does not

luhenry:
@martin-frbg do we have an idea or tooling for code coverage? That's one case where it would make sense to have to understand what test case(s) we might need to add.

martin-frbg:
Actually we don't - Reference-LAPACK has codecov set up in CI but I don't remember its reports being used (except as a source of annoyance whenever a technical issue caused a spike in spurious coverage change messages)

luhenry:
I'm having a look locally to get some reports on current code coverage for different architectures (focusing on x86_64, aarch64, riscv64) and different microarchitectures of each. I'll share the results in an issue. I can work on adding some shapes to the different tests then to make sure we have some more cases covered.

And I hear your pain on having it as a gate on CI, it can be quite brittle. My goal is to have the build system support and documentation as a first step. Then we can decide how we want to proceed from there. We can keep it as documentation and maintainers/contributors can regularly check what's the status and fix it accordingly. Or we can have some soft or hard warning signs on CI. Choice will be up to maintainers.

Issue #3808, opened 2022-11-08
Updated RISC-V vector support for review. 


ken-unger:
Hi Xianyi and Martin,

A colleague and I have worked on an updated RISC-V Vector (RVV) implementation the past few months and are now looking to contribute our work back.   Our motivation was to support OpenBLAS on the SiFive X280 which is RVV 1.0 compliant.  The existing C910V implementation has a number of limitations which we encountered, and as such we decided it was better to create *_rvv.c implementations within kernel/riscv64, leaving the existing *_vector.c implementation for C910V untouched.  

Some of the improvements with this RVV implementation:
-	Support vector ABS (apparently not supported on C910V)
-	Support segment load/stores for complex data types, for better performance.
-	Alternative (in our view better) implementation choices.  Rather than loops using a predetermined vector length (VL), and a subsequent tail,  we found that determining VL within the loop was actually more efficient (in addition to cleaner/smaller code).
-	Generally faster for most kernels.
-	Significantly faster version of GEMM, modeled after what was done for SVE (8xVLEN). 
-	All implementations using vector intrinsics,  even the faster GEMM.   No assembler required.
 
Prior to submitting a PR, for your and others review, what tests should be run?  All of the CTESTs currently pass.  In the course of our work, we’ve also run the benchmark suite many times.

Thank you,
Ken

P.S  Our eventual PR would add the following, all of which is copyright OpenBLAS 2022:

        kernel/riscv64/KERNEL.x280
        kernel/riscv64/amax_rvv.c
        kernel/riscv64/amin_rvv.c
        kernel/riscv64/asum_rvv.c
        kernel/riscv64/axpby_rvv.c
        kernel/riscv64/axpy_rvv.c
        kernel/riscv64/copy_rvv.c
        kernel/riscv64/dot_rvv.c
        kernel/riscv64/gemm_beta_rvv.c
        kernel/riscv64/gemm_ncopy_rvv_v1.c
        kernel/riscv64/gemm_tcopy_rvv_v1.c
        kernel/riscv64/gemmkernel_rvv_v1x8.c
        kernel/riscv64/gemv_n_rvv.c
        kernel/riscv64/gemv_t_rvv.c
        kernel/riscv64/iamax_rvv.c
        kernel/riscv64/iamin_rvv.c
        kernel/riscv64/imax_rvv.c
        kernel/riscv64/imin_rvv.c
        kernel/riscv64/izamax_rvv.c
[...]

PR #5819, opened 2026-05-24
docs: clarify RISC-V RVV target selection and GCC 14+ requirement for ZVL128B/ZVL256B

trg-rgb:
## Summary
 
Add two notes to the RISC-V section of the build instructions, before
the per-target entries, addressing a documented source of user confusion:
 
1. **Target selection:** `RISCV64_GENERIC` is intentionally scalar by
   design — `Makefile.riscv64` appends a scalar march override that
   takes precedence over any `-march=rv64gcv` flag. Users who supply
   the V extension without changing TARGET get a silently scalar build
   with no diagnostic. The correct targets for RVV 1.0 hardware are
   `RISCV64_ZVL128B` / `RISCV64_ZVL256B`, as documented in the design
   rationale at #3808.
2. **Compiler floor:** GCC 14+ is required for the current `_rvv.c`
   kernel set. GCC 13 builds complete and produce a library but routines
   using segmented load/store intrinsics (`__riscv_vsseg*`) fall back to
   scalar silently. Functional tests pass; only disassembly detects this.
   Verified on OpenBLAS 0.3.33: GCC 13 (scalar fallback), GCC 14
   (~12,691 RVV opcodes), GCC 15 (~14,355 RVV opcodes).
Both points are consistent with existing upstream documentation; this PR
surfaces them at the TARGET selection step where a user would first
encounter the choice.
 
Related: #3808
Community investigations: https://github.com/clusterchallenge/Hardware-Abstraction-Layer-Transitional-Libraries/issues/25 (trg-rgb), https://github.com/clusterchallenge/Hardware-Abstraction-Layer-Transitional-Libraries/issues/23 (Vaibhav805)
 
Co-authored-by: Vaibhav805 <Vaibhav805@users.noreply.github.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,8 @@ Please read `GotoBLAS_01Readme.txt` for older CPU models already supported by th
   make HOSTCC=gcc TARGET=x280 NUM_THREADS=8 CC=riscv64-unknown-linux-gnu-clang FC=riscv64-unknown-linux-gnu-gfortran
   ```
 
+(For RVV 1.0 target selection and compiler requirements, see the [RISC-V section in docs/install.md](docs/install.md#risc-v).)
+
 - **ZVL???B**: Level-3 BLAS and Level-1,2 including vectorised kernels targeting generic RISCV cores with vector support with registers of at least the corresponding width; ZVL128B and ZVL256B are available.
 e.g.:
   ```sh
--- a/docs/install.md
+++ b/docs/install.md
@@ -796,6 +796,55 @@ make BINARY=32 BINARY32=1 CC=$MTI_TOOLCHAIN-gcc AR=$MTI_TOOLCHAIN-ar FC="$MTI_TO
 ```
 
 
+### RISC-V
+
+OpenBLAS supports several RISC-V targets. Target selection determines whether
+the resulting build is vectorized.
+
+#### Target selection
+
+| Target              | Vectorization                  | Use when                                                |
+| ------------------- | ------------------------------ | ------------------------------------------------------- |
+| `RISCV64_GENERIC`   | None — scalar reference path   | Non-vector cores, or as a baseline reference build      |
+| `RISCV64_ZVL128B`   | RVV 1.0, VLEN >= 128 bits      | Generic RVV-1.0 hardware with 128-bit vector registers  |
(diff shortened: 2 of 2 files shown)


martin-frbg:
Thanks - actually the GENERIC or `ARCH_`GENERIC target maps to the same non-vectorized plain C code on _all_ platforms - it will result in a working build of libopenblas on cpus that have no specific optimized support, but is likely to be only a little faster than the original "netlib" Reference BLAS.

Maybe the compiler requirements for RISCV_ZVL... would be better placed in the "Building from source" documentation, probably near the end where there are architecture-specific notes for other platforms already ? The documentation is still kind of patchy, but I fear README.md is getting too bloated for something that is displayed on the github page by default.
(My experience with the RISCV platform suggests one would normally want to use the latest stable compiler version to keep up with the ever-changing extensions - hopefully we're moving past this phase now)

trg-rgb:
Thanks @martin-frbg — that GENERIC framing is right, and I've reworded.

v2 (just pushed):

- Moved the target-selection details and compiler-requirements note out of README.md and into a new `### RISC-V` subsection in `docs/install.md`, placed inside `## Building from source` between the MIPS and FreeBSD sections, so it lives with the other per-platform build notes.
- README.md now keeps a single one-line pointer to the new section at the spot where users encounter the RISC-V build commands, so the signposting is still there without bloating the page that GitHub renders by default.
- Reworded the GENERIC framing per your correction — it's now positioned as the general "GENERIC / ARCH_GENERIC maps to plain-C on all architectures" property, with the `Makefile.riscv64` `-march` override called out as the RISC-V-specific compounding factor.

The earlier CI failure (`RISCV64_GENERIC` runner, exit 140) looks unrelated to this docs-only change; happy to look more closely if it persists on this push.

trg-rgb:
@martin-frbg gentle nudge on this when you have a moment. v2 moved the target-selection and compiler notes out of README.md into a `### RISC-V` subsection in `docs/install.md` under *Building from source*, placed with the other per-platform notes (between the MIPS and FreeBSD entries), and left README.md with just a one-line pointer. I also reworded the GENERIC note per your correction, so it now describes the general behavior where GENERIC and ARCH_GENERIC both map to the non-vectorized plain C path on all platforms, with the `Makefile.riscv64` `-march` override called out as the RISC-V-specific factor. Does that placement look right, or is there anything else you'd like adjusted before it can go in? CI is green on the latest push (601ff0e).

PR #5827, opened 2026-06-04
Power10: Replace vector pair loads with __builtin_vsx_lxvp

amritahs-ibm:
Replace normal vector pair pointer dereferences with the optimized __builtin_vsx_lxvp builtin across DGEMM, ZGEMM, and DGEMV kernels.

Also done some identation corrections in dgemm_kernel_power10.c. This is done as part of POWER code cleanup and may not have any performance impact.


amritahs-ibm:
I have done functional and performance verification on a Power10 machine. There are no unit testcase failures and there is no performance regression.

martin-frbg:
Thank you

PR #5828, opened 2026-06-04
Fix incorrect inline assembly constraints in dcbt prefetch instructions

amritahs-ibm:
Corrected the register constraints for the PowerPC dcbt (Data Cache Block Touch) instruction in Power10 kernel implementations. The dcbt instruction has special behavior where if the first operand (RA) is r0, it uses the value 0 instead of the register contents. Therefore, RA must use the "b" constraint (any GPR except r0), while RB can use "r" (any GPR including r0).

Changes:
- Changed first operand constraint from "r" to "b" to exclude r0
- Changed second operand constraint from "b" to "r" for flexibility

This ensures correct prefetch behavior and compliance with PowerPC ISA specifications, preventing potential issues where r0 might be incorrectly used as the base address register.


martin-frbg:
Is this still a draft as suggested by its title ?

amritahs-ibm:
@martin-frbg Could you please merge this PR?

martin-frbg:
Done, thanks for the fix

PR #5835, opened 2026-06-12
Fix 32-bit integer overflow in LAPACKE nancheck index arithmetic

changangela:
The LAPACKE NaN checks compute array offsets in 32-bit `lapack_int`, which overflows for large matrices. Found while running Cholesky on a 47000x47000 matrix: `LAPACKE_dpotrf` with `n = lda >= 46341` crashes in `LAPACKE_dtr_nancheck` before the factorization even runs, because `j*lda` wraps negative. Same family of overflow as scipy/scipy#20229.

The `ge`/`gb`/`tp`/`tf` helpers already cast to `size_t`, so this applies the same fix to the ones that were missed:

- `?tr_nancheck`: cast `a[i+j*lda]` to `size_t` (covers the po/sy/he checks, i.e. potrf)
- `?_nancheck` (vector): the `n*inc` loop bound overflowed; iterate by count and index with `size_t`
- `?sp`/`?pp`/`?pf`/`?hp`/`?tp`/`?tf`: compute the packed length `n*(n+1)/2` in `size_t`
- `?tz_nancheck`: widen the part offsets to `int64_t`

Tested with UBSan at n = 47000: the old code overflows and segfaults (tr, tz) or silently returns "no NaN" on input that has one (vector, packed); the fixed code detects a NaN planted at the far end of the scanned region in all cases. Old and new code give identical results on an exhaustive sweep of small sizes (all layouts/uplo/diag/shapes, NaN at every position).

Reference-LAPACK master has the same bug in these files; I can submit the fix there as well.

### DIFF
--- a/lapack-netlib/LAPACKE/utils/lapacke_c_nancheck.c
+++ b/lapack-netlib/LAPACKE/utils/lapacke_c_nancheck.c
@@ -42,8 +42,8 @@ lapack_logical LAPACKE_c_nancheck( lapack_int n,
     if( incx == 0 ) return (lapack_logical) LAPACK_CISNAN( x[0] );
     inc = ( incx > 0 ) ? incx : -incx ;
 
-    for( i = 0; i < n*inc; i+=inc ) {
-        if( LAPACK_CISNAN( x[i] ) )
+    for( i = 0; i < n; i++ ) {
+        if( LAPACK_CISNAN( x[(size_t)i*inc] ) )
             return (lapack_logical) 1;
     }
     return (lapack_logical) 0;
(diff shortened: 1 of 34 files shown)


martin-frbg:
Thank you - and yes please submit this fix to Reference-LAPACK as well, we're only copying LAPACKE and (most of) LAPACK from there

changangela:
> Thank you - and yes please submit this fix to Reference-LAPACK as well, we're only copying LAPACKE and (most of) LAPACK from there

sounds good, here it is: https://github.com/Reference-LAPACK/lapack/pull/1294

PR #5838, opened 2026-06-14
Fix corruption due to lock sharding issues by centralizing locking

### LINKED ISSUE #5836: `gemm_driver` templating breaks GEMM locking when different GEMMs happen concurrently
c.f. https://github.com/numpy/numpy/issues/31618 and https://github.com/scikit-image/scikit-image/issues/8212 where this was originally found.

I used Claude to help debug this, I hope that's alright.

OpenBLAS uses templating to generate several variants of `gemm_driver`. If more than one of these variants are called concurrently this can corrupt internal state in OpenBLAS and the design intention that only one GEMM happens at a time.

OpenBLAS serializes concurrent multithreaded Level-3 calls with a mutex declared inside `gemm_driver()` ([driver/level3/level3_thread.c](https://github.com/ngoldbaum/scikit-image/blob/claude/gallant-goldberg-po5icy/driver/level3/level3_thread.c)):

```
static int gemm_driver(blas_arg_t *args, ...) {
  ...
  static pthread_mutex_t level3_lock = PTHREAD_MUTEX_INITIALIZER;   // ~line 590
  ...
  pthread_mutex_lock(&level3_lock);     // wraps job setup + exec_blas  ~line 659
```

But `gemm_driver` is template-compiled once per transpose combination — `GEMM_LOCAL` is #defined to `GEMM_NN`, `GEMM_NT`, `GEMM_TN`, `GEMM_TT`, … (lines 58–76), producing four separate functions `dgemm_thread_{nn,nt,tn,tt}`. A function-local static gets one instance per compiled function, so there are four independent `level3_lock` mutexes.

Consider the following C reproducer:

```C
#define _GNU_SOURCE
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <pthread.h>
#include <dlfcn.h>

typedef long bint;
typedef void (*dgemm_t)(int,int,int,bint,bint,bint,double,const double*,bint,
                        const double*,bint,double,double*,bint);
typedef void (*setnt_t)(int);
static dgemm_t dgemm;
enum { CblasRowMajor=101, CblasNoTrans=111, CblasTrans=112 };

#define M 262144          /* large enough that each GEMM goes multi-threaded */
static double *Att;       /* 4 x M, feeds the TransA=Trans GEMM (lda=M)      */
static double *Ann;       /* M x 4, feeds the TransA=NoTrans GEMM (lda=4)    */
static double Bmat[8], *ref;
static int NTHREADS = 8, NITER = 200, SAME_VARIANT = 0;
static long mismatches = 0;
static pthread_mutex_t mlock = PTHREAD_MUTEX_INITIALIZER;

/* GEMM #1: TransA=Trans, K=2  ->  dgemm_thread_{tn/nt} variant */
/* GEMM #2: either Trans (same variant) or NoTrans (different variant), K=4 */
static void compute(double *o1, double *o2) {
  dgemm(CblasRowMajor, CblasTrans, CblasNoTrans, M, 2, 2,
        1.0, Att, M, Bmat, 2, 0.0, o1, 2);
  if (SAME_VARIANT)
    dgemm(CblasRowMajor, CblasTrans, CblasNoTrans, M, 2, 4,
          1.0, Att, M, Bmat, 2, 0.0, o2, 2);   /* same Trans variant */
  else
    dgemm(CblasRowMajor, CblasNoTrans, CblasNoTrans, M, 2, 4,
          1.0, Ann, 4, Bmat, 2, 0.0, o2, 2);   /* different (NoTrans) variant */
}

static void *worker(void *unused) {
  (void)unused;
  double *o1 = malloc((size_t)M*2*sizeof(double));
  double *o2 = malloc((size_t)M*2*sizeof(double));
  long local = 0;
  for (int i = 0; i < NITER; i++) {
    compute(o1, o2);
    if (memcmp(o1, ref, (size_t)M*2*sizeof(double)) != 0) local++;
  }
  free(o1); free(o2);
  pthread_mutex_lock(&mlock); mismatches += local; pthread_mutex_unlock(&mlock);
  return NULL;
}

int main(int argc, char **argv) {
  if (argc > 1) NTHREADS = atoi(argv[1]);
  if (argc > 2) NITER    = atoi(argv[2]);
  if (argc > 3) SAME_VARIANT = atoi(argv[3]);

  void *h = dlopen(getenv("OB_LIB"), RTLD_NOW | RTLD_GLOBAL);
  if (!h) { fprintf(stderr, "set OB_LIB to the OpenBLAS .so (%s)\n", dlerror()); return 2; }
  /* try plain and scipy-suffixed symbol names */
  dgemm = (dgemm_t)dlsym(h, "cblas_dgemm");
  if (!dgemm) dgemm = (dgemm_t)dlsym(h, "scipy_cblas_dgemm64_");
  setnt_t setnt = (setnt_t)dlsym(h, "openblas_set_num_threads");
  if (!setnt) setnt = (setnt_t)dlsym(h, "scipy_openblas_set_num_threads64_");
  if (!dgemm) { fprintf(stderr, "no cblas_dgemm symbol\n"); return 2; }
  if (setnt) setnt(4);

  Att = malloc((size_t)4*M*sizeof(double));
  Ann = malloc((size_t)4*M*sizeof(double));
  for (size_t i = 0; i < (size_t)4*M; i++) { Att[i] = (double)(i % 512); Ann[i] = (double)((i*7) % 512); }
  for (int i = 0; i < 8; i++) Bmat[i] = 0.1*(i+1);

  ref = malloc((size_t)M*2*sizeof(double));
  { double *tmp = malloc((size_t)M*2*sizeof(double)); compute(ref, tmp); free(tmp); }

  pthread_t th[256];
  for (int i = 0; i < NTHREADS; i++) pthread_create(&th[i], NULL, worker, NULL);
  for (int i = 0; i < NTHREADS; i++) pthread_join(th[i], NULL);

  printf("%-26s %d threads x %d iters: %ld mismatches (%s)\n",
         SAME_VARIANT ? "same transpose variant:" : "different variants:",
         NTHREADS, NITER, mismatches, mismatches ? "RACE" : "clean");
  return mismatches ? 1 : 0;
}
```

You can compile it and run it like so:

```
cc -O2 openblas_gemm_variant_race.c -lpthread -ldl -o race 
```

```
goldbaum at Nathans-MBP in ~/Downloads
○  OB_LIB=/opt/homebrew/opt/openblas/lib/libopenblas.dylib ./race 8 200 0
different variants:        8 threads x 200 iters: 66 mismatches (RACE)

goldbaum at Nathans-MBP in ~/Downloads
○  OB_LIB=/opt/homebrew/opt/openblas/lib/libopenblas.dylib ./race 8 200 1
same transpose variant:    8 threads x 200 iters: 0 mismatches (clean)
```

ngoldbaum:
Fixes #5836.

I used both Claude and Codex to work on this. I typed this PR description by hand.

## Summary

Fixes the specific lock sharding issue described in #5836 as well as other related issues caused by the current locking strategy allowing shared overlapping calls of different kernels. The locks are function-local and so appear as different mutex objects in each separate compilation unit.

I fixed that by moving the locking to its own compilation unit with a new internal helper functions for serializing calls. This also centralizes the platform-specific locking logic into the new file.

This has the net effect of *reducing throughput* for workloads that make overlapping external calls into different threaded kernels, because those calls are now serialized consistently instead of using per-kernel lock state. I think the existing behavior is a bug rather than an intentional design choice but I wanted to raise that front and center.

It also fixes several other issues, more or less as a consequence of making the above change systematically:

* There is a behavior change in the OpenMP gemm3m implementation. Currently, locking is *skipped* for OpenMP gemm3m. I think this is a bug but maybe it's intentional? It would be a behavior change regardless.
* The old win32 path appears to have used a separate critical section per invocation, so it did not provide cross-call serialization - there is one critical section *per call* in the old implementation. The new file initializes a process-wide lock once and uses that.
* The old implementation caused substantial OpenMP oversubscription by duplicating [`parallel_section_left`](https://github.com/OpenMathLib/OpenBLAS/blob/7aa79fbdff522294a8386dfd5141075254edcea5/driver/level3/level3_thread.c#L575). Now there's only one version of this variable so there are far fewer OpenMP threads when one mixes kernels.

## Testing

To verify the correctness of the fix, I added a new multithreaded stress test based on the reproducer in #5836. I also enabled multithreaded stress testing for msys2 on a Windows host to test the Windows threading model.

Additionally there is a thread sanitizer test run. I manually verified that the new tests trigger validation errors and/or TSan race reports. I also verified the new mixed DGEMM stress test (no TSan) fails with incorrect results on an unpatched develop build and passes with this change. 

Right now there's only TSan testing for `OPENBLAS_NUM_THREADS=2` and the pthreads backend. TSan detected data races with `OPENBLAS_NUM_THREADS=4` and also with the OpenMP backend. I am intentionally leaving the TSan CI as-is in this PR. We'll need to look at other issues before setting up more thorough TSan CI.

### DIFF
--- a/.github/workflows/dynamic_arch.yml
+++ b/.github/workflows/dynamic_arch.yml
@@ -417,6 +417,10 @@ jobs:
             -DBUILD_SHARED_LIBS=ON
             -DBUILD_STATIC_LIBS=OFF
             -DBUILD_WITHOUT_LAPACK=ON
+            -DBUILD_SINGLE=OFF
+            -DBUILD_DOUBLE=ON
+            -DBUILD_COMPLEX=OFF
+            -DBUILD_COMPLEX16=OFF
             "-DDYNAMIC_ARCH=$dynamic_arch"
             -DNOFORTRAN=ON
             -DUSE_THREAD=ON
@@ -443,7 +447,9 @@ jobs:
           cmake "${cmake_args[@]}" ..
 
       - name: Build OpenBLAS
-        run: cd build && cmake --build .
+        run: |
+          cd build
+          cmake --build . --target dgemm_thread_safety dgemm_thread_safety_mixed dgemv_thread_safety
 
       - name: Show ccache status
         continue-on-error: true
@@ -535,7 +541,11 @@ jobs:
                 -DBUILD_SHARED_LIBS=ON \
                 -DBUILD_STATIC_LIBS=OFF \
                 -DBUILD_WITHOUT_LAPACK=ON \
-                -DDYNAMIC_ARCH=ON \
+                -DBUILD_SINGLE=OFF \
+                -DBUILD_DOUBLE=ON \
+                -DBUILD_COMPLEX=OFF \
+                -DBUILD_COMPLEX16=OFF \
+                -DDYNAMIC_ARCH=OFF \
                 -DNOFORTRAN=ON \
                 -DUSE_THREAD=ON \
                 -DUSE_OPENMP=OFF \
@@ -550,7 +560,9 @@ jobs:
                 ..
 
       - name: Build OpenBLAS
-        run: cd build && cmake --build .
+        run: |
+          cd build
+          cmake --build . --target dgemm_thread_safety dgemm_thread_safety_mixed dgemv_thread_safety
 
       - name: Show ccache status
         continue-on-error: true
@@ -567,6 +579,7 @@ jobs:
         timeout-minutes: 30
         run: |
           cd build
+          export PATH="$PWD/lib:$PATH"
           OPENBLAS_NUM_THREADS=8 OMP_NUM_THREADS=16 ctest -R 'dgemm_thread_safety|dgemm_thread_safety_mixed|dgemv_thread_safety' --output-on-failure
 
 
--- a/.github/workflows/dynamic_arch.yml
+++ b/.github/workflows/dynamic_arch.yml
@@ -334,6 +334,242 @@ jobs:
           echo "::endgroup::"
 
 
+  linux_thread_stress:
+    if: "github.repository == 'OpenMathLib/OpenBLAS'"
+    name: ${{ matrix.check-name }}
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - backend: pthread
+            check-name: "linux_thread_stress (pthread)"
+          - backend: openmp
+            check-name: "linux_thread_stress (openmp)"
+          - backend: tsan
+            check-name: linux_thread_sanitizer
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
+      - name: Install Dependencies
+        run: |
+          cat << EOF | sudo tee -a /etc/apt/apt.conf.d/01norecommend
+          APT::Install-Recommends "0";
+          APT::Install-Suggests "0";
+          EOF
+          sudo apt-get update
+          sudo apt-get install -y ccache cmake ninja-build
+          if [ "${{ matrix.backend }}" = "tsan" ]; then
+            sudo apt-get install -y clang llvm
+          fi
+
+      - name: Compilation cache
+        uses: actions/cache@v5
+        with:
+          path: ~/.ccache
+          key: ccache-${{ runner.os }}-thread-${{ matrix.backend }}-${{ github.ref }}-${{ github.sha }}
+          restore-keys: |
+            ccache-${{ runner.os }}-thread-${{ matrix.backend }}-${{ github.ref }}
+            ccache-${{ runner.os }}-thread-${{ matrix.backend }}
+
+      - name: Configure ccache
+        # Limit the maximum size and switch on compression to avoid exceeding the total disk or cache quota.
+        run: |
+          test -d ~/.ccache || mkdir -p ~/.ccache
+          echo "max_size = 250M" > ~/.ccache/ccache.conf
+          echo "compression = true" >> ~/.ccache/ccache.conf
+          ccache -s
+
+      - name: Configure OpenBLAS
+        run: |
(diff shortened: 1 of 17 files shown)


martin-frbg:
Thank you - beat me to it. (The single CI failure in Jenkins is an internal docker error related to the use of sudo for preparing a cmake-based build on zarch - I've restarted that job now)

h-vetinari:
Numpy started relying on this PR in a way that affects conda-forge: https://github.com/numpy/numpy/issues/31708

I've worked around this, so there's no specific urgency, but I was wondering when 0.3.34 is roughly planned to appear? Thanks. :)

martin-frbg:
@h-vetinari I had been hoping for either this weekend or the next, but now I'm no longer sure thanks to #5866 and another issue with the latest imports from Reference-LAPACK. I keep forgetting if there are readily available numpy wheels built against Reference-LAPACK that would let me check if these are genuine netlib bugs or aggravated by OpenBLAS FMA accuracy

h-vetinari:
> I keep forgetting if there are readily available numpy wheels built against Reference-LAPACK that would let me check if these are genuine netlib bugs or aggravated by OpenBLAS FMA accuracy

Not wheels AFAIU, but it's pretty easy to test numpy against reference lapack with the builds from conda-forge
```
mamba create -n test_env numpy libblas=*=*netlib hypothesis
mamba activate test_env
python -c "import numpy; numpy.test()"
```

rgommers:
I was just about to write the same thing. That can also be done when building numpy/scipy from source, using runtime switching of BLAS libraries against an existing `numpy` (or `scipy`) build: https://scipy.github.io/devdocs/dev/contributor/debugging_linalg_issues.html#setting-up-your-environment-for-switching-blas-libraries.

martin-frbg:
Thank you - I really really need to make a note of that this time.

PR #5842, opened 2026-06-20
bench.h: fix getsec() double time conversion on Apple Silicon

a-eridani:
On Apple Silicon, begin()/end() use clock_gettime_nsec_np(CLOCK_UPTIME_RAW), which already returns nanoseconds. getsec() then also multiplies by mach_timebase_info numer/denom — a conversion only valid for raw mach_absolute_time() ticks. On M-series numer/denom = 125/3, so every benchmark reports time ~41.67x too high (MFlops ~41.67x too low).

Repro (M-series): sleep 0.5s between begin()/end(), getsec() returns ~20.9s.

Drop the timebase scaling in the __APPLE__ branch.


martin-frbg:
Oops,looks like I conflated information gathered from the net, back when I didn't have access to M hardware.

PR #5843, opened 2026-06-20
AArch64: NEON omatcopy CT/RT kernels (s/d)

a-eridani:
AArch64 has no vectorized transpose copy - all variants hit the scalar generic. Adds NEON ct/rt kernels for s/d (register transpose + stnp)
Passes the utest extension suite (1460/1460)
Bench on Neoverse-N1 (domatcopy, 1thread): the scalar path degrades with matrix size while the NEON kernel stays flat, giving roughly 1.2x at 2k up to ~4.5x at 18k. Single precision gap is larger.


martin-frbg:
Thank you

martin-frbg:
(the addition probably needs copying to kernel/arm64/KERNEL to actually make it available on NEOVERSEN1 and others that don't `include KERNEL.ARMV8` - I'll see to it tomorrow after further testing)

PR #5844, opened 2026-06-23
fix data races in blas_server and level3_thread

kumaraditya303:
This PR fixes the remaining data races from https://github.com/OpenMathLib/OpenBLAS/pull/5838/ where previously it caused data races when the number of threads was higher than 2. With this PR, there are no such data races when running the tests even with 16 threads as I tested on my system.


ngoldbaum:
I think this should also let you increase `OPENBLAS_NUM_THREADS` in the TSan CI job. Currently it’s two, I saw races with four. Oversubscribing the runner a little bit probably makes races more likely.

martin-frbg:
Great, thanks a lot

PR #5858, opened 2026-06-24
Increase tolerance in ?LASD2 to improve ?BDSDC stability (Reference-LAPACK PR 1286)

martin-frbg:


martin-frbg:
This change creates a significant mismatch in the codspeed test that uses dgesdd - not sure if it is safe to increase our tolerance there (as the same repercussions might occur in numpy etc), or if I should raise this with the LAPACK team.
@rgommers ?
(thought I had already added this comment last night but somehow it didn't post - apologies if you see it twice)

rgommers:
I wouldn't particularly worry about `numpy` here, since if it's valid to bump tolerances here then it's equally valid to do so in the `numpy` test suite. However, the errors seems a bit large to me. For a size (1000, 222) array filled with random floats, the test tolerances were rtol=1e-07, atol=1e-05, and this isn't just a small increase:
```
E       Max absolute difference among violations: 2.336502e-05
E       Max relative difference among violations: 0.01933236
```
I don't see any discussion in https://github.com/Reference-LAPACK/lapack/pull/1286 or in the linked issue it closes that that's expected - so probably worth checking?

PR #5869, opened 2026-06-25
mkdocs: set repo_name

MaddyGuthridge:
That way the path to the GitHub repo is displayed instead of just the generic text "GitHub"


rgommers (approved):
LGTM, thanks @MaddyGuthridge.

Tested locally, changes from this:

Image

to this:

Image

It's minor, but it does look a little nicer.

I'm inclined to merge this without running O(90) more CI jobs; @martin-frbg please let me know if that doesn't sound right.

PR #5875, opened 2026-07-02
Allow benchmark targets to link against a user-specified OpenBLAS library

naoto-aoki-fy:
## Summary

This pull request makes the benchmark Makefile support linking benchmark binaries against a configurable OpenBLAS library.

It introduces the `OPENBLAS_LIB` variable, which defaults to the existing `../$(LIBNAME)` path, and replaces hard-coded references to `../$(LIBNAME)` in benchmark link targets with this variable.

## Motivation

The benchmark targets currently depend directly on `../$(LIBNAME)`, which makes it difficult to link them against OpenBLAS packages provided by Linux distributions or other externally installed OpenBLAS libraries.

By making the library path configurable, downstream packagers and users can more easily build benchmark binaries against a distro-provided OpenBLAS package, including shared-library installations. For example, benchmarks can be built with an externally provided library by passing `OPENBLAS_LIB` on the make command line:

```sh
make OPENBLAS_LIB=-lopenblaso
```

The default behavior is preserved when `OPENBLAS_LIB` is not overridden.

## Changes

* Add `OPENBLAS_LIB ?= ../$(LIBNAME)` to `benchmark/Makefile`
* Replace hard-coded `../$(LIBNAME)` dependencies with `$(OPENBLAS_LIB)` across `.goto` benchmark targets
* Allow benchmark builds to override the OpenBLAS link target, for example with `make OPENBLAS_LIB=-lopenblaso`
* Keep the existing default linkage behavior unchanged when `OPENBLAS_LIB` is not overridden

## Notes

This change is intended to improve build flexibility for benchmark targets, especially for packaging environments that prefer linking against system-provided OpenBLAS libraries.

### DIFF
--- a/benchmark/Makefile
+++ b/benchmark/Makefile
@@ -29,7 +29,9 @@
 TOPDIR	= ..
 include $(TOPDIR)/Makefile.system
 
-OPENBLAS_LIB ?= ../$(LIBNAME)
+ifndef OPENBLAS_LIB
+  OPENBLAS_LIB_DEP = ../$(LIBNAME)
+endif
 
 # ACML standard
 #ACML=/opt/acml5.3.1/gfortran64_mp/lib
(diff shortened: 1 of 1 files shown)


naoto-aoki-fy:
I noticed a mistake, so I'll close this for now.

PR #5874, opened 2026-07-02
Fix ZA tile slice indices in ssyrk SME direct kernel

jschueller:
The kernel_2x2 function uses 4 ZA tiles (0-3) each with svl slices. Tiles 0/1 handle rows 0..svl-1 with slice indices 0..svl-1. Tiles 2/3 handle rows svl..2*svl-1, so their slice indices must start at 0, i.e. (i - svl) instead of i.

Fix all three tile 2/3 access sites:
- C load into ZA (svwrite_hor_za32_f32_m)
- C writeback for UPPER (svst1_hor_za32)
- C writeback for LOWER (svst1_hor_za32)

Fixes #5873

PR #5880, opened 2026-07-03
Fix broken conditional preventing ARM64 SME compilation with AppleClang

martin-frbg:
(issue was noted in comment on #5874)


ogrisel:
I confirm this fixes the build on my Apple M4 macOS dev environment.

PR #5881, opened 2026-07-03
ci, docs: fix documentation text and skip docs-only CI

moluopro:
- Fix several documentation typos, wording issues, broken Markdown formatting, and links.
- Run the docs workflow only for documentation-related inputs:
  - `docs/**`
  - `mkdocs.yml`
  - `.github/workflows/docs.yml`
- Skip non-doc CI workflows when a change only touches documentation files:
  - `docs/**`
  - `**/*.md`

PR #5891, opened 2026-07-06
Update README and CI job matrix documentation

moluopro:
1. Update the CI job matrix to reflect the GitHub Actions jobs.

2. Replace stale README CI badges for inaccessible services.

3. Fix broken README links in the supported OS section and update the FreeBSD note.

4. Skip the FreeBSD and Azure CI workflows for docs-only changes, completing the follow-up from #5881.

### DIFF
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -4,9 +4,15 @@ on:
   push:
     branches:
       - develop
+    paths-ignore:
+      - 'docs/**'
+      - '**/*.md'
   pull_request:
     branches:
       - develop
+    paths-ignore:
+      - 'docs/**'
+      - '**/*.md'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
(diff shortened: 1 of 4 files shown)


martin-frbg:
Thanks. I guess the CI situation will always be a bit dynamic, it used to be easier to keep track of when this was a wiki page. 
Most of the qemu-based jobs take far too long, often hanging in the fork test that passes on actual hardware. 
I'm already thinking about removing the freebsd-on-arm one that replaced the freebsd on actual neoverse that Cirrus had. 
And the Jenkins jobs are just silly in that one can only see the job result while logged in to OSUOSL...

moluopro:
Yes, the current CI does seem to run into problems quite often. I think the CI process itself could be improved, for example by making the test scripts more robust, improving error reporting, and reducing unnecessary test triggers, as was already done for documentation-only changes.

In commercial RISC-V environments, companies usually maintain a board farm to support their CI pipelines, especially for performance testing, which works much better than QEMU.

As for documenting changes in the CI setup, it is admittedly tedious, but I think it is still manageable.

PR #5898, opened 2026-07-09
Fix ARM64 SME direct-kernel build and dispatch

moluopro:
- Add the missing CMake compile definitions for ARM64 SME direct-kernel variants to match the existing `kernel/Makefile.L3` rules.
- Restrict the STRMM direct fast path in `interface/trsm.c` to `TRMM` builds so STRSM cannot dispatch to STRMM direct kernels.
- Treat real-valued `CblasConjNoTrans` and `CblasConjTrans` as eligible for the corresponding STRMM direct paths.
- Add an `alpha == 0` fast zeroing path in `strmm_direct_arm64_sme1.c`.
- Avoid signed overflow in the `M * N * K` size heuristic in `sgemm_direct_performant.c`, and skip the direct path when `N <= 0` or `K <= 0`.
- Clean up the Windows/Clang condition for ARMV9SME and fix the surrounding indentation in `cmake/cc.cmake`.
- Normalize the relevant SME source files under `kernel/arm64` from CRLF to LF line endings.


moluopro:
This follows up on #5889. 

While working on the ARM64 SME TRMM direct path, I found a few related build and dispatch issues and fixed them here as well.

moluopro:
The `jenkins/pr-merge` pipeline seems to have failed on the last few PRs as well. Looks like Jenkins needs a good kick or two.

martin-frbg:
> The `jenkins/pr-merge` pipeline seems to have failed on the last few PRs as well. Looks like Jenkins needs a good kick or two.

Seems the Z arch build cannot find its base Ubuntu docker file recently. I will try to reach out to OSUOSL if the problem persists, but they're a small volunteer group too.

martin-frbg:
Messaged OSUOSL support now as their docker still fails to load images

martin-frbg:
Unfortunately the OSUOSL response just now has been to take the zarch node offline for an eventual rebuild

PR #5903, opened 2026-07-10
Add SiFive U74 target with a scalar 4×4 register-tiled GEMM kernel

hmeiland:
## Motivation

RV64GC application cores without the vector extension (SiFive U74/U54, Andes AX45MP, SiFive P550, …) currently fall back to the `RISCV64_GENERIC` target, whose GEMM uses a portable **2×2** C micro-kernel. On the StarFive VisionFive 2 (JH7110, quad-core U74) that leaves HPL at ≈3.1 GFLOPS — only ~26 % of the 12 GFLOPS double-precision peak.

The bottleneck is the micro-kernel, not the memory system. The U74's `fmadd.d` has a **7-cycle latency** at 1/cycle throughput on a single FP pipe; a 2×2 tile exposes only four independent accumulator chains — fewer than the latency — so the pipe stalls on the accumulator recurrence, while the 1:1 load:FMA ratio saturates the lone load/store port.

## What this adds

A dedicated **`U74`** target and a scalar **4×4 register-tiled** GEMM micro-kernel:

- **`kernel/generic/gemmkernel_4x4.c`** — portable C 4×4 kernel (16 accumulators). Raises in-flight chains above the FMA latency and lowers the load:FMA ratio to 1:2, fitting RV64G's 32 FP registers with no spill; handles all 4/2/1 edge cases.
- **`kernel/generic/kern_u74.S`** + **`kernel/generic/gemmkernel_4x4_u74.c`** — a hand-scheduled scalar **assembly** DGEMM micro-kernel (full operand double-buffering / P·Q ping-pong, load-before-FMA issue ordering matched to the dual-issue in-order front end), wired through a hybrid dispatcher: the asm serves full 4×4 / even-`bk` DGEMM blocks and falls back to the C kernel for ragged edges, odd `bk`, and TRMM. `kern_u74.S` is the readable source; the `.c` embeds it via top-level `__asm__` so it builds as a single kernel object with no build-system changes.
- **U74 target plumbing**: `getarch.c`, `cpuid_riscv64.c`, `TargetList.txt`, `KERNEL.U74`, `param.h`, `Makefile.riscv64`, `Makefile.prebuild`. Block sizes tuned for the 32 KiB L1D / 2 MiB shared L2 (`DGEMM_Q = 256`, aligning KC with a typical HPL `NB = 192`).

## Measured results (physical 8 GiB VisionFive 2, quad U74 @ 1.5 GHz)

Single-core DGEMM micro-kernel: **1.63 vs 1.00 GFLOPS** over the generic 2×2 (≈1.6×). Micro-benchmark probes show the assembly reaches the FP-pipe peak — 2.90 GFLOPS on L1-resident data (~16.5 cycles per 16 `fmadd.d`).

Four-core HPL, same board, matched N / NB=192 / 2×2 grid:

| N | generic 2×2 | 4×4 C | 4×4 C, Q=256 | + asm kernel |
|---:|---:|---:|---:|---:|
| 10 000 | 3.10 | 4.86 | 4.97 | 5.17 |
| 27 456 | 3.31 | 5.24 | 5.41 | **5.99** (~50 % of peak) |

End-to-end **+81 %** over the generic 2×2 at the memory-filling size; every run residual-PASSED.

## Correctness / validation

- The C 4×4 kernel is verified against a reference GEMM across **27,436** shape × scalar combinations (every 4/2/1 edge) with zero absolute error.
- A full `TARGET=U74` build passes the OpenBLAS BLAS **Level-3 test suite** — `DGEMM` 17,496 computational calls, 0 failures (DSYMM/DTRMM/DTRSM/DSYRK/DSYR2K all PASSED) — and HPL residual-PASSED.

## Notes

- The portable 4×4 tile is a safe default across the RVA20 scalar core family: it fits the guaranteed 32-register file and never worsens the load ratio, and is optimal on the dominant dual-issue in-order single-pipe profile.
- The residual gap between the FP-pipe peak and the streaming rate is memory latency feeding the packed panels through L2 (the JH7110 has no `Zicbop` software-prefetch), not the kernel schedule.
- All performance figures were measured on physical hardware. Happy to split this into smaller commits, rebase onto current `develop`, or drop the U74-specific assembly and keep only the portable C kernel if preferred.


hmeiland:
Update — addressed two things a reviewer would likely raise:

1. **CMake build support added.** `cmake -DTARGET=U74` now works: I mirrored the `RISCV64_GENERIC` blocks for U74 in `cmake/prebuild.cmake` (cache defines matching `getarch`), `cmake/system.cmake`, and `cmake/cc.cmake` (`-march=rv64imafdc_zba_zbb -mabi=lp64d -mtune=sifive-u74`). Validated on a VisionFive 2 — configure recognizes the target and a full `cmake --build` produces `libopenblas.a` with the U74 march/mtune flags applied to the kernels (0 errors), matching the Makefile path.

2. **Already current with `develop`.** The branch is based on the current `develop` HEAD (`7c99195`), so no rebase is needed and it merges cleanly.

Both the Makefile and CMake builds pass the full BLAS Level-3 test suite (DGEMM 17,496 calls, 0 failures) and HPL (residual PASSED) on physical U74 hardware.

martin-frbg:
Thank you. Can you please move the U74-specific gemm kernel (both gemmkernel_4x4_u74.c and kern_u74.S) to kernel/riscv64 rather than kernel/generic though ? It's wild enough that half the truly generic C kernels currently live in kernel/arm for "historic" reasons, having architecture-specific bits under "generic" seems a bit much :-)

I'll look into adding (Linux-specific) autodetection later, not sure why at least the U74 in the GCC Compile Farm identifies itself in /proc/cpuinfo via an "uarch" string rather than "model" but I haven't been following kernel patches that closely.

hmeiland:
Done in 5d7868f — moved both `gemmkernel_4x4_u74.c` and `kern_u74.S` to `kernel/riscv64/` (alongside `KERNEL.U74`) and pointed `DGEMMKERNEL` at `../riscv64/gemmkernel_4x4_u74.c`. The only other change is the moved kernel's `#include "conversion_macros.h"` → `"../generic/conversion_macros.h"`, since that header stays in `kernel/generic` (same form `kernel/wasm/gemmkernel_wasm128.c` already uses for it). The portable `gemmkernel_4x4.c` (SGEMMKERNEL) is untouched and stays under `kernel/generic`.

No functional change — `DGEMMKERNEL` still resolves through `$(KERNELDIR)`, so both the Makefile (`$(KERNELDIR)/$(DGEMMKERNEL)`) and CMake (`GenerateNamedObjects "${KERNELDIR}/${DGEMMKERNEL}"`) paths pick it up from the new location. Rebuilt `TARGET=U74` on riscv64 hardware to confirm: `dgemm_kernel.o` compiles from the new path and archives into `libopenblas` with 0 errors.

Thanks for taking this — and for looking into the Linux autodetection; happy to help test that on U74 hardware whenever you get to it.

PR #5904, opened 2026-07-10
Add a portable XERBLA handler API

moluopro:
After #5899 corrected the CTest registration, the Windows ARM64 shared-library job exposed failures in the `xerbla_*` extension tests: calls originating in the OpenBLAS DLL did not reach the test executable's `xerbla` override.

#5901 works around this Windows DLL symbol-resolution limitation by switching the CI job to a static build. This PR instead adds an explicit XERBLA handler API for shared-library users and tests, and prevents BLAS and CBLAS error-exit failures from passing silently.

## Changes

- Adds `openblas_set_xerbla`, with handler replacement and `NULL` reset support.
- Preserves the Fortran ABI and ELF strong-symbol interposition while normalizing routine-name lengths.
- Preserves callback types in prefixed and suffixed installs without changing existing Windows export ordinals.
- Uses explicit handlers where interposition is unavailable and propagates BLAS and CBLAS test failures through CTest and Make.

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -47,10 +47,7 @@ BLAS-like and conversion functions for `hfloat16` (available when OpenBLAS was c
 * `int openblas_set_affinity(int thread_index, size_t cpusetsize, cpu_set_t *cpuset)` sets the CPU affinity mask of the given thread
   to the provided cpuset. Only available on Linux, with semantics identical to `pthread_setaffinity_np`.
 * `openblas_set_thread_callback_function` overrides the default multithreading backend with the provided argument
-* `openblas_set_xerbla(openblas_xerbla_handler handler)` installs a process-wide XERBLA error handler and returns the
-  previous handler. Passing `NULL` restores the default handler. Handlers must be thread-safe because they may be called
-  concurrently. Replacing a handler is thread-safe but does not wait for in-progress calls to return, so the previous handler's
-  code must remain loaded until those calls finish. The `name` and `info` pointers are valid only during the callback. `name`
-  points to `name_length` valid bytes and need not be NUL-terminated; the length stops before the first NUL but otherwise includes
-  trailing spaces. Handler registration works normally on ELF platforms. For backward compatibility, an application-provided
-  strong `xerbla` symbol still overrides the OpenBLAS dispatcher and bypasses the registered handler.
+* `openblas_set_xerbla(openblas_xerbla_handler handler)` replaces the XERBLA handler for the current OpenBLAS
+  instance and returns the previous handler; passing `NULL` restores the default. Callbacks may be invoked concurrently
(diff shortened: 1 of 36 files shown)


docs/extensions.md
```diff
@@ -47,3 +47,9 @@ BLAS-like and conversion functions for `hfloat16` (available when OpenBLAS was c
 * `int openblas_set_affinity(int thread_index, size_t cpusetsize, cpu_set_t *cpuset)` sets the CPU affinity mask of the given thread
   to the provided cpuset. Only available on Linux, with semantics identical to `pthread_setaffinity_np`.
 * `openblas_set_thread_callback_function` overrides the default multithreading backend with the provided argument
+* `openblas_set_xerbla(openblas_xerbla_handler handler)` installs a process-wide XERBLA error handler and returns the previous handler.
+  Passing `NULL` restores the default handler. Handler replacement is thread-safe, but callbacks may run concurrently and must
+  therefore be thread-safe. The routine name and error information are valid only for the duration of the callback. The name is
+  valid for the supplied length, is not necessarily NUL-terminated, and retains any trailing spaces; when a NUL is present, the
+  reported length stops before the first NUL. Replacing a handler does not wait for callbacks already in progress. On ELF platforms,
+  a legacy strong `xerbla` definition takes precedence over this dispatcher.
```
imciner2: >  On ELF platforms, a legacy strong `xerbla` definition takes precedence over this dispatcher.

Does this mean that ELF platforms can't set an XERBLA handler? e.g., so this won't work on linux? If so, why is this limitation needed? It would be better IMHO to provide the same features on all platforms.

moluopro: `openblas_set_xerbla` works on Linux and other ELF platforms. Only an application-provided strong `xerbla` symbol takes precedence for backward compatibility. I’ll clarify this in the documentation.

martin-frbg:
While this looks like a very interesting addition, it still bothers me that it appears to be "only" the utest_ext job that fails to bind the matching xerbla, while the regular BLAS/CBLAS tests appear to manage just fine.

moluopro:
Good point. The regular BLAS/CBLAS tests had the same problem, but their failures were not propagated correctly.

I have prepared a fix and will push it shortly.

moluopro:
I've pushed changes addressing your concern.

martin-frbg:
Great, thank you very much. I guess it might make sense to upstream these changes to Reference-LAPACK eventually

moluopro:
Thanks for merging! I'll look into it.

imciner2:
>  I guess it might make sense to upstream these changes to Reference-LAPACK eventually

That could be good, and it would probably make more libraries implement the functionality. Right now I think only MKL and now OpenBLAS are the only two libraries that implement an XERBLA override.

(Not that you merged this, I'll add a method into the Julia LBT muxing library to support this as well).

PR #5907, opened 2026-07-12
Simplify `num_cpu_avail()` syntax

nh2:
* [ ] Based on #5808, merge that first.
* [ ] Get @martin-frbg clarification on comment from https://github.com/OpenMathLib/OpenBLAS/pull/5808#issuecomment-4943877382
  > so nothing to do with forcing OMP parallel jobs inside an OMP_PARALLEL region

### DIFF
--- a/cblas.h
+++ b/cblas.h
@@ -40,6 +40,39 @@ extern "C" {
 /*Set the number of threads on runtime.*/
 void openblas_set_num_threads(int num_threads);
 void goto_set_num_threads(int num_threads);
+/*
+ * Intent:
+ *   Sets the thread-local number of threads that OpenBLAS functions should request.
+ *   This thread-local number only affects the current execution thread
+ *   and, when set, overrides the global default.
+ *   This is useful for controlling nested parallelism.
+ *   For example when the caller's own (e.g. OpenMP) threads each invoke
+ *   OpenBLAS, `num_threads` determines how many threads OpenBLAS may use
+ *   per such call, instead of the process-wide default of
+ *   `omp_get_max_threads()`.
+ *
+ *   The intent of this was likely to match MKL's `mkl_set_num_threads_local()`
+ *   but implementation may not live up to the original intent yet.
+ *
+ * TODO:
+ *   Currently the implementation doesn't really do any of that:
+ *   * The setting is NOT thread-local.
+ *   * This function currently just calls `openblas_set_num_threads(num_threads)`,
+ *     changing the same process-wide thread count as that function,
+ *     and additionally records `num_threads` in an internal (global, not per-thread)
+ *     variable that is only consulted when OpenBLAS is invoked from within
+ *     a caller's OpenMP parallel region and the user has not made an
+ *     explicit thread-count choice.
+ *   * There is no special handling of `num_threads == 0`
+ *     (or other MKL-style reset values); the value is forwarded to
+ *     `openblas_set_num_threads()`, where values < 1 are treated
+ *     as "use the default" and large values are clamped to the number of cores.
+ *
+ * Returns the *previous* number of threads (as reported by
+ * `openblas_get_num_threads()`), so the caller can save and later restore it.
+ *
+ * On single-threaded OpenBLAS builds this is a no-op that returns 1.
+ */
 int openblas_set_num_threads_local(int num_threads);
 
 /*Get the number of threads on runtime.*/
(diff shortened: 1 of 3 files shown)


nh2:
@martin-frbg answering your https://github.com/OpenMathLib/OpenBLAS/pull/5808#issuecomment-4943877382

> Refactoring num_cpu_avail() _may_ be a good idea at some point, but it bears the risk of introducing new bugs and I have neither the time nor energy to review that now.

I added back 1 line I had originally removed, which simplifies the review.

The change is purely syntactical (beyond relying on `omp_get_max_threads()` having no side effect).

For review convenience:

# Step-by-step equivalence for `num_cpu_avail()` refactor

Shows that the change is purely syntactical (beyond relying on `omp_get_max_threads()` having no side effect).

## Starting point (old code)

```c
#ifdef USE_OPENMP
  /* ...unchanged prelude... */

int openmp_nthreads;
	openmp_nthreads=omp_get_max_threads();
	if (omp_in_parallel()) openmp_nthreads = blas_omp_threads_local;
#endif

#ifndef USE_OPENMP
  if (blas_cpu_number == 1
#else
     if (openmp_nthreads == 1
#endif
      ) return 1;

#ifdef USE_OPENMP
     if (openmp_nthreads > blas_omp_number_max){
     /* fprintf under #ifdef DEBUG */
     openmp_nthreads = blas_omp_number_max;
     }
     if (blas_cpu_number != openmp_nthreads) {
	  goto_set_num_threads(openmp_nthreads);
  }
#endif

  return blas_cpu_number;
```

To make review easier, **split into two independent cases** (one for `USE_OPENMP` defined, one not), then **rejoin** them at the end.

---

## Path A: `USE_OPENMP` is **NOT** defined

Applying preprocessor:

```c
  if (blas_cpu_number == 1
      ) return 1;

  return blas_cpu_number;
```

The early `return 1;` is redundant:

```c
  return blas_cpu_number;
```


## Path B: `USE_OPENMP` **IS** defined

Applying preprocessor:

```c
int openmp_nthreads;
	openmp_nthreads=omp_get_max_threads();
	if (omp_in_parallel()) openmp_nthreads = blas_omp_threads_local;

     if (openmp_nthreads == 1
      ) return 1;

     if (openmp_nthreads > blas_omp_number_max){
       /* fprintf under #ifdef DEBUG */
     openmp_nthreads = blas_omp_number_max;
     }
     if (blas_cpu_number != openmp_nthreads) {
	  goto_set_num_threads(openmp_nthreads);
  }

  return blas_cpu_number;
```

Turn first 3 lines into ternary; no change since `omp_get_max_threads()` has no side effect.
Also normalise whitespace:

```c
  int openmp_nthreads = omp_in_parallel() ? blas_omp_threads_local : omp_get_max_threads();

  if (openmp_nthreads == 1) return 1;

  if (openmp_nthreads > blas_omp_number_max) {
    /* fprintf under #ifdef DEBUG */
    openmp_nthreads = blas_omp_number_max;
  }
  if (blas_cpu_number != openmp_nthreads) {
    goto_set_num_threads(openmp_nthreads);
  }

  return blas_cpu_number;
```

## Rejoin

Joining the results of A and B:

```c
#ifdef USE_OPENMP
  /* ...unchanged prelude... */

  int openmp_nthreads = omp_in_parallel() ? blas_omp_threads_local : omp_get_max_threads();

  if (openmp_nthreads == 1) return 1;

  if (openmp_nthreads > blas_omp_number_max) {
    /* fprintf under #ifdef DEBUG */
    openmp_nthreads = blas_omp_number_max;
  }

  if (blas_cpu_number != openmp_nthreads) {
    goto_set_num_threads(openmp_nthreads);
  }
#endif

  return blas_cpu_number;
```

itamarst:
@nh2 https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2023-0/mkl-set-num-threads-local.html is I think the intent (not the reality) of what `openblas_set_num_threads_local` is supposed to do: instead of setting the number of threads for the whole process, it should set them for the current thread only. So it should work for non-OpenMP backends too, it's not an OpenMP thing. The implementation would involve thread locals.

This would be _extremely_ useful; I expect to write an article about why soon, but the short version is that if user code is creating its own threads, only having a process-wide setting is very frustrating.

nh2:
@itamarst Ah, OK. Well, th function really doesn't do any of that, not using thread-local variables at all.

I updated the comment with a write-up of that, so that outside contributors or users at least have an idea of what's going on.

PR #5913, opened 2026-07-14
Add transpose support for C in GEADD (fixes #4646)

### LINKED ISSUE #4646: GEADD is pointless without transposition
`?GEADD` has virtually no utility over `?AXPBY`, which is itself unoptimizable (i.e. the equivalent loops, optimized by a compiler, will perform at least as well in all cases).

Both `appleblas_?geadd` and `cublas?geam` support transposition, which is useful, in part because transposition is nontrivial to optimize.

`cblas_?omatcopy`, which is inspired by MKL, has transpose but does not support the generality of accumulation of the former.

It would be nice to have a GEAM-style routine in OpenBLAS.

aadityansha06:
Fixes #4646

Extends GEADD to support independent transposition of both A and C, matching the behavior of cuBLAS's geam and Apple's Accelerate geadd. Previously only A could be transposed.

- Add transc parameter across cblas.h, common_interface.h, common_level3.h, common_param.h
- Add transc handling to interface/geadd.c and interface/zgeadd.c
- Extend kernel/generic/geadd.c and kernel/generic/zgeadd.c with stride logic for transposed C
- Add transpose test coverage (hand-verified 2x2 cases and randomized large-matrix tests) for sgeadd, dgeadd, cgeadd, zgeadd

cc @martin-frbg

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -279,3 +279,8 @@ In chronological order:
 * Nathan Sircombe <nathan.sircombe@arm.com>
   * [2026-04-16] Add CPU ID for Neoverse V3
 hheei <hheei@users.noreply.github.com>
+
+* Aadityansha Verma <https://github.com/aadityansha06>
+  * [2026-07-14] Add independent transpose support for C in GEADD (sgeadd/dgeadd/cgeadd/zgeadd).
+
+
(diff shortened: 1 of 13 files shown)


aadityansha06:
The C910V test failure is the known timeout issue documented in PR #5900. 

- TEST 126/128 fork:safety consistently times out on C910V QEMU
- 104/105 tests pass successfully on all other platforms
- This is not related to the GEADD transpose changes

Awaiting either a re-run or PR #5900 merge to resolve.

martin-frbg:
yes, no forced push needed actually - I just can't spend all my time here...
C910 should be reasonable now, and Jenkins is currently borked due to some problems with the Docker nodes at OSUOSL

PR #5830, opened 2026-06-07
Enable RVV-optimized TRSM kernels for RISCV64_ZVL128B

Felix-Gong:
## Summary

- Enable RVV-optimized TRSM kernels (`trsm_kernel_{LN,LT,RN,RT}_rvv_v1.c`) for `RISCV64_ZVL128B` target
- These RVV kernel implementations already exist in the repository and are used by the `x280` target, but were not enabled for ZVL128B configuration
- Affects all four TRSM variants (S/D/C/Z) in both left/right and normal/transposed configurations

## Note on ZVL256B

ZVL256B is intentionally excluded from this change. Testing revealed that the RVV TRSM kernel produces incorrect results with VLEN=256 (e.g., 3x3 lower triangular solve returns wrong values). This needs to be investigated separately.

## Test plan

- [x] Built and tested on RISC-V hardware with RVV 1.0 (VLEN=128)
- [x] DTRSM correctness verified (n=2, n=100)
- [x] STRSM correctness verified
- [x] CI tests pass

PR #5928, opened 2026-07-18
riscv64: make RVV TRSM (_rvv_v1) kernels VLEN-agnostic

hmeiland:
### Problem

The `kernel/riscv64/trsm_kernel_{LN,LT,RN,RT}_rvv_v1.c` kernels tile packed-A along `M` by the **runtime** `VSETVL_MAX`, but the GEMM packing routine they interoperate with packs by the **compile-time** `GEMM_UNROLL_M`. When `VSETVL_MAX != GEMM_UNROLL_M`, the diagonal-block stride assumed by `solve()` no longer matches the packed layout, so the solve reads/writes the wrong blocks and the result is corrupted.

On **x280** the two happen to coincide (both 16), so the currently shipping target is unaffected — but the kernels are **not actually VLEN-agnostic**. Enabling them on any target where `VSETVL_MAX != GEMM_UNROLL_M` (e.g. `RISCV64_ZVL128B`, which packs at unroll 8) produces wrong results. This is exactly what surfaces in #5830, which enables these kernels for ZVL128B.

### Fix

Replace only the **outer control flow** so packed-A is tiled by `GEMM_UNROLL_M` (with a power-of-2 `M` remainder per `N`-block), matching the generic `kernel/generic/trsm_kernel_{LN,LT,RN,RT}.c` drivers. The RVV-vectorized `solve()` is retained unchanged. This makes the kernels correct on both unroll-8 (ZVL128B) and unroll-16 (x280) targets.

### Validation (SpaceMiT X60, VLEN=256, GCC 14.3.0)

Residual test `‖op(A)·X − B‖` / `‖X·op(A) − B‖` over `m,n ∈ {1,2,4,7,8,9,10,12,15,23}` × every Side/Uplo/Trans/Diag.

**Built as `RISCV64_ZVL128B` (unroll 8), TRSM wired to `_rvv_v1`:**

| Routine | Cases | Fails | Worst residual |
|---|---|---|---|
| STRSM | 1600 | 0 | 1.4e-7 |
| DTRSM | 1600 | 0 | 3.3e-16 |
| CTRSM (incl. ConjTrans) | 2400 | 0 | 3.0e-7 |
| ZTRSM (incl. ConjTrans) | 2400 | 0 | 5.5e-16 |

**Regression contrast, same harness, same target:**

| Kernel | DTRSM cases | Fails | Worst residual |
|---|---|---|---|
| Current `_rvv_v1` (`VSETVL_MAX` tiling) | 1600 | **408** | **0.89** |
| This PR (`GEMM_UNROLL_M` tiling) | 1600 | **0** | 3.3e-16 |

An unroll-16 (ZVL256B-params) build is being validated on the same hardware; I'll post the result as a follow-up. Since x280 already uses these kernels at unroll 16, the change is expected to be a no-op there (16 == 16) while hardening against future VLEN/UNROLL divergence.

### Notes

- Marked **draft** pending the unroll-16 confirmation run.
- This overlaps with the goal of #5830 (which enables `_rvv_v1` for ZVL128B). The fix here is what makes that enablement correct; happy to coordinate.

<sub>Validated on SpaceMiT X60 hardware.</sub>

### DIFF
--- a/kernel/riscv64/trsm_kernel_LN_rvv_v1.c
+++ b/kernel/riscv64/trsm_kernel_LN_rvv_v1.c
@@ -71,6 +71,30 @@ static FLOAT dm1 = -1.;
 #define GEMM_KERNEL   GEMM_KERNEL_N
 #endif
 
+#if GEMM_DEFAULT_UNROLL_M == 1
+#define GEMM_UNROLL_M_SHIFT 0
+#endif
+
+#if GEMM_DEFAULT_UNROLL_M == 2
+#define GEMM_UNROLL_M_SHIFT 1
+#endif
+
+#if GEMM_DEFAULT_UNROLL_M == 4
+#define GEMM_UNROLL_M_SHIFT 2
+#endif
+
+#if GEMM_DEFAULT_UNROLL_M == 6
+#define GEMM_UNROLL_M_SHIFT 2
+#endif
+
+#if GEMM_DEFAULT_UNROLL_M == 8
+#define GEMM_UNROLL_M_SHIFT 3
+#endif
+
+#if GEMM_DEFAULT_UNROLL_M == 16
+#define GEMM_UNROLL_M_SHIFT 4
+#endif
+
 #if GEMM_DEFAULT_UNROLL_N == 1
 #define GEMM_UNROLL_N_SHIFT 0
 #endif
@@ -91,7 +115,9 @@ static FLOAT dm1 = -1.;
 #define GEMM_UNROLL_N_SHIFT 4
 #endif
 
-// Optimizes the implementation in ../arm64/trsm_kernel_LN_sve.c
+// Packed A is tiled by GEMM_UNROLL_M (the GEMM micro-kernel / itcopy contract),
+// NOT by runtime VL: tiling by VSETVL_MAX only matches x280 and silently
+// corrupts on other targets (e.g. zvl128b, unroll 8). Only solve() vectorizes.
 
 #ifndef COMPLEX
 
@@ -217,10 +243,6 @@ int CNAME(BLASLONG m, BLASLONG n, BLASLONG k,  FLOAT dummy1,
   BLASLONG i, j;
   FLOAT *aa, *cc;
   BLASLONG  kk;
-  
-  size_t vl = VSETVL_MAX;
-
-    //fprintf(stderr, "%s , %s, m = %4ld  n = %4ld  k = %4ld offset = %4ld\n", __FILE__, __FUNCTION__, m, n, k, offset); // Debug
 
   j = (n >> GEMM_UNROLL_N_SHIFT);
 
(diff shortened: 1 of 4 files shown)


hmeiland:
### ✅ Unroll-16 (ZVL256B) confirmation — VLEN-agnostic proven

Follow-up to the note in the PR description. I built the same fix as **`RISCV64_ZVL256B`** (`GEMM_UNROLL_M = 16`, `-march=rv64imafdcv_zvl256b`) on the SpaceMiT X60 and re-ran the full residual harness. Same test matrix as before (`m,n ∈ {1,2,4,7,8,9,10,12,15,23}` × every Side/Uplo/Trans/Diag).

| Routine | Cases | Fails | Worst residual |
|---|---|---|---|
| STRSM | 1600 | 0 | 1.8e-7 |
| DTRSM | 1600 | 0 | 3.3e-16 |
| CTRSM (incl. ConjTrans) | 2400 | 0 | 3.6e-7 |
| ZTRSM (incl. ConjTrans) | 2400 | 0 | 6.7e-16 |

Combined with the ZVL128B (unroll 8) results in the description, this confirms the kernels are correct at **both** `GEMM_UNROLL_M = 8` and `GEMM_UNROLL_M = 16` — i.e. genuinely VLEN/UNROLL-agnostic rather than relying on the `VSETVL_MAX == GEMM_UNROLL_M` coincidence. Since x280 already uses these kernels at unroll 16, this is the expected no-op there while removing the latent divergence hazard.

Marking ready for review.

<sub>Validated on SpaceMiT X60 hardware, GCC 14.3.0.</sub>

hmeiland:
## Performance check — no regression (SpaceMiT X60, RVV 1.0, VLEN=256)

I benchmarked the fixed kernels against the original `_rvv_v1` kernels on hardware to confirm the VLEN-agnostic rewrite does not cost throughput. Harness is `benchmark/trsm.c`, single-thread, `L/U/N/N`, `OPENBLAS_LOOPS=20`, best-of-3 runs on an 8-core X60 @ EESSI GCC 14.3.0, both libs built `ZVL128B` (`GEMM_UNROLL_M=8`).

Δ = fixed vs original, MFlops:

| size | STRSM | DTRSM | CTRSM | ZTRSM |
|---|--:|--:|--:|--:|
| 8   | +4.3%  | +4.3% | −3.0%  | +0.3% |
| 16  | +32.6% | +0.2% | +31.9% | +2.6% |
| 32  | +40.9% | +0.4% | +29.2% | +1.6% |
| 64  | +35.6% | +1.4% | +21.4% | +1.5% |
| 128 | +26.0% | ≈0%   | +8.6%  | −2.6% |
| 256 | +17.3% | −3.9% | +6.6%  | −4.2% |
| 400 | +11.5% | −6.9% | −0.8%  | −1.5% |

- **STRSM / CTRSM**: consistent speedups (+10–40%), largest at small/mid sizes — `GEMM_UNROLL_M` tiling reuses the packed diagonal block better than the per-call `VSETVL_MAX` retiling.
- **DTRSM / ZTRSM**: within measurement noise (±~4%) on this loaded shared board. The scattered small negatives at large sizes sit inside the observed run-to-run variance — a 6-rep re-measure at DTRSM m=128 gave overlapping distributions (orig 1273–1412, fixed 1257–1371; equal medians), so the earlier single-run negatives there were jitter, not a systematic slowdown.

**Important caveat:** on `ZVL128B` the original `_rvv_v1` kernels are *numerically broken* (this is the bug the PR fixes — e.g. 408/1600 DTRSM cases fail, worst residual ~0.89), so their timings are not a legitimate baseline. The fix makes `ZVL128B` correct for the first time while being at least as fast, and usually faster, than the broken code it replaces. On `ZVL256B` / x280 (`VSETVL_MAX == GEMM_UNROLL_M == 16`) the tile width is unchanged, so the rewrite is a structural no-op for throughput there.

Correctness: `8000/0` pass on both `ZVL128B` (unroll 8) and `ZVL256B` (unroll 16) across S/D/C/Z.

PR #5944, opened 2026-07-26
Fix build error in zrotg.c with FUNCTION_PROFILE=1 from Makefile.rule

lujiaweics:
When FUNCTION_PROFILE is enabled in Makefile.rule, zrotg.c fails to compile because FUNCTION_PROFILE_START introduces an unmatched opening brace. This PR inserts FUNCTION_PROFILE_END and IDEBUG_END at the single exit point behind a do-while(0) wrapper.


martin-frbg:
Thank you - this was probably broken in my effort to realign ROTG with changes in the reference implementation a few years ago.

PR #5955, opened 2026-08-03
Fix RISC-V GEMM/GEMV forwarding in Make builds

moluopro:


moluopro:
Unrelated to this change, but I’ll look into the pipeline issue—I like to keep CI green.

martin-frbg:
Ouch, thanks. Weird that this appears to have been broken for two years, guess I need to step up my local riscv64 use.

moluopro:
This bug won’t actually cause any compile or runtime errors because there’s a fallback path. I’ve also fixed another pipeline issue affecting this PR in 5956.

PR #5956, opened 2026-08-04
Fix stale workspace pointers in reused thread queues

moluopro:
- Prevent thread queues from reusing stale workspace pointers.
- Apply the fix to the pthread, OpenMP, and Windows backends, with callback regression coverage.

### DIFF
--- a/common_thread.h
+++ b/common_thread.h
@@ -106,7 +106,15 @@ typedef struct blas_queue {
   blas_arg_t *args;
   void *range_m;
   void *range_n;
+
+  /*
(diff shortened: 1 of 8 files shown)


moluopro:
Reused thread queues may retain stale workspaces, causing intermittent computation errors and flaky CI failures.

martin-frbg:
Oops, would it be that simple ? This problem has been noticed already in #1851 (at least), and forced reverting of #4741 as I didn't manage to get my head around the buffer reuse happening in both level2 and level3 code

PR #5957, opened 2026-08-04
Fix pointer comparison in laed3.c threading threshold

uuiitwp:
## Bug: Pointer comparison instead of value comparison in `laed3.c` threading threshold

**File:** `interface/lapack/laed3.c`

**Line:** 76

**Current code:**
```c
if (n >= 64) nthreads = num_cpu_avail(4);
```

**Expected:**
```c
if (nval >= 64) nthreads = num_cpu_avail(4);
```


martin-frbg:
Thank you (I swear I already had this in my queue...)

PR #5961, opened 2026-08-05
Fix issues with the LoongArch & macOS pipeline

moluopro:
- Replace the existing LoongArch QEMU setup with Ubuntu-packaged QEMU 10.2.1.
- Increase the timeout to 100 minutes for slow Azure macOS jobs.

### DIFF
--- a/.github/workflows/loongarch64.yml
+++ b/.github/workflows/loongarch64.yml
@@ -51,7 +51,7 @@ jobs:
       - name: Install APT deps
         run: |
           sudo apt-get update && \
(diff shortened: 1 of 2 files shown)


moluopro:
The previous change did not actually replace the QEMU binary used by the tests. The lowercase `wget -o` option sets the log file, while the tests continued to run `qemu-loongarch64-static` from QEMU 8.2.2.

martin-frbg:
Ahh, frog eyes ...
And it looks like the mips job wants the same treatment

Issue #5958, opened 2026-08-04
KMeans segfaults on 344-logical-core box: concurrent single-threaded callers exhaust NUM_BUFFERS(256) even with threads pinned to 1; exhaustion path returns NULL instead of terminating


cakedev0:
Hi, I'm working on scikit-learn and testing/benchmarking it on big machines.

I used Claude to help investigate this and draft the report, I tried to make sure it was relevant and to make it digest.

## Summary

`sklearn.cluster.KMeans.fit()` reliably crashes (segfault or `corrupted size vs. prev_size`) on a 344-logical-core machine when linked against OpenBLAS's **pthreads** threading layer. scikit-learn is well aware of BLAS-oversubscription hazards and explicitly guards its Lloyd loop with `threadpoolctl`, forcing `openblas_get_num_threads() == 1` for the whole call. I verified that guard is genuinely active at the moment of the crash. It doesn't help, because the actual trigger doesn't involve OpenBLAS recruiting extra threads per call at all — it's simpler than that:

sklearn's own outer parallelism (one OpenMP worker per logical core — 344 here — in its `prange` distance-computation loop) means 344 threads hit `blas_memory_alloc()` at roughly the same time (OpenMP's fork/join barrier), each wanting one buffer for its own **single-threaded** BLAS call. `NUM_BUFFERS` (the size of the `memory[]` allocator table) is 256 on this build. 344 concurrent callers against 256 slots exhausts the table regardless of how many threads any individual call is allowed to use.

## Minimal reproducer

```c
/* ob_repro.c — build: gcc -fopenmp -O2 ob_repro.c -o ob_repro -l:libopenblas.so.0
 * run:  ./ob_repro <n_threads> <n_iters> <matrix_size>   e.g. ./ob_repro 344 50 128
 *
 * openblas_set_num_threads(1) mirrors scikit-learn's own
 * threadpool_limits(limits=1, user_api="blas") guard around its Lloyd loop.
 * Crashes anyway: the hazard is *how many threads call in*, not how many
 * threads OpenBLAS recruits per call. */
#include <stdio.h>
#include <stdlib.h>

extern void dgemm_(const char *transa, const char *transb,
                    const int *m, const int *n, const int *k,
                    const double *alpha, const double *a, const int *lda,
                    const double *b, const int *ldb,
                    const double *beta, double *c, const int *ldc);
extern void openblas_set_num_threads(int);

int main(int argc, char **argv) {
[...]

PR #5963, opened 2026-08-05
[WIP] Actually terminate OpenBLAS when running out of thread buffers

martin-frbg:
for #5958 - allowing program termination to be overriden by a user-supplied xerbla

### DIFF
--- a/driver/others/memory.c
+++ b/driver/others/memory.c
@@ -2765,7 +2765,7 @@ static int memory_overflowed = 0;
 void *blas_memory_alloc(int procpos){
 
   int i;
-
+  int buffer_fault = 999;
   int position;
 #if defined(WHEREAMI) && !defined(USE_OPENMP)
   int mypos = 0;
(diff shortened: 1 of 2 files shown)


imciner2:
I'm not a large fan of putting terminating statements inside library error handlers. It can lead to killing higher-level sessions, such as terminating the Julia REPL process instead of just throwing back an error to the user.

Is there no way to propagate the failure condition back through the function chain?

martin-frbg:
Yes, I don't like it much either. Probably the better alternative is to return a xerbla error on the individual BLAS call that failed, and hope that the caller heeds it. (Bad enough that there is no other error return mechanism in the standard, and even my makeshift error code 999 clashes will all those LAPACK routines that return something like the index of the first zero or otherwise improper diagonal element of a matrix)

martin-frbg:
Try 2 is in #5956 - return a XERBLA code -999 on the individual BLAS call without terminating the process, leaving it up to the caller if and how they handle the failed call. (Could add a wait-and-retry, but I think it would be unlikely to help in such cases of catastrophic oversubscription)

PR #5965, opened 2026-08-07
Return a specific XERBLA error to the caller when buffer assignment fails

### LINKED ISSUE #5958: KMeans segfaults on 344-logical-core box: concurrent single-threaded callers exhaust NUM_BUFFERS(256) even with threads pinned to 1; exhaustion path returns NULL instead of terminating

martin-frbg:
fixes #5958 without terminating the entire process

### DIFF
--- a/driver/level3/gemm_batch_thread.c
+++ b/driver/level3/gemm_batch_thread.c
@@ -88,6 +88,8 @@ int CNAME(blas_arg_t * args_array, BLASLONG nums){
 #define ERROR_NAME "QGEMM_BATCH "
 #elif defined(DOUBLE)
 #define ERROR_NAME "DGEMM_BATCH "
+#elif defined(BFLOAT16)
+#define ERROR_NAME "SBGEMM_BATCH "
 #else
 #define ERROR_NAME "SGEMM_BATCH "
 #endif
@@ -112,7 +114,7 @@ int CNAME(blas_arg_t * args_array, BLASLONG nums){
   if (!buffer) {
     info = -999;
     BLASFUNC(xerbla)(ERROR_NAME, &info, sizeof(ERROR_NAME));
-    return;
+    return(1); 
   }
 
   sa = (XFLOAT *)((BLASLONG)buffer +GEMM_OFFSET_A);
@@ -144,6 +146,8 @@ int CNAME(blas_arg_t * args_array, BLASLONG nums){
     queue=(blas_queue_t *)malloc((nums+1) * sizeof(blas_queue_t));
     if(queue == NULL){
       openblas_warning(0, "memory alloc failed!\n");
+      if (buffer) blas_memory_free(buffer);
+      BLASFUNC(xerbla)(ERROR_NAME, &info, sizeof(ERROR_NAME));
       return(1);
     }
     for(i=0; i<nums; i++){
--- a/driver/level3/gemm_batch_thread.c
+++ b/driver/level3/gemm_batch_thread.c
@@ -83,6 +83,25 @@ int CNAME(blas_arg_t * args_array, BLASLONG nums){
   int (*routine)(blas_arg_t *, void *, void *, XFLOAT *, XFLOAT *, BLASLONG);
   int i=0, /*j,*/ current_nums;
 
+#ifndef COMPLEX
+#ifdef XDOUBLE
+#define ERROR_NAME "QGEMM_BATCH "
+#elif defined(DOUBLE)
+#define ERROR_NAME "DGEMM_BATCH "
+#else
+#define ERROR_NAME "SGEMM_BATCH "
+#endif
+#else
+#ifdef XDOUBLE
+#define ERROR_NAME "XGEMM_BATCH "
+#elif defined(DOUBLE)
+#define ERROR_NAME "ZGEMM_BATCH "
+#else
+#define ERROR_NAME "CGEMM_BATCH "
+#endif
+#endif
+
+
 #ifdef SMP
   blas_queue_t * queue=NULL;
 #endif
@@ -90,6 +109,12 @@ int CNAME(blas_arg_t * args_array, BLASLONG nums){
   if(nums <=0 ) return 0;
 
   buffer = (XFLOAT *)blas_memory_alloc(0);
+  if (!buffer) {
+    info = -999;
+    BLASFUNC(xerbla)(ERROR_NAME, &info, sizeof(ERROR_NAME));
+    return;
+  }
+
   sa = (XFLOAT *)((BLASLONG)buffer +GEMM_OFFSET_A);
   sb = (XFLOAT *)(((BLASLONG)sa + ((GEMM_P * GEMM_Q * COMPSIZE * SIZE + GEMM_ALIGN) & ~GEMM_ALIGN)) + GEMM_OFFSET_B);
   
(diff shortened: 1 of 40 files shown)


imciner2:
Thanks for this! I know it's always hard to work on these old APIs, but this is certainly nicer for us so we don't kill user sessions.

PR #5966, opened 2026-08-08
Fix workspace size in ReLAPACK sytrf/hetrf

Ka-zam:
The `?sytrf`/`?hetrf` routines size their workspace as `n*(n/2)` and report that from the `lWork = -1` query, but the top level recursion addresses `Work` as an `n` by `xREC_SPLIT(n)` matrix with leading dimension `n`. `xREC_SPLIT` rounds up, so whenever `xREC_SPLIT(n) > n/2` the routine writes past the end of the buffer it asked for — silently wrong factorization and a corrupted heap.

For `zsytrf`, `ZREC_SPLIT(500) = 252` against the 250 columns the buffer holds.

Repro: build with `BUILD_RELAPACK=1 RELAPACK_REPLACE=1`, query `lWork`, allocate exactly that, call `zsytrf`. Before this patch 114 of 241 sizes in `[64,1024]` return garbage; sweeping every `n` in `[100,600]` aborts on heap corruption. After, a workspace-query/factorize/solve residual check over 12 routines x every `n` in `[64,400]` passes 4044/4044.

Sizes the workspace with the same `xREC_SPLIT` the recursion uses. Affects all 12 variants.

### DIFF
--- a/relapack/src/chetrf.c
+++ b/relapack/src/chetrf.c
@@ -19,8 +19,8 @@ void RELAPACK_chetrf(
     float *Work, const blasint *lWork, blasint *info
 ) {
 
-    // Required work size
-    const blasint cleanlWork = *n * (*n / 2);
+    // Required work size (CREC_SPLIT rounds up)
+    const blasint cleanlWork = *n * CREC_SPLIT(*n);
     blasint minlWork = cleanlWork;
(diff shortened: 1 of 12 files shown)


martin-frbg:
Great catch, thanks.

PR #5980, opened 2026-08-15
Fix: Small matrix `gemm_batch` with DYNAMIC_ARCH

ajz34:
Hi devs!

This PR will fix segfault, when batched gemm on small matrices with DYNAMIC_ARCH compiled OpenBLAS.

DYNAMIC_ARCH seems to require `SMALL_KERNEL_ADDR` to resolve dispatched small kernel.
For non-DYNAMIC_ARCH, since there is only one table and not required to be resolved, so the current code works.

------

This was found when I'm developing rust's numpy-like toolkit [RSTSR](https://github.com/RESTGroup/rstsr), when trying to use batched gemm for broadcasted matrix-multiplication implementation with OpenBLAS backend.
This bug was originally found in AI code agent session (Claude Code with model GLM-5.3).

### DIFF
--- a/interface/gemm_batch.c
+++ b/interface/gemm_batch.c
@@ -353,18 +353,18 @@ void CNAME(enum CBLAS_ORDER order, enum CBLAS_TRANSPOSE *  transa_array, enum CB
 #if !defined(COMPLEX)
       if(*(FLOAT *)(group_beta) == 0.0){
 	group_mode=mode | BLAS_SMALL_B0_OPT;
-	group_small_matrix_opt_routine=(void *)(gemm_small_kernel_b0[(group_transb<<2)|group_transa]);
+	group_small_matrix_opt_routine=SMALL_KERNEL_ADDR(gemm_small_kernel_b0, ((group_transb<<2)|group_transa));
       }else{
 	group_mode=mode | BLAS_SMALL_OPT;
-	group_small_matrix_opt_routine=(void *)(gemm_small_kernel[(group_transb<<2)|group_transa]);
+	group_small_matrix_opt_routine=SMALL_KERNEL_ADDR(gemm_small_kernel, ((group_transb<<2)|group_transa));
       }
 #else
       if(((FLOAT *)(group_beta))[0] == 0.0 && ((FLOAT *)(group_beta))[1] == 0.0){
 	group_mode=mode | BLAS_SMALL_B0_OPT;
-	group_small_matrix_opt_routine=(void *)(zgemm_small_kernel_b0[(group_transb<<2)|group_transa]);
+	group_small_matrix_opt_routine=SMALL_KERNEL_ADDR(zgemm_small_kernel_b0, ((group_transb<<2)|group_transa));
       }else{
 	group_mode=mode | BLAS_SMALL_OPT;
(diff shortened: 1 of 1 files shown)


ajz34:
This comment presents the reproduction of MWE.

<details>
<summary>
The C code to reproduce the MWE.
</summary>

```C
/*
 * MWE: grouped batched GEMM segfaults on small matrices in DYNAMIC_ARCH builds.
 *
 * Build: gcc bug2_dynarch_small_matrix_batch.c -o bug2 \
 *            -I<openblas>/build/generated -I<openblas>/build \
 *            -Wl,-rpath,<openblas>/build/lib -L<openblas>/build/lib -lopenblas
 *
 * Pass criterion: the call RETURNS (no SIGSEGV). Printed values are
 * informational only.
 *
 * Co-authored-by: Claude Code <noreply@anthropic.com>
 * Co-authored-by: glm-5.3 <service@zhipuai.cn>
 */
#include <stdio.h>
#include <cblas.h>

/* m*n*k <= 100^3 routes the batch through the SMALL_MATRIX_OPT kernels,
 * which is where the bug lives; sizes are otherwise arbitrary. */
#define M 8
#define N 12
#define K 16
#define COUNT 3 /* matrices, all in one group */

int main(void)
{
	setvbuf(stdout, NULL, _IONBF, 0); /* survive being killed mid-print */

	printf("openblas: %s\n", openblas_get_config());

	/* one col-major matrix per batch entry, fully in bounds */
	double a[COUNT][M * K], b[COUNT][K * N], c[COUNT][M * N];
	const double *ap[COUNT], *bp[COUNT];
	double *cp[COUNT];

	for (int t = 0; t < COUNT; t++) {
		for (int i = 0; i < M * K; i++)
			a[t][i] = i % 7 + 1;
		for (int i = 0; i < K * N; i++)
			b[t][i] = i % 5 + 1;
		for (int i = 0; i < M * N; i++)
			c[t][i] = 0.0;
		ap[t] = a[t];
		bp[t] = b[t];
		cp[t] = c[t];
	}

	/* single group of COUNT identical-shaped problems */
	CBLAS_TRANSPOSE ta[1] = { CblasNoTrans }, tb[1] = { CblasNoTrans };
	blasint m[1] = { M }, n[1] = { N }, k[1] = { K };
	blasint lda[1] = { M }, ldb[1] = { K }, ldc[1] = { M };
	blasint group_count = 1, group_size[1] = { COUNT };
	double alpha[1] = { 1.0 }, beta[1] = { 0.0 };

	cblas_dgemm_batch(CblasColMajor, ta, tb, m, n, k, alpha, ap, lda, bp,
			  ldb, beta, cp, ldc, group_count, group_size);

	for (int t = 0; t < COUNT; t++)
		printf("matrix %d: c[0]=%.1f\n", t, c[t][0]);
	printf("returned normally (no segfault)\n");
	return 0;
}
```

</details>

The code before PR will give Segmentation fault (exit=139).

The correct behavior should be something like
```
openblas: OpenBLAS 0.3.34.dev DYNAMIC_ARCH NO_AFFINITY USE_OPENMP Cooperlake MAX_THREADS=16
matrix 0: c[0]=168.0
matrix 1: c[0]=168.0
matrix 2: c[0]=168.0
```

martin-frbg:
Indeed, yes, thanks.

PR #5983, opened 2026-08-19
Add a 4x4 WASM SIMD128 GEMM microkernel

jjerphan:
## Summary

Follow-up to #5680 / #4023: replace the 2x2 shuffle GEMM for `WASM128_GENERIC` with a 4x4 outer-product kernel so `f32x4` / `f64x2` loads match packing width.

- SGEMM/DGEMM use `gemmkernel_4x4_wasm128.c` with matching `gemm_{n,t}copy_4` and generic TRMM 4x4.
- `KERNEL` TRMM defaults are wrapped in `ifndef` so `KERNEL.WASM128_GENERIC` packing width is not overwritten.
- `-mrelaxed-simd` plus `relaxed_madd` in the GEMM inner loop only; DOT/AXPY stay IEEE mul+add in `intrin_wasm.h`.
- Separate commit gates leftover WASM TRSM call counters behind `OPENBLAS_WASM_TRSM_PROFILE`.

## Benchmarks

Node / Emscripten, `TARGET=WASM128_GENERIC`, `USE_THREAD=0`, `COMMON_OPT=-O2`. Same machine, 5 warmup + 10 timed samples (median). Speedup > 1 means this branch is faster than `develop`.

Geomean vs `develop`: **SGEMM 1.44x**, **DGEMM 1.82x**, SSYRK 1.46x, DSYRK 1.73x, STRSM 1.33x, DTRSM 1.39x. Level-1/2 kernels are unchanged within noise.

| op | n | develop MFLOPS | 4x4 MFLOPS | speedup |
|---|---:|---:|---:|---:|
| sgemm | 32 | 14141.8 | 27612.0 | 1.953 |
| sgemm | 64 | 18077.5 | 26887.8 | 1.487 |
| sgemm | 128 | 20662.3 | 25570.7 | 1.238 |
| sgemm | 256 | 20922.9 | 28427.7 | 1.359 |
| sgemm | 512 | 22546.2 | 28625.6 | 1.270 |
| dgemm | 32 | 9728.9 | 17933.1 | 1.843 |
| dgemm | 64 | 10937.9 | 19025.1 | 1.739 |
| dgemm | 128 | 11355.1 | 20954.0 | 1.845 |
| dgemm | 256 | 11948.5 | 21770.6 | 1.822 |
| dgemm | 512 | 11893.9 | 22016.5 | 1.851 |
| ssyrk | 32 | 9938.9 | 17646.4 | 1.775 |
[...]

PR #5996, opened 2026-08-25
wasm: guard remaining KERNEL assignments so KERNEL.$(TARGET_CORE) overrides survive

### LINKED ISSUE #5993: guards missing in wasm/KERNEL file?
Hi all,

I am currently exploring SIMD usage via OpenBLAS enabled numpy in WASM. I am working on improving the performance of the binaries provided in [emscripten-forge](https://github.com/emscripten-forge). I specifically looked at ddot/sdot triggered by np.correlate usage. I was a bit surprised not to see more performance improvements through OpenBLAS [compiled with simd flag to WASM](https://github.com/emscripten-forge/recipes/blob/main/recipes/recipes_emscripten/openblas/build.sh). To explore deeper, I took [the produced binary](https://prefix.dev/channels/emscripten-forge-4x/packages/openblas) and did an object dump:

```
SO=lib/libopenblas_wasm128-r0.3.34.so
file $SO                          
# WebAssembly (wasm) binary version 0x1 (MVP module)

# 2. Confirm the shipped library's compile-time config (single command)
nix shell nixpkgs#binutils -c strings $SO | grep -E '^OpenBLAS 0\.'
# OpenBLAS 0.3.34 NO_AFFINITY WASM128_GENERIC

# 3. Locate the DOT / AXPY kernel symbols
nix shell nixpkgs#wabt -c wasm-objdump -x $SO | grep -E ' <(ddot_k|daxpy_k)>' | head
# - func[1480] sig=83 <daxpy_k>
# - func[1482] sig=91 <ddot_k>
# - func[1480] <daxpy_k> -> "daxpy_k"
# - func[1482] <ddot_k> -> "ddot_k"
# - func[1480] size=365 <daxpy_k>
# - func[1482] size=198 <ddot_k>

# 4. Full disassembly + relevant instructions
nix shell nixpkgs#wabt -c wasm-objdump -d $SO > /tmp/openblas.dis
for name in ddot_k daxpy_k sdsdot_k; do
  s=$(grep -n " <${name}>:" /tmp/openblas.dis | head -1 | cut -d: -f1)
  e=$(awk -v s=$((s+1)) 'NR>=s && /func\[[0-9]+\] </ {print NR; exit}' /tmp/openblas.dis)
  echo "=== ${name} (bytes: $((e-s)) lines) ==="
  sed -n "${s},${e}p" /tmp/openblas.dis \
    | grep -oE '\|[[:space:]]+[a-z0-9._]+' | awk '{print $2}' \
    | grep -E '^(f32|f64|v128|f[36][24]x[24]|i[0-9]+x[0-9]+)\.' | sort | uniq -c | sort -rn
done

#=== ddot_k (bytes: 110 lines) ===
#      6 f64.load
#      3 f64.mul
#      3 f64.add
#      1 f64.const
#=== daxpy_k (bytes: 197 lines) ===
#      6 f64.load
#      3 f64.store
#      3 f64.mul
#      3 f64.add
#      2 v128.load
#      1 v128.store
#      1 f64x2.splat
#      1 f64x2.mul
#      1 f64x2.add
#      1 f64.eq
#      1 f64.const
#=== sdsdot_k (bytes: 423 lines) ===
#     20 f64x2.promote_low_f32x4
#     16 f64x2.add
#     12 f64.promote_f32
#     12 f32.load
#     10 v128.load
#     10 i8x16.shuffle
#     10 f64x2.mul
#      9 f64.add
#      6 f64.mul
#      4 f64x2.extract_lane
#      1 f64.const
```

ddot, contrary to other functions, doesn't have any v128 instructions, indicating it's not SIMD accelerated and pure scalar. This is weird on the other hand because it seems that the CPU-variant kernel file https://github.com/OpenMathLib/OpenBLAS/blob/develop/kernel/wasm/KERNEL.WASM128_GENERIC#L58-L59 does point to a SIMD accelerated path https://github.com/OpenMathLib/OpenBLAS/blob/develop/kernel/generic/dot.c#L50 .

Exploring this further it seems that the generic KERNEL file in the wasm folder: https://github.com/OpenMathLib/OpenBLAS/blob/develop/kernel/wasm/KERNEL#L62-L63 points to RISCV which is *not* SIMD accelerated https://github.com/OpenMathLib/OpenBLAS/blob/develop/kernel/riscv64/dot.c .

Now, exploring further, it seems that, unexpectedly the generic KERNEL file *overrides* the specific KERNEL.WASM128_GENERIC one, a generic pattern throughout the codebase: https://github.com/OpenMathLib/OpenBLAS/blob/develop/kernel/Makefile#L119-L121

To get the correct behavior defined in the *specific* KERNEL.WASM128_GENERIC file, you would thus need to guard every line in the generic KERNEL file. And indeed, this seems to be done, for example,  in the case of arm: https://github.com/OpenMathLib/OpenBLAS/blob/develop/kernel/arm/KERNEL .

However, for WASM this seems not to be done for many kernels. So **my questions:**

- Is the above story correct? I am unsure since this is really my first interaction with the OpenBLAS codebase.
- **Should we guard everything in the wasm/KERNEL file to enable SIMD accelerated kernels for dot in WASM**? Was this simply forgotten or is there a deeper reason?
- Shouldn't this happen for other architectures as well? Most seem guarded appropriately but not all of them.
- What's the reason for not simply reversing import order in general and safe the guards?

MMesch:
Fixes #5993. Base `kernel/wasm/KERNEL` is included after `KERNEL.$(TARGET_CORE)` (`kernel/Makefile:119-121`), so unguarded `=` assignments in the base file clobber per-target overrides. #5984 (AXPY), #5983 (GEMM/TRMM), and #5990 (GEMV) fixed this family-by-family; this PR wraps every remaining assignment in `ifndef` so future WASM sub-targets can override any kernel without a matching guard edit here.

On `develop`, the still-unguarded families this PR moves are SDOT / DDOT / SROT / DROT. SIMD-opcode counts in `libopenblas.so` for these kernels (measured against the `v0.3.34` tag, since that predates the prior guard PRs):

| kernel  | simd_before | simd_after |
|---------|------------:|-----------:|
| sdot_k  |           0 |         27 |
| ddot_k  |           0 |         25 |
| srot_k  |          12 |         74 |
| drot_k  |          12 |         66 |

Downstream (emscripten-forge/recipes#6343): 1.5-2x speedup on `numpy.correlate` under Node/Emscripten (`TARGET=WASM128_GENERIC`, `USE_THREAD=0`).

Tested with `utest` and CBLAS `ctest` L1/L2/L3 under Node.

### DIFF
--- a/kernel/wasm/KERNEL
+++ b/kernel/wasm/KERNEL
@@ -1,44 +1,108 @@
+ifndef SAMAXKERNEL
 SAMAXKERNEL  = ../riscv64/amax.c
+endif
+ifndef DAMAXKERNEL
 DAMAXKERNEL  = ../riscv64/amax.c
+endif
+ifndef CAMAXKERNEL
 CAMAXKERNEL  = ../riscv64/zamax.c
+endif
+ifndef ZAMAXKERNEL
 ZAMAXKERNEL  = ../riscv64/zamax.c
+endif
 
+ifndef SAMINKERNEL
 SAMINKERNEL  = ../riscv64/amin.c
+endif
+ifndef DAMINKERNEL
 DAMINKERNEL  = ../riscv64/amin.c
+endif
+ifndef CAMINKERNEL
 CAMINKERNEL  = ../riscv64/zamin.c
+endif
+ifndef ZAMINKERNEL
 ZAMINKERNEL  = ../riscv64/zamin.c
+endif
 
+ifndef SMAXKERNEL
 SMAXKERNEL   = ../riscv64/max.c
+endif
+ifndef DMAXKERNEL
 DMAXKERNEL   = ../riscv64/max.c
+endif
 
+ifndef SMINKERNEL
 SMINKERNEL   = ../riscv64/min.c
+endif
+ifndef DMINKERNEL
 DMINKERNEL   = ../riscv64/min.c
+endif
 
+ifndef ISAMAXKERNEL
 ISAMAXKERNEL = ../riscv64/iamax.c
+endif
+ifndef IDAMAXKERNEL
 IDAMAXKERNEL = ../riscv64/iamax.c
+endif
+ifndef ICAMAXKERNEL
 ICAMAXKERNEL = ../riscv64/izamax.c
+endif
+ifndef IZAMAXKERNEL
 IZAMAXKERNEL = ../riscv64/izamax.c
+endif
 
+ifndef ISAMINKERNEL
 ISAMINKERNEL = ../riscv64/iamin.c
+endif
+ifndef IDAMINKERNEL
 IDAMINKERNEL = ../riscv64/iamin.c
+endif
+ifndef ICAMINKERNEL
 ICAMINKERNEL = ../riscv64/izamin.c
+endif
+ifndef IZAMINKERNEL
 IZAMINKERNEL = ../riscv64/izamin.c
+endif
 
+ifndef ISMAXKERNEL
 ISMAXKERNEL  = ../riscv64/imax.c
+endif
+ifndef IDMAXKERNEL
 IDMAXKERNEL  = ../riscv64/imax.c
+endif
 
+ifndef ISMINKERNEL
 ISMINKERNEL  = ../riscv64/imin.c
+endif
+ifndef IDMINKERNEL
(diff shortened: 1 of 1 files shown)


martin-frbg:
Thank you

PR #6003, opened 2026-08-28
Improve f_check when assembly has no vendor identity

jczhang07:
Vendor detection in f_check greps the probe assembly (ftest.s) for a compiler identity string. Some compilers, notably Intel ifx, emit no such string, so the primary `case "$data"` scan fails to set a vendor.

The name-based fallback only recognizes bare compiler names, so an ifx-backed wrapper (e.g. mpif90, mpifort or the Cray ftn wrapper) is not matched either, leaving the vendor empty and resulting in a false conclusion that no Fortran compilers are available.

Append the output of `$compiler --version` to $data so such compilers are identified by their version banner (ifx's contains "ifx", "Intel").

### DIFF
--- a/f_check
+++ b/f_check
@@ -76,6 +76,9 @@ else
     	if [ -z "$data" ]; then
     		data=`$compiler -O2 -S ftest.f > /dev/null 2>&1 && cat ftest.c && rm -f ftest.c`
     	fi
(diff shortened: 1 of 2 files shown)


martin-frbg:
Thank you (guess I'll also need to check why the Frankenstein f2c ?lamch fallback didn't work for you, but that may have already been fixed by the recent #5982)

PR #6020, opened 2026-09-08
wasm: do not enable relaxed SIMD by default

jjerphan:
## Summary

`Makefile.wasm` currently passes `-mrelaxed-simd` on every WASM object, so GEMM/DOT/ROT emit `relaxed_madd` into the final module. WebAssembly validates the whole module: engines that do not implement relaxed SIMD reject instantiate, even if those kernels are never called. Shipping Safari / JavaScriptCore is in that set ([feature status](https://webassembly.org/features/)).

- Keep portable SIMD128 (`-msimd128`) as the default.
- Opt in with `WASM_RELAXED_SIMD=1` (Make) or `-DWASM_RELAXED_SIMD=ON` (CMake). Kernels already switch on `__wasm_relaxed_simd__`.
- Document what relaxed SIMD is, why it is not portable, and how to enable it in `docs/install.md`.

## Test plan

After a WASM configure (`ARCH=wasm` in `Makefile.conf`), with `emcc` and `wasm-dis` on `PATH`:

```bash
./kernel/wasm/test_relaxed_simd.sh
```

That script checks:

- [x] Default compile line has `-msimd128` and not `-mrelaxed-simd`.
- [x] Default GEMM/DOT/ROT/TRMM objects and the linked module contain no `relaxed_madd`, and Node can instantiate that SIMD128-only module.
- [x] `WASM_RELAXED_SIMD=1` passes `-mrelaxed-simd` and those kernels emit `relaxed_madd`.


jjerphan:
Verified locally with Emscripten 4.0.9 (`TARGET=WASM128_GENERIC`, `emcc`).

Default `make` compile line has `-msimd128` only. `WASM_RELAXED_SIMD=1` adds `-mrelaxed-simd`.

Kernels compiled both ways (SGEMM/DGEMM, CGEMM/ZGEMM, SDOT/DDOT/DSDOT, SROT, DTRMM_LN), linked with those symbols exported, then `wasm-dis` + `WebAssembly.instantiate` in Node:

| build | `relaxed_madd` | `f32x4.mul` | `f64x2.mul` | instantiate |
|---|---:|---:|---:|---|
| default | 0 | 64 | 83 | OK |
| `WASM_RELAXED_SIMD=1` | 77 | 42 | 36 | OK (this engine has relaxed SIMD) |

Default SGEMM WAT is `f32x4.add` / `f32x4.mul`. Opt-in SGEMM WAT is `f32x4.relaxed_madd`. Every listed kernel that uses `v_muladd` / `MADD_*` switched the same way.

PR #6050, opened 2026-09-17
kernel/riscv64: fold the complex MAC into the accumulators in CGEMM

6eanut:
The CGEMM micro-kernel for ZVL256B built every complex product into a temporary and then added it to the accumulator:

    tmp = vfmul(Ai, Bi); tmp = VFMACC(tmp, Br, Ar); ACC = vfadd(ACC, tmp)

That is three vector instructions per accumulator and k-step. The two FMAs write the same destination, so applying them straight to ACC yields the same per-k-step sum for two instructions instead:

    ACCr = VFMACC_RR( ACCr, Bi, Ai);
    ACCr = VFMACC_RR( ACCr, Br, Ar);
    ACCi = FOLD_OPA( ACCi, Br, Ai);
    ACCi = FOLD_OPB( ACCi, Bi, Ar);

`VFMACC_RR`/`VFMACC_RI` already encode the +/- conjugation of the variant being compiled, but the imaginary fold needs an op sequence that has no existing macro, so it is spelled out per variant as `FOLD_OPA`/`FOLD_OPB` next to them.

Two FMAs compose to a plain accumulation because `vfmsac(x,u,v)=u*v-x` and `vfmsac(vfmsac(a,u,v),w,y)=a+w*y-u*v`, and likewise for `vfmacc`/`vfnmsac`. The value added per k-step is therefore unchanged and the per-column accumulation order is preserved. Only the FP rounding order differs, by a few ULP; measured against an exact double reference the patched kernel has the same error as the unpatched one.

A packed A element has its real and imaginary halves adjacent, so one strided segment-2 load replaces the two separate strided loads.

### Performance

Measured on a SpaceMiT X100 (2.2 GHz, single thread, ZVL256B), median of three runs: `cblas_cgemm` +44.3% at 256, +45.0% at 512, +42.3% at 1024.

### Testing

`make tests` returns 0; 125/125 utest and 1473/1473 extension tests pass; the CBLAS L1/L2/L3 suites show no new failures.


martin-frbg:
Thank you

